PR #10, opened 2018-01-24
spelling

fblanqui:


rlepigre:
Note that I'm trying to keep the comments "justified" (with an equal number of characters), because I think it looks better (that's probably an OCD). So I tend to add spaces here and there to ensure that it is the case. I am not going to force everyone to keep this invariant, but to at least not break this by removing spaces. I'll merge the PR and revert some of the changes accordingly.

fblanqui:
Ok.


Le 24/01/2018 à 17:00, Rodolphe Lepigre a écrit :

PR #20, opened 2018-03-13
Split

rlepigre:


rlepigre:
The for_all2 function was already written in the Extra module.

PR #37, opened 2018-04-09
Fixed subject reduction problem

rlepigre:
I think the subject reduction problem is solved in this branch, but this needs reviewing.

### DIFF
--- a/src/sr.ml
+++ b/src/sr.ml
@@ -46,7 +46,7 @@ let subst_from_constrs : constrs -> tvar array * term array = fun cs ->
             in
             build_sub acc cs
         | (Symb(Def(sa)), Symb(Def(sb))) when sa == sb ->
-            wrn "%s may not be injective...\n%!" sa.def_name;
+            (*wrn "%s may not be injective...\n%!" sa.def_name;*)
             build_sub acc cs
         | (Vari(x)      , _            ) when argsa = [] ->
             build_sub ((x,b)::acc) cs
--- a/src/sr.ml
+++ b/src/sr.ml
@@ -53,7 +53,7 @@ let subst_from_constrs : constrs -> tvar array * term array = fun cs ->
         | (_            , Vari(x)      ) when argsb = [] ->
             build_sub ((x,a)::acc) cs
         | (a            , b            ) ->
-            wrn "Cannot use constraint [%a ~ %a]...\n" pp a pp b;
+            wrn "Ignored constraint [%a ~ %a]...\n" pp a pp b;
             build_sub acc cs
   in
   let (xs, ts) = List.split (build_sub [] cs) in
--- a/src/sr.ml
+++ b/src/sr.ml
@@ -56,8 +56,8 @@ let subst_from_constrs : constrs -> tvar array * term array = fun cs ->
             wrn "Cannot use constraint [%a ~ %a]...\n" pp a pp b;
             build_sub acc cs
   in
-  let sub = build_sub [] cs in
-  (Array.of_list (List.map fst sub), Array.of_list (List.map snd sub))
+  let (xs, ts) = List.split (build_sub [] cs) in
+  (Array.of_list xs, Array.of_list ts)
 
 (** [eq_modulo_constrs cs t u] checks  whether the terms [t] and [u] are equal
     modulo rewriting and a list of (valid) constraints [cs]. *)
--- a/src/sr.ml
+++ b/src/sr.ml
@@ -41,8 +41,11 @@ let subst_from_constrs : constrs -> tvar array * term array = fun cs ->
         let (hb,argsb) = get_args b in
         match (unfold ha, unfold hb) with
         | (Symb(Sym(sa)), Symb(Sym(sb))) when sa == sb ->
-            let cs = try List.combine argsa argsb @ cs with _ -> cs in
-            build_sub acc cs
+            let args =
+              try List.combine argsa argsb @ cs with _ ->
+              fatal "Unsatisfiable constraint [%a ~ %a]...\n" pp a pp b
+            in
+            build_sub acc (args @ cs)
         | (Symb(Def(sa)), Symb(Def(sb))) when sa == sb ->
             build_sub acc cs
         | (Vari(x)      , _            ) when argsa = [] ->
@@ -50,7 +53,7 @@ let subst_from_constrs : constrs -> tvar array * term array = fun cs ->
         | (_            , Vari(x)      ) when argsb = [] ->
             build_sub ((x,a)::acc) cs
         | (a            , b            ) ->
-            wrn "Not implemented [%a] [%a]...\n%!" pp a pp b;
+            wrn "Cannot use constraint [%a ~ %a]...\n" pp a pp b;
             build_sub acc cs
   in
   let sub = build_sub [] cs in
--- a/src/sr.ml
+++ b/src/sr.ml
@@ -41,12 +41,9 @@ let subst_from_constrs : constrs -> tvar array * term array = fun cs ->
         let (hb,argsb) = get_args b in
         match (unfold ha, unfold hb) with
         | (Symb(Sym(sa)), Symb(Sym(sb))) when sa == sb ->
-            let cs =
-              try List.combine argsa argsb @ cs with Invalid_argument _ -> cs
-            in
+            let cs = try List.combine argsa argsb @ cs with _ -> cs in
             build_sub acc cs
         | (Symb(Def(sa)), Symb(Def(sb))) when sa == sb ->
-            (*wrn "%s may not be injective...\n%!" sa.def_name;*)
             build_sub acc cs
         | (Vari(x)      , _            ) when argsa = [] ->
             build_sub ((x,b)::acc) cs
--- a/src/unif.ml
+++ b/src/unif.ml
@@ -38,6 +38,8 @@ let instantiate : meta -> term array -> term -> bool = fun u env a ->
   let b = Bindlib.bind_mvar (Array.map to_var env) (lift a) in
   Bindlib.is_closed b && (set_meta u (Bindlib.unbox b); true)
 
+(** [unify_list l] tests equality between all the pairs of terms of [l], while
+    possibly instantiating metavariables. *)
 let rec unify_list : (term * term) list -> bool = fun l ->
   match l with
   | []                   -> true
@@ -60,8 +62,7 @@ let rec unify_list : (term * term) list -> bool = fun l ->
       | (Abst(a1,t1)  , Abst(a2,t2)  ) ->
           let (_,t1,t2) = Bindlib.unbind2 mkfree t1 t2 in
           unify_list ((a1,a2)::(t1,t2)::l)
-      | (Appl(t1,u1)  , Appl(t2,u2)  ) ->
-          unify_list ((t1,t2)::(u1,u2)::l)
+      | (Appl(t1,u1)  , Appl(t2,u2)  ) -> unify_list ((t1,t2)::(u1,u2)::l)
       | (Meta(u1,e1)  , Meta(u2,e2)  ) when u1 == u2 ->
           let l = ref l in
           Array.iter2 (fun a b -> l := (a,b)::!l) e1 e2;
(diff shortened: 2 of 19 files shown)


src/sr.ml
```diff
@@ -41,12 +41,9 @@ let subst_from_constrs : constrs -> tvar array * term array = fun cs ->
         let (hb,argsb) = get_args b in
         match (unfold ha, unfold hb) with
         | (Symb(Sym(sa)), Symb(Sym(sb))) when sa == sb ->
-            let cs =
-              try List.combine argsa argsb @ cs with Invalid_argument _ -> cs
-            in
+            let cs = try List.combine argsa argsb @ cs with _ -> cs in
```
fblanqui: Wouldn't it be better to write:
   ```with Invalid_argument _ -> assert false (* cannot happen *)```
?

fblanqui: What if sa and sb are not applied to the same number of arguments (even if sa==sb)? This could happen.

fblanqui: Line 57: use List.split?

fblanqui: Line 53: put a more explicit warning?

rlepigre: As you mentioned, this can happen, so we cannot use an `assert false` here. If the number of arguments is different, then the constraint is no good for building a substitution, and there is nothing we can do with it so it is ignored. In fact, we could fail because this means that the constraints cannot be solved.

An any case, this is not really a bug, because it just means that we accept to type-check rules that we will never be able to apply.

fblanqui: Line 72: rename t into l and u into r?

src/unif.ml
```diff
@@ -37,30 +37,44 @@ let instantiate : meta -> term array -> term -> bool = fun u env a ->
...
   let b = Bindlib.bind_mvar (Array.map to_var env) (lift a) in
   Bindlib.is_closed b && (set_meta u (Bindlib.unbox b); true)
-    
+
+(** [unify_list l] tests equality between all the pairs of terms of [l], while
+    possibly instantiating metavariables. *)
+let rec unify_list : (term * term) list -> bool = fun l ->
+  match l with
+  | []                   -> true
+  | (a,b)::l when a == b -> unify_list l
```
fblanqui: How often can a==b happen?

src/sr.ml
fblanqui: About assert false: Then, it would be good to give more explanations about what this function does.

src/unif.ml
rlepigre: Surprisingly often... ;)

I just tried to count this case on Matita, and this is used one million time after two minutes.

src/unif.ml
```diff
@@ -102,28 +116,63 @@ let unify_modulo : term -> term -> bool = fun a b ->
...
+    | (a,b)::l                ->
+    let a = whnf a in
+    let b = whnf b in
+    let rec get_args args a b =
+      match (unfold a, unfold b) with
+      | (Appl(t1,u1), Appl(t2,u2)) -> get_args ((u1,u2)::args) t1 t2
+      | (Meta(u1,e1), Meta(u2,e2)) when u1 == u2 -> (a,b,args)
+      | (Meta(m,e)  , b          ) when instantiate m e b -> get_args args a b
+      | (a          , Meta(m,e)  ) when instantiate m e a -> get_args args a b
+      | (_          , Meta(_,_)  ) -> assert false
```
fblanqui: Why is there an assert false here? And not for the symmetric case? Comment required.

src/unif.ml
```diff
@@ -37,30 +37,44 @@ let instantiate : meta -> term array -> term -> bool = fun u env a ->
...
+      | (Symb(Def(s1)), Symb(Def(s2))) -> s1 == s2 && unify_list l
+      | (Vari(x1)     , Vari(x2)     ) ->
+          Bindlib.eq_vars x1 x2 && unify_list l
+      | (Prod(a1,b1)  , Prod(a2,b2)  ) ->
+          let (_,b1,b2) = Bindlib.unbind2 mkfree b1 b2 in
+          unify_list ((a1,a2)::(b1,b2)::l)
+      | (Abst(a1,t1)  , Abst(a2,t2)  ) ->
+          let (_,t1,t2) = Bindlib.unbind2 mkfree t1 t2 in
+          unify_list ((a1,a2)::(t1,t2)::l)
+      | (Appl(t1,u1)  , Appl(t2,u2)  ) -> unify_list ((t1,t2)::(u1,u2)::l)
```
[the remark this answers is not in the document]
rlepigre: This function only makes things equal. Here, injectivity is only a problem when we learn constraints in `unify_modulo`, I think.

src/unif.ml
```diff
@@ -102,28 +116,63 @@ let unify_modulo : term -> term -> bool = fun a b ->
...
+      | _    -> unify
+    in
+    let (a,b,args) = get_args [] a b in
+    let args = List.rev args in
+    match (unfold a, unfold b) with
+    | (Wild         , _            )
+    | (_            , Wild         )
+    | (ITag(_)      , _            )
+    | (_            , ITag(_)      )
+    | (Type         , Type         ) -> assert (args = []); unify_list l
```
fblanqui: Why do you call unify_list here, and not unify_modulo?

src/unif.ml
rlepigre: No idea... I probably forgot to remove this line at some point...

src/unif.ml
fblanqui: Why do use asserts? Shouldn't we generate some error instead?

src/unif.ml
rlepigre: For the whole Matita, the physical equality case is used `840 000 000` times!

src/unif.ml
```diff
@@ -102,28 +116,63 @@ let unify_modulo : term -> term -> bool = fun a b ->
...
+    let rec get_args args a b =
+      match (unfold a, unfold b) with
+      | (Appl(t1,u1), Appl(t2,u2)) -> get_args ((u1,u2)::args) t1 t2
+      | (Meta(u1,e1), Meta(u2,e2)) when u1 == u2 -> (a,b,args)
+      | (Meta(m,e)  , b          ) when instantiate m e b -> get_args args a b
+      | (a          , Meta(m,e)  ) when instantiate m e a -> get_args args a b
+      | (_          , Meta(_,_)  ) -> assert false
+      | (_          , _          ) -> (a,b,args)
+    in
+    let unif =
```
fblanqui: Why not always using unify_modulo? Comment required.

src/unif.ml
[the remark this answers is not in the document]
fblanqui: I disagree. If you unify with f(X) and f(Y) with X and Y distinct metavariables and f non injective, then you will instantiate X by Y (or Y by X), which may not be correct.

tests/OK/vectDependant.dk
```diff
@@ -29,8 +29,7 @@ def S_bis : Nat -> Nat.
 cns_bis : n:Nat -> Vec n -> Vec (S_bis n).
 
 def id_bis : n:Nat -> Vec (S_bis n) -> Vec (S_bis n).
-[n,l] id_bis _ (cns_bis n l) --> cns_bis n l.
-[m,n,l] id_bis m (cns_bis n l) --> cns_bis n l.
+[n,l] id_bis n (cns_bis n l) --> cns_bis n l.
 
 (; Does not work anymore... (ill-typed rule since S_bis is definable). ;)
```
fblanqui: Remove "Does not work anymore": this example should not be accepted.

src/sr.ml
```diff
@@ -41,23 +41,23 @@ let subst_from_constrs : constrs -> tvar array * term array = fun cs ->
...
         | (Symb(Def(sa)), Symb(Def(sb))) when sa == sb ->
-            wrn "%s may not be injective...\n%!" sa.def_name;
             build_sub acc cs
         | (Vari(x)      , _            ) when argsa = [] ->
             build_sub ((x,b)::acc) cs
         | (_            , Vari(x)      ) when argsb = [] ->
             build_sub ((x,a)::acc) cs
         | (a            , b            ) ->
-            wrn "Not implemented [%a] [%a]...\n%!" pp a pp b;
+            wrn "Cannot use constraint [%a ~ %a]...\n" pp a pp b;
```
fblanqui: Cannot -> Do not

src/unif.ml
fblanqui: Are a and b just constants or bigger terms? Because we do no hashconsing...

src/unif.ml
rlepigre: Using `unify_modulo` is necessary when not in constraint mode. However, it should not be used in constraint mode, as otherwise we somehow bypass the injectivity check for constraints.

src/unif.ml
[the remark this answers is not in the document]
rlepigre: So you mean that I should implement `unify` (or rather `unify_list`) in the same way as `unify_modulo`?

That is, separating the function in head position from its arguments, an only allowing unifications in the arguments when the function is injective (i.e., a static symbol).

src/unif.ml
```diff
@@ -102,28 +116,62 @@ let unify_modulo : term -> term -> bool = fun a b ->
...
+      | (a          , Meta(m,e)  ) when instantiate m e a -> get_args args a b
+      | (_          , _          ) -> (a,b,args)
+    in
+    let unif =
+      match !constraints with
+      | None -> (fun a b -> unify_modulo [(a,b)])
+      | _    -> unify
+    in
+    let (a,b,args) = get_args [] a b in
+    let args = List.rev args in
```
fblanqui: Is it really necessary to reverse the list?

rlepigre: Yes, because in the last case we need to reattach the arguments.

rlepigre: Actually, I'm not sure. Maybe It should not be reversed at all.

rlepigre: That was a bug, I removed it...

src/unif.ml
```diff
@@ -102,28 +116,62 @@ let unify_modulo : term -> term -> bool = fun a b ->
...
+        unify_modulo l
+    | (Meta(u1,e1)  , Meta(u2,e2)  ) ->
+        u1 == u2 &&
+        Array.for_all2 (fun a b -> unif a b || add_constraint a b) e1 e2 &&
+        List.for_all (fun (a,b) -> unif a b || add_constraint a b) args &&
+        unify_modulo l
+    | (a            , b            ) ->
+        match !constraints with
+        | None ->
+            unify a b &&
```
fblanqui: Why are the possibilities for [unify a b] to hold here?

Moreover, if the Appl case in unify is not fixed, injectivity may be implicitly used here again.

src/unif.ml
rlepigre: Hashconsing is not the only way for things to be equal. There is also substitution, and the instantiation of unification variables. In any case, this is an harmless optimisation.

src/unif.ml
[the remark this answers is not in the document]
fblanqui: Yes, I think so. This is the main problem in your algorithm I think.

src/unif.ml
rlepigre: There is the definable constant case (with non-empty list of arguments).

src/unif.ml
fblanqui: I see. That's interesting. So, I should add this in the pa branch too.

src/unif.ml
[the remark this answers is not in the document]
rlepigre: OK, I'll experiment in this direction then. I find these injectivity questions really frustrating, I hope I'll get my mind around it soon!

Anyway, thanks for the remarks and suggestions.

fblanqui:
Then, do the reverse at this moment only.


Le 09/04/2018 à 15:51, Rodolphe Lepigre a écrit :

rlepigre:
@fblanqui what do you think about this:

https://github.com/rlepigre/lambdapi/blob/57ecd4899291889c27c7126ca1d5f305413b21c1/src/unif.ml#L41-L114

The commented parts are attempts at handling the cases where a meta-variables faces an application. Right now, this is handled in the last few cases of the `unify_list` function. Removing the commented part of `unify_list` prevents unifications like `?42[...] = Vec zero`, and many (most) examples fail.

fblanqui:
I don't understand why you are using eq instead of unif. Why not working 
on infer2.ml instead?


Le 10/04/2018 à 18:32, Rodolphe Lepigre a écrit :

rlepigre:
You told me not to unify when we are not sure that the head of the application is injective. Wasn't this what you meant?

I don't want to move too fast to `src/infer2.ml` because there are still basic things that need to be fixed. I cannot work on the `pa` branch because too many things are broken, including features that have nothing to do with the proof assistant (for example separate compilation), and I cannot identify bugs efficiently. My plan is to integrate the things that do work in `master` so that we can identify the problems in the relevant parts.

PR #58, opened 2018-07-27
Rewrite Tactic

arisPapadop:
Initial implementation of rewrite tactic in src/rewrite.ml. Some additions in examples/proof.dk and small changes to accommodate for the new tactic.


rlepigre:
I think you are a couple commits behind master. You should merge. Note that the `#QED` problem is fixed, so you can fix examples accordingly. Also, I modified the Travis script for actually failing when tests don't go through (this weirdly wasn't the case before).

examples/proof.dk
```diff
@@ -73,41 +73,92 @@ eq : !a, T a -> T a -> Prop.
...
+(; Case S ;)
 #SIMPL.
+#REFINE n => ih => _.
+#REWRITE ih.
+#REWRITE add_succ_r m n0.
+#REFINE refl nat (s (add m n0)).
+#QED.
+
+
+(; At the moment, we can unfortunately prove anything in the proof mode
```
rlepigre: This will not work any more when you merge, so it will probably need to be removed.

src/lambdapi.odocl
```diff
@@ -6,10 +6,10 @@ Extra
 Files
 Handle
 Lambdapi
-Parser
```
rlepigre: I think this line should remain here.

src/rewrite.ml
```diff
@@ -0,0 +1,235 @@
...
+  let b = Bindlib.unbox b in
+  g_meta.meta_value := Some b ;
+  (*Since #REWRITE does not change g_hyps etc. we just update the necessary
+   * parameters in the original goal. *)
+  let thm =
+    {thm with t_goals = {g with g_meta = new_m ; g_type = new_type}::gs} in
+  theorem := Some thm ;
+
+  begin
+    print_endline " -------------- #REWRITE information -------------- " ;
```
rlepigre: Use `wrn` and a newline here. Note that you should only use the functions of the `Console` module for outputs. You can also use the logging function defined at the top of `src/handle.ml`. If you do so, the output will only be enabled if you give the corresponding character to the `--debut` option.

src/rewrite.ml
```diff
@@ -0,0 +1,235 @@
...
+    new goal, where all occurrences of left are replaced by the new term.
+    Use: This will be called with a fresh Vari term to build a product term,
+    using the bindlib lilfting functions. *)
+let match_subst : term -> term -> term -> term = fun g_type l t ->
+  let rec matching_aux : term -> term = fun cur ->
+    let cur = unfold cur in
+    if term_match cur l then t else match cur with
+    | Vari _ | Type | Kind | Symb _ -> cur
+    | Appl(t1, t2) ->
+        let t1' = matching_aux t1 and t2' = matching_aux t2 in Appl(t1', t2')
```
rlepigre: This `let` is useless, inline in the constructor.

src/rewrite.ml
```diff
@@ -0,0 +1,235 @@
...
+
+open Timed
+open Terms
+open Print
+open Console
+open Proofs
+
+(****************************************************************************)
+(* Error function. *)
+let fail_to_match = fun n ->
```
rlepigre: I would avoid using this function as it is not useful, and it is really easy mix up messages. You should inline the calls to `fatal_no_pos`.

examples/proof.dk
arisPapadop: We removed this when we added rewrtie.ml to the project. Some warnings of the sort

+ ocamlfind ocamldoc -dump src/parser.odoc -hide-warnings -charset utf-8 -package timed -package earley.str -package earley -package bindlib -package unix -pp pa_ocaml -I src src/parser.ml
File "src/parser.ml", line 15, character 0:
lexing: empty token
File "src/parser.ml", line 18, character 0:
lexing: empty token
...

appear when making the docs. If that's OK I'll add it back.

rlepigre: If you are talking about the `Parser` module in `src/lambdapi.odocl` then you are right. I removed the line in master, so you can merge again.

rlepigre:
@arisPapadop Note that Travis build fails, due to a problem while checking the file `examples/rewrite.dk`. Make sure to run `make tests`.

rlepigre:
Tanks, I'll merge now.

PR #63, opened 2018-08-27
Rewrite

arisPapadop:
Rewrite tactic with quantifiers implemented.


src/rewrite.ml
```diff
@@ -40,27 +158,17 @@ let bind_match : pattern -> term -> (term, term) Bindlib.binder = fun p t ->
...
   in
-  Bindlib.unbox (Bindlib.bind_var x (lift_subst t))
+  Bindlib.unbox (Bindlib.bind_var x (lift_subst t2))
 
 (** [handle_rewrite t] rewrites according to the equality proved by [t] in the
     current goal. The term [t] should have a type corresponding to an equality
     (without any quantifier for now). All instances of the LHS are replaced by
     the RHS in the obtained goal. *)
 let handle_rewrite : term -> unit = fun t ->
-  (* Obtain the required symbols from the current signature. *)
```
rlepigre: I like this version better, as this code will eventually be changed. In particular, `find_sym` is hackish, and it should not be used anywhere, so I'd rather it be local.

src/rewrite.ml
```diff
@@ -77,27 +185,10 @@ let handle_rewrite : term -> unit = fun t ->
     | None    ->
         fatal_no_pos "Cannot infer the type of [%a] (given to rewrite)." pp t
   in
-
-  (* Check that the type of [t] is of the form “P (Eq a l r)”. *)
```
rlepigre: Similarly, I think it is best to inline the `break_eq` function, as it is very specific. However, the `is_symb` function can be take out. It could even be moved to the `src/terms.ml` file (e.g., right before the definition of `eq`).

src/rewrite.ml
```diff
@@ -152,3 +244,24 @@ let handle_rewrite : term -> unit = fun t ->
   log_rewr "  pred           = [%a]" pp pred;
   log_rewr "  new goal       = [%a]" pp goal_type;
   log_rewr "  produced term  = [%a]" pp term
+
+(* --------------- TODO (From the meetings on 26/07/2018) ------------------ *)
```
rlepigre: I don't think these comments should be included in the main repository.

src/rewrite.ml
```diff
@@ -106,9 +197,10 @@ let handle_rewrite : term -> unit = fun t ->
         fatal_no_pos "Rewrite expects a goal of the form “P t” (found [%a])."
           pp g.g_type
   in
-
-  (* Build the predicate by matching [l] in [g_term]. *)
-  (* TODO keep track of substitutions in the first instance of l in g. *)
+  (***************************************************************************)
```
rlepigre: It is best not to use this separation style, for uniformity reasons. You can do as above: skip a line, and add a comment explaining what happens in this small block of code.

src/rewrite.ml
```diff
@@ -5,27 +5,145 @@ open Terms
...
+open Solve
 
 (** Logging function for the rewrite tactic. *)
 let log_rewr = new_logger 'w' "rewr" "informations for the rewrite tactic"
 let log_rewr = log_rewr.logger
 
-(** For now, a pattern is just a term. *)
-type pattern = term
+(** A substitution is a mapping from metavariables to terms. *)
+type substitution = (term * term) list
```
rlepigre: You could ensure by the type that the first argument is a metavariable. Anyway, I'm not sure if this is the right representation for a substitution. Using Bindlib variables would probably be better than using metavariables. You don't have to fix that now, and we can discuss that later.

PR #66, opened 2018-09-02
Rewrite re-implemented with Bindlib

arisPapadop:


rlepigre:
I only had a quick look, but here is a small comment (I'll give more feedback tomorrow). I think that you should not use an association list when building a substitution, but rather an array of variables and an array of terms with corresponding positions.

The idea would be that `break_prod` would return an array of variables (an not a list) as the domain of the substitution, and you would then build an array of terms (with `Kind` or something similar as a default value, but not an option type). To associate a term to a variable, you would then need to find its index in the array returned by `break_prod`, and write the corresponding term at the same index in the array of terms you build. This way, the array of terms returned when the substitution is build can already be used for `Bindlib.msubst`. You thus avoid to manipulate too many different data structures.

rlepigre:
Another small comment: I find the type definitions at the beginning useless. It is more informative to write `tvar list` than `var_list` for example (`tvar` is defined in `src/terms.ml`). It is better to have comments in the related function to explain what is going on, and how the types are used (what are the invariants, ...).

src/rewrite.ml
```diff
@@ -13,26 +13,37 @@ let log_rewr = log_rewr.logger
...
     with fresh metas from make_meta all the quantified variables in t_type.
     At the same time it applies the new metavariables tot the equality proof,
     so that afterwards they are substituted with the right terms. *)
-let break_prod : term -> term * var_list = fun t_type ->
-  let rec aux : term -> var_list -> term * var_list = fun t_type vars ->
-    match t_type with
-    | Prod(_,b) -> let (v,b) = Bindlib.unbind b in aux b (v::vars)
-    | _         -> (t_type, List.rev vars)
-  in aux t_type []
+let break_prod : term -> term * term Bindlib.var list = fun t ->
```
rlepigre: Use `tvar` instead of `term Bindlib.var`, it is shorter.

PR #85, opened 2018-10-05
[lsp] Migrate to new language format.

ejgallego:
It works (TM), however it needs a patch on dedukti-editor as to
recognize the `.lp` extension. For the moment, use the `.dk` one.


rlepigre:
OK, cool. Should I merge this or not? And if yes, should we just remove the `Pure` module (and rename `New_pure` into `Pure`)?

PR #90, opened 2018-10-11
[pure] Beginnings of unit testing for Pure.

ejgallego:
We move the `command` type to a module and add a small unit test.

I've chosen a quite popular solution, but of course we can discuss
about it.

I am not sure I got the OPAM magic work but that shouldn't be too hard
to fix if bad.


rlepigre:
Do we need another package for that? Travis is complaining.

PR #91, opened 2018-10-12
[ci] Better testing in Travis

ejgallego:
We run the unit tests and add an opam-building job.

`dune-release` will always run the unit tests, but we also run them in
the regular builds just in case.

We tweak a few unit tests on the way. Note, support for 4.04.0 had to
be dropped due to what it looks like an incompatibility with certain
ppx. IMHO that is not so serious as 4.04 users should be in 4.04.2 [or
even better in 4.05].


ejgallego:
Merci !

PR #96, opened 2018-10-18
`let` in old Dk syntax

GuillaumeGen:
To be consistent with https://github.com/Deducteam/Dedukti/pull/131


rlepigre:
Is this PR ever going to be merged, or can we remove it? It seems that the extension has never been merged on the Dedukti side. Is there only plan on that side @Gaspi?

To comments for @GuillaumeGen if we are to merge this:
 - it would be nice to have a test file with an example of the syntax (in, e.g., `tests/OK/let.dk`),
 - you should probably use the `P_LLet` constructor instead of building a redex.

fblanqui:
I close this PR for the moment. It can be reopened if necessary.

PR #97, opened 2018-10-19
Assert tactic

GuillaumeGen:
Creation of a tactic allowing the user to introduce an intermediary lemma.

### DIFF
--- a/src/handle.ml
+++ b/src/handle.ml
@@ -77,15 +77,20 @@ let handle_tactic : sig_state -> Proof.t -> p_tactic -> Proof.t =
   | P_tac_print
   | P_tac_proofterm
   | P_tac_focus(_)      -> assert false (* Handled above. *)
-  | P_tac_assert(t)     ->
+  | P_tac_assert(h,t)     ->
      begin
        (* Obtaining the goals environment, type and meta variable. *)
        let env = fst (Proof.Goal.get_type g) in
        let t = fst (Scope.scope_term StrMap.empty ss env t) in
        let m = Proof.Goal.get_meta g in
        (* Enriching the environnement with a var of type t *)
-       let var = Bindlib.new_var Terms.mkfree "n" in
-       let vs = Array.of_list ((List.map (fun (_,(x,_)) -> x) env)@[var]) in
+       let h = match h with
+         | None   -> "h"
+         | Some x -> x.elt
+       in
+       let var = Bindlib.new_var Terms.mkfree h in
+       let vs = Array.of_list (var::(List.map (fun (_,(x,_)) -> x) env)) in
+       (* The current goal is enriched with an hypothesis *)
        let m2val = match !(m.meta_value) with
          | None     -> None
          | Some(mb) ->
@@ -93,23 +98,12 @@ let handle_tactic : sig_state -> Proof.t -> p_tactic -> Proof.t =
             Some(Bindlib.unbox (Bindlib.bind_mvar vs mbox))
        in
        let sz_ctx = m.meta_arity in
-       let rec add_last_prod n a b =
-         if n = 0
-         then
-           let a_bind = Bindlib.unbox (Bindlib.bind_var var (lift a)) in
-           Prod(b,a_bind)
-         else
-           match a with
-           | Prod(x,y) ->
-              let (v,y) = Bindlib.unbind y in
-              let prod_b_y = Bindlib.box (add_last_prod (n-1) y b) in
-              Prod(x,Bindlib.unbox (Bindlib.bind_var v prod_b_y))
-           | _         -> assert false
-       in
-       let m2typ = add_last_prod sz_ctx !(m.meta_type) t in
+       let mtyp = Bindlib.unbox (Bindlib.bind_var var (lift !(m.meta_type))) in
+       let m2typ = Prod(t,mtyp) in
        let m2 = fresh_meta m2typ (sz_ctx+1) in
        m2.meta_value := m2val;
        let g_enrich = Proof.Goal.of_meta_decomposed m2 in
+       (* A new goal is created with the same context *)
        let rec replace_goal_prod n a b =
          if n = 0
          then b
--- a/src/handle.ml
+++ b/src/handle.ml
@@ -116,7 +116,7 @@ let handle_tactic : sig_state -> Proof.t -> p_tactic -> Proof.t =
            | _         -> assert false
        in
        let m3typ = replace_goal_prod sz_ctx !(m.meta_type) t in
-       let m3 = fresh_meta m3typ m.meta_arity in
+       let m3 = fresh_meta m3typ sz_ctx in
        let new_goal = Proof.Goal.of_meta_decomposed m3 in
        Proof.({ps with proof_goals = new_goal :: g_enrich :: gs})
      end
--- a/src/handle.ml
+++ b/src/handle.ml
@@ -85,8 +85,8 @@ let handle_tactic : sig_state -> Proof.t -> p_tactic -> Proof.t =
       (* Enriching the environnement with a var of type t *)
       let h =
         match h with
-        | None       -> "h"
-        | Some((),x) -> x.elt
+        | None   -> "h"
+        | Some x -> x.elt
       in
       let var = Bindlib.new_var Terms.mkfree h in
       let vs = Array.of_list (var::(List.map (fun (_,(x,_)) -> x) env)) in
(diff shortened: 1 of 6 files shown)


src/parser.ml
```diff
@@ -205,16 +205,18 @@ let parser rw_patt = "[" r:rw_patt_spec "]" -> in_pos _loc r
...
+  | _intro_ xs:ident+            -> Pos.in_pos _loc (P_tac_intro(xs))
+  | _apply_ t:term               -> Pos.in_pos _loc (P_tac_apply(t))
+  | _simpl_                      -> Pos.in_pos _loc P_tac_simpl
+  | _rewrite_ p:rw_patt? t:term  -> Pos.in_pos _loc (P_tac_rewrite(p,t))
+  | _refl_                       -> Pos.in_pos _loc P_tac_refl
+  | _sym_                        -> Pos.in_pos _loc P_tac_sym
+  | _focus_ i:nat_lit            -> Pos.in_pos _loc (P_tac_focus(i))
+  | _print_                      -> Pos.in_pos _loc P_tac_print
+  | _proofterm_                  -> Pos.in_pos _loc P_tac_proofterm
+  | _assert_ t:term              -> Pos.in_pos _loc (P_tac_assert(None   ,t))
```
rlepigre: You can achieve the same thing in one line:
```ocaml
  | _assert_ t:term h:{_as_ h:ident}? -> Pos.in_pos _loc (P_tac_assert(h,t))
```

src/pretty.ml
```diff
@@ -102,6 +102,7 @@ let pp_p_tactic : p_tactic pp = fun oc t ->
   | P_tac_focus(i)           -> out "focus %i" i
   | P_tac_print              -> out "print"
   | P_tac_proofterm          -> out "proofterm"
+  | P_tac_assert(_,t)        -> out "@[<hov 2>assert@ %a@]" pp_p_term t
```
rlepigre: The `as` should probably be printed as well.

src/syntax.ml
```diff
@@ -86,6 +86,8 @@ type p_tactic_aux =
   (** Print the current goal. *)
   | P_tac_proofterm
   (** Print the current proof term (possibly containing open goals). *)
+  | P_tac_assert of (ident option * p_term)
```
rlepigre: I'm guessing you don't want parentheses here. The memory representation is actually not the same with and without parentheses. Without them, both elements of the product are stored in the constructor, but with parentheses the constructor only stores a pointer to a couple (thus there is an extra indirection).

Note also that the order of the two arguments should be reversed to match the syntax (the as comes second).

src/syntax.ml
```diff
@@ -86,6 +86,8 @@ type p_tactic_aux =
...
+  (** [P_tac_assert(t) adds an hypothesis of type [t] and a subgoal of proving [t] *)
```
rlepigre: There is a missing square bracket in `[P_tac_assert(t)]`.

src/handle.ml
```diff
@@ -77,6 +77,49 @@ let handle_tactic : sig_state -> Proof.t -> p_tactic -> Proof.t =
   | P_tac_print
   | P_tac_proofterm
   | P_tac_focus(_)      -> assert false (* Handled above. *)
+  | P_tac_assert(h,t)     ->
+     begin
```
rlepigre: Using `begin` / `end` is not necessary here. Moreover, indentation is not correct (see other cases).

src/handle.ml
```diff
@@ -77,6 +77,49 @@ let handle_tactic : sig_state -> Proof.t -> p_tactic -> Proof.t =
...
+       (* Obtaining the goals environment, type and meta variable. *)
+       let env = fst (Proof.Goal.get_type g) in
+       let t = fst (Scope.scope_term StrMap.empty ss env t) in
+       let m = Proof.Goal.get_meta g in
+       (* Enriching the environnement with a var of type t *)
+       let h = match h with
```
rlepigre: To be consistent with the rest of the code, the `match` should be on the following line (same thing below).

src/handle.ml
```diff
@@ -77,6 +77,49 @@ let handle_tactic : sig_state -> Proof.t -> p_tactic -> Proof.t =
...
+         | None   -> "h"
+         | Some x -> x.elt
+       in
+       let var = Bindlib.new_var Terms.mkfree h in
```
rlepigre: I'm wondering if you took the simplest approach in the following (I did not really look at the details yet). What I would have done is call refine with a term of the form: `(λ(h:t),_) _`, where the first `_` corresponds to the previous goal with the extra hypothesis `h`, and the second `_` corresponds to the proof of `h`. I can explain more in person Monday if you want.

fblanqui:
We should have a general discussion on tactics before merging anything.

Le 19/10/2018 à 21:59, Rodolphe Lepigre a écrit :

rlepigre:
@fblanqui, I don't think there is any harm in extending the set of tactics (or at least if they just amount to a call to refine). They are not critical features in any way, and they can be easily modified or removed if need be.

Can you elaborate on what you think should be discussed?

ejgallego:
Well my very small 2cents is that I think this formulation of assert is fairly cumbersome to use as compared to for example ssreflect's `have`. Particularly interesting topics are generalization for example.

Talking about forward-reasoning tactics I would certainly indeed start to design from the ones in math-comp.

GuillaumeGen:
Rodolphe, I made the changes you required (but did not change the implementation of the tactic, even if it might be a good idea to seek for uniformity and rely on refine).
I had a small issue on the parser, when using the syntax `h:{_as_ h:ident}?`, the  `h` is considered of type `(unit * ident) option` and not simply of type `ident option`.

rlepigre:
Thanks @GuillaumeGen! The problem you are having with the parser is that you need to use `h:{_:_as_ ident}?` rather  than what I gave you in the comments (sorry, my mistake). Note that with earley, what you give in brackets is a sub-grammar. Its semantic value can be given with an arrow, as you can do at the top level of a parser. For instance, you could also write `h:{_as_ h:ident -> h}?`. However, when no arrow is given, the product of the semantic values of the components is given, except for "boring" values (the parsing of a fixed string for example). You can discard one of the values using `_` as a name.

src/handle.ml
```diff
@@ -77,6 +77,49 @@ let handle_tactic : sig_state -> Proof.t -> p_tactic -> Proof.t =
   | P_tac_print
   | P_tac_proofterm
   | P_tac_focus(_)      -> assert false (* Handled above. *)
+  | P_tac_assert(t,h)     ->
```
rlepigre: Here is an alternative implementation for the tactic:
```ocaml
       (* Build the proof term [(λ (h : t), _) _]. *)
       let h = match h with None -> Pos.none "H" | Some(h) -> h in
       let t = Pos.none (P_Appl(Pos.none (PAbst([(h,t)], Pos.none P_Wild)), Pos.none P_Wild)) in
       (* Scoping the term in the goal's environment. *)
       let env = fst (Proof.Goal.get_type g) in
       let t = fst (Scope.scope_term StrMap.empty ss env t) in
       (* Refine using the built term. *)
       handle_refine t
```
Note that it does not make use of any low-level Bindlib stuff. There is still one thing wrong here: there could well be a variable called ["H"] in the environment. This should be checked, and a "fresh" variable name should be generated (say, the first "Hn", with "n" a natural number, such that "Hn" is not in the environment).

src/handle.ml
```diff
@@ -77,6 +77,49 @@ let handle_tactic : sig_state -> Proof.t -> p_tactic -> Proof.t =
...
+      in
+      let var = Bindlib.new_var Terms.mkfree h in
+      let vs = Array.of_list (var::(List.map (fun (_,(x,_)) -> x) env)) in
+      (* The current goal is enriched with an hypothesis *)
+      let m2val =
+        match !(m.meta_value) with
+        | None     -> None
+        | Some(mb) ->
+           let mbox = Bindlib.box (snd (Bindlib.unmbind mb)) in
+           Some(Bindlib.unbox (Bindlib.bind_mvar vs mbox))
```
rlepigre: Note that here, you are binding variables that have no reason to appear in `mbox` (if they did not already appear free in it before the call to `unbind` on the previous line). What were you trying to do here?

rlepigre:
I think we can close this PR for now, since details still need to be figured out. Moreover, it has become hard to merge, and this work can be done again fairly easily. But thanks @GuillaumeGen!

PR #106, opened 2018-10-29
Adding beginning of lib, making [apply t] adding as many _ as needed by [t]'s type, and making refine in a proper function

FranckS:
[apply t] now does refine (t _ _ ... _ _) with as many _ as needed by [t]'s type. The number of _ can be zero, making [apply] usable as an [exact] tactic.


src/apply.ml
```diff
@@ -0,0 +1,31 @@
+open Basics
```
rlepigre: I'm not too happy with moving stuff to new files in this way. First, it does not really make sense to have 30 lines files, and second, this introduces code duplication. The same changes would only introduce three or four lines in the current `src/handle.ml`.

Note that `src/handle.ml` will soon be split in two, but I'd rather do these changes myself since I have a general plan in mind. But at least, we should discuss this before working in two different directions.

src/refine.ml
```diff
@@ -0,0 +1,35 @@
+open Extra
```
rlepigre: Same comment as for `src/handle.ml`. This is not the right time to do that.

src/apply.ml
```diff
@@ -0,0 +1,31 @@
...
+open Console
+open Terms
+open Refine
+
+(** [nbArgsOf t] counts the number of arguments of the type [t], i.e.
+    it counts the number of Prod binders located at the front of [t]. *)
+let rec nbArgsOf : term -> int = fun t ->
```
rlepigre: I think `count_head_products` would be a better name. Note that we use the OCaml naming convention (using `_`) not the Haskell convention (capitalizing first letters of each words).

src/apply.ml
```diff
@@ -0,0 +1,31 @@
...
+  match unfold t with
+  | Prod(_,b)   -> let (_,b) = Bindlib.unbind b in 1 + nbArgsOf b
```
rlepigre: I'd write `1 + snd (Bindlib.unbind b)` here, it is slightly shorter. Also, It is better to align the arrows of the pattern matching for readability.

rlepigre:
Note also that `List.init` is not available on all version of OCaml that we support (this is the Travis error).

FranckS:
Thanks for these suggestions Rodolphe. I'll re-do this in a very minimalistic way (not moving things in different files, etc), so that it doesn't interfere with your current plans for handle.ml.

rlepigre:
@FranckS: Any progress on that front?

FranckS:
I'll open another pull request.

PR #111, opened 2018-11-05
Implementing beginning of standard lib for Dedukti 2

FranckS:
This is a very early work in progress, but including it now to the main repository would help for discussing with others about the kind of library that we need, as suggested by Frédéric.


rlepigre:
Another thing: I don't know if you have seen, but I added the possibility to define infix operators. It is probably a good idea to use them for addition, multiplication, less than, ...

Note that you cannot define a symbol for (polymorphic) equality yet. However, nothing prevents you to define a local equality symbol `eq_nat` such that `eq_nat &m &n` reduces to `eq Nat &m &n`, and to define a symbol for `eq_nat`. To achieve better we will need implicit arguments, and a more general notion of operator. For now, an operator only corresponds to a symbol (not an arbitrary, closed term).

FranckS:
Damn, I was sure that I did run the tests. Apologies! I'll correct this.

FranckS:
I'll open another pull request to make things cleaner.

PR #121, opened 2018-11-28
[opam] Remove the lambdapi-lsp OPAM package.

### LINKED ISSUE #120: Have a single opam package
@ejgallego: do you mind if I remove the lp-lsp package? It don't feel that it makes much sense to have a separate package since we can install the binary together with `lambdapi`. I don't think that such a small tool deserves more than one package on the central Opam repository anyway. What do you think?

If you agree, what should I do with `lp-lsp/README.md`? The version numbers won't make sense with respect to `lambdapi`, as it was not really versioned before...

ejgallego:
We know include the server in the main package.

Fixes #120


CHANGES.md
```diff
@@ -1,3 +1,7 @@
+### 1.1 ()
```
rlepigre: I guess this should be part of the 1.0 release, that I'm planning to issue as soon as the work on LSP is somewhat ready, and the installation of the emacs mode is done properly.

ejgallego: Oh, I saw the date thus I assumed the release was done. Amending.

rlepigre: Yes sorry, I was planning to release today, and I changed my mind with this single-opam-package thing!

PR #128, opened 2018-12-13
Termination checker output

GuillaumeGen:
A new option --termination outputing a file for the termination competition format (XTC).
However, since there is no category including dependent types yet, I had to enrich the XTC syntax to handle it.


src/lambdapi.ml
```diff
@@ -95,6 +108,9 @@ let spec =
     ; ( "--confluence"
       , Arg.String (fun cmd -> confluence_checker := Some(cmd))
       , "<cmd> Runs the given confluence checker." )
+    ; ( "--termination"
```
fblanqui: It should be explained somewhere in lambdapi.ml or USER_MANUAL.md what happens if there are several --confluence or --termination flags.

src/menhir_parser.mly
```diff
@@ -282,4 +282,8 @@ term:
   | x=ID COLON a=aterm FARROW t=term {
       make_pos $loc (P_Abst([(make_pos $loc(x) x, Some(a))], t))
     }
+  | L_PAR x=ID COLON t=aterm DEFEQ b=aterm R_PAR FARROW u=term {
```
fblanqui: Why do you change the grammar? It doesn't seem related to termination checking.

GuillaumeGen: Indeed, I branched on the wrong commit, I will change this right now.

GuillaumeGen:
I close this PR and open a better one.

fblanqui:
You do not need to close the PR, do you? You can just update your branch.

Le 13/12/2018 à 16:09, GuillaumeGen a écrit :

GuillaumeGen:
I indeed could have suppress unrelated modifications, but it would have induced more commits and a dirty tree. I know that having a clean tree is not our main, but since it cost me almost nothing to do it cleanly.

rlepigre:
No, you can force-push. That is the usual way of doing it, but only on a personal branch.

PR #144, opened 2019-02-10
[lsp] Fix processing of proofs.

ejgallego:
There are some remaining problems tho, but they should be easy to fix:

- No way to get the goals from a proof state.
- `Pure.end_proof` produces an exception instead of returning a more
  informative result.

Tested with emacs, seems to work modulo the problems with the exceptions in `end_proof`.

### DIFF
--- a/lp-lsp/lp_doc.ml
+++ b/lp-lsp/lp_doc.ml
@@ -30,27 +30,56 @@ type doc = {
   nodes : doc_node list;
 }
 
+let option_default o1 d =
+  match o1 with | None -> d | Some x -> x
+
 let mk_error ~doc pos msg =
   LSP.mk_diagnostics ~uri:doc.uri ~version:doc.version [pos, 1, msg, None]
 
+let process_pstep (pstate, diags) tac =
+  let open Pure in
+  let tac_loc = Tactic.get_pos tac in
+  match handle_tactic pstate tac with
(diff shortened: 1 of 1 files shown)


ejgallego:
Ok, this is ready to merge despite the problem with the highlighting of `end_proof`. Actually I'd be cool if the AST provided some extra locations for:
```
theorem foo : B ⇒ B
proof
   tacs...
qed
```
I'd be great to get the location of `qed` and of `theorem foo : .... proof` as to provide good highlighting.

rlepigre:
Thanks @ejgallego, I'll merge and see what I can do to fix the remaining issues tomorrow!

rlepigre:
@ejgallego positions are now available for the theorem statement and for the terminator (using functions taking as input a proof state). I also fixed the problem with the exception. I'll take care of the missing things later.

ejgallego:
Looks good thanks!

rlepigre:
@ejgallego: I added a function te extract goals from the proof state in 17780d2cfdb1a8b65289e4112398c99edb3d43d5. You can have a look at the implementation of `Proof.pp` to see how to extract the relevant data. Alternatively, I could give you a function with type `proof_state -> ((string * term) * term) list` if you prefer.

ejgallego:
It is good like this, thanks!

PR #146, opened 2019-02-11
[lsp] Improve proof and goal display.

ejgallego:
Thanks to recent improvements. `theorem` is green if the statement
typechecks, `qed` is green when the proof has been finished.

### DIFF
--- a/lp-lsp/lp_doc.ml
+++ b/lp-lsp/lp_doc.ml
@@ -41,8 +41,7 @@ let process_pstep (pstate, diags) tac =
   let tac_loc = Tactic.get_pos tac in
   match handle_tactic pstate tac with
   | Tac_OK pstate ->
-    (* Fixme *)
-    let goals = None in
+    let goals = Some (current_goals pstate) in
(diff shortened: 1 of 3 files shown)


rlepigre:
Stupid question: is there any way to display the goal using the Emacs mode?

ejgallego:
> Stupid question: is there any way to display the goal using the Emacs mode?

Not as of today; `eglot` doesn't like non-standard fields in the diagnostics.

We could either:
- use the more complete `emacs-lsp` mode and add support for the goals field;
- add goal information to hover

That second choice is fairly easy to do although not optimal.

rlepigre:
Maybe we should use the second option as a first step then. That is clearly not great, but that's a start. Did you manage to get Atom to work again?

ejgallego:
Indeed atom does work now; however I don't see the goals [they are sent but atom doesn't print them]

It must be something trivial but I won't have time to look at Atom today.

rlepigre:
OK. I personally don't think that Atom is worth fixing, especially if it is gonna die soon. I guess it is more important to have some support for emacs, even if it is only through hover.

PR #155, opened 2019-02-13
split USER_MANUAL in various files

fblanqui:


rlepigre:
I think we should continue having a `README.md` that give more information on Lambdapi. For instance, what it is, how to install it, and even a short example. In other words, it should not be the index of the user manual. But it should of course point to it.

GuillaumeGen:
I agree with Rodolphe, that the `README` should be more than a list of hypertext links.

fblanqui:
Why? Which information do you want to see first in the README?

rlepigre:
Because when I just have a look at a repository out of curiosity, I expect to find enough information in the README for me to know what the project is about. And then if I am interested, I dig deeper. And I think many people work like that.

Plus, some informations that were (only) in the previous README have been completely removed. We should do something about it, especially the installation instruction "from source".

GuillaumeGen:
I agree with Rodolphe about the fact that the `README` is a showcase of the project for newcomers, and a list of links is not an attractive showcase.
PS : The installation from sources is described in the `compiling and profiling` section.

fblanqui:
Le 13/02/2019 à 13:25, Rodolphe Lepigre a écrit :
>
> Because when I just have a look at a repository out of curiosity, I 
> expect to find enough information in the README for me to know what 
> the project is about. And then if I am interested, I dig deeper. And I 
> think many people work like that.
>
Isn't the title enough?


> Plus, some informations that were (only) in the previous README have 
> been completely removed. We should do something about it, especially 
> the installation instruction "from source".
>
Nothing has been removed or lost, just moved around. Compilation from 
sources is in devel.md.

rlepigre:
OK I see, and no that is far from enough for me. I think we need at least:
 - a paragraph explaining the aim of the project
 - installation instruction from source and with opam
 - an example of source file containing a motivating example
 - a links to the documentation, to related papers, and how to get involved in the project.

rlepigre:
Here are a couple of examples:
 - https://github.com/coq/coq
 - https://github.com/ocaml/ocaml
 - https://github.com/neovim/neovim

fblanqui:
Ok. Do it please.

Le 13/02/2019 à 13:39, Rodolphe Lepigre a écrit :

rlepigre:
OK, that is mostly done. The README is not perfect yet, but others can improve on that (I recovered part of the old README). Note that I moved some documentation things around. For instance the `syntax.bnf` file is now in `doc`, where it should be. And a created a `sections` directory to cleanup the `doc` directory so that it only contains "entry points". Links have been fixed, so things should work as expected.

PR #157, opened 2019-02-13
Add a pygmentize lexer for Dedukti

francoisthire:
Add a file `tool/dedukti.py` that can be used with the LaTeX package `minted`.


fblanqui:
Hi François. Thank you very much! Could you please add some sample Latex file to test your script?

francoisthire:
Which folder should I use for that?

rlepigre:
Just create a directory `tools/latex_minted`. Ideally, it would be great to have a `.sty` defining a new environment so that we can do
```latex
\begin{dedukti}
...
\end{dedukti}
```
And similarly with `lambdapi`.

francoisthire:
I tried something, let me know if this is what you expect.

rlepigre:
Just put all files in the `tools/latex_minted` directory, and that will be perfect. We can add a Lambdapi syntax later. And you can also remove the dedukti file, since its contents is in the tex fil already.

francoisthire:
That should be it.

Issue #147, opened 2019-02-12
Add a pre-commit hook in github?


fblanqui:
Hi Rodolphe. With https://github.com/Deducteam/lambdapi/commit/b835172eea912434f799e53d6d145109784113d6, you added some script doing some sanity checks and a Makefile target. This could be done automatically with git.

rlepigre:
I known you can do that locally using `git`, but the file cannot be shared since it is in the `.git` directory. What we could do is add the check to the CI, but for now the script I wrote always returns successfully (with exit code 0), so failure cannot be tested.

fblanqui:
A discussion on file indentation started on https://github.com/Deducteam/lambdapi/pull/159#issuecomment-463662632 .

GuillaumeGen:
In PR #165 , I suggest running `make sanity_check` with continuous integration. I agree with @fblanqui it is required if we want everyone to follow these rules.

rlepigre:
PR #165 has been merged. This is a start, but it only does its job if every commit goes through a PR, and that it is only merged after CI is successful. Can we close this issue?

GuillaumeGen:
Otherwise, you receive a mail saying that build on `master` are broken.

GuillaumeGen:
And who has the right to commit on `master` without doing a PR?

rlepigre:
Only @fblanqui and I can push to `master` directly, I think. But we should try to avoid that.

PR #159, opened 2019-02-13
parser: allow arguments in symbol declarations

fblanqui:


src/handle.ml
```diff
@@ -65,7 +65,7 @@ let handle_cmd_aux : sig_state -> command -> sig_state * proof_data option =
       in
       (* Open the module. *)
       (open_sign ss sign, None)
-  | P_symbol(ts, x, a)         ->
+  | P_symbol(ts, x, xs, a)         ->
```
rlepigre: Remove 4 whitespaces before the arrow, to align it with the corresponding ones.

fblanqui:
This is possible in Coq for theorems and used a lot in MathComp I 
believe. Well, this PR is about symbol declarations and not about 
theorems. I need to do it for theorems too. This is coherent with what 
is already allowed for definition and let. I see no reason to allow it 
for definition and not for symbol and theorem since a definition is just 
a symbol declaration + a definition.

Le 13/02/2019 à 18:05, Rodolphe Lepigre a écrit :

fblanqui:
Thanks @rlepigre to validate this PR.

src/syntax.ml
```diff
@@ -262,18 +263,19 @@ let eq_p_cmd : p_cmd eq = fun c1 c2 ->
       m1 = m2 && eq_require_mode r1 r2
   | (P_open(m1)                  , P_open(m2)                  ) ->
       m1 = m2
-  | (P_symbol(l1,s1,a1)          , P_symbol(l2,s2,a2)          ) ->
-      l1 = l2 && eq_ident s1 s2 && eq_p_term a1 a2
+  | (P_symbol(l1,s1,al1,a1)      , P_symbol(l2,s2,al2,a2)      ) ->
+     l1 = l2 && eq_ident s1 s2 && eq_p_term a1 a2
```
rlepigre: Indentation is not correct here, missing one space before `l1 = l2 && ...` and on the following line.

src/syntax.ml
```diff
@@ -145,13 +145,14 @@ type p_cmd =
...
-  | P_symbol     of symtag list * ident * p_type
+  | P_symbol     of symtag list * ident * p_arg list * p_type
   (** Symbol declaration. *)
   | P_rules      of p_rule list
   (** Rewriting rule declarations. *)
   | P_definition of bool * ident * p_arg list * p_type option * p_term
   (** Definition of a symbol (unfoldable). *)
-  | P_theorem    of (ident * p_type) loc * p_tactic list * p_proof_end loc
+  | P_theorem
+    of (ident * p_arg list * p_type) loc * p_tactic list * p_proof_end loc
```
rlepigre: We should define a type `p_statement` for the first argument:
```ocaml
type p_statement = (ident * p_arg list * p_type) loc
```
And this way the declaration will fit again on one line.

src/syntax.ml
ejgallego: Note that both `ocp-indent` and `ocamlformat` choose instead to remove the extra before `l1 = l2` [2 minus than what you use]. This style makes sense as you are 4 spaces right of the `|`, and helps save very precious horizontal space.

ejgallego:
> And one remark: In the current state, the "arguments" are not automatically introduced. However, we might want these intros to happen automatically (I think this is the case in Coq). What do you think? In any case, this should not be done as part of this PR since it is not clear what is the right way to do that.

I would avoid doing as in Coq, which needs to call the `intros` tactic to elaborate stuff. IMVHO desugaring at parsing-time is the way to go.

ejgallego:
> I would avoid doing as in Coq, which needs to call the `intros` tactic to elaborate stuff. IMVHO desugaring at parsing-time is the way to go.

This helps keeping the constructor as it originally was, however you may want to add some info for better printing [such as the number of args to the left] but that info remains purely for debug purposes.

src/syntax.ml
rlepigre: I know that some people do that, but I don't like this style since it hurts readability in my opinion. I largely prefer having well-identified block, and I rely on other methods to shorten the length of lines (avoiding too many levels of nesting, ...).

rlepigre:
> > And one remark: In the current state, the "arguments" are not automatically introduced. However, we might want these intros to happen automatically (I think this is the case in Coq). What do you think? In any case, this should not be done as part of this PR since it is not clear what is the right way to do that.
> 
> I would avoid doing as in Coq, which needs to call the `intros` tactic to elaborate stuff. IMVHO desugaring at parsing-time is the way to go.

Actually, this I was thinking of inserting lambdas in front of the goal to extend its scope. There is no need to call any tactic for that in Lambdapi, and that could well be done at scoping time.

fblanqui:
One of the problem is that we do not use the same editors and not the 
same config for indentation. I use tuareg. I don't know if this is 
possible to modify its behavior so that we get the same indentation rules.

Le 14/02/2019 à 15:57, Rodolphe Lepigre a écrit :

ejgallego:
> One of the problem is that we do not use the same editors and not the same config for indentation. I use tuareg. I don't know if this is possible to modify its behavior so that we get the same indentation rules.

If you are using `tuareg + ocp-indent` [which is the recommended setup] indeed we can add a configuration file to the repository so indentation is consistent.

ejgallego:
But more and more people are just leaving formatting to `ocamlformat` and thus don't worrying about indentation at all.

I like this approach even if the tool still needs some tweaks / options.

fblanqui:
I didn't have ocp-indent. I just installed it with opam. What else do I 
have to do?

Please, go on and update the CONTRIBUTING files. :-)

Le 14/02/2019 à 16:12, Emilio Jesús Gallego Arias a écrit :

fblanqui:
How to enforce this with git? Perhaps, we should continue this 
discussion on https://github.com/Deducteam/lambdapi/issues/147.

Le 14/02/2019 à 16:13, Emilio Jesús Gallego Arias a écrit :

rlepigre:
I tried setting up things in a satisfactory way with `ocp-indent` once, and it just does not do what I want. This actually is the best way to mess up the repository.

PR #160, opened 2019-02-13
[lsp] Add preliminary VSCode mode.

ejgallego:
Many thanks to Vicent Laporte who extracted the relevant parts from
VSCoq.

The mode works, however we are back at the game of `Pure` raising
`assert false`.

I am very much afraid that the current model is unworkable; the LSP
API and the internal programming model are essentially different so it
is futile to manually ensure that they are in sync. Indeed, almost
every time I get back to the code, I tend to find some new assert that
is triggered by the LSP code.

I have added a workaround for the lack of location data on a "file not
found by require" error, but this should be fixed upstream.


rlepigre:
Great! Thanks @ejgallego! I love `assert false`! This is probably just a small bug, can you tell me how I can reproduce? It is not very surprising that you discover bugs in the `Pure` interface, since you are the only one using it right now. I don't think there is any fundamental problem with this approach, it is more of a philosophical problem I guess.

ejgallego:
My point is that we should use typing to ensure this kind of stuff; if `parse` cannot raise unlocated assertions then the type should reflect that; similarly for the state, it gets out of sync once we load the full project, some parts of the signature leak from one document to another :/

I am much afraid that if we have these kind of problems on this small scale imagine when the codebase grows. I am confident on the LSP code as we are torturing it big time with Coq and never found a leaking problem [so far, of course I'll have its own bugs]

The best way to reproduce is to use VSCode which is miles ahead of any other UI in terms of LSP support.

In this case, the unlocated exception is raised when trying to `require open` a non-existing file, hinted by the use of `fatal_no_pos` in parser.

Cheers!

ejgallego:
Leaking of signature state can be reproduced by just playing around with the files, it could well be that the exceptions are not interacting properly with the imperative state, as it tends to happen when I introduce errors on the `require open`.

rlepigre:
Oh right, I see what the problem is. This is due to changes in the way we load dependencies, I completely forgot about that. Of course, errors with no locations can now be triggered. I should have anticipated the problem in `Pure`. In that precise case, we should be able to give the position of the `require`.

There cannot be any bad interaction between references and state, since rollbacks are only based on saved times. Exceptions are only having an effect on the control flow.

rlepigre:
By the way, can I merge?

ejgallego:
Sure! This is working except for the signature state leaks.

ejgallego:
> There cannot be any bad interaction between references and state, since rollbacks are only based on saved times. Exceptions are only having an effect on the control flow.

I dunno what is going on, but indeed at least in the current version of this PR signature state leaks / is not restored properly.

ejgallego:
> I dunno what is going on, but indeed at least in the current version of this PR signature state leaks / is not restored properly.

My point is that so far this has become a very common problem here, so maybe we should use something that ensures correctness by construction? The codebase is small now, but I am afraid how it would look like when it is larger.

rlepigre:
OK, I'll see what I can do on that front. But feel free to send me suggestions! ;)

rlepigre:
OK, I fixed the missing position problem with 355cd9d7c8ffaa04cbb6020e8c32251f4b8e8439. Now error will be reported on the `require` command. I also found another missing position in the parser. I'll try to see what I can do to use more type safety here, but I first want to finish the parser and possibly cleanup a couple of things.

rlepigre:
@ejgallego: I don't understand how to install the vscode extension. Should I run `npm install` in the `lp-lsp` directory? And if yes, what should be the effect of that command? I am getting:
```
npm WARN saveError ENOENT: no such file or directory, open '/home/rodolphe/dev/lambdapi/lp-lsp/package.json'
npm WARN enoent ENOENT: no such file or directory, open '/home/rodolphe/dev/lambdapi/lp-lsp/package.json'
npm WARN lp-lsp No description
npm WARN lp-lsp No repository field.
npm WARN lp-lsp No README data
npm WARN lp-lsp No license field.

up to date in 1.051s
found 0 vulnerabilities
```

I am using `Code - OSS 1.31.0` (if I understand correctly that is an open source subset of vscode, but is that suitable?).

ejgallego:
See upcoming PR.

PR #161, opened 2019-02-14
[lsp] Enable goal display on hover.

ejgallego:
This provides goal display on hover which while not perfect allows one
to complete a proof.

Note the VSCode interprets the goal display as markdown so we want to
add a different printing mode for improved readability.


ejgallego:
Improved install instructions are here @rlepigre you need to build in the vscode dir, then symlink it to `~/.vscode/extensions`

I use `dune exec code` to get the right env.

rlepigre:
OK, I see. Thank you!

PR #164, opened 2019-02-15
Sanity

GuillaumeGen:
I made a mistake. The sanity checker does not accept the current state of master, here is a correction.

PS: Wouldn't it be possible to check it during CI, to avoid it to happen again (I am often too hurried to push) ?


GuillaumeGen:
Ok, Rodolphe already did it, I am closing this PR.

PR #190, opened 2019-03-08
parser: allow many arguments of the same type

fblanqui:


rlepigre:
What you did looks good, but I would prefer to reflect the actual syntax in the [Syntax] module. Of course, the changes are slightly more invasive (although the parser changes are extremely minimal), but I don't think that this is a problem.

I will propose an alternative implementation so that you can have a look.

fblanqui:
If you can do it quickly, that's fine. Because it's currently terrible to make proofs and debug tactics. If you cannot do it quickly, then I propose you to merge this PR and make yourself another one later.

PR #191, opened 2019-03-08
print: do not print implicit arguments, and print infix symbols infix

fblanqui:


fblanqui:
After that, we could add some flags to print or not print implicits, type annotations in abstractions or type annotations in products.

fblanqui:
I agree but this is also very urgent to print the goals in a way similar to the user (do not print implicit arguments and print infix symbol application infix) because it is currently terrible to make proofs and debug tactics. So, I would prefer to merge this PR right now, and improve the code later (including point 2), by adding a way to print every thing explicitly if needed.

About 1. Note sure it makes things more readable than simply writing (plus a1 a2 a3 l r). And this would require new parsing code (what we print should be parsable). So, I would prefer not to do it, at least for the moment: this is not urgent.

fblanqui:
done in https://github.com/Deducteam/lambdapi/pull/194

PR #194, opened 2019-03-08
Better pretty-printing + boolean flags.

rlepigre:
Several improvements on pretty-printing (not optimal, but already a lot better):
 1) Infix symbols are printed back as infix symbols (note that if you type `add n m` in the code, you will get `add n m`, but you will get `n + m` if you type `n + m`).
 2) the λ-abstractions are factorized, meaning that we always print `λx y z, t` instead of `λx, λy, λz, t`, and similarly for dependent products,
 3) a flag controls the printing (or not) of the domains of λ-abstractions (they are not printed by default).

The latter point is based on a new API in [Console] that allows registering boolean flags. Roughly, you declare a new flag by calling `Console.register_flag key default_value`, which gives you back a reference to a boolean. You can then use this reference in your code to perform actions depending on its value. Then the value of the flag can be changed by a new configuration command of the form `set flag "key" on` / `set flag "key" off` to enable / disable the flag.

To resume the printing of the domains of the λ-abstractions in some file, you thus need to write the following line.
```
set flag "print_domains" on
```

Things that could be improved later:
 - print less parentheses around symbols by taking into account their given associativity and precedence (not completely trivial),
 - ~~add another flag to disable the printing of implicit arguments.~~

Note that this is an alternative to #191 with some additions (contraction of binders, flag mechanism), ~~and some missing pieces (printing of implicits)~~.

### DIFF
--- a/src/console.ml
+++ b/src/console.ml
@@ -1,6 +1,7 @@
 (** Output and debugging utilities. *)
 
 open Timed
+open Extra
 
 (** Short name for a standard formatter. *)
 type 'a outfmt = ('a, Format.formatter, unit) format
@@ -131,3 +132,20 @@ let out : int -> 'a outfmt -> 'a = fun lvl fmt ->
   let fmt = if !log_enabled then mag fmt else fmt ^^ "%!" in
   if lvl > !verbose then Format.ifprintf Pervasives.(!out_fmt) fmt
   else Format.fprintf Pervasives.(!out_fmt) fmt
+
+(** List of registered boolean flags. *)
+let boolean_flags : bool ref StrMap.t Pervasives.ref =
+  Pervasives.ref StrMap.empty
+
+(** [register_flag id d] registers a new boolean flag named [id], with default
+    value of [d]. Note the the name should not have been used previously. *)
+let register_flag : string -> bool -> bool ref = fun id default ->
+  if StrMap.mem id Pervasives.(!boolean_flags) then
+    invalid_arg "Console.register_flag: already registered";
+  let r = ref default in
+  Pervasives.(boolean_flags := StrMap.add id r !boolean_flags); r
+
+(** [set_flag id b] sets the value of the flag named [id] to be [b], or raises
+    [Not_found] if no flag with this name was registered. *)
+let set_flag : string -> bool -> unit = fun id b ->
+  StrMap.find id Pervasives.(!boolean_flags) := b
--- a/src/print.ml
+++ b/src/print.ml
@@ -12,6 +12,9 @@ open Terms
 (** Flag controling the printing of the domains of λ-abstractions. *)
 let print_domains : bool ref = Console.register_flag "print_domains" false
 
+(** Flag controling the printing of implicit arguments. *)
+let print_implicis : bool ref = Console.register_flag "print_implicits" true
+
 (** [pp_symbol h oc s] prints the name of the symbol [s] to channel [oc] using
     the printing hint [h] to decide qualification. *)
 let pp_symbol : pp_hint -> sym pp = fun h oc s ->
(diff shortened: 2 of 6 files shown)


ejgallego:
Are the flags syncronized wrt functional state?

From the document model point of view, controlling printing with flags has proven to be very problematic for that exact reason. It is much better to have on demand, stateless printing so that clients do always pass the flags they want to use for display.

rlepigre:
Indeed, flags are synchronized with the functional state, so they should not pose any problem. Note that it would be very simple to provide, e.g., a printing function depending on a value for flags or other debugging features.

rlepigre:
I added another flag to avoid printing the implicit arguments. Unlike domains of λ-abstractions, implicit arguments are printed by default. It will be possible to change that in the future, but it is better stick to that for now (at least until we have reliable unification heuristics).

fblanqui:
Thanks Rodolphe. I will check your code on Monday afternoon. But I agree with Emilio: printing flags should be local. It would be painful to do: set flag; print; unset flag each time you want to print something explicitly for debug. We would like to be able to just do print flag.

fblanqui:
Do not print implicits should be the default behavior, even for debug. It's only locally when necessary you want to print implicit arguments.

rlepigre:
Having global options does not prevent you from having local configuration. In the editor, queries would do that for you by setting arguments before running the query, and rolling back to the state prior to the query afterwards.

I think that it would be a mistake to hide implicit arguments by default in debugging messages right now. We can do that when we know that they work as expected. If you want to remove them (mostly) everywhere, you only need one line at the top of your file.

ejgallego:
It is fine to have global options in the toplevel if you want, but those should be managed from the toplevel itself. The interface to the printing functions itself should be fully-functional, as for example to allow proper unit testing.

Global flags lead to a mess than you always have to clean up the moment the project grows after a certain size.

src/console.ml
```diff
@@ -131,3 +132,20 @@ let out : int -> 'a outfmt -> 'a = fun lvl fmt ->
   let fmt = if !log_enabled then mag fmt else fmt ^^ "%!" in
   if lvl > !verbose then Format.ifprintf Pervasives.(!out_fmt) fmt
   else Format.fprintf Pervasives.(!out_fmt) fmt
+
+(** List of registered boolean flags. *)
+let boolean_flags : bool ref StrMap.t Pervasives.ref =
+  Pervasives.ref StrMap.empty
+
+(** [register_flag id d] registers a new boolean flag named [id], with default
+    value of [d]. Note the the name should not have been used previously. *)
```
fblanqui: the the

src/print.ml
```diff
@@ -5,9 +5,16 @@
...
 
+open Timed
 open Extra
 open Terms
 
+(** Flag controling the printing of the domains of λ-abstractions. *)
+let print_domains : bool ref = Console.register_flag "print_domains" false
+
+(** Flag controling the printing of implicit arguments. *)
+let print_implicis : bool ref = Console.register_flag "print_implicits" true
```
fblanqui: print_implicis -> print_implicits

src/print.ml
```diff
@@ -27,43 +34,94 @@ let pp_meta : meta pp = fun oc m ->
...
+  let out = Format.fprintf in
   let rec pp (p : [`Func | `Appl | `Atom]) oc t =
-    let pp_func = pp `Func in
-    let pp_appl = pp `Appl in
-    let pp_atom = pp `Atom in
+    let (h, args) = Basics.get_args t in
+    let args =
+      if !print_implicis then args else
+      let impl =  match h with Symb(s,_) -> s.sym_impl | _ -> [] in
+      let rec filter_impl impl args =
```
fblanqui: Why not putting this generic function on lists in Extra like in my PR?

rlepigre: We could, but it is unlikely that this function will be required somewhere else. And I think that it is easier to see what we are doing here with the implementation of the function next to its use. If you insist, I can move it! :)

PR #200, opened 2019-03-25
Define the relation used in our algorithm for checking subject reduction

wujuihsuan2016:
The following code defines the relation used in our algorithm for checking subject reduction.
There might be some modifications after since some details in the algorithm still need to be clarified.

### DIFF
--- a/src/subject_reduction.ml
+++ b/src/subject_reduction.ml
@@ -1,0 +1,43 @@
+open Terms
+open Print
+open Basics
+
+module Eset = Set.Make(
+  struct
+    type t = (term * term) 
+    let compare = Pervasives.compare
+  end)
+
+exception No_relation
+
+(** [find_relation t] calculates a pair of {!type:term} [A] and {!type:Eset.t}
+    [E] such that t:A[E] holds and raises the exception [No_relation] if
+    such a pair does not exist. Note that such a pair is unique if it exists
+    **)
+let find_relation : term -> (term * Eset.t) = fun t ->
+  let t = unfold t in
+  match t with
+  | Symb (s, _) -> (!s.sym_type, Eset.empty)
+  | Appl (_, _) ->
+      (* Check if t is of the form ft₁...tn” *)
+      let rec appl_aux : term list -> term -> (sym * term list) = fun acc t ->
+        match t with 
+        | Appl (t1, t2) -> appl_aux (t2 :: acc) t1
+        | Symb (s, _)   -> (s, acc)
+        | _             -> raise No_relation in 
+      let (s, ts) = appl_aux [] t in
+      let rs = List.map find_relation ts in
+      let rec cal_relation t tlist rs es = match tlist with
+        | []       -> (t, es)
+        | t1 :: ts -> begin
+            match t with
+            | Prod (a, b) ->
+                let t' = Bindlib.subst b t1 in
+                let (a1, e1) :: rs' = rs in
+                let es' = Eset.add (a1, a) (Eset.union es e1) in
+                cal_relation t' ts rs' es'
+            | _           -> raise No_relation 
+            end
+      in
+      cal_relation s.sym_type ts rs Eset.empty
+  | _           -> raise No_relation


src/subject_reduction.ml
```diff
@@ -0,0 +1,43 @@
+open Terms
+open Print
+open Basics
+
+module Eset = Set.Make(
+  struct
+    type t = (term * term)
```
fblanqui: Parentheses are useless here.

fblanqui:
Hi Jui-Hsuan. Thanks for your first PR!

Note that this PR should modify the file sr.ml.

fblanqui:
The set of equations is already computed by the type inference algorithm. You don't have to compute it again. The main problem is to turn this set of equations into a confluent and terminating rewrite systems using completion. To this end, you should export this set of equations to some external tool, call it, and get back the convergent rewrite system if any. Then, you can check that the RHS has the type of the LHS modulo these rewrite rules using the current type-checking function.

src/subject_reduction.ml
```diff
@@ -0,0 +1,43 @@
+open Terms
+open Print
+open Basics
+
+module Eset = Set.Make(
+  struct
+    type t = (term * term) 
+    let compare = Pervasives.compare
```
rlepigre: Note that `Pervasives.compare` will not work to compare terms, since their representation contains closures (in Bindlib Binders).

wujuihsuan2016:
> The set of equations is already computed by the type inference algorithm. You don't have to compute it again. The main problem is to turn this set of equations into a confluent and terminating rewrite systems using completion. To this end, you should export this set of equations to some external tool, call it, and get back the convergent rewrite system if any. Then, you can check that the RHS has the type of the LHS modulo these rewrite rules using the current type-checking function.

I took a look at the MKBtt tool (http://cl-informatik.uibk.ac.at/software/mkbtt/index.php) today. According to the documentation, yices 1.11 headers and libs are required in order to compile the source code but I haven't found them online yet. I also tried to use the binary executable file which is available online. It works well for the generation of complete systems. However, I always get the message "the process terminated with exit code 127" when using the option -cert in order to get certifiable proofs and I am still trying to figure out where it comes from. Besides, it seems that there exists already a function in confluence.ml that allows to print in the TRS format and it can be adapted for our purpose (adding equations).

rlepigre:
> The main problem is to turn this set of equations into a confluent and terminating rewrite systems using completion. To this end, you should export this set of equations to some external tool, call it, and get back the convergent rewrite system if any. 

Wouldn't it be easier to implement completion directly? That is not very hard, and it would be a lot faster than having to call an external tool and converting the data both ways.

wujuihsuan2016:
> Wouldn't it be easier to implement completion directly? That is not very hard, and it would be a lot faster than having to call an external tool and converting the data both ways.

In fact, in the traditional Knuth-Bendix completion algorithm, a reduction ordering on terms is required and it is in general given by users, which is not very convenient in practice. In the MKBtt tool, some termination tools are used to deal with this issue. There are more details here (http://cl-informatik.uibk.ac.at/software/mkbtt/papers/SWKM08.pdf). I am not sure if this feature can be easily implemented.

fblanqui:
Rodolphe, completion is very hard in general (harder than termination itself!). And, as we can already export rules to the TRS (for confluence) and XTC (for termination) formats, it should not be too difficult to use these exports to call external completion tools too. But, you are right, as we only have closed equations, we could also implement an adhoc optimized completion algorithm using any total reduction ordering on terms (e.g. LPO over an arbitrary total ordering on symbols). This is fine if, in the end, we get no critical pairs with user-defined rewriting rules. Otherwise, we may have to rely on a general completion tool. More theoretical work would be useful in this case. So, it may be interesting to have both.

fblanqui:
By the way, I see that the exports to TRS and XTC are not documented. @GuillaumeGen : could you please add a PR with at the beginning of termination.ml some details about the corresponding export functions? In particular, it should explain what is done with abstractions and products as TRS is an untyped first-order setting. As for XTC, it should explain how dependent types are handled. Is it the current simply-typed version of XTC or its extension with dependent types? @wujuihsuan2016 : you can probably do the same thing for confluence.ml by reverse-engineering the code.

fblanqui:
@wujuihsuan2016 : there exist various completion tools. It would be good to try several.

wujuihsuan2016:
In termination.ml, a qualified identifier dir1. ... .dirn.id is translated to c_dir1_..._dirn_id. It is impossible to retrieve information of symbols using their translations (c_a_b_id is the translation of both a.b.id and a_b.id for example). Thus, it might be useful to keep a mapping : translation(sym) -> sym.

fblanqui:
Is there any restriction on identifiers in the XTC format? If not, as I suspect for XTC is an XML format, then there is no reason for changing names.

fblanqui:
Is there any restriction on identifiers in the XTC format? If not, as I suspect for XTC is an XML format, then there is no reason for changing names. This should be fixed.

It's different perhaps for the TRS format. Question: what are the restrictions on identifiers in the TRS format? In particular, is "." forbidden? If not, then there is no reason for changing names and should be fixed.

fblanqui:
In the TRS format (see http://project-coco.uibk.ac.at/problems/trs.php), we have:

>id is any nonempty sequence of characters not containing
>    whitespace
>    the characters (   )   "   ,   |   \
>    the sequences ->   ==   COMMENT   VAR   RULES

So, dots can be used in names.

In the XTC format (see http://cl2-informatik.uibk.ac.at/mercurial.cgi/TPDB/file/tip/xml/xtc.xsd), there are no restrictions on names.

rlepigre:
For the record, I implemented the translation to the TRS format based on what was done in Dedukti, hence using the same translation conventions. It would indeed be great if we could preserve more identifiers. To rule out any problem with variables named `COMMENT`, `VAR` or `RULES`, I don't see any other solution than to prefix every identifier coming from Lambdapi with some sequence of character (e.g., `c_`). And that also prevents name clashes with the locally defined symbol (see the prelude).

For the XTC format, one has to be careful with special characters (as in any XML format), but they can be easily escaped.

In both cases, somewhat arbitrary sequences of characters can arise in Lambdapi through the `{|...|}` syntax. Thus some care should be taken, and a first approximation could be to fail when the translation reaches such an identifier.

For exchanging arbitrary identifiers, one solution is to use some bijective encoding. As an example, sequences of characters like `{|...|}` could be base-64-encoded, and then prefixed with some reserved prefix. And decoding would then be the obvious thing. In any case, it may be a good idea to define encoding / decoding functions that would always be used by the translation.

src/completion.ml
```diff
@@ -0,0 +1,147 @@
+(** Completion procedure *)
```
fblanqui: Precise that this is a completion procedure for closed equations.

fblanqui:
General remark: for computing critical pairs, do not use the unification algorithm implemented in unif.ml, but define your own syntactic first-order unification algorithm.

src/completion.ml
```diff
@@ -0,0 +1,147 @@
...
+
+open Terms
+open Basics
+open Extra
+
+(** total strict order *)
+type 'a order = 'a -> 'a -> bool
```
fblanqui: It is perhaps more convenient to use comparison functions:
```
type 'a cmp = 'a -> 'a -> int
(* [cmp x y] returns
- [0] when x=y
- [r<0] when x<y
- (r>0] when x>y *)
```
Then, there is no need to introduce gt, ge, etc.
```

File: src/completion.ml [hunk 5]
```diff
@@ -0,0 +1,147 @@
...
+
+let gt : 'a order -> 'a -> 'a -> bool = fun ord a b -> ord a b
+
+(** [ge ord] calculates the reflexive closure of [ord] *)
+let ge : 'a order -> 'a -> 'a -> bool = fun ord a b -> a = b || ord a b
+
+(** [ord_lex ord] calculates the lexicographic order corresponding to the
+    alphabetical order [ord] *)
+let ord_lex : 'a order -> ('a list) order = fun ord l1 l2 ->
```
fblanqui: parentheses are not needed

src/completion.ml
```diff
@@ -0,0 +1,147 @@
...
+  let rec elim_comm l1 l2 = match (l1, l2) with
```
fblanqui: Please, follow the style and indentation of master files.

fblanqui: What is the meaning of "elim_comm"?

src/completion.ml
fblanqui: Moving from ord to cmp, we need:
```
let lex : 'a cmp -> 'a list cmp'
```

fblanqui: Idem for lpo.

src/completion.ml
```diff
@@ -0,0 +1,147 @@
...
+      build_prod k p
+  in
+  let mk_typ = Bindlib.unbox (build_prod k _Type) (*FIXME?*) in
+  let mk = fresh_meta mk_typ k in
+  let tk = _Meta mk (Array.map _Vari vs) in
+  Bindlib.unbox (build_prod k tk)
+
+type rule_ = sym * pp_hint * rule
+
+(** [cp builtins r1 r2] calculates the critical pairs of the rules r1 and r2
```
fblanqui: calculates -> computes

src/completion.ml
```diff
@@ -0,0 +1,147 @@
...
+      let b = Bindlib.bind_var vs.(k) p in
+      let p = Bindlib.box_apply2 (fun a b -> Prod(a,b)) tk b in
+      build_prod k p
+  in
+  let mk_typ = Bindlib.unbox (build_prod k _Type) (*FIXME?*) in
+  let mk = fresh_meta mk_typ k in
+  let tk = _Meta mk (Array.map _Vari vs) in
+  Bindlib.unbox (build_prod k tk)
+
+type rule_ = sym * pp_hint * rule
```
fblanqui: 1. rule_ is not nice
2. pp_hint doesn't seem useful. Please remove it.

Then, you can do:
```
type rule = sym * Term.rule
```

src/completion.ml
```diff
@@ -0,0 +1,147 @@
...
+
+(** [cp builtins r1 r2] calculates the critical pairs of the rules r1 and r2
+    *)
+let cp : sym StrMap.t -> rule_ -> rule_ -> (term * term) list
+  = fun builtins (s1, h1, r1) (s2, h2, r2) ->
+  let arity1 = Bindlib.mbinder_arity r1.rhs in
+  let arity2 = Bindlib.mbinder_arity r2.rhs in
+  let cps = ref [] in
+  let rec to_m : int -> (meta option) array -> term -> tbox = fun k metas t ->
```
fblanqui: If this function is taken from somewhere else, it would be better to not duplicate the code, but factorize it instead.

src/completion.ml
```diff
@@ -0,0 +1,147 @@
...
+    | false ->
+        if f = g then
+          List.for_all (fun x -> gt (lpo ord) t1 x) args' &&
+          ord_lex (lpo ord) args args'
+        else false
+
+(** [build_meta_type k] builds the type “∀(x₁:A₁) (x₂:A₂) ⋯ (xk:Ak), B” where
+    “x₁” through “xk” are fresh variables, “Ai = Mi[x₁,⋯,x(i-1)]”, “Mi” is a
+    new metavar of arity “i-1” and type “∀(x₁:A₂) ⋯ (x(i-1):A(i-1), TYPE”. *)
+let build_meta_type : int -> term = fun k ->
```
fblanqui: If this function is taken from somewhere else, it would be better to not duplicate the code, but factorize it instead.

src/completion.ml
```diff
@@ -0,0 +1,147 @@
...
+      let xs = Bindlib.new_mvar mkfree xs in
+      let e = Array.map _Vari xs in
+      TE_Some (Bindlib.unbox (Bindlib.bind_mvar xs (_Meta m e)))
+    in
+    let te_envs1 = Array.map fn metas1 in
+    let te_envs2 = Array.map fn metas2 in
+    let rhs1 = Bindlib.msubst r1.rhs te_envs1 in
+    let rhs2 = Bindlib.msubst r2.rhs te_envs2 in
+    let to_solve = [(t, lhs2)] in
+    match Unif.(solve builtins false {no_problems with to_solve}) with
```
fblanqui: As said in a previous comment, finally, this is not a good idea for the moment to use the unification algorithm defined in unif.ml because it is doing unification modulo rewriting. You need to use syntactic unification here. So it is better to write your own unification algorithm.

fblanqui:
For the sake of simplicity and modularity, I propose to proceed in clearly distinct steps as follows:
1. Translate rule lhs's into terms by replacing pattern variables by fresh metavariables. Translate rhs's by substituting bound TEnv's by those metavariables as well.
2. Compute critical pairs on such terms by using a first-order syntactic unification algorithm failing when there is an abstraction (hence all metavariables of interest will be of arity 0). This already provides a way to check local confluence.
3. For doing completion, we further need to check that rhs's are first-order too (no abstraction/product or metavariable applications).

src/completion.ml
fblanqui: See https://github.com/Deducteam/lambdapi/pull/200#issuecomment-481692245.

src/completion.ml
```diff
@@ -0,0 +1,147 @@
...
+    of the rewrite system rs *)
+let rec critical_pairs : sym StrMap.t -> rule_ list -> (term * term) list
+  = fun builtins rs ->
+  match rs with
+  | [] -> []
+  | r :: rs' ->
+      let cps = critical_pairs builtins rs in
+      List.fold_left
+        (fun acc r' ->
+           cp builtins r r' @ cp builtins r' r @ acc)
```
fblanqui: It would be more efficient to have cp take an accumulator as well to avoid append's.

src/sr.ml
```diff
@@ -69,6 +70,64 @@ let build_meta_type : int -> term = fun k ->
   let tk = _Meta mk (Array.map _Vari vs) in
   Bindlib.unbox (build_prod k tk)
 
+(** Translation from Xml-light
```
fblanqui: The translation is not from Xml-light but from CPF. Add URL http://cl-informatik.uibk.ac.at/software/cpf/. xml is just the data structure used to represent the CPF file.

fblanqui: It is probably better to put this function in a separate cpf.ml file.

wujuihsuan2016: I am not sure if this is a good idea since what I am doing is not for general use. This is only used for translating CPF files generated by external completion tools.

fblanqui: Just make this clear at the beginning of the file.

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+open Terms
+open Basics
+open Extra
+open Timed
+
+type 'a cmp = 'a -> 'a -> int
+
+(** [ord_lex ord] computes the lexicographic order corresponding to the
+    alphabetical order [ord] *)
+let rec ord_lex : 'a cmp -> 'a list cmp = fun ord l1 l2 ->
```
fblanqui: Thanks to replace ord by cmp everywhere.

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+      end
+  | _              -> assert false
+
+(** [to_term r] translates the rule [r] into a pair of terms *)
+let to_term : rule -> term * term = fun (s, r) ->
+  let arity = Bindlib.mbinder_arity r.rhs in
+  let metas = Array.init arity (fun _ -> None) in
+  let lhs = List.map (fun p -> Bindlib.unbox (to_m 0 metas p)) r.lhs in
+  let lhs = add_args (symb s) lhs in
+  (* [to_term_env m] computes the term with environment correspoding to the
```
fblanqui: typo: correspo+n+ding

src/completion.ml
```diff
@@ -0,0 +1,292 @@
+(** Completion procedure for closed equations *)
+
+open Terms
+open Basics
+open Extra
+open Timed
+
+type 'a cmp = 'a -> 'a -> int
```
fblanqui: For the sake of completeness, please add a comment on the semantics of the output (same as the one of OCaml std lib).

src/completion.ml
fblanqui: Should probably be moved to Extra.

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+  match (l1, l2) with
+  | [], []             -> 0
+  | [], _              -> -1
+  | _, []              -> 1
+  | h1 :: t1, h2 :: t2 ->
+      match ord h1 h2 with
+      | 0 -> ord_lex ord t1 t2
+      | x -> x
+
+let get_sym : term -> sym = fun t -> match unfold t with
```
fblanqui: Should probably be moved to Terms or Basics.

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+      let (x, t) = Bindlib.unbind t in
+      _Abst (to_m 0 metas a) (Bindlib.bind_var x (to_m 0 metas t))
+  | Appl (t, u)    -> _Appl (to_m (k + 1) metas t) (to_m 0 metas u)
+  | Patt (i, n, a) ->
+      begin
+        let a = Array.map (to_m 0 metas) a in
+        let l = Array.length a in
+        match i with
+        | None   ->
+            let m = fresh_meta ~name:n (build_meta_type (l + k)) l in
```
fblanqui: For confluence, we do not care of types. Simply use Kind instead of build_meta_type.

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+  let mk_typ = Bindlib.unbox (build_prod k _Type) (*FIXME?*) in
+  let mk = fresh_meta mk_typ k in
+  let tk = _Meta mk (Array.map _Vari vs) in
+  Bindlib.unbox (build_prod k tk)
+
+type rule = sym * Terms.rule
+
+(** [to_m k metas t] computes a new (boxed) term by replacing every pattern
+    variable in [t] by a fresh meta-variable and store the latter in [metas],
+    where [k] indicates the order of the term obtained *)
```
fblanqui: I don't understand what is k. Please, give more explanations.

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+let rec to_m : int -> (meta option) array -> term -> tbox = fun k metas t ->
```
fblanqui: Why having a meta option array and not a meta array?
Why the meta array is an input and not an output?

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+  | Symb(s1, _), Symb(s2, _)      ->
+      if s1 == s2 then unif eqs else raise Unsolvable
+  | Vari _, _
+  | _, Vari _
+  | Prod _, _
+  | _, Prod _
+  | Abst _, _
+  | _, Abst _ -> assert false
+  | Appl (t1, u1), Appl (t2, u2)  -> unif ((t1, t2) :: (u1, u2) :: eqs)
+  | Meta (m1, ts1), Meta (_, ts2) ->
```
fblanqui: Isn't this case a particular case of the next one?

src/completion.ml
wujuihsuan2016: When we consider a rule, the arity of rhs can be directly obtained by the function Bindlib.mbinder_arity. It is thus useful to take an empty array as an input when we apply to_m. Since we could have several occurrences of the same meta-variable in a term, we use a meta option array to prevent reinitialization.

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+  | _                             -> raise Unsolvable
+
+(** [unif eqs] attempts to solve the unification problem [eqs] *)
+and unif : unif_constrs -> unit = fun eqs ->
+  match eqs with
+  | []              -> ()
+  | (t1, t2) :: eqs -> unif_aux t1 t2 eqs
+
+(** [cps rs] computes the critical pairs of the rewrite system [rs] *)
+let cps : rule list -> (term * term) list = fun rs ->
```
fblanqui: Please, put this function in a separate file, say confluence.ml, as it will be used for checking local confluence too later.

src/completion.ml
wujuihsuan2016: Indeed, but I think we need to check that both meta-variables are applied to nothing.

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+            match metas.(i) with
+            | Some m -> _Meta m a
+            | None   ->
+                let m = fresh_meta ~name:n (build_meta_type (l + k)) l in
+                metas.(i) <- Some m;
+                _Meta m a
+      end
+  | _              -> assert false
+
+(** [to_term r] translates the rule [r] into a pair of terms *)
```
fblanqui: Please give more details on what kind of terms you get: how are translated pattern variables and term_env's?

src/completion.ml
fblanqui: Ok, I see why the meta array must be given as input. But I think that you do not need the option, do you?

fblanqui: We could even define the meta array before and not update it later as metavariable names are not very important.

src/completion.ml
wujuihsuan2016: In fact, this function is also used in sr.ml when we attempt to infer the type of the LHS of a rule. In this case, the types of the metavariables count (see term_of_meta in terms.ml).

fblanqui:
Please reply to my comments on 
https://github.com/Deducteam/lambdapi/pull/200/files. Otherwise we 
cannot follow what you are talking about and whether this is fixed or not.

Le 16/04/2019 à 09:51, Wu Jui-Hsuan a écrit :

fblanqui:
Ignore this message. I didn't look at the right part of the code... :-(

Le 16/04/2019 à 11:20, Frédéric Blanqui a écrit :

src/completion.ml
fblanqui: Right. But this was the point of a previous comment: don't copy and paste, but factorize! This function build_meta_type should be defined only once in Basics, and should be renamed by something more appropriate like build_prod_meta.

src/completion.ml
fblanqui: Hello @rlepigre . Could you please explain what is the k argument for in https://github.com/Deducteam/lambdapi/blob/33809c57b1b39aeb65bb0583b2139bdec46b2e4e/src/sr.ml#L80 ?

src/completion.ml
fblanqui: Right. You need to add a when.

fblanqui: Or simply replace ts1 by [||].

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+and unif : unif_constrs -> unit = fun eqs ->
+  match eqs with
+  | []              -> ()
+  | (t1, t2) :: eqs -> unif_aux t1 t2 eqs
+
+(** [cps rs] computes the critical pairs of the rewrite system [rs] *)
+let cps : rule list -> (term * term) list = fun rs ->
+  let acc = ref [] in
+  let cps_aux : rule -> rule -> unit = fun r1 r2 ->
+    let lhs1, rhs1 = to_term r1 in
```
fblanqui: The translation of rules should be done only once in cps before entering cps_aux. Here, you do it n²/2 times...

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+      iter_meta reset_meta lhs2;
+      try
+        unif [(t1, t2)];
+        acc := (rhs1, rhs2) :: !acc
+      with Unsolvable -> () in
+    let _, args1 = get_args lhs1 in
+    let _, args2 = get_args lhs2 in
+    find_cp lhs1 lhs2;
+    List.iter (iter (find_cp lhs1)) args2;
+    List.iter (iter (fun t -> find_cp t lhs2)) args1 in
```
fblanqui: It seems to be incomplete. For computing all critical pairs, you need to try to unify a lhs l with every non-variable subterm of every lhs (including l itself).

fblanqui: You can use Terms.iter.

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+  let rec cps rs =
+    match rs with
+    | []       -> ()
+    | r :: rs' -> List.iter (cps_aux r) rs; cps rs' in
+  cps rs;
+  !acc
+
+(** Rewriting for closed rules and closed terms *)
+
+type ctxt = term -> term
```
fblanqui: Please, do not overload ctxt already defined in Ctxt.t.

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+
+(** [match_rule ctxt t (s, r)] attempts to rewrite [ctxt[t]] by appling the
```
fblanqui: You should not write yet another normalization function but use Eval.snf instead by replacing metavariables in rules by fresh function symbols with no rule (function gamma).

src/completion.ml
```diff
@@ -0,0 +1,292 @@
...
+(** [to_rule (lhs, rhs)] translates the pair [(lhs, rhs)] into a rule for
+    closed lhs and rhs. *)
+let to_rule : term * term -> rule = fun (lhs, rhs) ->
+  let (s, args) = get_args lhs in
+  let s = get_sym s in
+  let vs = Bindlib.new_mvar te_mkfree [||] in
+  let rhs = Bindlib.unbox (Bindlib.bind_mvar vs (lift rhs)) in
+  s, { lhs = args ; rhs = rhs ; arity = List.length args ; vars = [||] }
+
+(** [completion eqs ord] applies Huet's completion procedure on the set of
```
fblanqui: Note that we do not need a so complex algorithm (with marked rules, etc.) because we only have closed equations.

src/completion.ml
wujuihsuan2016: The reason why I didn't choose to use Eval.snf is that when we attempts to apply rules in Eval.find_rule, we only consider the rules stored in s.sym_rules However, in the algorithm of completion, we may need to compute the normal forms of a term in different rewrite systems. I don't think it's a good idea to update the field s.sym_ruies every time and that's why I implemented an ad hoc function for computing normal forms of closed terms.

src/completion.ml
fblanqui: Note in particular that, since equations are closed, there is no need for unification.

Yet, unification is useful for checking the local confluence in the general case. That's why I propose to move unif and cps to a separate file dedicated to confluence.

Now, let's consider subject reduction. We start from a set of general unification constraints computed by type inference. We could proceed as follows:
1. Extract first-order equations from these constraints.
2. Replace metavariables by fresh function symbols with no type and no rule.
3. Orient the resulting equations by LPO. This gives a set R0 of rewriting rules. Update the rules of each symbol accordingly.
4. Here, we could compute critical pairs but, since equations are closed, we just need to do some normalization.
Assume for instance that we have a rule g->d and a rule l[g]->r. Then, we have a critical pair (l[d],r), but l[d] is just a reduct of l[g]. So, when we have a rule l->r, we could as well normalize l into l' with the rules distinct from l->r, and similarly normalize r into r'.
- If l'=r' then we can remove the rule l->r.
- If l=l' then there is no critical pair and thus nothing to do.
- If l<>l'>r' in LPO, then we can add the rule l'->r'.
- If l<>l'<r', then we can add the rule r'->l'.
Then, we start again with the new set of rules until there is no critical pair. Note that each new rule is obtained by reducing an initial equation by rules including in LPO. Therefore, this terminates.

src/subject_reduction.ml
```diff
@@ -0,0 +1,43 @@
...
+  end)
+
+exception No_relation
+
+(** [find_relation t] calculates a pair of {!type:term} [A] and {!type:Eset.t}
```
fblanqui: No need to rewrite yet another type inference algorithm. Just use Typing.infer_constr like in the current sr.ml file. The difference with the new algorithm is after this point: instead of extracting a substitution from the constraints, we extract a set of rules.

src/completion.ml
fblanqui: No, there is no problem in updating the sym_rules field. We can use the Timed module to backtrack if necessary.

src/basics.ml
```diff
@@ -112,6 +112,13 @@ let is_symb : sym -> term -> bool = fun s t ->
   | Symb(r,_) -> r == s
   | _         -> false
 
+(** [get_symb t] returns [Some s] if [t] is of the form [Symb (s , _)].
+    Otherwise, it returns [None]. *)
+let get_symb : term -> sym option = fun t ->
```
fblanqui: Rename into get_sym has the type is sym.

src/basics.ml
```diff
@@ -201,6 +208,103 @@ let has_metas : term -> bool =
   let exception Found in fun t ->
   try iter_meta (fun _ -> raise Found) t; false with Found -> true
 
+(** [build_meta_type k] builds the type “∀(x₁:A₁) (x₂:A₂) ⋯ (xk:Ak), B” where
+    “x₁” through “xk” are fresh variables, “Ai = Mi[x₁,⋯,x(i-1)]”, “Mi” is a
+    new metavar of arity “i-1” and type “∀(x₁:A₂) ⋯ (x(i-1):A(i-1), TYPE”. *)
```
fblanqui: Parenthesis missing after A(i-1)

fblanqui: Description of B missing. Should be rename into A(k+1)?

fblanqui: metavar -> metavariable

src/basics.ml
```diff
@@ -201,6 +208,103 @@ let has_metas : term -> bool =
...
+let build_meta_type : int -> term = fun k ->
```
fblanqui: Rename into build_prod

src/basics.ml
fblanqui: Ignore my previous comment: there are other functions using symb instead of sym already.

src/basics.ml
```diff
@@ -201,6 +208,103 @@ let has_metas : term -> bool =
...
+(** [new_symb name t] returns a new function symbol of name [name] and of
+    type [t]. *)
+let new_symb name t =
+  { sym_name = name ; sym_type = ref t ; sym_path = [] ; sym_def = ref None
+  ; sym_impl = [] ; sym_rules = ref [] ; sym_mode = Const }
+
+(** [to_m k metas t] computes a new (boxed) term by replacing every pattern
+    variable in [t] by a fresh metavariable and store the latter in [metas],
+    where [k] indicates the order of the term obtained *)
+let rec to_m : int -> (meta option) array -> term -> tbox = fun k metas t ->
```
fblanqui: Rename to_m into something more informative like replace_patt_by_meta.
Parentheses around meta option are useless.

src/basics.ml
```diff
@@ -201,6 +208,103 @@ let has_metas : term -> bool =
...
+            match metas.(i) with
+            | Some m -> _Meta m a
+            | None   ->
+                let m = fresh_meta ~name:n (build_meta_type (l + k)) l in
+                metas.(i) <- Some m;
+                _Meta m a
+      end
+  | _              -> assert false
+
+exception Not_FO
```
fblanqui: Comment missing! What does that mean FO? Use a more informative name.

Where is it used?

src/basics.ml
```diff
@@ -201,6 +208,103 @@ let has_metas : term -> bool =
...
+
+(** [to_closed symbs t] computes a new (boxed) term by replacing every
+    pattern variable in [t] by a fresh symbol [c_n] of type [t_n] ([t_n] is
+    another fresh symbol of type [Kind]) and store [c_n] the latter in
+    [symbs]. *)
+let rec to_closed : (sym option) array -> term -> tbox
```
fblanqui: Rename into replace_patt_by_symb. Remove parentheses around sym option.

src/basics.ml
```diff
@@ -201,6 +208,103 @@ let has_metas : term -> bool =
...
+                let c_n = new_symb ("{c_" ^ n) term_t_n in
+                symbs.(i) <- Some c_n;
+                _Symb c_n Nothing
+      end
+  | Patt _            -> raise Not_FO
+  | _                 -> assert false
+
+(** [is_new_symb s] checks if [s] is a function symbol created for checking
+    SR. *)
+let is_new_symb s = s.sym_name.[0] = '{'
```
fblanqui: This is too adhoc. When you replace pattern variables par symbols, you get an array with the new symbols. Use this array instead.

src/basics.ml
```diff
@@ -239,3 +343,52 @@ let term_of_rhs : rule -> term = fun r ->
     TE_Some(Bindlib.unbox (Bindlib.bind_mvar vars p))
   in
   Bindlib.msubst r.rhs (Array.mapi fn r.vars)
+
+(** [to_terms r] translates the rule [r] into a pair of terms. The pattern
+    variables in the LHS are replaced by metavariables and the terms with
+    environment in the RHS are replaced by their corresponding metavariables.
+    *)
+let to_terms : sym * rule -> term * term = fun (s, r) ->
```
fblanqui: rename into replace_patt_by_meta

fblanqui: Isn't it possible to factorize the codes of the functions replace_patt_by_*?

src/basics.ml
```diff
@@ -239,3 +343,52 @@ let term_of_rhs : rule -> term = fun r ->
...
+  let lhs = add_args (symb s) lhs in
+  let to_term_env : sym option -> term_env = fun s ->
+    let s = match s with Some s -> s | None -> assert false in
+    TE_Some (Bindlib.unbox (Bindlib.bind_mvar [||] (_Symb s Nothing))) in
+  let terms_env = Array.map to_term_env symbs in
+  let rhs = Bindlib.msubst r.rhs terms_env in
+  (lhs, rhs)
+
+(** [check_fo t] checks that [t] is a first-order term. *)
+let rec check_fo : term -> unit = fun t ->
```
fblanqui: When a function may raise an exception, this must be said in the comment.
Replace fo by first_order.
Why raising an exception and not returning a boolean? This should be explained.

src/basics.ml
```diff
@@ -239,3 +343,52 @@ let term_of_rhs : rule -> term = fun r ->
...
+  match t with
+  | Type
+  | Kind
+  | Symb _
+  | Wild
+  | Patt _                      -> ()
+  | Meta (_, ar) when ar = [||] -> ()
```
fblanqui: replace this when clause by | Meta (_,[||])

src/basics.ml
fblanqui: The name is misleading or the code is wrong: this function does not check that the term is first-order. You should specify what is first-order. Why not using iter to implement it?

src/extra.ml
```diff
@@ -130,6 +130,27 @@ module Array =
       Array.length a1 = Array.length a2 && for_all2 eq a1 a2
   end
 
+module Ord =
```
fblanqui: Why using a module here?

src/completion.ml
```diff
@@ -0,0 +1,228 @@
+(** Completion procedure for closed equations *)
+
+open Terms
+open Basics
+open Extra
+open Timed
+
+(** [lpo ord] computes the lexicographic path ordering corresponding to
+    the strict total order [ord] on the set of all function symbols. *)
+let rec lpo : sym Ord.cmp -> term Ord.cmp = fun ord t1 t2 ->
```
fblanqui: Get rid of "Ord."

src/completion.ml
```diff
@@ -0,0 +1,228 @@
...
+  let f, args = get_args t1 in
+  let f = get_symb f in
```
fblanqui: This let can be removed/unfold.

src/completion.ml
```diff
@@ -0,0 +1,228 @@
...
+                else -1
+            | 0            ->
+                if List.for_all (fun x -> lpo ord t1 x > 0) args' then
+                  Ord.ord_lex (lpo ord) args args'
+                else -1
+            | _            -> -1
+
+(** [to_rule (lhs, rhs)] tranlates the pair [(lhs, rhs)] of closed terms into
+    a rule. *)
+let to_rule : term * term -> sym * rule = fun (lhs, rhs) ->
```
fblanqui: rename into make_rule?

src/completion.ml
```diff
@@ -0,0 +1,228 @@
...
+  let (s, args) = get_args lhs in
+  let s = get_symb s in
```
fblanqui: remove this let

src/completion.ml
```diff
@@ -0,0 +1,228 @@
...
+      let vs = Bindlib.new_mvar te_mkfree [||] in
+      let rhs = Bindlib.unbox (Bindlib.bind_mvar vs (lift rhs)) in
+      s, { lhs = args ; rhs = rhs ; arity = List.length args ; vars = [||] }
+
+(** [find_deps eqs] returns a pair [(symbs, deps)], where [symbs] is a list of
+    (names of) new symbols and [deps] is a list of pairs of symbols. (s, t)
+    is added into [deps] iff there is a equation (l, r) in [eqs] such that
+    (l = s and t is a proper subterm of r) or (r = s and t is a proper subterm
+    of l). In this case, we also add s and t into [symbs]. *)
+let find_deps : (term * term) list -> string list * (string * string) list
```
fblanqui: This seems to be related to the computation of dependency pairs. Check with @GuillaumeGen whether you could factorize/share some code.

src/completion.ml
```diff
@@ -0,0 +1,228 @@
...
+                add_symb sym'.sym_name;
+            | _                           -> () in
+          let _, args = get_args t2 in
+          List.iter (Basics.iter check_root) args
+    | None     -> () in
+  List.iter
+    (fun (t1, t2) -> find_dep_aux (t1, t2); find_dep_aux (t2, t1)) eqs;
+  (!symbs, !deps)
+
+module StrMap = Map.Make(struct
```
fblanqui: This is already defined in basics.

src/completion.ml
```diff
@@ -0,0 +1,228 @@
...
+  type t = string
+  let compare = compare
+end)
+
+exception Not_DAG
```
fblanqui: Comment missing!

src/basics.ml
wujuihsuan2016: In the algorithm for checking SR, we only consider left-algebraic rule. However, the type inference works for any term, so I left the possibility of applying the algorithm to any rule with LHS containing no metavariable of arity > 0 but the algorithm might not work in this case.

src/completion.ml
```diff
@@ -0,0 +1,228 @@
...
+
+(** [topo_sort symbs edges] computes a topological sort on the directed graph
+    [(symbs, edges)]. *)
+let topo_sort : string list -> (string * string) list -> int StrMap.t
```
fblanqui: This should be defined in Extra.

fblanqui: More explanations required. What is the resulting map?

fblanqui: You should add more comments inside the code for explaining what you are doing.

src/completion.ml
fblanqui: Complexity of the algorithm?

src/completion.ml
fblanqui: Complexity of the algorithm?

fblanqui: You should say that this function may raise some exception. When? Why?

src/completion.ml
```diff
@@ -0,0 +1,228 @@
...
+          if s1 == s2 then 0
+          else begin
+            try
+              StrMap.find s2.sym_name ord - StrMap.find s1.sym_name ord
+            with _ -> Pervasives.compare s1.sym_name s2.sym_name
+          end
+      | true, false  -> 1
+      | false, true  -> -1
+      | false, false -> Pervasives.compare s1.sym_name s2.sym_name
+  with Not_DAG ->
```
fblanqui: This exception is raised by topo_sort only, right? If so, you should instead write:
```
let ord = try topo_sort ... with Pervasives.compare in fun s1 s2 -> ....
```

src/confluence.ml
```diff
@@ -0,0 +1,78 @@
...
+      if s1 == s2 then unif eqs else raise Unsolvable
+  | Vari _, _
+  | _, Vari _
+  | Prod _, _
+  | _, Prod _
+  | Abst _, _
+  | _, Abst _                     -> assert false
+  | Appl (t1, u1), Appl (t2, u2)  -> unif ((t1, t2) :: (u1, u2) :: eqs)
+  | Meta (m1, ts1), Meta (_, ts2) ->
+      assert (ts1 = [||] && ts2 = [||]);
```
fblanqui: replace by | Meta(m1,[||]), Meta(_,[||]) ->

fblanqui: Why do you have this special case? Why can't this be done by the following case?

src/confluence.ml
```diff
@@ -0,0 +1,78 @@
...
+      let vs = Bindlib.new_mvar mkfree [||] in
+      let bt2 = Bindlib.bind_mvar vs (lift t2) in
+      set_meta m1 (Bindlib.unbox bt2);
+      unif eqs
+  | (Meta (m1, ts1), t2)
+  | (t2, Meta (m1, ts1))          ->
+      if occurs m1 t2 then raise Unsolvable
+      else
+        assert (ts1 = [||]);
```
fblanqui: Should be in the pattern.

src/confluence.ml
```diff
@@ -0,0 +1,78 @@
...
+        let vs = Bindlib.new_mvar mkfree [||] in
+        let bt2 = Bindlib.bind_mvar vs (lift t2) in
+        set_meta m1 (Bindlib.unbox bt2);
```
fblanqui: Can't this be simplified/optimized @rlepigre ?

src/confluence.ml
```diff
@@ -0,0 +1,78 @@
...
+
+open Terms
+open Basics
+open Timed
+
+exception Unsolvable
+
+(** [unif_aux t1 t2 eqs] attempts to solve the first-order unification
+    problem [(t1 = t2) :: eqs]. *)
+let rec unif_aux : term -> term -> unif_constrs -> unit = fun t1 t2 eqs ->
```
fblanqui: Should be in basics.

src/confluence.ml
fblanqui: @rlepigre : In addition, as references are timed, and all these meta and meta substitutions are only used for checking subject reduction, shouldn't we through all these references away after subject reduction using save/restore to limit memory usage? Idem for TRef instantiations used in eq when calling the tactic rewrite.

src/handle.ml
```diff
@@ -125,15 +125,16 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
       in
       let rs = List.map handle_rule rs in
       (* Adding the rules all at once. *)
       let add_rule (s,h,r) =
         out 3 "(rule) %a\n" Print.pp_rule (s,h,r.elt);
         Sign.add_rule ss.signature s r.elt
       in
-      List.iter add_rule rs; (ss, None)
+      List.iter add_rule rs;
+      List.iter (Sr.check_rule ss.builtins) rs; (ss, None)
```
fblanqui: SR must be checked BEFORE adding the rules. So these modifications should be reverted.

src/handle.ml
```diff
@@ -208,6 +209,7 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
             wrn cmd.pos "Proof aborted."; ss
         | P_proof_admit ->
             (* If the proof is finished, display a warning. *)
+
```
fblanqui: remove this useless modification to minimize the diff

src/sr.ml
```diff
@@ -2,9 +2,11 @@
 
 open Timed
 open Console
-open Terms
 open Print
 open Extra
+open Terms
```
fblanqui: Minimize the diff

src/sr.ml
```diff
@@ -38,127 +40,114 @@ let subst_from_constrs : (term * term) list -> subst = fun cs ->
...
+  fatal r.pos "Unable to prove SR for rule [%a]." pp_rule (s, h, r.elt)
 
-let build_meta_type : int -> term = fun k ->
-  let m' = new_meta Type (*FIXME?*) k in
-  let m_typ = Meta(m',[||]) in
-  let m = new_meta m_typ k in
-  Meta(m,[||])
-*)
+let sr_not_preserved_msg (s, h, r) =
+  fatal r.pos "Rule [%a] does not preserve typing." pp_rule (s, h, r.elt)
```
fblanqui: does not -> may not

src/sr.ml
```diff
@@ -38,127 +40,114 @@ let subst_from_constrs : (term * term) list -> subst = fun cs ->
...
-  Bindlib.unbox (build_prod k tk)
+  List.exists (eq_comm eq) eqs
 
-(** [check_rule builtins r] check whether rule [r] is well-typed. The program
-    fails gracefully in case of error. *)
-let check_rule : sym StrMap.t -> sym * pp_hint * rule Pos.loc -> unit
+(** [check_rule symbs builtins r] checks whether rule [r] is well-typed. The
+    program fails gracefully in case of error. *)
+let check_rule :
+  sym StrMap.t -> sym * pp_hint * rule Pos.loc -> unit
```
fblanqui: Is the pp_hint argument useful?

src/unif.ml
```diff
@@ -237,7 +237,7 @@ and solve_aux : term -> term -> problems -> unif_constrs = fun t1 t2 p ->
    injective. Currently, it only handles a specific case: when s is the
    builtin P. *)
   let solve_inj s ts v =
-    if !(s.sym_rules) = [] then error ()
+    if !(s.sym_rules) = [] then add_to_unsolved ()
```
fblanqui: This modification is not directly related to SR. If you think that this is a bug, this should done in a separate oneline PR.

src/handle.ml
wujuihsuan2016: When we check SR for rule fl -> r, we should consider all the rules defining f. That is why the SR should be checked in this way.

src/confluence.ml
rlepigre: We could maybe avoid a lifting/unboxing here. Is the term `t2` guaranteed to be closed here?

For the second point, I don't think this would make much difference. The memory is collected for every undo that is not reachable from a reachable saved time, so I would not worry too much about that.

src/unif.ml
wujuihsuan2016: In the algorithm for checking SR, we introduce new function symbols (c_x, t_x) which are not like user-defined symbols. When we attempt to unify (c_x, t), where t is not headed by a symbol, we shouldn't raise an error like we did here. Maybe we should check if s is one of these special symbols but the code will become more complicated by doing so.

src/basics.ml
fblanqui: Ok, give a more informative name then.

src/handle.ml
fblanqui: I don't understand. Checking SR for a rule or set of rules should not depend on the rules themselves.

src/unif.ml
fblanqui: Why shouldn't we raise an error? How can you have c_x = t if c_x is undefined and t is not headed by a function symbol?

wujuihsuan2016: To check SR for the rule l -> r, we check Γ |- lθ : T => Γ |- rθ : T. Here c_x actually corresponds to the term xθ and t_x to the type of any occurrence of xθ in the derivation of Γ |- lθ : T as I understand from the algorithm.

fblanqui: Yes, but Γ |- rθ : T has to be done in a system where l->r has not been added yet. It would be strange to need l->r (in type conversions) to check that l->r preserves typing...

wujuihsuan2016: Yes, this has been corrected but I think you might have pasted this comment in the wrong place.

fblanqui: What do you mean? Your last commit is on May 3...

fblanqui:
make sanity_check fails. Thanks to fix it.

src/unif.ml
wujuihsuan2016: I think your last comment refers to the modification I made in handle.ml, but not this part of code...

fblanqui: Yes, there seems to be some confusion. So, what has been corrected?

fblanqui: There are two points to clarify:
1. in handle: add a rule in the signature AFTER successful SR check only
2. in unif: solve_inj should fail if the head symbol has no rule

wujuihsuan2016: https://github.com/Deducteam/lambdapi/pull/200#discussion_r290650646 This has been corrected but I haven't made the commit yet.

fblanqui: Ok, thanks.

fblanqui: It would be good to commit more often so that I know what is fixed or not, and can close some of the discussions.

wujuihsuan2016: For the point 2, see https://github.com/Deducteam/lambdapi/pull/200#discussion_r290701911 or consider the following example:
symbol eta : (nat => nat) => nat => nat
rule eta &g -> \x. &g[x]
In the algorithm for checking SR, we first replace the metavariable &g with a new function symbol c_g of type t_g  and then try to infer the type of modified LHS, i.e. eta c_g. The type inference of LHS will generate the constraint t_g ~ nat => nat, which corresponds to the constraint T(xθ) ~ nat => nat obtained from the derivation Γ |- lθ : T (T(xθ) is the type of any occurrence of xθ in this derivation).

src/extra.ml
wujuihsuan2016: I created the module for that since most of the code in extra.ml is organized in this fashion. It just seems a bit weird to include a separate function in a file written in this style.

src/basics.ml
wujuihsuan2016: Knowing that there are lots of intermediate functions written for the completion procedure, I am not sure if it is a good idea to use the array of symbols. Each of these functions should take the array as an argument in order to to this. Another idea is to write a local function for it instead of including it in basics.ml as it is only used in three functions.

src/unif.ml
fblanqui: This rule should be rejected at the scope level because g is of arity 0 in the LHS and 1 in the RHS. The arity should be constant.
Now, yes, when trying to type the LHS, you get the constraint t_g = nat=>nat. Then, it means that you should check that the RHS (say g) is of type nat=>nat using the rule t_g --> nat=>nat, but not the rule eta g --> g itself!

src/extra.ml
fblanqui: Right but, it seems to me, not for the same reasons. String, List and Arrays are extensions of already existing modules. Option is not but, as it provides functions with common names, it is not intented to be opened but used to qualify those common names.

src/unif.ml
wujuihsuan2016: I think there must be some misunderstanding. When we try to type LHS, we first try to unify t_g and nat => nat. This is the origin of the constraint t_g = nat => nat. The original version of the unification function raises a fatal error in this case. However, instead of raising an error, we should add this equation into the unsolved part. I was not reasoning about using the rule itself to check SR...

src/basics.ml
fblanqui: But I don't understand why you need to distinguish between old and new symbols in the first place... You need to clarify this first.

src/unif.ml
fblanqui: I see. But my idea was initially to proceed differently:
Given a rule l --> r,
1. in l, replace pattern variables by metavariables, giving l'
2. apply the standard type inference algorithm on l' (because unification acts on metavariables) giving a type T and a set of constraints E
3. build a map M from the (unsolved) metavariables of l' or their types to new symbols
4. complete M(E) into a convergent rewrite system R
5. try to check that M(r) has type M(T) modulo R

fblanqui:
This PR is probably too old now. It should be completely rewritten after #328.

PR #204, opened 2019-04-02
fix #197

### LINKED ISSUE #197: Grammar: allow ∀x y:N,...
That is, remove the need for parentheses.

fblanqui:


src/parser.ml
```diff
@@ -287,9 +287,13 @@ let parser term @(p : prio) =
   (* Products. *)
   | "∀" xs:arg+ "," b:(term PFunc)
       when p >= PFunc -> in_pos _loc (P_Prod(xs,b))
+  | "∀" xs:arg_ident+ ":" a:(term PFunc) "," b:(term PFunc)
+     when p >= PFunc -> in_pos _loc (P_Prod([xs,Some(a),false],b))
```
rlepigre: The indentation of the `when` is not correct, but otherwise I can't see any problem.

fblanqui:
Yes, your example is hard to read but the user can always add parentheses to make things more readable when this is necessary, which is the point.

src/parser.ml
fblanqui: Fixed in 6dfeeee506300c83218b2f1d245b99158a08be72

PR #207, opened 2019-04-03
make require/open commands take several files

fblanqui:


src/syntax.ml
```diff
@@ -296,10 +285,12 @@ let eq_p_config : p_config eq = fun c1 c2 ->
...
-  | (P_require(m1,r1)            , P_require(m2,r2)            ) ->
-      m1 = m2 && eq_require_mode r1 r2
-  | (P_open(m1)                  , P_open(m2)                  ) ->
-      m1 = m2
+  | (P_require(ps1)              , P_require(ps2)              )
+  | (P_require_open(ps1)         , P_require_open(ps2)         )
+  | (P_open(ps1)                 , P_open(ps2)                 ) ->
+     (try List.for_all2 (=) ps1 ps2 with Invalid_argument _ -> false)
+  | (P_require_as(p1,id1)  , P_require_as(p2,id2)  ) ->
+     p1 = p2 && id1.elt = id2.elt
```
rlepigre: Indentation is wrong.

src/syntax.ml
```diff
@@ -296,10 +285,12 @@ let eq_p_config : p_config eq = fun c1 c2 ->
...
 let eq_p_command : p_command eq = fun c1 c2 ->
   match (c1.elt, c2.elt) with
-  | (P_require(m1,r1)            , P_require(m2,r2)            ) ->
-      m1 = m2 && eq_require_mode r1 r2
-  | (P_open(m1)                  , P_open(m2)                  ) ->
-      m1 = m2
+  | (P_require(ps1)              , P_require(ps2)              )
+  | (P_require_open(ps1)         , P_require_open(ps2)         )
+  | (P_open(ps1)                 , P_open(ps2)                 ) ->
+     (try List.for_all2 (=) ps1 ps2 with Invalid_argument _ -> false)
```
rlepigre: This is very wrong! Polymorphic equality should not be used here. And equality should be modulo positions. Moreover, you can use `List.equal` (defined in `Extra`).

src/syntax.ml
```diff
@@ -153,21 +153,17 @@ type p_config =
...
-  | P_require_open
-  (** Require the module and open it. *)
-  | P_require_as of ident
-  (** Require the module and aliases it with the given indentifier. *)
-
 (** Parser-level representation of a single command. *)
 type p_statement = (ident * p_arg list * p_type) loc
 
 type p_command_aux =
-  | P_require    of module_path * require_mode
```
rlepigre: You should use a boolean to say whether the modules must be opened. That would save a constructor.

src/pretty.ml
```diff
@@ -156,14 +156,14 @@ let pp_p_tactic : p_tactic pp = fun oc t ->
 let pp_command : p_command pp = fun oc cmd ->
   let out fmt = Format.fprintf oc fmt in
   match cmd.elt with
-  | P_require(m, P_require_default) ->
```
rlepigre: You should output code as it was given by the used, so only one `require`. You can use `List.pp`. And the same comment holds for the other constructors (but see my comment about `require open` in `src/syntax.ml` first).

src/parser.ml
```diff
@@ -440,16 +440,18 @@ let require : (Files.module_path -> unit) Pervasives.ref = ref (fun _ -> ())
 
 (** [cmd] is a parser for a single command. *)
 let parser cmd =
-  | _require_ m:{_open_ -> P_require_open}?[P_require_default] p:path
```
rlepigre: Rather use something like `_require_ o:{_open_ -> true}[false] ps:path+` to combine `require` and `require open` as suggested in my comment on `src/syntax.ml`.

src/handle.ml
```diff
@@ -39,6 +39,51 @@ type proof_data =
   ; pdata_finalize : sig_state -> Proof.t -> sig_state (* Finalizer. *)
   ; pdata_term_pos : Pos.popt (* Position of the proof's terminator. *) }
 
+(** [handle_require pos ss p] handles the command [require p] with [ss] as the
+   signature state. On success, an updated signature state is returned. *)
+let handle_require : popt -> sig_state -> Path.t -> sig_state =
```
rlepigre: To be combined with `handle_require_open` with an extra boolean argument.

src/syntax.ml
fblanqui: What do you mean modulo positions? What should I put instead of (=)? Because ps1 and ps2 are of type [Path.t list].

rlepigre: I was convinced that there were positions involved. If not, then there should be since we want errors to be reported for a single module, not the whole list. The argument of `P_require` should thus be of type `module_path loc`, and the equality should be `(fun mp1 mp2 -> mp1.elt = mp2.elt)`.

rlepigre: Alternatively, we can just give the position of the whole statement, and print the module name for which there is a problem. This is probably simpler as a first step, since there is nothing to change. And that is probably good enough.

fblanqui:
I fixed the problems you mentioned. The only thing to improve perhaps is to make P_require to take an argument of type Path.t loc list instead of Path.t list. But I would prefer that you do it yourself if you don't mind after merging this PR.

src/pretty.ml
```diff
@@ -156,10 +156,10 @@ let pp_p_tactic : p_tactic pp = fun oc t ->
 let pp_command : p_command pp = fun oc cmd ->
   let out fmt = Format.fprintf oc fmt in
   match cmd.elt with
-  | P_require(ps)                   ->
```
rlepigre: You could also do the following, which is shorter.
```
  | P_require(o,ps)                  ->
    out "require%s %a" (if o then " open" else "") (List.pp pp_path " ") ps
```

rlepigre:
I'll have a look when I have a moment for the position thing. But it may well be good enough as it is right now, since error reporting will contain the name of the module that failed to be loaded / could not be found.

PR #216, opened 2019-04-19
XTC update

GuillaumeGen:
After @wujuihsuan2016's suggestion, replace the underscore by a dot in the output of constants to have an injective export.
Also remove the `c_` before constant name, since it is useless to have both this prefix and a `v_` before variable name.


fblanqui:
This PR also changes the translation of application but this is not documented. There is no <funapp> and <arg> generated anymore. Is it correct to do this?

fblanqui:
@GuillaumeGen : any answer to https://github.com/Deducteam/lambdapi/pull/216#issuecomment-495220387 ?

GuillaumeGen:
> This PR also changes the translation of application

Indeed, with dependent types, functions do not have arity anymore, hence it seems more reasonable to me not to use <funapp>, since in Wanda (such a restriction is not documented in the definition of the XTC format) it is enforced that <funapp>f<arg>...</arg></funapp> always has the same number of arguments.

> There is no and generated anymore

I don't understand this sentence. If "and" should be substituted by "funapp", the justification is above.

src/xtc.ml
```diff
@@ -7,11 +7,11 @@ open Extra
 open Timed
 open Terms
 
-(** [print_sym oc s] outputs the fully qualified name of [s] to [oc]. The name
-    is prefixed by ["c_"], and modules are separated with ["_"], not ["."]. *)
+(** [print_sym oc s] outputs the fully qualified name of [s] to [oc]. Modules
+    are separated with ["."]. *)
```
fblanqui: 'Modules are separated with ["."]': can be removed

fblanqui:
I cannot remember ;-) Ignore

fblanqui:
The parameters i and s in printing functions are not documented. What are they for?

fblanqui:
Why variable names need to be prefixed by v_?

rlepigre:
@GuillaumeGen ping.

PR #223, opened 2019-06-19
[lsp] Fix bad handling of errors in Pure.

ejgallego:
This case does happen when errors on require.

Not a fan of the assert false, would it be possible to ensure things
work using types?

### DIFF
--- a/src/pure/pure.ml
+++ b/src/pure/pure.ml
@@ -29,6 +29,14 @@ type state = Time.t * Scope.sig_state
 (** Exception raised by [parse_text] on error. *)
 exception Parse_error of Pos.pos * string
 
+let dummy_pos : Pos.pos = Lazy.from_val Pos.(
+  { fname      = None
+  ; start_line = 1
+  ; start_col  = 0
+  ; end_line   = 1
+  ; end_col    = 80
+  })
+
 let parse_text : state -> string -> string -> Command.t list * state =
     fun (t,st) fname s ->
   let old_syntax = Filename.check_suffix fname Files.legacy_src_extension in
@@ -42,7 +50,8 @@ let parse_text : state -> string -> string -> Command.t list * state =
   with
   | Fatal(Some(Some(pos)), msg) -> raise (Parse_error(pos, msg))
   | Fatal(Some(None)     , _  ) -> assert false (* Should not produce. *)
-  | Fatal(None           , _  ) -> assert false (* Should not produce. *)
+  | Fatal(None           , msg) -> raise (Parse_error(dummy_pos, msg))
+
 
 type proof_finalizer = Scope.sig_state -> Proof.t -> Scope.sig_state
 type proof_state = Time.t * Scope.sig_state * Proof.t * proof_finalizer


rlepigre:
I fixed the problem at its source with dc3df6619a5b0998b685b0c197847d3de0cac190, which takes care of several problems at once. Now the `assert false` really is unreachable! And no, I don't think it would be easy to ensure things with types. This would only move a bug around: you might not have a crash, but bad error reporting wouldn't be avoided. Anyway, that should not be a problem anymore.

Can I close?

ejgallego:
Thanks a lot @rlepigre !

I would like to say the problem is not gonna happen again but this is like the 3rd time it comes back and we lose a lot of time each time we debug it; so indeed I would prefer not to have any assert false.

What's the point of it anyways? If these cases are unreachable then they should not have a constructor.

rlepigre:
It's arguably not perfect, but the whole error-reporting mechanism has proved to work quite well until now. There is no simple way to remove constructors here since they are all useful at some level:
 - the `None` case is used for errors that do not have any relevant source-code position (e.g., a file not existing),
 - the `Some None` case is used for reporting an error on a term that was generated (and has thus no source code position),
 - the `Some (Some p)` case is the ideal one.

Of course, we may decide to attach the position of the current command to an error as a last resort. The parsing function behaves in this way, although I initially forgot to repack the error reports coming from the compilation of dependencies.

One thing that would really help for avoiding such errors is to have a test suite for the LSP server. Ideally, `make test` would ensure that things don't change in a bad way over time. We could maybe have sequences of queries with expected answers for typical runs, with decent coverage. What do you think?

ejgallego:
I think the test suite is a good idea, but complementary to the problem here. Indeed I do agree that having `fatal_no_pos` et al is good, the problem I see is why they are then handled as `assert false` ?

If there is no location just report no location to the UI, not an assert false.

And indeed it seems there are many uses of `fatal_no_pos` in the codebase, so I am positive that at some point the invariant will be broken again; maybe you can maintain it but that's not going to scale once you have more people contributing to the codebase.

Likely a good way to handle this would be to repack in the top-level command handler, introducing a different exception, I can give it a go, tho preserving stack traces requires some care.

rlepigre:
I'm not sure that it would make sense to repack exceptions here (meaning in the main parsing functions), because the issue only comes up with the `Pure` API. Error handling is otherwise completely uniform, and I would prefer not to introduce any extra complexity.

From my point of view, having `assert false` in these two cases makes sense because we definitely want to maintain the invariant that errors coming from the parser have a position. Another solution would be, as you proposed, to have some dummy position when that is not the case. However I find it not very satisfactory because we probably would not notice if the invariant is broken.

I got a bit carried away while writing this, and I made a few modifications to make it easier to enforce the invariant: c01b734c6ceb94d7b83dc06540cb33c062ab5ca5. It was actually violated in one place in the legacy parser, but now I'm very confident that every error reported by either parsers have attached positions. I also made it harder for anyone to mistakenly raise the `Fatal` exception with no positions in the parser files.

ejgallego:
Thanks @rlepigre , I'll have a look ASAP!

Indeed I guess I was a bit unlucky but the 3 last times I went back to lambdapi [as you know I work intermittently on the LSP side] I got somehow bitten by this, so IMHO if we can make the invariants actually hold by typing that would be great.

Having "hidden" invariants may difficult working on the implementation as even ourselves may forget about them.

I'd like to try to give the code a go, then I'll close / update this PR as appropriate.

ejgallego:
Ok, after review I think I'm OK with the current solution, tho IMVHO it is not very satisfying; I think the re-raise as done in `handle.ml` will for example lose backtraces, and of course having `assert false` defeats the goal of typing in many cases.

I have found in practice that catching exceptions to re-raise tends to produce a bit of mess in the error handling if the code grows big.

I guess one solution would be to refine the exceptions, make them private, and provide some combinators. However this is quite a lot of work for little gain [moreover LP doesn't ship .mli files] so let's close this for now.

PR #228, opened 2019-07-20
Fixes #225, but loses performances.

### LINKED ISSUE #225: Bug in higher-order matching
I am sorry, I don't have time to minimize the example. I am aware of a type checking issue in Lambdapi which is probably related to Higher-Order matching. The error can be found in the Librairies repository. There are two files in play. The first one encodes Cedille's core theory: [cedille.dk](https://github.com/Deducteam/Libraries/blob/master/theories/cedille.dk) . The second one encodes a basic example with it: [nat.dk](https://github.com/Deducteam/Libraries/blob/master/examples/cedille/nat.dk).

The first file type checks but not the second in Lambdapi (but it does in Dedukti and Dedukti is right on this one). I know that the error commes from either [this rule](https://github.com/Deducteam/Libraries/blob/024cb976dbb55f05e69a5e9c5ccca95e1ecee15d/theories/cedille.dk#L93) or [this rule](https://github.com/Deducteam/Libraries/blob/024cb976dbb55f05e69a5e9c5ccca95e1ecee15d/theories/cedille.dk#L99) which are both Higher-Order rules. One of these rules is not applied while it should.

rlepigre:
This fixes several issues with higher-order rewriting, that previously weren't handled completely correctly. However, this induces a quite big loss of performance. I could try and fix that, but I don't want to waste my time on code that is going to be replaced by decision tree stuff. @fblanqui I let you decide whether you want to merge this or not.

On of the issues that this also fixes is related to the case where the pattern variable has no argument. What I used to do here was to construct a trivial multiple binder (with no arguments), and I did not check any variable constraints since I thought that every term should be closed at this point for some reason. Since that is not the case, we need to box the term even in that case to know if the variables that should not be captured (those coming form lambdas in the patter) appear in the matched term.

### DIFF
--- a/src/eval.ml
+++ b/src/eval.ml
@@ -133,40 +133,39 @@ and find_rule : sym -> stack -> (term * stack) option = fun s stk ->
     let rec match_args ps ts =
       match (ps, ts) with
       | ([]   , _    ) -> Some(Bindlib.msubst r.rhs ar, ts)
-      | (p::ps, t::ts) -> if matching ar p t then match_args ps ts else None
+      | (p::ps, t::ts) -> if matching [] ar p t then match_args ps ts
+                          else None
       | (_    , _    ) -> assert false (* cannot happen *)
     in
     match_args r.lhs stk
   in
   List.map_find match_rule Timed.(!(s.sym_rules))
 
-(** [matching ar p t] checks that term [t] matches pattern [p]. The values for
-    pattern variables (using the [ITag] node) are stored in [ar], at the index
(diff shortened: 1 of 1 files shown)


rlepigre:
Actually, my fix does not work for the same reason that I thought @gabrielhdt's would not work: free variables can appear during reduction. I thought I was wrong since you seemed to say that @gabrielhdt's branch fixed the problem, but I don't see how it possibly could without changing changing more things.

PR #229, opened 2019-07-21
Fixes #225, but loses performances.

### LINKED ISSUE #225: Bug in higher-order matching

rlepigre:
The semantics seems to be as expected in this branch, but efficiency is still a problem.

### DIFF
--- a/src/basics.ml
+++ b/src/basics.ml
@@ -99,6 +99,25 @@ let eq : term -> term -> bool = fun a b -> a == b ||
   in
   try eq [(a,b)]; true with Not_equal -> false
 
+(** [is_closed t] tells whether the term [t], which should not contain pattern
+    variables or term environments, is closed. Note that this function has NOT
+    a constant time complexity: it requires traversing every branch of a term,
+    until a binder is found. *)
+let rec is_closed : term -> bool = fun a ->
+  match unfold a with
+  | Patt(_,_,_)
+  | TEnv(_,_) -> assert false
+  | Prod(a,b)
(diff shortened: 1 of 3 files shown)


francoisthire:
I don't understand the loss of performances. Currently, we not really need HO rules for the librairies we have in Dedukti/Lambdapi : Matita/HOL/Verine/Focalize. Only first-order is used (ok, the theory file in matita declares an HO rule, but it is not needed in practice).

rlepigre:
The problem is that everything goes through the higher-order mechanism, unlike in Dedukti. To regain efficiency we could have a special handling of symbols that do not have any higher-order in their rewriting rules.

Issue #169, opened 2019-02-21
Handling implicit arguments or _ in RHS?


fblanqui:
In M_LHS mode, we could replace them by a fresh pattern variable like if it were a P_Wild. The problem is with the M_RHS mode. What to do in this case?... It is like if we allow _ in a RHS. This should generate a fresh meta and we would need to check that SR instantiates all these generated metas. For the moment, we could just fail when they are implicit arguments in RHS, and leave this for future work.

GuillaumeGen:
In #222 , I discovered that non-annotated lambdas are allowed on the rhs of a rule declared by `:=` (a definition). Is there any reason justifying such a heterogeneous behaviour between two ways to declare rules?

rlepigre:
Yes, immediate definitions are handled differently since the definition is a closed term. In particular, there is no subject reduction concern here.

fblanqui:
Well, there is some type-checking if a type is given. So,this is like checking SR, but it is much simpler.

fblanqui:
@francoisthire , this issue is fixed now.

francoisthire:
That's great, thank you.

PR #232, opened 2019-07-29
fix issue #169 (allow metas in RHS)

fblanqui:


rlepigre:
Note that I just rebased your branch on `master` so that it can be merged. You need to sync it with your local branch before you can do anything.

Issue #226, opened 2019-07-19
Option --beautify does not sanatize names


francoisthire:
Here an example where the --beautify output cannot be parsed.

```lambdapi
(; CoC encoding ;)
(; The Higher-order product is missing because it is not useful (yet!) ;)


Sort : Type.

star : Sort. (; ⋆ ;)
box  : Sort. (; □ ;)

(; Compute R(s1,s2) since we are in a functional PTS ;)
def rule : Sort -> Sort -> Sort.
[s] rule _ s --> s.
```

The output is 

```lambdapi
symbol const Sort : TYPE

symbol const star : Sort

symbol const box : Sort

symbol rule : Sort ⇒ Sort ⇒ Sort

rule rule _ &s → &s
```

rlepigre:
Yeah, we should probably try to escape keywords.

rlepigre:
Note that ddeea4770795c38751f122bb7433cd6fa6a23308 introduce the following behavior: if an invalid Lambdapi identifier is encountered with `--beautify` on the legacy syntax then the program fails. To translate, first fix up your source file with a couple of search and replace.

PR #237, opened 2019-08-02
fix issue #226

fblanqui:


francoisthire:
Don't we need just to sanitize names when there is a conflict? Sanitization may need to keep a list of identifiers already saw though.

rlepigre:
> Don't we need just to sanitize names when there is a conflict? Sanitization may need to keep a list of identifiers already saw though.

I'd prefer not to have to do that. Seriously, we don't translate that many things from the old syntax to the new one anyway. On thing we could do though, is fail on printing if we find a keyword. That way you know that you need to rename stuff in the original file before translating, and the target of the translation is always correct.

fblanqui:
About {|...|} in DK: this is another example showing that this is really annoying to keep adding features to DK!!!

fblanqui:
Ok for failing. After all, this is not the job of LP to take care/rename names!

fblanqui:
So, we could close this PR and the corresponding issue (won't fix)?

rlepigre:
Yes, I guess so. But what you did in this PR is a good start. I'll add a couple of comments with things for you to change.

rlepigre:
Actually let's close that. I'll take care of the change it is going to be faster.

fblanqui:
You can push to my branch now if you want.

rlepigre:
I'll push on master when I'm done, I just realized that there is a little problem with module paths.

Issue #211, opened 2019-04-03
Add a --version option


fblanqui:
To know which version is currently installed.

rlepigre:
That's a good point.

PR #238, opened 2019-08-02
fix issue #211

fblanqui:


rlepigre:
I can't push to your branch, so I'll just merge and push my change to master.

rlepigre:
With my changes, if you pin the repository then `--version` has the expected behavior. In case of development version it gives the commit hash, and in release mode it gives the version number. Note however that the version number remains weird if you run `dune exec -- lambdapi --version` since the `%%VERSION%%` tag is not substituted at this point.

PR #251, opened 2019-10-18
Test also OCaml 4.08 and 4.09

XVilka:
Lets see if it breaks something...


rlepigre:
It will: `earley` is not yet available for these versions.

XVilka:
Right, forgot about that.

rlepigre:
Subsumed by #314.

PR #252, opened 2019-10-21
Conversion modulo eta

01mf02:
To quote Makarius Wenzel:

> Isabelle proof terms are always understood modulo beta-eta.

This commit adds an option to Lambdapi allowing typechecking modulo eta, thus allowing checking of Isabelle proof terms.

### DIFF
--- a/src/eval.ml
+++ b/src/eval.ml
@@ -134,7 +134,7 @@ and eq_modulo : term -> term -> bool = fun a b ->
     | (Abst(_ ,b ), t          )
     | (t          , Abst(_ ,b )) when !eta_equality ->
         let (x,b) = Bindlib.unbind b in
-        eq_modulo ((b,(Appl(t, Vari(x))))::l)
+        eq_modulo ((b,Appl(t, Vari(x)))::l)
     | (Appl(t1,u1), Appl(t2,u2)) -> eq_modulo ((u1,u2)::(t1,t2)::l)
     | (Meta(m1,a1), Meta(m2,a2)) when m1 == m2 ->
         eq_modulo (if a1 == a2 then l else List.add_array2 a1 a2 l)
--- a/src/eval.ml
+++ b/src/eval.ml
@@ -33,7 +33,7 @@ let log_conv = new_logger 'c' "conv" "conversion"
 let log_conv = log_conv.logger
 
 (** Convert modulo eta. *)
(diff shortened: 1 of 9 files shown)


rlepigre:
Looks great to me. However it would be nice to fix the test since it is currently not run with the new option. I propose adding a folder `tests/OK+eta-conv` in which all files should expect the `--eta-conv` option. If you do that, then you'll need to modify `GNUmakefile` accordingly, following what is already done.

src/eval.ml
```diff
@@ -128,6 +131,10 @@ and eq_modulo : term -> term -> bool = fun a b ->
     | (Abst(a1,b1), Abst(a2,b2)) ->
         let (_,b1,b2) = Bindlib.unbind2 b1 b2 in
         eq_modulo ((a1,a2)::(b1,b2)::l)
+    | (Abst(_ ,b ), t          )
+    | (t          , Abst(_ ,b )) when Pervasives.(! eta_conv) ->
+        let (x,b) = Bindlib.unbind b in
+        eq_modulo ((b,(Appl(t, Vari(x))))::l)
```
fblanqui: The number of parentheses can be minimized.

fblanqui:
Instead of using command line options, wouldn't it better to add some command in a file to tell LP to use eta or not? Like `set flag "eta" on/off`.

rlepigre:
> Instead of using command line options, wouldn't it better to add some command in a file to tell LP to use eta or not? Like `set flag "eta" on/off`.

I'm not sure this is a good idea, because this would allow mixing code that relies on eta and code that does not. But we can of course do that. Since this is still a quite experimental feature, I would vote for keeping things as they are for now.

fblanqui:
What's the problem with mixing code using eta and code not using eta? Note that eta could be seen as a rewriting rule if we allowed rules headed by abstractions: `rule λx,M x → M`.

01mf02:
Following Frédéric's suggestion, I made equality modulo eta a flag instead of a command-line option. This follows the intuition that this option should always be enabled for files that require it, and it solves the issue of the test failing.
I named the flag `eta_equality`, following [the equivalent flag in Agda](https://github.com/agda/agda/issues/2436).

01mf02:
Sorry, forgot to integrate Frédéric's suggestion to reduce the number of parentheses.

rlepigre:
No problem, thanks @01mf02!

Issue #260, opened 2019-10-23
Print infix symbols in an infix way


fblanqui:
Currently, the printing of a symbol depends on its pp_hint. Rewriting and tactics build symbols with pp_hint = Nothing. Hence, in proofs, infix symbols are often not printed in an infix way, or in a mixed way (some occurrences are infix and some others are infix) which is worse. In case of pp_hint = Nothing (or a new pp_hint for generated symbols), we should check whether the symbol is infix or not. This could also be done at the creation time of the symbol, by redefining the function Terms.symb.

fblanqui:
I am wondering: is this pp_hint really a good idea? It seems to me that we could get rid of it and have the printing function depend on the current signature (open modules, aliases, builtins and infix symbols).

rlepigre:
The idea is to print things as they were parsed. The notion of open module, aliases and all does not exist in the internals. This is all handled at scoping time. And it is the same story for implicit arguments. And anyway, through reduction the scope of a term can change so I really don't see how to do things better.

However I know how to print infix symbols correctly, with the least amount of parentheses. The main thing is precisely to extend the printing hints so that they contain the priority and associativity parameters in the infix case. Then we all all the informations we need in the terms, and we only need to do a better work in `src/print.ml`.

fblanqui:
I don't understand what you mean by "the scope of a term can change".

Improving the the printing of infix symbols with less parentheses is the topic of #210 .

The point of this issue is that it happens that, in a goal with several occurrences of an infix, some occurrences of the infix are printed infix, while others are not. This is not consistent.

Instead of relying on inconsistent pp_hint's, I would prefer to print terms in a uniform way by adding in the signature what are the open modules, aliases and infix, and use this information for printing symbols and terms.

rlepigre:
Imagine you have a module `m1` containing
```
open lib.nat
definition add x y = x + y
```
And then you have a module `m2` with
```
eval add lib.nat.zero lib.nat.zero
```
How should you print the result of the evaluation?
Currently we would get `lib.nat.zero + lib.nat.zero` which is not great since `lib.nat` is not open, so you should not get the infix symbol I think.

fblanqui:
This is a matter related to require. Require could import symbols and rules only, and not notations. Only open would import notations as well.

rlepigre:
> This is a matter related to require. Require could import symbols and rules only, and not notations. Only open would import notations as well.

Well, that's exactly the current behavior.

[...]

Issue #243, opened 2019-10-10
expected relative path


ckeller:
Hi

I would like to call lambdapi on a file which is not in a subdirectory of my working directory, say `/tmp/foo.dk`. However, lambdapi fails with the following error:
```shell
$ lambdapi /tmp/foo.dk 
[unknown location] Invalid path for [/tmp/foo.dk] (expected relative path).
```

I am using the opam package but a quick look at the master branch let me think that it would produce the same answer.

Thank you for your help.

fblanqui:
Hello Chantal. This is on purpose. For the moment, LP does not accept file names with absolute paths. Why not running LP in /tmp, or moving your file foo.dk somewhere else?

ckeller:
I am calling lambdapi from another tool so
- I do not control the working directory from which lambdapi is called
- I do not want to pollute user's space
Why such a restriction?

fblanqui:
Workaround: instead of calling LP directly, use a script that will run LP on the filename part only from the dirname part.
Why such a restriction? @rlepigre shall give you some argument soon hopefully.

rlepigre:
The justification for this behavior is that this enforces a correspondence between file path and module path. This simplifies many things, but there are also limitations as you just discovered @ckeller. Of course, this is not satisfactory, and we need some mechanism that:
1) is more permissive,
2) retains some of the good properties we have now for the dependency analysis.

There are several options here:
1) Having a `PATH`-like mechanism (a list of directories to explore in order when looking for a module).
2) Introducing mappings between arbitrary directories and module path prefixes (that is what Coq does to some extend I believe, but it is not always great).

Could you please elaborate a bit on your use case @ckeller:
1) Are your Lambdapi files completely self-contained? If that is the case, then there is an easy workaround.
2) Would a `PATH`-like mechanism suffice for your application?

francoisthire:
What are the issues with Coq system? I have the impression that it is
always better to make the distinction between the file system and the
logical path.

Le jeu. 10 oct. 2019 à 22:44, Rodolphe Lepigre <notifications@github.com> a
écrit :

ckeller:
Hi @rlepigre 
Thank you for the explanation.

>     1. Are your Lambdapi files completely self-contained? If that is the case, then there is an easy workaround.

No, the objective is to have a prelude in some directory and some other files in other directories. We can as a first step be ok with a self-contained file but not on the longer term. What would the workaround be?

[...]

PR #262, opened 2019-10-28
[WIP] Introducing privacy

gabrielhdt:
Mirroring [privacy in Dedukti](https://github.com/Deducteam/Dedukti/pull/141), although the semantics are not really clear.  The main idea is to *restrict* the use of a private symbol out of its module.  A *protected* symbol may not appear on the right hand side of a rewrite rule.  This last constraint might be relaxed into: a private symbol may not appear on the right hand side of a rewrite rule *defined on a public symbol*.

This PR is motivated by the encoding of proof irrelevance in the case of PVS (and predicate subtyping).

We additionally provide *privacy* which is stronger than protection: a private symbol cannot be used outside of the module where it is defined; a private symbol is not in the exported signature. Privacy is motivated by https://github.com/Deducteam/Dedukti/issues/177.

Note that the terms used in Dedukti 2.7 were _local_ instead of _private_ and _private_ instead_ of _protected_.


src/parser.ml
```diff
@@ -160,6 +160,7 @@ let _assertnot_  = KW.create "assertnot"
 let _const_      = KW.create "const"
 let _inj_        = KW.create "injective"
 let _TYPE_       = KW.create "TYPE"
+let _prv_        = KW.create "private"
```
fblanqui: _private_

src/scope.ml
```diff
@@ -398,6 +398,18 @@ let scope_rule : sig_state -> p_rule -> sym * pp_hint * rule loc = fun ss r ->
   (* We scope the LHS and add indexes in the environment for metavariables. *)
   let lhs = Bindlib.unbox (scope (M_LHS(map)) ss Env.empty p_lhs) in
   let (sym, hint, lhs) =
+    let rec privacy t =
+      (* Ensure that there are no foreign private symbols used. *)
```
fblanqui: How could this happen if private symbols are not exported?

src/syntax.ml
```diff
@@ -28,6 +28,7 @@ type unop = string * priority * qident
 (** Representation of a binary operator. *)
 type binop = string * assoc * priority * qident
 
+
```
fblanqui: spurious newline

src/terms.ml
```diff
@@ -94,6 +94,8 @@ type term =
   (** The symbol may have a definition or rewriting rules (but NOT both). *)
   | Injec
   (** Same as [Defin], but the symbol is considered to be injective. *)
+  | Priva
```
fblanqui: There is no uniformity in the names of constructors. In one file, this is const, inj or prv. In another one, this is Injec or Priva. We should uniformize all this. Why not simply write Injective and Private everywhere?

src/unif.ml
```diff
@@ -314,6 +314,7 @@ and solve_aux : term -> term -> problems -> unif_constrs = fun t1 t2 p ->
           if !(s1.sym_rules) <> [] || List.same_length ts1 ts2
           then add_to_unsolved ()
           else error ()
+       | Priva -> failwith "implement me"
```
fblanqui: There should be no Private sym_mode.

tests/OK/prv.lp
```diff
@@ -0,0 +1,3 @@
+symbol const A : TYPE
+
+symbol private f : A ⇒ A
```
fblanqui: I don't like this. private is not the same kind of property as const or injective. It's about scoping. We should write: `private symbol f:A ⇒ A`. The semantics is: f is not exported.

fblanqui:
I think that what you should do is to add at https://github.com/Deducteam/lambdapi/blob/30e497d017a7a80650f765a1751059431874e3ca/src/compile.ml#L65 some code to expurge the signature of private symbols so that there is no way to refer to those symbols later.

francoisthire:
A private symbol should be exported as long as the rules which defines it.
However it should not be appear except if it's definition is extended.
Private symbols play a major role during conversion.

fblanqui:
@francoisthire I don't understand your first (unfinished) sentence. The second one neither.

fblanqui:
@gabrielhdt where in the code decision trees are recomputed at loading time?

fblanqui:
@francoisthire it may be possible in LP to not export private symbols while still keeping their defining rules in memory.

francoisthire:
1. Rules of a private symbol should always be exported.
2. A rewrite rule can be added on a private symbol in another file. In that
case, this private symbol is allowed to appear on the right hand side of
the rule.

fblanqui:
Do you have an example of 2?

gabrielhdt:
@francoisthire is **not** allowed to appear on the right hand side?

francoisthire:
Ask to Gaspard. But it is mainly about subtyping with inductive types.

fblanqui:
@Gaspi ?

gabrielhdt:
@fblanqui the dtrees are generated line 129 of sign.ml:

https://github.com/gabrielhdt/lambdapi/blob/fb394ba7476db82c69a6af44e0d8942731c5716b/src/sign.ml#L129

Gaspi:
1. Private symbols were initially a way to define conversion using auxiliary symbols and "smart(er) public constructors" (for instance forgetting their argument):
```
[s,s',a] lift s s' _ a --> lift' s s' a.
[s,a] lift' s s a --> a.
```
where `lift'` is private and should not be "used" to write proofs.

So I think, at the time, the expected behavior was:
- no occurrence of private symbols from external module
- no restriction in the module they are defined
- rules using private symbols are exported and used everywhere like any other rule, their main point is precisely to extend the conversion (while hiding the details to the user)

2. We then realized then that we needed to use `lift'` in the destructor of inductives:
```
match_prod _ (x => y => lift' s _ (P x y)) --> match_prod s (x => y => P x y).
```
where private symbol `lift'` occurs on the LHS but (luckily) no private symbol occur on the RHS.
We cannot use the public version `lift` here because it would never match since it reduces to the private constructor.

It seems that this issue arises naturally, hence it would probably be a good idea to:
- allow private symbols in LHS patterns
- forbid private symbols on the RHS of rules, including in `def c := ...` definition, in type annotations and (especially) in proofs
This seems safe since these rules referencing private symbols would never *introduce* them, only match on them.

3. Their could be slightly different needs in the future so it would probably be best to allow several level of privacy: strictly no occurrence, occurrence allowed on RHS under conditions, untyped private symbols (purely computational objects), etc

In particular, the rules needed for template polymorphic inductive subtyping now look like this:
```
[i,j,a,b] pair i j a b --> pair' (code i a) (code j b).
```
where private symbol `code` occurs on the right but is "guarded" by a new locally private symbol `pair'`.

It is not clear to me whether the definition of a new inductive type
- is part of the theory and can be assimilated with the "private" part of the encoding
- is part of the proof development in which case it should probably avoid using private symbols

src/terms.ml
gabrielhdt: Should I modify it in this PR?

tests/OK/prv.lp
gabrielhdt: Regarding the syntax: this was the easiest way; but I agree it is arguable.

src/unif.ml
fblanqui: Yes

tests/OK/prv.lp
gabrielhdt: And what about using a `set` command? Like `set f private`?

src/unif.ml
gabrielhdt: I assume the 'yes' concerns the remark above...

fblanqui: Yes :-)

fblanqui:
Thanks @Gaspi for your explanations and examples! In your first example, why don't you write `[s,s',a] lift s s' _ a --> a` directly?

tests/OK/prv.lp
rlepigre: I think it is best to keep it as it is now. It might make sense to combine several modifiers and the syntax is more uniform that way.

francoisthire:
Because the rule is also needed for lift' because after a computation we
can have an identity lift'.

tests/OK/prv.lp
fblanqui: I do not agree. Private could be applied to other things than just symbols.

fblanqui:
Sorry, I didn't see that in the second rule the two arguments were equal.

fblanqui:
Then, I suggest to replace https://github.com/Deducteam/lambdapi/blob/cfee85433a7207c2716c9b8b6dfe45744a611668/src/sign.ml#L20 by ```{ sign_symbols  : (sym * Pos.popt * bool) StrMap.t ref```.

tests/OK/prv.lp
gabrielhdt: Like private rules?

fblanqui: Why not. I was thinking in private tactics, notations, etc.

src/scope.ml
```diff
@@ -415,6 +415,19 @@ let scope_rule : sig_state -> p_rule -> sym * pp_hint * rule loc = fun ss r ->
...
     Bindlib.unbox (Bindlib.bind_mvar vars t)
   in
+  let _, t = Bindlib.unmbind rhs in
+  let rec privacy t =
+    (* Ensure that there are no foreign private symbols used. *)
+    let h, args = Basics.get_args t in
+    begin match unfold h with
+      | Symb({sym_path; sym_mode=Priva; _},_) ->
+        if sym_path <> ss.signature.sign_path
+        then fatal p_lhs.pos "Foreign private symbol not allowed in RHS."
```
fblanqui: It's not only in RHS that external private symbols are forbidden: it's every where except in the arguments of a rule left-hand side.

fblanqui: The function that should be modified is find_sym, which should take an additional boolean argument saying that external private symbols are allowed or not. By default, this argument is false except when scoping the arguments of a rule left-hand side.

src/scope.ml
```diff
@@ -42,39 +42,44 @@ let open_sign : sig_state -> Sign.t -> sig_state = fun ss sign ->
...
+            try PathMap.find mp Timed.(!Sign.loaded)
+            with Not_found -> assert false (* Should not happen. *)
+          in
+          (* Look for the symbol. *)
+          try (Sign.find sign s, Qualified) with Not_found ->
+          fatal pos "Unbound symbol [%a.%s]." Files.pp_path mp s
+        end
+  in
+  if s.sym_expo = Local && s.sym_path <> st.signature.sign_path
+  then fatal pos "Foreign local symbol used";
```
fblanqui: This case cannot happen as compile filters local symbols out.

src/scope.ml
```diff
@@ -415,6 +420,19 @@ let scope_rule : sig_state -> p_rule -> sym * pp_hint * rule loc = fun ss r ->
...
+    let h, args = Basics.get_args t in
+    begin match unfold h with
+      | Symb({sym_path; sym_expo=Private; _},_) ->
+        if sym_path <> ss.signature.sign_path
+        then fatal p_lhs.pos "Foreign private symbol not allowed in RHS."
+      | _                                       -> ()
+    end;
+    List.iter privacy args
+  in
+  privacy t; (* Checking for private symbols in rhs. *)
```
fblanqui: So, here, you make 2 traversals of the rhs. Moreover, you only check for private symbols in RHS. But my understanding is that private symbols should be checked every where except in LHS... Hence, my proposal to add some additional boolean argument to find_sym. In this case there is no need for a second traversal with this privacy function.

gabrielhdt: Yes, it is not complete yet, I'll add an optional argument to find_sym which defaults to false

src/scope.ml
gabrielhdt: Compile only concerns signatures written to disk, doesn't it?  We need this line to satisfy the test `tests/KO/localsymb.lp`.

fblanqui: That's not nice... We need to change that. I think that we should slightly modify https://github.com/Deducteam/lambdapi/blob/1f6ec9864a08110925d5fa2a7f22ed4067fd6c63/src/compile.ml#L52-L67 . After https://github.com/Deducteam/lambdapi/blob/1f6ec9864a08110925d5fa2a7f22ed4067fd6c63/src/compile.ml#L64 and before https://github.com/Deducteam/lambdapi/blob/1f6ec9864a08110925d5fa2a7f22ed4067fd6c63/src/compile.ml#L65, we should filter out of sign the local symbols.

src/compile.ml
```diff
@@ -62,7 +62,15 @@ let rec compile : bool -> Files.module_path -> unit = fun force path ->
...
       in
       ignore (List.fold_left handle sig_st (parse_file src));
-      if Pervasives.(!gen_obj) then Sign.write sign obj;
+      if Pervasives.(!gen_obj) then
+        begin
+          let not_local _ (sym, _) = Terms.(sym.sym_expo) <> Terms.Local in
+          let sign_symbols =
+            ref (StrMap.filter not_local !(sign.sign_symbols))
+          in
+          let sign = { sign with sign_symbols } in
```
fblanqui: Following my comment in https://github.com/Deducteam/lambdapi/pull/262#discussion_r344095185, this should be done in all cases even if gen_obj is not set.

src/scope.ml
gabrielhdt: Indeed, it is a way better idea

src/compile.ml
```diff
@@ -62,6 +62,9 @@ let rec compile : bool -> Files.module_path -> unit = fun force path ->
             data.pdata_finalize ss st
       in
       ignore (List.fold_left handle sig_st (parse_file src));
+      let not_local _ (sym, _) = Terms.(sym.sym_expo) <> Terms.Local in
+      (* FIXME strip bultins, unops and binops as well? *)
```
fblanqui: builtins

src/parser.ml
```diff
@@ -145,40 +145,42 @@ module KW = Keywords.Make(
...
-let _require_    = KW.create "require"
-let _open_       = KW.create "open"
-let _as_         = KW.create "as"
-let _let_        = KW.create "let"
-let _in_         = KW.create "in"
-let _symbol_     = KW.create "symbol"
-let _definition_ = KW.create "definition"
-let _theorem_    = KW.create "theorem"
-let _rule_       = KW.create "rule"
+let _TYPE_       = KW.create "TYPE"
```
fblanqui: why so many changes in this file?

src/pretty.ml
```diff
@@ -40,6 +40,12 @@ let pp_symtag : symtag pp = fun oc tag ->
 let pp_symtags : symtag list pp = fun oc ->
   List.iter (Format.fprintf oc " %a" pp_symtag)
 
+let pp_expostag : expostag pp = fun oc tag ->
+  match tag with
```
fblanqui: factorize Format.fprintf?

src/parser.ml
gabrielhdt: I sorted the lines alphabetically

fblanqui: Ok. Good.

src/why3_tactic.ml
```diff
@@ -180,7 +180,9 @@ let handle : Pos.popt -> Proof.proof_state -> sig_state -> string option ->
   (* Get the meta type of the current goal (with quantified context). *)
   let trm = !((Proof.Goal.get_meta g).meta_type) in
   (* Add the axiom to the current signature. *)
-  let a = Sign.add_symbol ss.signature Const (Pos.none axiom_name) trm [] in
+  let a = Sign.add_symbol ss.signature Public Const (Pos.none axiom_name) trm
+      []
```
fblanqui: not very nice ;-) could you try to get it on 1 line? (rename trm into t, or add let p = Pos.none ...)

tests/KO/private_rhs.dk
```diff
@@ -0,0 +1,5 @@
+#REQUIRE tests.OK.private.
```
fblanqui: I would prefer not to add more dk files...

tests/OK/prv_lhs.lp
```diff
@@ -0,0 +1,3 @@
+require tests.OK.prv
+
+rule tests.OK.prv.f &X → &X
```
fblanqui: Shouldn't this be rejected?

fblanqui: Yes, scope_rule needs to be modified.

tests/KO/private_rhs.dk
gabrielhdt: Checking retro compatibility might be still relevant since there are problems with the earley one...

src/compile.ml
fblanqui: typo not fixed yet

src/compile.ml
```diff
@@ -62,6 +62,11 @@ let rec compile : bool -> Files.module_path -> unit = fun force path ->
             data.pdata_finalize ss st
       in
       ignore (List.fold_left handle sig_st (parse_file src));
+      let not_local_fst _ (sym, _) = Terms.(sym.sym_expo) <> Terms.Local in
+      let not_local _ sym = Terms.(sym.sym_expo) <> Terms.Local in
```
fblanqui: should be the other way around: first define not_local, then not_local_fst using not_local to avoid code duplication

src/scope.ml
```diff
@@ -208,12 +221,14 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
...
   (* Scoping function for head terms. *)
-  and scope_head : env -> p_term -> tbox = fun env t ->
+  and scope_head : env -> p_term -> tbox = fun  env t ->
     match (t.elt, md) with
     | (P_Type          , M_LHS(_) ) ->
         fatal t.pos "[%a] is not allowed in a LHS." Print.pp Type
     | (P_Type          , _        ) -> _Type
-    | (P_Iden(qid,_)   , _        ) -> find_qid ss env qid
+    | (P_Iden(qid,_)   , _        ) ->
+      let allow_priv = match md with M_LHS(_) -> true | _ -> false in
```
fblanqui: as the outer match is already on md (for its second component), it would be better to have two clauses in the outer matching:
```
| (P_Iden(qid,_)   , M_LHS(_)        ) -> find_qid true ss env qid
| (P_Iden(qid,_)   , _        ) -> find_qid false ss env qid (* as before *)
```

src/scope.ml
```diff
@@ -401,9 +416,15 @@ let scope_rule : sig_state -> p_rule -> sym * pp_hint * rule loc = fun ss r ->
     let (h, args) = Basics.get_args lhs in
     let is_const s = s.sym_mode = Const in
```
fblanqui: This function should be moved to Terms (or Basics) where there is already is_inj.

gabrielhdt: Should a symbol with no rewrite rules be considered constant, even if it hasn't been marked `const`?

fblanqui: No.

src/pretty.ml
```diff
@@ -33,13 +33,21 @@ let pp_path : Pos.popt -> p_module_path pp = fun pos ->
...
+  Format.pp_print_string oc
+    (match tag with
+     | Sym_const -> "const"
+     | Sym_inj   -> "injective")
 
 let pp_symtags : symtag list pp = fun oc ->
   List.iter (Format.fprintf oc " %a" pp_symtag)
 
+let pp_expostag : expostag pp = fun oc tag ->
+  Format.fprintf oc
```
fblanqui: Use pp_print_string instead like in pp_symtag.

fblanqui:
I am happy with this PR. Thank you @gabrielhdt . @rlepigre do you have any other comment? Otherwise I will merge it tomorrow.

gabrielhdt:
There might still be an issue: perhaps we should ensure that a public symbol can't rewrite to local ones.

fblanqui:
Right. This should be extended to definitions as well (including symbols defined interactively with a proof when this feature will be added).

rlepigre:
@fblanqui I don't think this PR is ready to be merged, and I don't have time to review it now. In particular, it would be great to have some documentation motivating the need for local and private symbols (including examples).

@gabrielhdt  Something that I find very weird at a first glance is that what you call `local` should really be called `private` (i.e., non-exported symbols). By the way, how do you ensure that such symbols do not escape the corresponding abstraction barrier (i.e., the module they are defined in)? And also, what is the use case for what you call `private` symbols? I really can't see what that can be useful for (but I might just be too puzzled by the `private` name).

fblanqui:
Concerning the motivation, this has already been discussed. @EmilieGrienenberger needs this feature.

fblanqui:
@rlepigre the names are well chosen I think. Well, at least they correspond to what is done in OCaml. In Ocaml, private constructor are exported but can be used in pattern maching clauses only.

fblanqui:
Private symbols are necessary for many things and in particular for encoding proof irrelevance (in Matita, PVS, etc.).

rlepigre:
> Concerning the motivation, this has already been discussed. @EmilieGrienenberger needs this feature.

I'm not saying that it is not useful, I would just like to understand what they are so that I can review properly. A minimal example in the standard Lambdapi syntax would help, and it would be useful anyway to document the feature.

> @rlepigre the names are well chosen I think. Well, at least they correspond to what is done in OCaml. In Ocaml, private constructors are exported but can be used in pattern matching clauses only.

OCaml is really messed up for this part. The usual notion of privacy (that you call locality here) can only be accessed by masking definitions with an `.mli` file. We should not replicate the mistake they made by following their terminology. That said, I'm not sure what word could be used in place of `local`.

gabrielhdt:
@rlepigre I do agree that the naming scheme is a bit unfortunate.  Concerning the escape, this is why I suggest that we ought to ensure that public and private symbol don't rewrite to local ones; may the rewriting be introduced by a `rule` or a `definition`.  Types of public or private symbols shouldn't involve local symbols as well.  In the end it seems that local symbols might only be used in _complete_ proofs of theorems (since they are forgotten).

Concerning examples, I have written some.  I completed the documentation as well in the symbols section.

gabrielhdt:
@rlepigre perhaps `protected` instead of `private`?

fblanqui:
You are right. We should find another word for private (and then rename local into private). I'm afraid that protected has another meaning in OO programming. We should try to find something else.

fblanqui:
`protected` is not so bad after all. Unless someone else has a better idea, we could adopt it.

gabrielhdt:
A quick search for English synonyms provides:
- shielded,
- guarded,
- secured,
- sheltered,
- covered,
- safe.

fblanqui:
What about `abstract` like abstract classes/methods in OOP?

gabrielhdt:
If I'm not mistaken, `abstract` means there are no instances of such a class, whereas here, there are instances, but we can't build them; it might be misleading.

gabrielhdt:
If we want to make parallel with OO, `protected` seems more suited, looking at [C++](https://en.cppreference.com/w/cpp/language/access),

> A protected member of a class `Base` can only be accessed

fblanqui:
Ok, let's use protected instead of private, and private instead of local. @gabrielhdt could you please update your PR accordingly?

gabrielhdt:
In fine I have coded the following rules
1. a public symbol can't be typed by private symbols
2. a public symbol can't rewrite to private symbols
3. foreign protected symbols can be used in the left-hand side of rewrite rules exclusively
4. rewrite rules can't be defined on foreign private symbols

src/scope.ml
```diff
@@ -33,83 +33,108 @@ let open_sign : sig_state -> Sign.t -> sig_state = fun ss sign ->
...
   {ss with in_scope; builtins}
 
-(** [find_sym b st qid] returns the symbol and printing hint corresponding  to
-    the qualified identifier [qid].  If [fst qid.elt] is empty,  we search for
-    the name [snd qid.elt] in the opened modules of [st]. The boolean [b] only
-    indicates if the error message should mention variables, in the case where
-    the module path is empty and the symbol is unbound. This is reported using
-    the [Fatal] exception. *)
-let find_sym : bool -> sig_state -> qident -> sym * pp_hint = fun b st qid ->
+(** [find_sym ?allow_protected b st qid] returns the symbol and printing hint
```
fblanqui: ?allow_private missing

src/scope.ml
```diff
@@ -208,15 +245,24 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
     in
     aux env xs
   (* Scoping function for head terms. *)
-  and scope_head : env -> p_term -> tbox = fun env t ->
+  and scope_head : ?allow_private:bool -> env -> p_term -> tbox =
+    fun ?allow_private env t ->
     match (t.elt, md) with
-    | (P_Type          , M_LHS(_) ) ->
+    | (P_Type          , M_LHS(_)         ) ->
```
fblanqui: @rlepigre I don't like this idea of indenting the right parenthesis in matching clauses because it introduces spurious diffs in commits. I propose not to continue to do this.

src/scope.ml
```diff
@@ -137,24 +162,36 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
       _Patt None name (Env.to_tbox env)
   in
   (* Toplevel scoping function, with handling of implicit arguments. *)
-  let rec scope : env -> p_term -> tbox = fun env t ->
+  let rec scope : ?allow_private:bool -> env -> p_term -> tbox =
```
fblanqui: Why do you take an optional argument here? Isn't it possible to use a boolean instead? The semantics would be clearer.

src/sign.ml
```diff
@@ -176,12 +176,15 @@ let unlink : t -> unit = fun sign ->
...
-    (which should not already be used in [sign]) and with the type [a], in the
-    signature [sign]. The list [impl] tells whether the first arguments of the
-    symbol are set to be implicit. The created symbol is returned. *)
-let add_symbol : t -> sym_mode -> strloc -> term -> bool list -> sym =
-    fun sign sym_mode s a impl ->
+(** [add_symbol sign ?sym_expo mode name a impl] creates a fresh symbol with
+    name [name] (which should not already be used in [sign]) and with the type
+    [a], in the signature [sign]. The exposition is {!constructor:Public} by
+    default, unless [?sym_expo] is precised. The list [impl] tells whether the
+    first arguments of the symbol are set to be implicit. The created symbol
```
fblanqui: whether the first arguments of the symbol are set to be implicit -> which arguments are implicit

src/handle.ml
```diff
@@ -209,7 +218,7 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
       (* Obtaining the implicitness of arguments. *)
       let impl = Scope.get_implicitness a in
       (* Scoping the type (statement) of the theorem. *)
-      let a = scope_basic a in
+      let a = scope_basic Public a in (* Theorems are public *)
```
fblanqui: No, private theorems should be allowed too. You could want to have auxiliary lemmas that you do not want to export.

src/parser.ml
```diff
@@ -566,12 +573,12 @@ let parser cmd =
...
+  | e:exposition? _symbol_ l:symtag* s:ident al:arg* ":" a:term
+      -> P_symbol(Option.get e Symex_public,l,s,al,a)
   | _rule_ r:rule rs:{_:_and_ rule}*
       -> P_rules(r::rs)
-  | _definition_ s:ident al:arg* ao:{":" term}? "≔" t:term
-      -> P_definition(false,s,al,ao,t)
+  | e:exposition? _definition_ s:ident al:arg* ao:{":" term}? "≔" t:term
+      -> P_definition(Option.get e Symex_public,false,s,al,ao,t)
   | st:statement (ts,e):proof
       -> P_theorem(st,ts,e)
```
fblanqui: Theorems should have an exposition too.

fblanqui:
Gabriel wrote: 

> In fine the following rules have been implemented
> 1. a public symbol can't be typed by private symbols
> 2. a public symbol can't rewrite to private symbols
> 3. foreign protected symbols can be used in the left-hand side of rewrite rules exclusively
> 4. rewrite rules can't be defined on foreign private symbols

I would formulate differently:
1. the type of a public symbol cannot contain locally defined private symbols
2. the right-hand side of a rewriting rule defining a public symbol cannot contain locally defined private symbols
3. externally defined protected symbols cannot appear at the head of a left-hand side or in a right-hand side of a rule

I don't understand the point 4 as there cannot be any foreign (= externally defined) private (= not exported) symbols.

gabrielhdt:
My mistake, I, of course, meant protected.

doc/sections/commands.md
```diff
@@ -84,6 +86,22 @@ We recommend to start types and predicates by a capital letter.
...
+- `protected`: the symbol can only be used in left-hand side of rewrite rules
+  outside of its module.
+
+Exposition obeys the following rules
+- a public symbol can't by typed by private symbols
+- a public symbol can't rewrite to private symbols
+- foreign protected symbols can be used in the left-hand side of rewrite rules
+  exclusively
+- rewrite rules can't be defined on foreign private symbols.
+
```
fblanqui: Should be clarified. See https://github.com/Deducteam/lambdapi/pull/262#issuecomment-553010109 . Start by saying "Inside a module, ..."

fblanqui:
doc/syntax.bnf should be updated too.

fblanqui:
ok, so my n°3 = your n°3 + your n°4

gabrielhdt:
For the sake of being picky, locally private is a pleonasm, isn't it?

src/scope.ml
gabrielhdt: You can choose to not display space related changes via a tickbox in the github interface

src/scope.ml
gabrielhdt: It would be a boolean option: on the first scoping of a rewrite rule, this option is set to `Some(b)` to indicate whether private symbols are allowed. Therefore, the first run of `scope` is performed without indicating the argument.

src/scope.ml
rlepigre: I would prefer to continue doing that, because this makes the code more readable. As Gabriel said, there are ways to ignore space changes when looking at diffs.

rlepigre:
Looking at the syntax, I don't think it makes sense to have "modifiers" on both sides of the `symbol` keyword. Why not use, e.g., the following syntax?
```
const symbol z : N
private injective symbol s : type
```
I think I used that syntax at the beginning since I already thought it made more sense.  In English you would say something like "we define a private, injective symbol s" so that feels more natural.

gabrielhdt:
@rlepigre I propose to do that in an other PR, right after this one is merged, since it involves modifying many test files.

src/scope.ml
fblanqui: I didn't know. Thanks!

src/parser.ml
fblanqui: But TYPE should come later then... Please add at the beginning that it should be kept in alphabetical order.

src/scope.ml
```diff
@@ -33,83 +33,106 @@ let open_sign : sig_state -> Sign.t -> sig_state = fun ss sign ->
...
+          try (fst (StrMap.find s st.in_scope), Nothing) with Not_found ->
+          let txt = if b then " or variable" else "" in
+          fatal pos "Unbound symbol%s [%s]." txt s
+        end
+    | [m] when StrMap.mem m st.aliases -> (* Aliased module path. *)
+        begin
+          (* The signature must be loaded (alias is mapped). *)
+          let sign =
+            try PathMap.find (StrMap.find m st.aliases) Timed.(!Sign.loaded)
+            with _ -> assert false (* cannot fail. *)
```
fblanqui: cannot fail -> Should not happen, like a few lines later

src/scope.ml
```diff
@@ -33,83 +33,106 @@ let open_sign : sig_state -> Sign.t -> sig_state = fun ss sign ->
...
+        end
+  in
+  (* Reject (if [allow_protected] is false) protected symbols from other
+     signatures. *)
+  begin
+    match (prt, prv, s.sym_expo) with
+    | (false, _    , Protected) ->
+        if s.sym_path <> st.signature.sign_path then
+          fatal pos "Protected symbol not allowed."
+    | (_    , false, Private  ) -> fatal pos "Private symbol not allowed."
```
fblanqui: allowed -> allowed here (on both lines)

src/parser.ml
gabrielhdt: Capital letters come first I guess (I only used M-x sort-lines on emacs...)

src/scope.ml
fblanqui: I don't understand. Could you please elaborate? Why are you talking of several scoping runs/passes? Note that, after lines 255-256, None seems to be equivalent to Some(true).

fblanqui:
Concerning https://github.com/Deducteam/lambdapi/pull/262#issuecomment-553294233, I agree but, as Gabriel said, in another PR. Then, we should change const into constant as well.

src/parser.ml
fblanqui: I guess that Emacs is using ASCII code ordering. This is not very human friendly. I propose to use the standard lexicographic ordering ignoring whether letters are small or big.

src/scope.ml
fblanqui: When we handle a definition or a theorem, allow_private is given by the expostag of the definition/theorem: allow_private=true if expostag=Private, and false otherwise.
When we handle a tactic or a query, we can take allow_private=true.
When we handle a rule l->r, we could start by checking that l is of the form f l1 .. ln with f a function symbol, without calling scope. Then, scoping l1 .. ln with allow_private=true and r with allow_private=(f is Private).

gabrielhdt: The first attempt was to get `f` itself to see its exposition, but it seems that because of infix symbols, this appeared tricky. What is done now is:
- scope `l -> r`
- the call to `scope_head` allows to retrieve the root symbol; so we allow private symbol to this first recursive call
- the rest of the rule is scoped with the exposition of the root symbol.

My "scoping runs" referred to the successive recursive calls of the `scope` function.

fblanqui: I don't understand why infix symbols would be a problem. Looking more at the code, I feel that there is something strange concerning unary and binary symbol applications @rlepigre . Syntax.get_args does not really return the head term nor all the arguments. If the term is a P_BinO or P_UnaO, it does not decompose it. This is not documented. Now, P_UnaO((op,_,f),t) should be the same as P_Appl(Symb(f,Unary(op)),t), and similarly for P_BinO. Well, in fact, the parser could certainly replace one by the other so that P_UnaO and P_BinO could be removed from p_term. Why don't we do that?

doc/syntax.bnf
```diff
@@ -1,38 +1,40 @@
-reserved "require"
-reserved "open"
-reserved "as"
-reserved "let"
-reserved "in"
-reserved "symbol"
-reserved "definition"
-reserved "theorem"
-reserved "rule"
+reserved "TYPE"
```
fblanqui: Thanks to put TYPE later (standard lexicographic ordering).

src/pretty.ml
```diff
@@ -33,13 +33,21 @@ let pp_path : Pos.popt -> p_module_path pp = fun pos ->
...
-  | Sym_inj   -> Format.pp_print_string oc "injective"
+  Format.pp_print_string oc
+    (match tag with
+     | Sym_const -> "const"
+     | Sym_inj   -> "injective")
 
 let pp_symtags : symtag list pp = fun oc ->
   List.iter (Format.fprintf oc " %a" pp_symtag)
 
+let pp_expostag : expostag pp = fun oc tag ->
```
fblanqui: @gabrielhdt would you mind to replace expostag by expotag?

src/scope.ml
rlepigre: It would be a really bad idea to remove these constructors. They are useful for many good reason, for example to be able to print back the AST (although this is not yet done correctly). During scoping, unary and binary symbols should be desugared early enough so that they do not interfere with further computations.

fblanqui: I don't understand why you want to be able to print out terms as they have been entered. This is not important to me. What is important is to print terms in a uniform and predictable way. This is related to #260.

rlepigre: Because you want the user to be able to recognize their own terms! For example if there is a notation in scope that the user does not know about or does not want to use it is not a good idea to force them to see it when printing subterms of what the user typed.

fblanqui: Let's continue the discussion on notations in #260.

fblanqui: @rlepigre for what other reasons the constructors P_UnaO and P_BinO are useful?

fblanqui: @gabrielhdt could you please explain why it would be tricky with infix symbols? Is it because Syntax.get_args is not correct? As this function is used only in scope, we could change it.

rlepigre: Another reason why they are useful is to have correct positions in scoping.

fblanqui: I don't understand. Could you elaborate?

rlepigre: Well, in the current representation, if you have a term representing `e1 + e2` then you have the position of the whole expression and you also have access to the positions of `e1` and `e2`. On the other hand, if you build a term representing `(add e1) e2` from a parsed infix notation then you have no correct source code position for the subterm `add e1`, nor for `add` itself.

fblanqui: Ok but, currently, this is not used much. Remark: the position of add could be set to that of +.

gabrielhdt: @fblanqui it was precisely because of `Syntax.get_args`, I expected it to do like `Basics.get_args`, which for `3 + 2` would return `(+, [3; 2])` (at the level of `p_term`) but apparently it does not.

fblanqui: Ok. What if you fix Syntax.get_args? Could you try to use a bool instead of a bool option in this case?

gabrielhdt: Indeed I could, I would scope the root symbol alone, get its privacy and then scope the rest of the lhs with the boolean set accordingly.

rlepigre: It would actually make sense to move `Syntax.get_args` inside the definition of the scoping function (it is only useful there anyway). That way you can actually move the code that unsugars the unary and binary symbols into the definition of `get_args`, and so the unary and binary symbol constructors are then eliminated early.

fblanqui:
Thanks @gabrielhdt !

rlepigre:
Why was this merged already?! It is not a good idea to rush things, I did not even have time to look at the code yet.

fblanqui:
@rlepigre you cannot review everything in full details! all the more so since there are other things to do like #243

rlepigre:
Well, I think it would make sense for me to review PRs that touch at big chunks of code that I wrote (in this case scoping and signature stuff). What if this code introduces some bugs that I could have spotted had you given me time to review? The PR is still marked as work in progress, so I did not even know that it was ready for me to review.

fblanqui:
I reviewed carefully this code. One careful review is enough I think. Many people are waiting for PRs to be merged and LP to be improved to make further progress in their work: Chantal, Michael, Gaspard, Rehan, etc. We have to go forward. We cannot wait so long.

rlepigre:
I'm not the one responsible for the time it takes, I always review things fairly quickly (assuming they are in a reasonable state). We should not confuse work in progress with code that is ready to be merged: I only have time to review things that are deemed ready to be merged.

fblanqui:
@EmilieGrienenberger , @francoisthire and @Gaspi : private and protected symbols are now available in LP. Be careful that Dedukti.private = Lambdapi.protected. See https://github.com/Deducteam/lambdapi/blob/master/doc/sections/commands.md for the doc.

Issue #264, opened 2019-10-30
Comment missing in the code


fblanqui:
@gabrielhdt , in https://github.com/Deducteam/lambdapi/blob/cfee85433a7207c2716c9b8b6dfe45744a611668/src/tree.ml#L431 there is a List.hd that is not inside a try ... with. Is it the case that the clauses argument is always non empty? If so, please add a comment. Otherwise, thanks to fix this line.

rlepigre:
Actually, it is in a `try ... with Not_found -> ...` block.

gabrielhdt:
Ah indeed that one is a bit tricky: there is always at least one element because empty lines have been filtered out by the `Check_stack` instruction three lines above.  If the `rule_order` flag is set up, then we filter out empty rules

fblanqui:
@rlepigre , List.hd does not raise Not_found. @gabrielhdt thanks to add a comment in the code.

rlepigre:
> @rlepigre , List.hd does not raise Not_found. @gabrielhdt thanks to add a comment in the code.

Oh yes of course! :)

fblanqui:
Fix in ec2ee4e6c05d8fc5ad200b609eaf7eb2724f82e9

PR #265, opened 2019-10-30
Documentation issue 264

gabrielhdt:
See #264

### DIFF
--- a/src/tree.ml
+++ b/src/tree.ml
@@ -428,10 +428,12 @@ module CM = struct
     then Check_stack else
     try
       if Pervasives.(!rule_order) then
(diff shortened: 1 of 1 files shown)


src/tree.ml
```diff
@@ -428,10 +428,12 @@ module CM = struct
     then Check_stack else
     try
       if Pervasives.(!rule_order) then
+        (* There is no empty rule in this branch of the [if]. *)
```
fblanqui: Be more informative: explain why.

gabrielhdt: Better?

fblanqui:
Thanks!

PR #266, opened 2019-11-06
doc minor corrections

Rehan-MALAK:
documentation minor corrections

### DIFF
--- a/doc/sections/commands.md
+++ b/doc/sections/commands.md
@@ -327,7 +327,7 @@ and identifiers, but they are by no means sufficient (it is still possible to
 break the parser by defining well-chosen tokens).
 
 **equality-related builtins** In order to use tactics related to
-Leibinz equality, one first has to define a number of builtin symbols
+Leibniz equality, one first has to define a number of builtin symbols
 as follows:
 
 ```
--- a/doc/sections/commands.md
+++ b/doc/sections/commands.md
@@ -257,7 +257,7 @@ set debug -s
 
 Each functionality is represented by a single character. For instance,
 `i` stands for type inference. To get the list of debuggable functionalities,
-do `lambdapi -h`.
+do `lambdapi -help`.
 
 **flags** The user can set/unset some flags:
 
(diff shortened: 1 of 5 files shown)


lp-lsp/README.md
```diff
@@ -17,9 +17,11 @@ The resulting binary will be in `../_build/install/default/bin/lp-lsp`
 
 There is an extension for VSCode derived from VSCoq. To install do:
 
-$ cd editors/vscode/
-$ npm install
-$ ln -s `pwd` ~/.vscode/extensions/
+```
```
rlepigre: We could do even better, and use bash syntax coloring (see the main `README.md` for an example).

doc/sections/commands.md
```diff
@@ -257,7 +257,7 @@ set debug -s
 
 Each functionality is represented by a single character. For instance,
 `i` stands for type inference. To get the list of debuggable functionalities,
-do `lambdapi -h`.
+do `lambdapi -help`.
```
fblanqui: -h is an option that works. -h and -help are added by ocaml but only -help is listed. I don't know if it is possible to change that. -help is not standard though. long option names should be prefixed by -- like --help. So, I propose to use --help instead.

rlepigre: No, `-h` does not work. Like with every other unknown options, an error message is shown and the list of valid options is given. We do have `-help` and `--help` which are both added by default (see [here](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Arg.html)), and that behavior is hard to change.

fblanqui: Ah right. I didn't see the error message (shouldn't we put them in red?). So we have to use --help indeed.

rlepigre: I don't think we easily can since the `Arg` module produces the error message. But yes, we can document `--help` instead.

tools/rec_to_lp/README.md
```diff
@@ -11,7 +11,7 @@ REC files can be retrieved from
 <https://scm.gforge.inria.fr/anonscm/svn/rec>.  The haskell files are
 in `rec/2019-CONVECS/HASKELL/`.  Then use the awk script,
 
-``` shell
+```bash
```
Rehan-MALAK: ``bash`` is an alias for ``shell`` https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

PR #240, opened 2019-08-30
[lsp] Add vscode plugin with goal display support

houdaM97:
Hi, 
this is the last version of my work on vscode extension for lambda pi (LSP client/server), it still to upload merge and upload to market place

PR #268, opened 2019-11-08
Vscode lsp client install

Rehan-MALAK:
Dear all, I don't know what's best if I want to suggest a PR (mine) on a PR (houdaM97), but the last one is from a student not working anymore on the project.

So @ejgallego @rlepigre @fblanqui , how about we review the code of vscode plugin here instead, so I can work on it without being obliged to PR somewhere else ?

https://github.com/Deducteam/lambdapi/pull/240


fblanqui:
Compilation works now. Thanks!

fblanqui:
But I have two problems:
1) No goals are printed.
2) When I open a file A.lp containing require B.lp and B.lpo exists, I get the error: `Error when loading module [Set]. File [Set.lpo] is incompatible with current binary...`

Rehan-MALAK:
It's not working correctly right now.
And the Goals window is opening even if we are not working on a lambdapi source file.
I have to learn a bit about VScode extensions.

rlepigre:
> 2. When I open a file A.lp containing require Set.lp and Set.lpo exists, I get the error: `Error when loading module [Set]. File [Set.lpo] is incompatible with current binary...`

That one is normal, currently the `lambdapi` and LSP server binary cannot share `.lpo` files. There are basically two ways to solve the problem:
 1. Use our own serialization library instead of using the `Marshal` module,
 2. Integrate the LSP server into the `lambdapi` binary (through an option `--lsp-server` for example), which I think would make sense anyway.

fblanqui:
Ok for 2. If I understand well, it consists in calling Lp_lsp.main() from lambdapi.ml when the option --lsp-server is set.

ejgallego:
Hey folks, I'm just back to Paris; I will take care of the plugin this week. A few comments:
- I have a cleaned up version of Houda branch [that I did with her], I propose I update the original PR, test, and merge. Goal display etc... should work.
- Regarding the marshall error, I think that is really strange, `lp-lsp` should be compatible with `lp` as it doesn't modify any data type. I don't think merging the binaries is the right solution to this problem [I will comment more on the corresponding PR]
- Using a serialization library tends to have many advantages over `Marshall` IMO, but it is also not a pressing need I think.

fblanqui:
@Rehan-MALAK do you want to do another PR based on master to include the things you have done that have not been included in https://github.com/Deducteam/lambdapi/pull/240?

Rehan-MALAK:
There isn't much to keep. I'll open a PR for the Makefile idea and maybe another one to update editors/vscode/vscode.proposed.d.ts
After that I could investigate on this 1.40 problem

Issue #277, opened 2019-11-14
Improve documentation of graphviz files


fblanqui:
Change bound variable prefix from "tr" to "v" or "var", and add comment about that in tree_graphviz.ml and in the doc too.
Add also in those files that a file is generated for each function symbol following the schema file.symbol_name.gv.

PR #279, opened 2019-11-15
Variables in tree representation, documentation

gabrielhdt:
Answering to #277, documentation completion and better names for variables in decision trees.


src/tree.ml
```diff
@@ -693,7 +703,7 @@ let rec compile : CM.t -> tree = fun ({clauses ; positions ; slot} as pats) ->
       (* Abstraction specialisation*)
       let abstraction =
         if List.for_all (fun (x, _) -> x <> TC.Abst) cons then None else
-        let var = Bindlib.new_var mkfree "tr" in
+        let var = Bindlib.new_var mkfree (make_fresh_varname ()) in
```
fblanqui: Isn't make_fresh_varname duplicating Bindlib code @rlepigre ?
If you want variable names to start with "var0", you can probably do once an `ignore(Bindlib.new_var mkfree "var")` somewhere.

rlepigre: If you have not already done it, read the Bindlib paper or documentation.

gabrielhdt: It is for the display of decision trees, to be able to distinguish matched variables (the name collision of Bindlib works like a charm, but isn't visible).

rlepigre: I don't get it, they should already be distinguished if there is a name collision.

gabrielhdt: In the graphviz file, we see only "var" for all variables.  Adding the integers enables to *see* "var1", "var2" in the generated graphviz. It is only used for the sake of being more informative in the visible trees.

rlepigre: But aren't the names correct otherwise? What do you mean by "more informative"?

gabrielhdt: I guess two images is worth two thousand words in that case, compare
flip_ok
with this "less informative" image
flip_KO

gabrielhdt: which is still a bit confusing since the rhs uses `x` and `y`, but closedness constraints would use the `var1` and `var2` variables. And as you suggested @rlepigre, this is not essential, it is only used for prettyness.

rlepigre: I see, but in that case why even give names to a variable? The only useful information here seems to be that we have an abstraction constructor. Why not just print the lambda?

gabrielhdt: It _can_ have meaning in two cases:
- conditional nodes
dx
- matching on variables
vm

rlepigre: I see. And what do these examples if your hack is removed? (I say it is a hack because that is definitely not the code you want to write with Bindlib. Actually, I wonder if using Bindlib variables here makes sense.)

gabrielhdt: You would only see `var` instead of `var1`.

gabrielhdt: Another way would be to have some kind of (imperative) hashtable in `tree_graphviz.ml` mapping each bindlib variable to a textual representation.

gabrielhdt: Another way would be to use an (imperative) hashtable in `tree_graphviz` to map bindlib vars to strings

rlepigre: OK. I think I would prefer that solution. Note that if you do that you'll need to use `Hashtbl.Make` to create custom hashtable module (using `Bindlib.eq_vars` and `Bindlib.hash_var`).

gabrielhdt: Ah I am crippled by my `TC.Vari` which only use strings in `tree_types`, so it means that the modification has to be done in the variable name itself. Or the `TC.Vari` type must be changed. I'll see if I manage to get something interesting.

rlepigre: OK. I never liked the fact that we use string there, but I don't see any other obvious solution (my understanding of this code is still not perfect).

gabrielhdt: So now a hash table is used to map a numeric id to each variable (instead of storing this id into the name of the variable).

src/tree.ml
```diff
@@ -211,6 +201,28 @@ module CP = struct
...
+(** Hashable term variables. *)
+module TvarHash =
+struct
+  type t = tvar
+  let equal = Bindlib.eq_vars
+  let hash = Bindlib.hash_var
+end
+
+(** Hashtable of term variables. *)
+module TvarHashtbl = Hashtbl.Make(TvarHash)
```
rlepigre: You can unfold `TvarHash` here:
```ocaml
module TvarHashtbl = Hashtbl.Make(
  struct
    type t = tvar
    let equal = Bindlib.eq_vars
    let hash = Bindlib.hash_var
  end)
```

src/tree.ml
```diff
@@ -211,6 +201,28 @@ module CP = struct
...
+
+(** Associates to each variable an identifier to be used by
+    {!constructor:Tree_types.TC.Vari}. *)
+let varmap : int TvarHashtbl.t = TvarHashtbl.create 0
+
+(** Prefix of bound variables. *)
+let var_prefix = "var"
+
+(** A counter for identifiers. TODO hide it in a [store] function. *)
+let vid : int Pervasives.ref = Pervasives.ref 0
```
rlepigre: Is the hash table and the identifier ever reset? Otherwise you will always have growing number which will not be easily readable when printing.

With the hash table, you will also be leaking memory since the variables, even when not used anymore, will still be accessible (and hence not collected by the GC). To work around, you can use instead a weak hash table (`Weak` module). But a better solution would be to reset the hash table and the counter when you have finished building a decision tree (so that next time it starts fresh, from 0).

gabrielhdt: But the ids might be accessed when evaluating the decision trees, in `eval.ml`, so we don't want to reset the hash table

rlepigre: OK. Having one per tree would make more sense indeed.

fblanqui: All this looks over complicated to me. @rlepigre what's wrong with Gabriel's first proposal? Note that all this is just to get clearer graphs. It's only for debug purposes.

rlepigre: What is wrong is that it makes a bad use of Bindlib. I don't want the code to contain hacks like these because it might encourage others to do similar things (thinking that it is the right way to use Bindlib).

gabrielhdt: I think it tends to be clearer to explicitly declare the identifier used to relate the variables than hiding this identifiers in their names (which can be seen as hackish).

fblanqui: I understand but, in this case, is it worth doing that? it will increase memory and slow down tree construction, unless this mapping is done only when one wants to print the trees.

rlepigre: Hash table operations are roughly constant time. And the tree construction is not efficiency critical anyway: only matching is.

gabrielhdt: But this seems tricky since it makes the dtree type depend on a functor, doesn't it force to raise everything to the functor level?

rlepigre: What you are missing here is that the custom hash table module is only created once (using the functor) at the toplevel of the file. But then you will just manipulate distinct hash tables (all created with `TvarHashtbl.create`).

gabrielhdt: But if you want one hash table per tree, don't you want to modify the `dtree` type to contain the hash table?

rlepigre: Sure, but it is only gonna include an element of type `int TvarHashtbl.t` somewhere. No need to apply the functor more than once (at the top level of the file).

src/tree.ml
```diff
@@ -452,12 +479,17 @@ module CM = struct
...
           Some(TC.Symb(arity, sym_name, sym_path), e)
-      | Abst(_, _)                           -> Some(Abst, e)
-      | Vari(x)                              ->
-          Some(Vari(Bindlib.name_of x), e)
-      | _                                    -> None
+      | Abst(_)                          -> Some(Abst, e)
+      | Vari(x)                          ->
+          let id =
+            try TvarHashtbl.find varmap x
+            with Not_found -> assert false
```
fblanqui: could perhaps be written on one line

src/eval.ml
```diff
@@ -261,9 +275,11 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
                 with Not_found -> default ()
               end
           | Vari(x)    ->
-              let cons = TC.Vari(Bindlib.name_of x) in
               begin
                 try
+                  let id = Tree.TvarHashtbl.find vid x in
```
fblanqui: wouldn't be better to write `try ... with Not_found -> assert false`?

src/tree.ml
```diff
@@ -201,6 +201,33 @@ module CP = struct
...
+    needed. *)
+let store_bound : tvar -> unit =
+  let open Pervasives in
+  let vid : int ref = ref 0 in
+  fun v ->
+    if not (TvarHashtbl.mem varmap v) then
+    begin
+      incr vid;
+      TvarHashtbl.add varmap v !vid
+    end
```
fblanqui: could be written on one or two lines

src/tree_graphviz.ml
```diff
@@ -54,6 +60,16 @@ open Extra
...
+(** [name_of v] returns a friendly and unique name of variable [v] using
+    {!val:Tree.varmap}. *)
+let name_of : tvar -> string =
+  fun v ->
+  let id =
+    try Tree.TvarHashtbl.find Tree.varmap v
+    with Not_found -> assert false
+  in
+  Tree.var_prefix ^ (string_of_int id)
+
```
fblanqui: please shorten the number of lines, and remove parentheses around string_of_int

src/tree_graphviz.ml
```diff
@@ -54,6 +60,16 @@ open Extra
 open Terms
 open Tree_types
 
+(** [name_of v] returns a friendly and unique name of variable [v] using
+    {!val:Tree.varmap}. *)
+let name_of : tvar -> string =
```
fblanqui: why not moving this function in tree.ml?

src/eval.ml
```diff
@@ -163,6 +176,7 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
   let (lazy capacity, lazy tree) = tree in
   let vars = Array.make capacity Kind in (* dummy terms *)
   let bound = Array.make capacity TE_None in
+  let vid = Tree.TvarHashtbl.create 0 in
```
fblanqui: rename vid by something more informative like "var_hashtbl"

src/eval.ml
```diff
@@ -274,7 +290,11 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
                 match abstraction with
                 | None         -> default ()
                 | Some(fv, tr) ->
-                    let nfv = Bindlib.new_var mkfree (Bindlib.name_of fv) in
+                    let nfv = Bindlib.new_var mkfree Tree.var_prefix in
```
fblanqui: Why using Tree.var_prefix here? Shouldn't Tree.var_prefix use in graphviz rendering only? If so, then Tree.var_prefix should be move to tree_graphviz.ml too.

src/eval.ml
```diff
@@ -274,7 +290,11 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
...
+                    (* Get the id set during tree build. *)
+                    let id = Tree.TvarHashtbl.find Tree.varmap fv in
```
fblanqui: use `try ... with Not_found -> assert false`

fblanqui:
Hi. What's the situation with this PR?

The code is very confusing to me. Tree declares a global htable varmap used in tree construction to map Bindlib vars (of the LHS) to unique numbers. But in the abstraction field of the Node constructor, one still uses Bindlib vars... Then, during evaluation, when matching on an abstraction, one retrieves the number mapped by varmap to the abstracted variable (why isn't this done at construction time?) and uses yet another hash table vid to map the construction-time abstracted variable to a new one, and substitute the matched abstracted variable by this new one! 

I am wondering: do the tree type really need to use Bindlib vars (in Leaf and Node abstraction)? Couldn't the tree type use unique numbers for variables that would be mapped to actual tvar's at evaluation time?

src/tree_graphviz.ml
```diff
@@ -57,7 +63,7 @@ open Tree_types
 (** Printing hint for conversion to graphviz. *)
 type dot_term =
   | DotDefa (* Default case *)
-  | DotAbst of tvar
+  | DotAbst of tvar * int
```
fblanqui: Why do you need to keep a tvar argument here? Can't you remove it?

src/tree_graphviz.ml
```diff
@@ -73,18 +79,18 @@ let to_dot : string -> sym -> unit = fun fname s ->
     let pp_dotterm : dot_term pp = fun oc dh ->
       let out fmt = Format.fprintf oc fmt in
       match dh with
-      | DotAbst(v)           -> out "λ%a" Print.pp_tvar v
+      | DotAbst(v,id)        -> out "λ%s%d" (Bindlib.name_of v) id
```
fblanqui: Why not `out "λ%s%d" Tree.var_prefix i` like in the case `DotCons(Vari(i))`?

src/tree_types.ml
```diff
@@ -90,7 +87,7 @@ type ('term, 'rhs) tree =
           right hand side, are for constraint checks. *)
       ; children : ('term, 'rhs) tree TCMap.t
       (** Subtrees representing the matching of available constructors. *)
-      ; abstraction : ('term Bindlib.var * ('term, 'rhs) tree) option
+      ; abstraction : ('term Bindlib.var * int * ('term, 'rhs) tree) option
```
fblanqui: Can't you remove the tvar argument?

src/tree.ml
```diff
@@ -201,6 +201,8 @@ module CP = struct
     if res = None then choose_vf pools else res
 end
 
+(** Prefix of bound variables. *)
+let var_prefix = "var"
```
fblanqui: Can probably be removed once the tvar arguments are removed from the tree type.

src/tree_graphviz.ml
gabrielhdt: It seems that if I don't keep the tvar, the substitution can't be done properly in `eval`. To see this, comment line 203 of `eval.ml` and run
```
dune exec -- lambdapi tests/OK/miller2.dk --verbose 3
```
to obtain the term
```
λx0 y0, pair 0 (p (d (λx0, ch x) x) (d (λx, sh y) x))
```

src/tree_graphviz.ml
gabrielhdt: It should be rigorously the same

src/tree_graphviz.ml
fblanqui: Which line in the current version? Please give a permalink.

gabrielhdt: https://github.com/gabrielhdt/lambdapi/blob/8d59ad2462bb54c1fca12fd0c05e6deb6f0b4173/src/eval.ml#L203

But I figured out what happens: during the tree build, the env_builder is built regarding the variables used. To wipe out the variables from the TC, I'd have to modify the tree build. I'll give it a try later on, but we might still want to merge what has been done here (for I think the code is already better).

fblanqui: I prefer to get rid of tvar's completely first.

gabrielhdt:
The variables have been completely removed. @fblanqui, @rlepigre, you can review the code, only the documentation, and perhaps some code simplification regarding tree types are yet to be done.

src/tree_types.ml
```diff
@@ -11,10 +11,8 @@ module TC =
     type t =
       | Symb of int * string * Files.module_path
```
fblanqui: Would you mind to put those arguments in reverse order, `Symb of Files.module_path * string * int`, or does this imply too many changes?

src/tree.ml
fblanqui: I propose to remove this definition as there is no reason to share it between tree.ml and tree_graphviz.ml.

src/tree.ml
```diff
@@ -686,18 +706,25 @@ let rec compile : CM.t -> tree = fun ({clauses ; positions ; slot} as pats) ->
...
         let ncm = CM.{clauses ; slot ; positions} in
         if CM.is_empty ncm then None else Some(compile ncm)
       in
       (* Abstraction specialisation*)
       let abstraction =
-        if List.for_all (fun (x, _) -> x <> TC.Abst) cons then None else
-        let var = Bindlib.new_var mkfree "tr" in
+        let is_abst = function Abst(_) -> true | _ -> false in
+        if List.for_all (fun x -> not (is_abst x)) column then None else
+        let var = Bindlib.new_var mkfree var_prefix in
```
fblanqui: Inline var_prefix here.

src/tree_graphviz.ml
```diff
@@ -73,18 +79,18 @@ let to_dot : string -> sym -> unit = fun fname s ->
     let pp_dotterm : dot_term pp = fun oc dh ->
       let out fmt = Format.fprintf oc fmt in
       match dh with
-      | DotAbst(v)           -> out "λ%a" Print.pp_tvar v
+      | DotAbst(id)          -> out "λ%s%d" Tree.var_prefix id
```
fblanqui: Inline Tree.var_prefix here.

src/tree_graphviz.ml
```diff
@@ -73,18 +79,18 @@ let to_dot : string -> sym -> unit = fun fname s ->
...
       | DotDefa              -> out "*"
       | DotCons(Symb(a,n,_)) -> out "%s<sub>%d</sub>" n a
-      | DotCons(Vari(s))     -> out "%s" s
-      | DotCons(Abst)        -> assert false
+      | DotCons(Vari(i))     -> out "%s%d" Tree.var_prefix i
```
fblanqui: Inline Tree.var_prefix here too.

src/eval.ml
```diff
@@ -154,6 +154,19 @@ and eq_modulo : term -> term -> bool = fun a b ->
...
+       introduced;
+    3. a {!constructor:Tree_types.TC.t.Vari} which is a simplified
+       representation of a variable for trees.
+
+    Regarding reentrancy: each time an abstraction is matched, a fresh
+    variable is created. If fresh variable [w] represents the variable [v]
+    stored in the tree, the equality of the tree constructors for [v] and [w]
+    must be ensured (for the correctness of the matching). *)
+
 (** [tree_walk tree stk] tries to apply a rewriting rule by matching the stack
```
fblanqui: The comment should be updated with the new arguments.

src/eval.ml
```diff
@@ -199,22 +212,23 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
           | CondNL(i, j) ->
               if eq_modulo vars.(i) vars.(j) then ok else fail
           | CondFV(xs,i) ->
-              let xs = Array.map (fun e -> VarMap.find e fresh_vars) xs in
+              let xs = Array.map (fun e -> IntMap.find e id_vars) xs in
+              let fn b =
```
fblanqui: Try to use more informative names like:
```
              let id_vars_notin b =
                let fn _ x = not (Bindlib.occur x b) in IntMap.for_all fn id_vars
```

src/eval.ml
```diff
@@ -257,33 +272,35 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
...
-                | Some(fv, tr) ->
-                    let nfv = Bindlib.new_var mkfree (Bindlib.name_of fv) in
-                    let bound = Bindlib.subst b (mkfree nfv) in
-                    let u = bound :: args in
-                    let fresh_vars = VarMap.add fv nfv fresh_vars in
-                    walk tr (List.reconstruct left u right) cursor fresh_vars
+                | None        -> default ()
+                | Some(id,tr) ->
+                    let bound, body = Bindlib.unbind b in
+                    let u = body :: args in
```
fblanqui: could be inlined

src/tree.ml
```diff
@@ -686,18 +706,25 @@ let rec compile : CM.t -> tree = fun ({clauses ; positions ; slot} as pats) ->
...
+        let id = !count in
```
fblanqui: you can avoid this let by incrementing first

src/tree_types.ml
gabrielhdt: But this order allow the pervasive comparison to start comparing the int, which should be faster (although one can argue that we should favour code readability)

src/tree.ml
```diff
@@ -651,17 +663,22 @@ let harvest : term array -> rhs -> CM.env_builder -> int -> tree =
 
 (** [compile m] translates the pattern matching problem encoded by the matrix
     [m] into a decision tree. *)
-let rec compile : CM.t -> tree = fun ({clauses ; positions ; slot} as pats) ->
+let compile : CM.t -> tree = fun m ->
+  let count = ref 0 in
+  let rec compile : var_indexing -> CM.t -> tree =
```
fblanqui: why not putting vars_id in a ref too?
```
  let vars_id = ref VarMap.empty in
  let rec compile ({clauses; positions; slot} as pats) =
    ...
  in compile m
```

src/tree.ml
```diff
@@ -495,10 +504,11 @@ module CM = struct
...
-          | Some(i) -> (slot, (i, Array.map to_tvar e)) :: r.env_builder
+          | Some(i) ->
+              (slot, (i, Array.map (index_var vi) e)) :: r.env_builder
           | None    -> r.env_builder
         in
-        { r with env_builder ; cond_pool }
+        {r with env_builder; cond_pool}
     | _             -> r
 
   (** [specialize pat col pos cl] filters and transforms LHS of [cl] assuming
```
fblanqui: It would be nice that in all these functions (specialize, abstract, etc.) the arguments name in the comment and in the code are the same.

src/eval.ml
gabrielhdt: But the `if ... then ... else` won't be a one liner

src/tree.ml
gabrielhdt: Isn't it better to have a constant rather that several times the "var" string written in the code?

src/tree.ml
gabrielhdt: I don't think I can since, in the last line, I need the `id` variable, as well as the counter updated, for the new call to `compile`

gabrielhdt: To be more precise, the value of the counter must be saved to be written in the resulting tuple `Some(id, compile...)`. But the call to `compile` which seems to be performed before the evaluation of the `id` variable (caml reads arguments from right to left it seems) modifies the counter.

src/tree_types.ml
fblanqui: Are you sure that OCaml compares the arguments from left to right? If you want to be sure of that, it seems better to define your own comparison function.

src/eval.ml
fblanqui: Perhaps but understandability is more important.

src/tree.ml
fblanqui: But var_prefix is used only here in tree.ml... Using a constant usable outside suggests that the choice of this prefix is important which is not the case. In fact you could take any string here, including "".

src/tree.ml
fblanqui: I see.

src/tree_types.ml
gabrielhdt: Indeed, I'll change this

src/eval.ml
```diff
@@ -143,17 +143,23 @@ and eq_modulo : term -> term -> bool = fun a b ->
   let res = try eq_modulo [(a,b)]; true with Exit -> false in
   if !log_enabled then log_conv (r_or_g res "%a == %a") pp a pp b; res
 
-(** {b NOTE} that matching with trees involves three collections of terms.
+(** {b NOTE} that matching with trees involves two collections of terms.
```
fblanqui: Shouldn't this comment be inside the code of tree_walk, to comment the auxiliary function walk?

src/eval.ml
```diff
@@ -163,12 +169,12 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
   let (lazy capacity, lazy tree) = tree in
   let vars = Array.make capacity Kind in (* dummy terms *)
   let bound = Array.make capacity TE_None in
-  (* [walk t s c m] where [s] is the stack of terms to match and [c] the
-     cursor indicating where to write in the [env] array described in
-     {!module:Terms} as the environment of the RHS during matching.  [m]
-     maps the free variables contained in the tree to the free variables
-     used in this evaluation. *)
-  let rec walk tree stk cursor fresh_vars =
+  (* [walk t s c v i] where [s] is the stack of terms to match and [c] the
```
fblanqui: Using the same names in the comment and in the code helps.

src/eval.ml
```diff
@@ -187,34 +193,44 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
...
-    | Cond({ ok ; cond ; fail })                          ->
+    | Cond({ok; cond; fail})                              ->
         let next =
           match cond with
           | CondNL(i, j) ->
               if eq_modulo vars.(i) vars.(j) then ok else fail
           | CondFV(xs,i) ->
-              let xs = Array.map (fun e -> VarMap.find e fresh_vars) xs in
+              let allowed = Array.map (fun e -> IntMap.find e id_vars) xs in
+              (* FIXME some comparisons can be avoided. *)
```
fblanqui: I remember we discussed this together but, for the record, it would be nice to add some example.

src/tree.ml
```diff
@@ -44,10 +44,13 @@ let rule_order : bool Pervasives.ref = Pervasives.ref false
 type rhs = (term_env, term) Bindlib.mbinder
 
 (** Representation of a branching condition (see {!type:Terms.tree_cond}). *)
-type tree_cond = term Tree_types.tree_cond
+type tree_cond = Tree_types.tree_cond
```
fblanqui: can be inlined and the definition removed

src/tree.ml
```diff
@@ -44,10 +44,13 @@ let rule_order : bool Pervasives.ref = Pervasives.ref false
...
 
 (** Representation of a tree (see {!type:Terms.tree}). *)
-type tree = (term, rhs) Tree_types.tree
+type tree = rhs Tree_types.tree
+
+(** Type used to relate introduced variables to a unique identifier. *)
+type var_indexing = int VarMap.t
```
fblanqui: why not inlining?

src/tree.ml
```diff
@@ -651,32 +661,41 @@ let harvest : term array -> rhs -> CM.env_builder -> int -> tree =
...
   if CM.is_empty pats then Fail else
   match CM.yield pats with
-  | Yield({c_rhs ; env_builder ; c_lhs ; _}) ->
+  | Yield({c_rhs; env_builder; c_lhs ; _}) ->
       if c_lhs = [||] then Leaf(env_builder, c_rhs) else
-      harvest c_lhs c_rhs env_builder slot
-  | Condition(cond)                          ->
+      harvest c_lhs c_rhs env_builder vars_id slot
+  | Condition(cond)                        ->
+      let compile = compile count vars_id in
```
fblanqui: factorize by writing `fun count vars_id ({clauses; positions; slot} as pats) -> let compile_cv = compile count vars_id`?

src/eval.ml
gabrielhdt: It won't be visible in the generated doc, or will it be?

src/eval.ml
gabrielhdt: The question is: is an `Array.mem` faster than a `Bindlib.occur`?

src/eval.ml
fblanqui: I don't know. I am not expert in ocamldoc. But this is confusing to talk about things that are not arguments of the following function. Or else you speak of [walk] inside the comment for [tree_walk].

gabrielhdt: Indeed it won't.

src/eval.ml
fblanqui: Could you please remind me the example we discussed together? I cannot remember it right now.

rlepigre: > The question is: is an Array.mem faster than a Bindlib.occur?

Should be roughly the same.

gabrielhdt: @fblanqui, the current code evinces the redundancy, it concerns the variables held by the condition node

gabrielhdt: As of 7ad8397826e3ebc80ad17acb2175912628ad6ca9, the map on which we iterate to check for variable occurrence is freed from the variables held by the constraint

src/eval.ml
```diff
@@ -187,34 +196,56 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
...
           | CondNL(i, j) ->
               if eq_modulo vars.(i) vars.(j) then ok else fail
           | CondFV(xs,i) ->
-              let xs = Array.map (fun e -> VarMap.find e fresh_vars) xs in
+              let allowed =
+                let _, dummy = IntMap.choose id_vars in
+                Array.make (Array.length xs) dummy in
+              (* Remove variables that are allowed from the context; and fill
+                 the environment array. *)
+              let ligthened_map =
```
fblanqui: typo

src/eval.ml
```diff
@@ -163,12 +171,13 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
   let (lazy capacity, lazy tree) = tree in
   let vars = Array.make capacity Kind in (* dummy terms *)
   let bound = Array.make capacity TE_None in
-  (* [walk t s c m] where [s] is the stack of terms to match and [c] the
-     cursor indicating where to write in the [env] array described in
-     {!module:Terms} as the environment of the RHS during matching.  [m]
-     maps the free variables contained in the tree to the free variables
-     used in this evaluation. *)
-  let rec walk tree stk cursor fresh_vars =
+  (* [walk tree stk cursor vars_id id_vars] where [s] is the stack of terms to
```
fblanqui: s -> stk, c -> cursor, etc.

src/eval.ml
```diff
@@ -187,34 +196,55 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
...
+    | Cond({ok; cond; fail})                              ->
         let next =
           match cond with
           | CondNL(i, j) ->
               if eq_modulo vars.(i) vars.(j) then ok else fail
           | CondFV(xs,i) ->
-              let xs = Array.map (fun e -> VarMap.find e fresh_vars) xs in
+              let allowed =
+                (* Term variables allowed in the term. *)
+                let _, dummy = IntMap.choose id_vars in
```
fblanqui: What if id_vars is empty?

fblanqui: Why not defining a dummy var independent of id_vars?

src/eval.ml
```diff
@@ -187,34 +196,55 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
...
+                Array.make (Array.length xs) dummy in
+              let lightened_map =
```
fblanqui: Why building a map since the values of the domain are ignored in the function [fn] used in [only_allowed]?

src/eval.ml
```diff
@@ -187,34 +196,55 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
...
+                (* *Lightened* map containing variables from the context
+                   without the variables in the constraint (the [xs]). This
+                   avoids checking occurrence of variables that have been
+                   bound in the term (via [bind_mvar]).
+                   Aditionally, the [allowed] array is filled imperatively. *)
+                let c = Pervasives.ref 0 in
```
fblanqui: What is [c] for? It is used nowhere.

src/eval.ml
gabrielhdt: id_vars shouldn't be empty since constraints concern variables that appear in the term.

src/tree_types.ml
```diff
@@ -39,14 +37,14 @@ module TCMap = Map.Make(TC)
...
-type 'term tree_cond =
+type tree_cond =
   | CondNL of int * int
   (** Are the terms at the given indices convertible? We enforce the invariant
       that the first element is a point of reference, which appears in all the
       convertibility conditions involving a given non-linear variable. *)
-  | CondFV of 'term Bindlib.var array * int
-  (** Are the bound variables (which are free at the time of the checking) of
-      the term at the given index in the array? *)
+  | CondFV of int array * int
```
fblanqui: Gabriel, would you mind to permute these arguments, that is, to take `CondFV of int * int array`?

src/eval.ml
gabrielhdt: Ah indeed there is an error: it ought to be used to fill the [allowed] array, instead of the [i]. It's strange that it works with this.

src/eval.ml
gabrielhdt: Which domain? [only_allowed] doesn't ignore anything on its own.

src/eval.ml
fblanqui: I suggest to simplify the code as follows:
```
let allowed_vars = Array.map (fun id -> try IntMap.find id id_vars with Not_found -> assert false) xs in
let other_vars = IntMap.fold (fun id var other -> if Array.mem id xs then other else var::other) id_vars [] in
let no_other_vars b = not (List.exists (fun v -> Bindlib.occur v b)) in
let b = Bindlib.bind_mvar allowed (lift vars.(i)) in
if no_other_vars b
then (bound.(i) <- TE_Some(Bindlib.unbox b); ok) else
...
```

src/tree.ml
```diff
@@ -92,7 +89,7 @@ module CP = struct
     (** Set of convertibility constraints [(i,j)] with [i < j]. The constraint
         [(i,j)] is satisfied if the terms stored at indices [i] and [j] in the
         [vars] array of the {!val:Eval.tree_walk} function are convertible. *)
-    ; fv_conds : (tvar array) IntMap.t
+    ; fv_conds : (int array) IntMap.t
```
fblanqui: parentheses necessary?

src/tree.ml
```diff
@@ -607,24 +609,27 @@ module CM = struct
     List.filter (fun r -> r.c_lhs <> [||])
 end
 
-(** [harvest l r e s] exhausts linearly the LHS [l] composed only of pattern
-    variables with no constraints, to yield a leaf with RHS [r], environment
-    builder [e] completed. *)
-let harvest : term array -> rhs -> CM.env_builder -> int -> tree =
-    fun lhs rhs env_builder slot ->
+(** [harvest lhs rhs eb vi sl] exhausts linearly the LHS [lhs] composed only
```
fblanqui: eb -> env_builder, sl -> slot

src/tree.ml
```diff
@@ -651,53 +656,67 @@ let harvest : term array -> rhs -> CM.env_builder -> int -> tree =
 
 (** [compile m] translates the pattern matching problem encoded by the matrix
     [m] into a decision tree. *)
-let rec compile : CM.t -> tree = fun ({clauses ; positions ; slot} as pats) ->
+let compile : CM.t -> tree = fun m ->
+  (* [compile c vi cm] compiles clause matrix [cm] which contains [c]
```
fblanqui: -> compile count vars_id pats
(you can also change pats to cm if you prefer)

src/tree_graphviz.ml
```diff
@@ -72,19 +78,20 @@ let to_dot : string -> sym -> unit = fun fname s ->
   let output_tree oc tree =
     let pp_dotterm : dot_term pp = fun oc dh ->
       let out fmt = Format.fprintf oc fmt in
+      let var_px = "var" in
```
fblanqui: now, we can perhaps change it to something shorter like "v"

src/tree_graphviz.ml
```diff
@@ -72,19 +78,20 @@ let to_dot : string -> sym -> unit = fun fname s ->
...
       | DotSuccess           -> out "✓"
       | DotFailure           -> out "✗"
     in
-    let pp_tcstr : term tree_cond pp = fun oc cstr ->
+    let pp_tcstr : tree_cond pp = fun oc cstr ->
       let out fmt = Format.fprintf oc fmt in
       let pp_ar oc ar =
-        Format.pp_print_list Print.pp_tvar oc (Array.to_list ar)
+        Format.pp_print_list Format.pp_print_int oc
+          (Array.to_list ar)
```
fblanqui: Why not using Extra.Array.pp?

src/eval.ml
gabrielhdt: Here you traverse the [xs] array several times (with the [Array.mem]), whereas my code traverses it only once. My objective was to create the [allowed_vars] and the [other_vars] values in one traversal.

fblanqui: Perhaps but [xs] is usually very short (size <= 1 in 99% of cases). So readability and conciseness is more important here.

gabrielhdt: Fair enough.

src/eval.ml
```diff
@@ -163,12 +171,13 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
...
   let vars = Array.make capacity Kind in (* dummy terms *)
   let bound = Array.make capacity TE_None in
-  (* [walk t s c m] where [s] is the stack of terms to match and [c] the
-     cursor indicating where to write in the [env] array described in
-     {!module:Terms} as the environment of the RHS during matching.  [m]
-     maps the free variables contained in the tree to the free variables
-     used in this evaluation. *)
-  let rec walk tree stk cursor fresh_vars =
+  (* [walk tree stk cursor vars_id id_vars] where [stk] is the stack of terms
+     to match and [curesor] the cursor indicating where to write in the [vars]
```
fblanqui: curesor -> cursor

src/eval.ml
```diff
@@ -187,34 +196,47 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
...
+              let allowed =
+                (* Variables that are allowed in the term. *)
+                let fn id =
+                  try IntMap.find id id_vars with Not_found -> assert false
+                in
+                Array.map fn xs
+              in
+              (* Ensure there are no variables from [forbidden] in [b]. *)
+              let no_forbidden b =
+                let forbidden =
```
fblanqui: This is independent of b; it should be computed only once:
```
              let no_forbidden =
                 (* Term variables forbidden in the term. *)
                 let forbidden =         
                  IntMap.filter (fun id _ -> not (Array.mem id xs)) id_vars
                in
                fun b -> IntMap.for_all (fun _ x -> not (Bindlib.occur x b)) forbidden
              in
  ```

Comment (gabrielhdt) [reply src/eval.ml]: It must be computed once with the term as it is and once, in case of failure with the snf'd term

Comment (fblanqui) [reply src/eval.ml]: but forbidden does not depend on b... so it should be defined before introducing b. (it's possible that the compiler does this optimization but it's easy to do it by hand)

File: src/eval.ml [hunk 45]
```diff
@@ -187,34 +196,47 @@ and tree_walk : dtree -> stack -> (term * stack) option = fun tree stk ->
...
+                  (* Term variables forbidden in the term. *)
+                  IntMap.filter (fun id _ -> not (Array.mem id xs)) id_vars
+                in
+                IntMap.for_all (fun _ x -> not (Bindlib.occur x b)) forbidden
```
fblanqui: Why do you prefer `IntMap.for_all (fun _ x -> not (f x)) m` with O(m) calls to `not` instead of `not (IntMap.exists (fun _ -> f) m)` with 1 call to `not`?

gabrielhdt: Ah it just seemed more natural to me, but this is very personal. I'll change this.

fblanqui:
Thanks @gabrielhdt !

Issue #271, opened 2019-11-13
Put symbol properties before the name


fblanqui:
https://github.com/Deducteam/lambdapi/pull/262#issuecomment-553294233

fblanqui:
I am taking care of this.

PR #280, opened 2019-11-15
put symbol property before symbol name + other simplifications

fblanqui:
implement issue #271


fblanqui:
Any volunteer to review this PR today? I would like to merge it quickly.

src/terms.ml
```diff
@@ -12,6 +12,20 @@ open Timed
...
+  (** The symbol is constant. *)
+  | Injective
+  (** The symbol is injective. *)
+
+(** Exposition tags. *)
+type expo =
+  | Protected
+  (** Exposed but usable in LHS arguments only. *)
+  | Private
+  (** Not visible outside of its module. *)
```
rlepigre: It is a really bad idea not to include the "normal" case in these types. Going through an indirection with the option type induces extra pointer dereferencing everywhere. Just add constructors `Defin` and `Public`, respectively. Also, I would stick with shorter names to avoid longer lines in the code. The names do not have to exactly match what is in the syntax.

src/parser.ml
```diff
@@ -321,14 +321,14 @@ let parser path = m:path_elem ms:{"." path_elem}* $ -> m::ms
...
-  | _const_ -> Sym_const
-  | _inj_   -> Sym_inj
+let parser property =
+  | _constant_  -> Terms.Constant
+  | _injective_ -> Terms.Injective
 
 (** [exposition] parses the exposition tag of a symbol.*)
 let parser exposition =
-  | _protected_ -> Symex_protected
-  | _private_   -> Symex_private
```
rlepigre: I think we should only have one syntactic entity for "modifiers", including all of these properties. I originally intended for that kind of extension (hence the star that was on the `symtag` grammar).  We should have a rule of the form:
```ocaml
ms:modifier* _symbol_ s:ident al:arg* ":" a:term
```
And a corresponding type of modifiers in `src/syntax.ml` that contains all the constructors.

rlepigre:
Also, CI is broken, you should investigate that @fblanqui! :)

src/terms.ml
fblanqui: Should be fixed in 053cc3646c21fa4000d6ddfca1ab1b02f257994e .

fblanqui:
CI should be fixed by 7013c7af20f2f0e48f8ba36e615e96f4e7ecf050 .

src/parser.ml
fblanqui: I see no good reason at the moment to mix visibility markers that could apply to objects that are not symbols, and property markers that apply on function symbols only. And currently there is no need for other visibility/property markers.

fblanqui:
I still have problems with CI. What can I do locally to find out what is the problem? It seems that make is not enough.

fblanqui:
The CI problem comes from pure.ml which include tests with some invalid LP file.
It doesn't seem to be a very good idea to hard-code LP texts in strings. @rlepigre do you want to keep those tests in pure.ml?

rlepigre:
> The CI problem comes from pure.ml which include tests with some invalid LP file.
> It doesn't seem to be a very good idea to hard-code LP texts in strings.
> @rlepigre do you really want to keep those tests in pure.ml?

I did not add these tests, @ejgallego did.

src/parser.ml
rlepigre: I was thinking of always allowing them at parse time, but reporting an error when they are used in a way that is not allowed when handling the command. That way an error message could explain why some specific modifier is not allowed in certain places.

fblanqui: I understand but I see no necessity at the moment.

fblanqui:
Hello @ejgallego . I don't understand what's going on with CI and pure.ml. I hope you can help.

rlepigre:
> Hello @ejgallego . I don't understand what's going on with CI and pure.ml. I hope you can help.

There is a parse error: be careful with the syntax! It is `constant`, not `const` anymore (though I'm not sure I'm a fan of the longer version).

fblanqui:
Thanks @rlepigre I didn't do the replacement every where! :-(

doc/sections/commands.md
```diff
@@ -57,13 +57,13 @@ modifier or an exposition marker. In the following example, `const` is a
...
+const symbol zero : Nat
+const symbol succ (x:Nat) : Nat
 symbol add : Nat ⇒ Nat ⇒ Nat
-symbol const list : Nat ⇒ TYPE
-symbol const nil : List zero
-symbol const cons : Nat ⇒ ∀n, List n ⇒ List(succ n)
+const symbol list : Nat ⇒ TYPE
+const symbol nil : List zero
+const symbol cons : Nat ⇒ ∀n, List n ⇒ List(succ n)
 private symbol aux : ∀n, List n ⇒ Nat
```
rlepigre: `constant`, not `const`.

src/extra.ml
```diff
@@ -71,14 +71,21 @@ module Option =
       | (None    , None    ) -> true
       | (Some(e1), Some(e2)) -> eq e1 e2
       | (_       , _       ) -> false
+
+    (** [pp pp_elt oc o] prints on the channel [oc] the element [e] with
+       [pp_elt] if [o = Some e]. *)
+    let pp : 'a pp -> 'a option pp = fun pp_elt oc o ->
+      match o with
+      | None   -> ()
+      | Some e -> pp_elt oc e
```
rlepigre: This seems to be unused. If yes, please, remove.

src/extra.ml
```diff
@@ -71,14 +71,21 @@ module Option =
...
   end
 
 module List =
   struct
     include List
 
-    (** [pp pp_e sep oc l] prints the list [l] on the channel [oc] using [sep]
-        as separator, and [pp_e] for printing the elements. *)
+    (** [pp pp_elt sep oc l] prints the list [l] on the channel [oc] using
+       [sep] as separator, and [pp_elt] for printing the elements. *)
```
rlepigre: This change has nothing to do with the current PR and it breaks consistency with the rest of the file.

src/why3_tactic.ml
```diff
@@ -180,7 +180,8 @@ let handle : Pos.popt -> Proof.proof_state -> sig_state -> string option ->
   (* Get the meta type of the current goal (with quantified context). *)
   let trm = !((Proof.Goal.get_meta g).meta_type) in
   (* Add the axiom to the current signature. *)
-  let a = Sign.add_symbol ss.signature Const (Pos.none axiom_name) trm [] in
+  let a = Sign.add_symbol ss.signature
+            Privat Const (Pos.none axiom_name) trm [] in
```
rlepigre: This is not the way to indent a `let`-binding. Please, change to:
```
let a =
  Sign.add_symbol ss.signature Privat Const
    (Pos.none axiom_name) trm []
in
```
or make it fit on one line by defining `Pos.non axiom_name`.

src/scope.ml
```diff
@@ -257,7 +257,8 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
         fatal t.pos "[%a] is not allowed in a LHS." Print.pp Type
     | (P_Type          , _                ) -> _Type
     | (P_Iden(qid,_)   , M_LHS(_,p)       ) -> find_qid true p ss env qid
-    | (P_Iden(qid,_)   , M_Term(_,Private)) -> find_qid false true ss env qid
+    | (P_Iden(qid,_)   , M_Term(_,Privat )) ->
+        find_qid false true ss env qid
```
rlepigre: No need for a separate line here.

src/scope.ml
```diff
@@ -84,11 +84,11 @@ let find_sym : prt:bool -> prv:bool -> bool -> sig_state -> qident ->
...
     match (prt, prv, s.sym_expo) with
-    | (false, _    , Protected) ->
+    | (false, _    , Protec) ->
         if s.sym_path <> st.signature.sign_path then
           (* Fail if symbol is not defined in the current module. *)
           fatal pos "Protected symbol not allowed here."
-    | (_    , false, Private  ) ->
+    | (_    , false, Privat) ->
         fatal pos "Private symbol not allowed here."
     | _                         -> ()
```
rlepigre: The arrow is not aligned.

src/scope.ml
```diff
@@ -468,9 +469,9 @@ let scope_rule : sig_state -> p_rule -> sym * pp_hint * rule loc = fun ss r ->
   let names = Array.of_list (List.map fst map) in
   let vars = Bindlib.new_mvar te_mkfree names in
   let rhs =
-    let prv = sym.sym_expo = Private in
     let map = Array.map2 (fun n v -> (n,v)) names vars in
-    let t = scope (M_RHS(Array.to_list map, prv)) ss Env.empty p_rhs in
+    let t = scope (M_RHS(Array.to_list map, is_private sym))
+              ss Env.empty p_rhs in
```
rlepigre: This is not proper indentation.

rlepigre:
I'll do a quick last check and merge when you fix the last minor issues I just highlighted! :)

src/why3_tactic.ml
fblanqui: fixed in d7b99a6f1252f9d51d90833cc70c9ff9774a01b8

src/scope.ml
fblanqui: Fixed in d7b99a6f1252f9d51d90833cc70c9ff9774a01b8.

src/scope.ml
fblanqui: Fixed in d7b99a6f1252f9d51d90833cc70c9ff9774a01b8.

src/scope.ml
fblanqui: Fixed in d7b99a6f1252f9d51d90833cc70c9ff9774a01b8.

src/extra.ml
fblanqui: Right. But I updated the comment so that it uses the same names as the ones in the code.

src/extra.ml
fblanqui: Right but it could be useful later.

doc/sections/commands.md
fblanqui: Fixed in 28a70b39044793af24c859077a1aac32331171ee.

src/scope.ml
```diff
@@ -470,8 +469,9 @@ let scope_rule : sig_state -> p_rule -> sym * pp_hint * rule loc = fun ss r ->
   let vars = Bindlib.new_mvar te_mkfree names in
   let rhs =
     let map = Array.map2 (fun n v -> (n,v)) names vars in
-    let t = scope (M_RHS(Array.to_list map, is_private sym))
-              ss Env.empty p_rhs in
+    let t =
+      scope (M_RHS(Array.to_list map, is_private sym))
+        ss Env.empty p_rhs in
```
rlepigre: The `in` should be on its own line.

src/why3_tactic.ml
```diff
@@ -180,8 +180,9 @@ let handle : Pos.popt -> Proof.proof_state -> sig_state -> string option ->
   (* Get the meta type of the current goal (with quantified context). *)
   let trm = !((Proof.Goal.get_meta g).meta_type) in
   (* Add the axiom to the current signature. *)
-  let a = Sign.add_symbol ss.signature
-            Privat Const (Pos.none axiom_name) trm [] in
+  let a =
+    Sign.add_symbol ss.signature
+      Privat Const (Pos.none axiom_name) trm [] in
```
rlepigre: Same here.

src/extra.ml
rlepigre: Well, it does not use the same name as the rest of the file. You should rather have used the same name in your new code!

src/extra.ml
rlepigre: If you really want to keep that function (and I think we should not maintain dead code), then please fix the indentation in the comment.

src/extra.ml
rlepigre: Please, fix the indentation in the comment.

src/extra.ml
```diff
@@ -212,10 +219,15 @@ module Array =
         let f x y = if not (f x y) then raise Done in
         try iter2 f a1 a2; true with Done -> false
 
-    (** [pp pp_e sep oc a] prints the array list [a] on the channel [oc] using
-        [sep] as separator, and [pp_e] for printing the elements. *)
+    (** [pp pp_elt sep oc a] prints the array list [a] on the channel [oc]
+        using [sep] as separator, and [pp_e] for printing the elements. *)
     let pp : 'a pp -> string -> 'a array pp = fun pp_elt sep oc a ->
-      List.pp pp_elt sep oc (to_list a)
+      let n = Array.length a in
```
rlepigre: I think the following would be slightly more readable (and one line shorter):
```ocaml
let n = Array.length a in
if n > 0 then pp_elt oc a.(0);
for i = 1 to n - 1 do
  Format.fprintf oc "%s%a" sep pp_elt a.(i)
done
```

rlepigre: But for the record, I think that using `List.pp` was fine. The complexity of the function was the same, and it's not like input/output functions are critical for efficiency here.

PR #281, opened 2019-11-15
[lsp] add Makefile

Rehan-MALAK:
- cf https://github.com/coq-community/vscoq/blob/master/Makefile
- installed symbolic link is now called ~/.vscode/extensions/lp-vscode


Rehan-MALAK:
I think the only thing to keep from (closed) #260

fblanqui:
Thanks @Rehan-MALAK

Issue #275, opened 2019-11-14
Rules where the LHS is not a pattern


francoisthire:
The last version of Lambdapi accets the rewrite rule in the example below but it cannot be used. This can be seen when someone prints the decision tree with the option `--write-tree`.

```
symbol const A : TYPE

symbol const a : A

symbol f : (A ⇒ A) ⇒ A

rule f (λ x, &X &Y) → a

symbol x : A ⇒ A

symbol y : A

compute f (λ z, (x y))
```

Hence, is that kind of rewrite rule should be accepted? Is this normal that it does not compute?

rlepigre:
I guess that since we introduced decision trees that kind of rule cannot be used anymore. (We lost the syntactic matching feature that I liked.)

We should now probably reject the patterns that cannot be handled after scoping.

francoisthire:
Once could recover this feature using dkmeta which exists only for Dedukti currently. I have to admit that from my point of view, syntactic pattern matching has no ther purpose that meta programmation so this is not a big loss.

fblanqui:
Not sure that we should accept this kind of rule as this breaks many properties. See counter-example 2.3 in http://doi.org/10.4230/LIPIcs.RTA.2010.17 . Anyway decision trees should be able to handle that. @gabrielhdt should have a look at it to see what's wrong and perhaps fix it.

GuillaumeGen:
To store a discussion I had with @francoisthire, with β, the rule `f (λ x, &A &B) -> &B` makes the conversion trivial, since for all `t` and `u`, `t <- f (λ x, (λ y, 0) t) ->β f (λ x, 0) <-β f (λ x, (λ y, 0) u) -> u`.

rlepigre:
> Anyway decision trees should be able to handle that

No they cannot handle that case because there is no constructor to mach on in the body of the function.

fblanqui:
The generated graph has an edge labeled "λtr". This "tr" is really strange. Why do we get "tr"?

fblanqui:
We need to merge #262 before fixing this issue to avoid conflicts.

gabrielhdt:
For the above comment, see the PR #279.

fblanqui:
Rejecting non-patterns has been added in https://github.com/Deducteam/lambdapi/pull/288 since current decision trees cannot handle non-patterns. We do not plan to change this for the moment. So I close this issue.

PR #283, opened 2019-11-16
[WIP] Rejecting ill formed patterns

gabrielhdt:
In answer to #275, a procedure to check whether a term is a valid pattern is introduced.

For now, a pattern is rejected if it is a pattern variable applied to another pattern variable.


src/basics.ml
```diff
@@ -240,6 +240,21 @@ let distinct_vars : term array -> bool =
     | _ -> raise Not_unique_var
   in try Array.iter check ts; true with Not_unique_var -> false
 
+(** [is_pattern t] tells whether term [t] is a valid pattern. *)
+let rec is_pattern : term -> bool = fun t ->
```
rlepigre: I think what we want is to reject rules in which a pattern variable is applied to something.
To do that, you just need to use `Basics.get_args`, and check that the argument list is empty if the head is a pattern variable. And then call the checking function recursively on the arguments if that is not the case. I don't remember if that's enough though. You might want to lookup the exact definition of Miller patterns.

tests/KO/275.lp
```diff
@@ -0,0 +1,8 @@
+symbol const A : TYPE
+symbol const a : A
+symbol f : (A ⇒ A) ⇒ A
+rule f (λ x, &X &Y) → a
+ and f (&X &Y) → a
```
rlepigre: This line is not tested because the previous one will fail I guess. It might be a good idea to have several test files.

src/basics.ml
gabrielhdt: We could allow something like `&F x y` if `x` and `y` are two bound variables. But we can enforce writing `F[x,y]`.

gabrielhdt:
François also pointed out https://github.com/Deducteam/Dedukti/issues/97, should we give it a try?

src/basics.ml
gabrielhdt: Ah but it breaks compatibility with legacy syntax.

rlepigre: The idea was always to compile applications of symbol `F` to distinct bound variables `x` and `y` (in the legacy syntax) as `&F[x,y]`, not as `&F x y`.

rlepigre: If that is not what is done, that's a bug.

gabrielhdt: OK, I'll have a look.

gabrielhdt: It looks like there is a bug indeed, if I print the obtained lhs in `scope.ml`, I obtain, on `55.dk`
```
h &aa (λx, &aa x)
h (λx, &aa x) &aa
```

fblanqui: No, that's normal. In the dk file, aa is applied once to no argument, and another time to one argument. In this case, we replace aa by a pattern variable &aa of the smallest arity (0 here). Changing this would require to eta-expand some patterns.

fblanqui:
https://github.com/Deducteam/Dedukti/issues/97 doesn't seem very urgent. It would be better to do it in another PR the day we really need it. You can add an issue about that for the record.

src/basics.ml
```diff
@@ -240,6 +240,16 @@ let distinct_vars : term array -> bool =
...
+  match get_args (unfold t) with
+  | (Patt(_), []  ) -> true (* Pattern variables are not applied *)
+  | (Patt(_), args) -> distinct_vars (Array.of_list args)
+  | (Abst(_,b), []) ->
+      let _, b = Bindlib.unbind b in
+      is_pattern b
```
fblanqui: all this can be written on one line

src/basics.ml
```diff
@@ -240,6 +240,16 @@ let distinct_vars : term array -> bool =
     | _ -> raise Not_unique_var
   in try Array.iter check ts; true with Not_unique_var -> false
 
+(** [is_pattern t] tells whether term [t] is a valid pattern. *)
+let rec is_pattern : term -> bool = fun t ->
+  match get_args (unfold t) with
```
fblanqui: no need for unfold, it is done by get_args

src/basics.ml
```diff
@@ -240,6 +240,16 @@ let distinct_vars : term array -> bool =
...
+  | (Patt(_), []  ) -> true (* Pattern variables are not applied *)
+  | (Patt(_), args) -> distinct_vars (Array.of_list args)
+  | (Abst(_,b), []) ->
+      let _, b = Bindlib.unbind b in
+      is_pattern b
+  | (_      , args) -> List.for_all is_pattern args
```
fblanqui: don't you need to check that you have a function symbol at the head? can your trees handle a pattern like (\x, x &F)?

src/basics.ml
```diff
@@ -240,6 +240,16 @@ let distinct_vars : term array -> bool =
     | _ -> raise Not_unique_var
   in try Array.iter check ts; true with Not_unique_var -> false
 
+(** [is_pattern t] tells whether term [t] is a valid pattern. *)
+let rec is_pattern : term -> bool = fun t ->
+  match get_args (unfold t) with
+  | (Patt(_), []  ) -> true (* Pattern variables are not applied *)
+  | (Patt(_), args) -> distinct_vars (Array.of_list args)
```
fblanqui: should be `| (Patt(_,_,ts), []) -> distinct_vars ts`

src/basics.ml
```diff
@@ -240,6 +240,16 @@ let distinct_vars : term array -> bool =
     | _ -> raise Not_unique_var
   in try Array.iter check ts; true with Not_unique_var -> false
 
+(** [is_pattern t] tells whether term [t] is a valid pattern. *)
+let rec is_pattern : term -> bool = fun t ->
+  match get_args (unfold t) with
+  | (Patt(_), []  ) -> true (* Pattern variables are not applied *)
```
fblanqui: to be removed

src/basics.ml
rlepigre: Eta-expanding some patterns would seem more reasonable to me, since that would follow the Lambdapi approach more closely. Are we even sure that matching behaves correctly on such patterns?

src/basics.ml
rlepigre: Some of the checks are already done in scoping. Maybe that would be a better place to extend the tests.

src/basics.ml
fblanqui: Let's keep it as it is now. There is no need to change that at the moment.

src/basics.ml
gabrielhdt: This covers the case `&F[x, y]`, but not the case `&F x y` which apparently can happen.

src/basics.ml
```diff
@@ -242,13 +242,11 @@ let distinct_vars : term array -> bool =
...
-  match get_args (unfold t) with
-  | (Patt(_), []  ) -> true (* Pattern variables are not applied *)
-  | (Patt(_), args) -> distinct_vars (Array.of_list args)
-  | (Abst(_,b), []) ->
-      let _, b = Bindlib.unbind b in
-      is_pattern b
-  | (_      , args) -> List.for_all is_pattern args
+  match get_args t with
+  | (Patt(_,_,ts), []  ) -> distinct_vars ts
+  | (Patt(_)     , args) -> distinct_vars (Array.of_list args)
```
rlepigre: I would have a FIXME after the definition of the function about this. I think we should disallow patterns like `&X x y` in the future.

src/basics.ml
```diff
@@ -242,13 +242,11 @@ let distinct_vars : term array -> bool =
...
+  | (Abst(_,b)   , []  ) -> is_pattern (snd (Bindlib.unbind b))
```
rlepigre: Here you also need to add (I think):
```ocaml
| (Abst(_,b)   , _   ) -> fatal_no_pos "Pattern not in beta-normal-form."
```

(I think that this checking code should be part of scoping so that we have positions to report.)

fblanqui:
What is the tree you get with a pattern like \x,\y,&F x y?

Le 18/11/2019 à 13:25, Gabriel Hondet a écrit :

gabrielhdt:
An unfinished tree...
tree f

rlepigre:
Yeah, so I guess we really need to make sure to reject patterns like `&X x y`, and do some eta-expansion to handle the legacy syntax.

fblanqui:
I think that the only change that needs to be done to the current code (in addition to fixing Syntax.get_args) is to add before https://github.com/Deducteam/lambdapi/blob/589f83bf23a1b144f2e216761801f4586d5d29cb/src/scope.ml#L187:
```
(* Check that pattern variables are not applied. *)
begin
  match h.elt with
  | P_Patt(_,_) -> if args <> [] then fatal t.pos "Pattern variables cannot be applied."
  | _ -> ()
end;
```

fblanqui:
@rlepigre do you agree? It should take just a few minutes to change this and close this issue.

rlepigre:
> @rlepigre do you agree? It should take just a few minutes to change this and close this issue.

I'm not completely sure, and I don't have time to look in more details. However, I stick with what I said earlier. It is not a good idea to keep compiling the legacy patterns in the wrong way. The matching procedure was not thought with that idea in mind, and we might well be breaking invariants that I had in mind while designing the core of the system.

fblanqui:
I'm working on this right now. I shall submit a new PR soon.

fblanqui:
PR subsumed by https://github.com/Deducteam/lambdapi/pull/288.

PR #288, opened 2019-11-22
check that LHS are well-formed + eta-expand legacy pattern variables

fblanqui:


src/menhir_parser.mly
```diff
@@ -59,31 +60,30 @@ let translate_old_rule : old_p_rule -> p_rule = fun r ->
...
                 fatal a.pos "Annotation in legacy pattern."
             | [([Some x],None   ,_)] ->
                 compute_arities (x.elt::env) t
-            | _                      ->
-                fatal h.pos "Invalid legacy pattern lambda."
+            | _                      -> assert false
           end
-      | P_Patt(_,_)     -> fatal h.pos "Pattern in legacy rule."
       | P_Impl(_,_)     -> fatal h.pos "Implication in legacy pattern."
-      | P_LLet(_,_,_,_) -> fatal h.pos "Let expression in legacy rule."
```
gabrielhdt: There is a let in legacy Dedukti, so perhaps we should keep the error message for it

fblanqui: fixed in c73b6ac

fblanqui:
Why applying a pattern in a RHS would be a problem?

gabrielhdt:
No indeed, I made a mistake.

fblanqui:
Hello @rlepigre . Do you agree with this PR?

src/menhir_parser.mly
```diff
@@ -59,31 +60,30 @@ let translate_old_rule : old_p_rule -> p_rule = fun r ->
             begin
               try
                 let n = Hashtbl.find arity x in
-                if nb_args < n then Hashtbl.replace arity x nb_args
+                if nb_args > n then Hashtbl.replace arity x nb_args
               with Not_found -> Hashtbl.add arity x nb_args
             end
       | P_Wild          -> ()
-      | P_Type          -> ()
```
rlepigre: That change should be reverted, that error should be caught at scoping time (which it is).

src/menhir_parser.mly
```diff
@@ -59,31 +60,30 @@ let translate_old_rule : old_p_rule -> p_rule = fun r ->
...
-      | P_Wrap(_)       -> fatal h.pos "Wrapping constructor in legacy rule."
-      | P_Expl(_)       -> fatal h.pos "Explicit argument in legacy rule."
+      | P_Meta(_,_)
+      | P_Patt(_,_)
+      | P_NLit(_)
+      | P_UnaO(_,_)
+      | P_BinO(_,_,_)
+      | P_Wrap(_)
+      | P_Expl(_)       -> assert false
     end;
```
rlepigre: The changes done here are not a great idea: why remove all these decent error messages? Yes they may never happen, but a nice error message is better than an assert failure. Especially in the middle of this disgusting code that is very error-prone. Plus, this has nothing to do with the current PR.

src/scope.ml
```diff
@@ -180,9 +192,9 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
   (* Toplevel scoping function, with handling of implicit arguments. *)
   let rec scope : env -> p_term -> tbox = fun env t ->
     (* Extract the spine. *)
-    let (p_head, args) = Syntax.get_args t in
+    let (p_head, args) = get_args t in
     (* Scope the head and obtain the implicitness of arguments. *)
-    let h = scope_head env p_head in
+    let h = scope_head env p_head args in
```
rlepigre: Here it is weird to pass an extra argument to `scope_head`. What I would do instead is add a line with something like the following (where `is_patt` is similar to `Basics.is_symb`):
```ocaml
if is_patt h && args <> [] then
  fatal t.pos "Pattern variables cannot be applied.";
```
Note that the position would be better because it would correspond to the whole application as well.

src/scope.ml
```diff
@@ -164,6 +164,18 @@ let get_implicitness : p_term -> bool list = fun t ->
   in
   get_impl t
 
+(** [get_args t] decomposes the parser level term [t] into a spine
+   [(h,args)], when [h] is the term at the head of the application and [args]
+   is the list of all its arguments. WARNING: it does not decompose P_LLet,
+   P_UnaO, P_BinO and P_NLit. *)
```
rlepigre: The indentation of the comment is bad, why did you not keep the one we had?
Also, the constructor names are not in brackets. I suggest using the following (which avoids listing constructors because that list may evolve):
```ocaml
(** [get_args t] decomposes the parser level term [t] into a spine [(h,args)],	
    when [h] is the term at the head of the application and [args] is the list	
    of all its arguments. Note that sugared applications (e.g., infix symbols)
    are not expanded, so [h] may still be unsugared to an application. *)
```

src/menhir_parser.mly
```diff
@@ -100,13 +100,31 @@ let translate_old_rule : old_p_rule -> p_rule = fun r ->
...
+             let qid = Pos.make p ([], prefix ^ string_of_int (i-k)) in
+             Pos.make p (P_Iden (qid, false))
+         in
+         let args = Array.init max arg in
+         let patt = Pos.make p (P_Patt(Pos.make h.pos x, args)) in
+         let rec vars i =
+           if i >= nb_exp then []
+           else Some (Pos.make p (prefix ^ string_of_int i)) :: vars (i+1)
+         in
+         Pos.make t.pos (P_Abst([vars 0, None, false], patt))
```
rlepigre: I'll review that part later, I need to open the file for that (I don't have time now).

rlepigre: I can already say that there can be some unintended variable capture here: you should use special name, not accessible to the user for the bound variables here. For example you could take `"@x"` for the prefix. Or at least check for collisions when following your naming convention (skipping the indices that are already used).

src/menhir_parser.mly
fblanqui: Why this one should be caught at scoping time and not P_Prod?

src/menhir_parser.mly
fblanqui: Because if this assertion fails then there is a bug. It's not a user error.

src/menhir_parser.mly
rlepigre: Well, it would be best to also handle `P_Prod` at scoping time, but in that case the code for the translation is annoying to write. In any case, my main point was that these changes are not necessary.

src/menhir_parser.mly
rlepigre: Well, using the legacy syntax is a user error as far as I'm concerned... :)

Frankly, it is not a big deal. I think the previous code was better, but if you want to keep it that way then at least add an `assert false` for every matched pattern. You want to known which pattern triggered the error in the case of a bug.

src/menhir_parser.mly
```diff
@@ -40,10 +41,10 @@ let translate_old_rule : old_p_rule -> p_rule = fun r ->
     (if !verbose > 1 then Option.iter fn ao); x
   in
   let ctx = List.map get_var ctx in
-  (* Find the minimum number of arguments a variable is applied to. *)
   let is_pat_var env x =
     not (List.mem x env) && List.exists (fun y -> y.elt = x) ctx
   in
+  (* Find the maximum number of arguments a variable is applied to. *)
```
rlepigre: This comment does not make more sense here as it did above. Again, please, avoid unnecessary changes.

src/scope.ml
fblanqui: Fixed in 5b0793f

src/scope.ml
fblanqui: To change the indentation please push directly on my branch.

src/menhir_parser.mly
fblanqui: Fixed in cd1a1fd

src/menhir_parser.mly
fblanqui: You are right. I am not in favor of using "@..." because this is not parsable.

fblanqui: Should be fixed in 376bf6e

fblanqui:
@rlepigre do you have any new comment?

rlepigre:
> @rlepigre do you have any new comment?

I'm gonna push some simplification, but I'd like to test first. What file did you use to check that eta-expansion works?

fblanqui:
There is tests/OK/55.dk:
```
18:59 ~/lambdapi (pattern) l tests/OK/55.dk
A : Type.
def h : (A -> A) -> (A -> A) -> A -> A.
[aa] h aa (x => aa x) --> aa.
[aa] h (x => aa x) aa --> aa.

18:59~ ~/lambdapi (pattern) lp --beautify tests/OK/55.dk
constant symbol A : TYPE
symbol h : (A ⇒ A) ⇒ (A ⇒ A) ⇒ A ⇒ A
rule h (λ v0, &aa[v0]) (λ x, &aa[x]) → λ v0, &aa[v0]
rule h (λ x, &aa[x]) (λ v0, &aa[v0]) → λ v0, &aa[v0]
```

We should probably add more.

rlepigre:
I'm good for merging once the CI has passed.

Issue #274, opened 2019-11-14
option --version is buggy


francoisthire:
lambdapi --version prints

```
Lambdapi %%VERSION%%
```

rlepigre:
That's not a bug, that's a feature of Dune. Only releases have their version substituted.

francoisthire:
Well If I do `make install` I expect to have something better than `%%%VERSION%%%`.

francoisthire:
This is very frustrating because right now there is no easy way to know the current version of `lambdapi`.

rlepigre:
I agree, I'm as pissed of as you are by how crappy dune is. I would expect a version number like `devel-<commit hash>`, but apparently that's not possible.

rlepigre:
@francoisthire: I just discovered that if you install using `opam pin` then you get the version number.

ejgallego:
You need to use the `build-info` library https://dune.readthedocs.io/en/stable/dune-libs.html#build-info

rlepigre:
I find the idea of carrying build-system dependent code around in my project kind of worrying...

But anyway, is there any doc available for that library? I can't even find its source code in the middle of the whole dune repository. I'm hoping that it also provides installation paths.

ejgallego:
> I find the idea of carrying build-system dependent code around in my project kind of worrying... But thanks @ejgallego! I did not know that already existed.

The version problem is not an easy problem; in particular it is hard to maintain a proper incremental build. You don't the binary rebuilt on every hash change.

You can use `build-info` or else you can roll your own wrapper to `git describe` too, but then you have to be very careful as to avoid the rebuild problem.

Code for build-info is here BTW https://github.com/ocaml/dune/tree/master/otherlibs/build-info

rlepigre:
OK, do since build-info does not allow to do what I want, I don't think it is worth using it. I'm personally OK with the current situation where releases get a version number, and you also get a version number if you pin the repository. I'd rather not implement the version stuff myself.

As for the other limitations, I think I'm gonna have to hack around.

ejgallego:
> since build-info does not allow to do what I want

What would you like it to do? At least for the original request in this issue, it should work.

rlepigre:
Yeah, it would. But I also want to know where the `lib` directory used for the installation is (i.e., the result of `opam var lib`).

ejgallego:
[...]

PR #289, opened 2019-11-25
File management and module mapping improvement

rlepigre:
This PR adresses #243 and several other previous limitations.

Here is the rough idea of the new system (though the work is not finished):
 - There is a fixed "library root" directory that default to some reasonable place under the `~/.opam` directory  (or some system-wide directory if opam is not used).
 - The root directory can be altered using the `--lib-root` flag. This is important because the root directory should always be a valid directory, and that is typically not the case prior to installation. So we must now use something like `dune exec -- lambdapi check --lib-root lib` so that the `lib` directory in the source tree is considered the library root. Note that to recover the previous behavior we must run with `.` as the root.
 - There is the `--map-dir` option that can be used to map a module path to a specific directory (that may or may not be under the library root). For instance, we might do `--map-dir matita:libraries/matita` so that the files under the `libraries/matita` folder correspond to module paths starting with `matita` and then continuing with the underlying directory structure.
 - There is no more limitation to relative paths. There is now a piece of C code that calls `realpath` to get an absolute file name. Note that symbolic links are removed, so we cannot use a symbolic link to include a sub-tree under some mapped module path.
 - This PR subsumes #269, this change is useful for several good reasons. The main advantage is that the initialisation code is forced to be consistent, and so the server now knows about the path for the installed libraries.

Things that remain to be done:
 - [x] More testing with corner cases.
 - [x] Make sure that the libraries still compile, possibly cleaning up the scripts.
 - [x] Create a minimal standard library under `lib/std`, `lib` being intended as the library root. As a consequence, the modules of the standard library would be accessed with `std.bool` for example (file `lib/std/bool.lp`). The `lib` directory should then be installed as the library root of the system.
 - [x] Make sure that the LSP server still works (it is probably broken now since some initialization code is missing).
 - [x] Fix the documentation and ensure that the editors work with the right binary. Now it needs to be something like `lambdapi lsp --standard-lsp`.
 - [x] Write documentation for the module system and the intended installation procedure for third-party libraries.


fblanqui:
Thanks @rlepigre for your PR. I don't understand why you need a --lib-root option. It seems to me that the --map option subsumes it.

rlepigre:
No, it is not subsumed because it would mean being able to map an "empty" module path, which is not the case. Plus, the library root has a very particular role: every installed library are aimed to be installed under it on the system.

src/files.ml
```diff
@@ -3,6 +3,211 @@
...
+    let empty = Node(None, StrMap.empty)
+
+    exception Already_mapped
+
+    let set_root path (Node(po, m)) =
+      match po with
+      | None    -> Node(Some(path), m)
+      | Some(_) -> raise Already_mapped
+
+    let rec singleton ks path =
```
fblanqui: What's the type of this function and what is it doing?

rlepigre: The local module has a signature.

src/files.ml
```diff
@@ -3,6 +3,211 @@
...
+    (** [get mp m] obtains the file path corresponding to the module path [mp]
+        in [m] (with no particular extension). The exception [Root_not_set] is
+        raised if the root of [m] was not previously set with [set_root]. *)
+    val get : Path.t -> t -> file_path
+
+    (** [iter fn m] calls function [fn] on every binding stored in [m]. *)
+    val iter : (Path.t -> file_path -> unit) -> t -> unit
+  end =
+  struct
+    type t = Node of file_path option * t StrMap.t
```
fblanqui: The Node constructor seems useless. Why do you use it?

rlepigre: It cannot be left out without the `-rectypes` option.

rlepigre: And it does not cost anything.

src/files.ml
rlepigre: And this function is not part of the interface (it is only an auxiliary function).

fblanqui:
It's not clear to me. Please provide an example.

src/files.ml
fblanqui: Ok but I don't understand what you try to represent and why it is so complicated. An example would be useful here.

rlepigre: Well, it's just a tree, with an optional directory path on each node, and with strings (or rather, module path members) on its branches. But I'll document more later anyway since for now this is still a work in progress.

rlepigre:
> It's not clear to me. Please provide an example.

So, let us assume that we have some library root directory `<root>`, and that we also have the following mappings:
 - `foo.bar:<dir1>` (i.e., module path `foo.bar` is mapped to directory `<dir1>`),
 - `baz:<dir2>` (i.e., module path `baz` is mapped to directory `<dir2>`).

Globally, this is represented by the following tree:
```
Some(<root>)
├─foo── None ──bar── Some(<dir1>)
└─baz── Some(<dir2>)
```

The search for files corresponding to different module paths would then yield:
 - `foo.bar.baz.mod` mapped to `<dir1>/baz/mod`,
 - `a.b.c` mapped to `<root>/a/b/c`,
 - `foo.baz.mod` mapped to `<root>/foo/baz/mod`.

There are several important things here:
 - This process does not fail as long as the root of the tree contains `Some(_)` (this requirement is explicitly stated in the actual lookup code). This is enforced when calling `init_lib_root`, which registers the root.
 - We fall back on the last `Some(_)` encountered during the tree traversal to pick a base directory, and we then complete it by taking the structure of the module path as the directory sub-structure. This is what happens with the `foo.baz.mod`: there is no branch with the `baz` label under first branch, so we fall back to the `<root>` directory.

fblanqui:
Why not taking the current directory as the root? I was personally thinking in something different:

lambdapi --map r1:d1 .. --map rk:dk file.lp

where ri is a non-qualified ident called a root and di is an arbitrary path

some roots could also be given through an environment variable $DK_PATH

there is also an implicit root r0 mapped to the current directory

then in file.lp we can have:

require a.b // refers to ./a/b.lp
require ri:a.b // refers to ri/a/b.lp

rlepigre:
The root is an actual root: it is the entry point of the module system. So it does not make sense to have the current directory be the root since installed libraries are not installed under the current directory. Of course, the current directory is special, and it should be mapped under a module path that is chosen for the project (maybe automatically by looking at the name of the current directory name, or using a meta data file). I insist that what I proposed here is still work in progress, so everything is not optimal for usability yet.

What you propose could maybe work, but I don't see that as a standard approach. (I did not really think your proposal through as much as I did mine, so it's hard for me to understand its limitations.) With what I propose, we will be prepared for having proper tooling (in the line of what Rust has with Cargo), and it also accounts for installation of libraries. Moreover, I don't really like adding yet another syntactic element (your roots). These are not useful since they can just be represented using a normal module path. The location of the corresponding files alone can tell whether a module path corresponds to a library or the a local module.

fblanqui:
Why do you want to map module paths to directories and not just roots?

fblanqui:
How to handle locally installed libraries with your system?

rlepigre:
> Why do you want to map module paths to directories and not just roots?

I don't understand that question: what you call "roots" are directories right?

> How to handle locally installed libraries with your system?

What do you mean by "locally installed libraries"?

fblanqui:
In a previous example, you proposed to map foo.bar to some directory. Why would you like to map foo.bar and not just foo?

fblanqui:
By locally installed, I mean "manually" installed, that is, not installed using a system like opam.

rlepigre:
> In a previous example, you proposed to map foo.bar to some directory. Why would you like to map foo.bar and not just foo?

Oh, I see. The reason is simple: we may want to have a more complex structure in our libraries that just a flat one. For example, we may want to have the following structure in our installed libraries:
- `std` is the entry point of the standard library, with e.g., `std.bool`, `std.nat` and `std.list` modules.
- `std.list` could contain some modules for algorithms on lists. For example, we could have modules `std.list.sorting.qsort` and `std.list.sorting.msort` for quick- and merge-sort.
- `imported` containing imported libraries like `imported.matita` or `imported.verine`.

It just gives more options for placing things in relevant directories.

> By locally installed, I mean "manually" installed, that is, not installed using a system like opam.

Ideally, it should be impossible to install a library "manually", otherwise you get the same mess as with OCaml. But in any case, the approach I propose enforces that libraries are installed under the system's library root. And there is no difference whether they are installed "manually" or automatically. It is just a convention that everyone must respect.

fblanqui:
Do you have an example of use of mapping std to dir1 and std.list to a directory that is not a sub-directory of dir1 (even transitively)?

rlepigre:
> Do you have an example of use of mapping std to dir1 and std.list to a directory that is not a sub-directory of dir1 (even transitively)?

For `std` no, because I guess that the standard library would be a single package. But for `imported`, the example that I gave above makes sense.

fblanqui:
Isn't it a problem to enforce the use of a system like opam? What if the user wants to experiment with some libraries he modified and compiled somewhere in its home directory, for instance, in some git repos?

src/files.ml
```diff
@@ -3,6 +3,211 @@
...
+  with Sys_error(_) | Invalid_argument(_) ->
+    exit_with "Invalid library root (no such file or directory [%s])." path
+
+(** [default_lib_root ()] returns the default library root curently configured
+    for the system. It depends on the current Opam switch (if any), and it may
+    or may not correspond to an existing directory. *)
+let default_lib_root : unit -> string = fun _ ->
+  match run_process "opam var lib" with
+  | Some([l]) -> Filename.concat l "lambdapi/lib_root"
+  | _         -> "/usr/local/lib/lambdapi/lib_root"
```
fblanqui: Isn't it a problem to depend on some external system like opam?

fblanqui: Or to hard-code paths like "/usr/local/lib/lambdapi/lib_root" which is probably OS dependent?

rlepigre: > Isn't it a problem to enforce the use of a system like opam? What if the user wants to experiment with some libraries he modified and compiled somewhere in its home directory, for instance, in some git repos?

Well, that's what the `--map` option is for!

In any case, enforcing the use of a proper tool (nothing like opam, which is crap) is great. It makes everything simpler and more uniform, and the user does not have to care about completely useless details. And also, everything is predictable.

rlepigre:
> Isn't it a problem to enforce the use of a system like opam? What if the user wants to experiment with some libraries he modified and compiled somewhere in its home directory, for instance, in some git repos?

Well, that's what the `--map` option is for!

In any case, enforcing the use of a proper tool (nothing like opam, which is crap) is great. It makes everything simpler and more uniform, and the user does not have to care about completely useless details. And also, everything is predictable.

src/files.ml
rlepigre: Note that the installation of our `lambdapi` binary also depends on opam! There is nothing I can do about that.

ejgallego: That's an optional dependency indeed. I'd recommend a variation on set of rules dune uses, c.f. https://github.com/ocaml/dune/pull/1226 so for example:

```
    let opam_prefix = Env.get env "OPAM_SWITCH_PREFIX" in
```
can be used for opam2, and also you allow the user to override that using a config step.

ejgallego: > Note that the installation of our `lambdapi` binary also depends on opam! There is nothing I can do about that.

What do you mean? It doesn't depend on OPAM AFAICS; for emacs files you could just add the proper `(install ...)` rules and `dune install` would take care of it.

rlepigre: Well, what I meant is that everything is gonna be installed under `opam` since dune uses its data. And since `dune` is not yet able to give me access to the location of the `lib` directory at compile time I don't have many choices here. I would indeed prefer to pick the `lib` directory during a config step.

rlepigre: > That's an optional dependency indeed. I'd recommend a variation on set of rules dune uses, c.f. [ocaml/dune#1226](https://github.com/ocaml/dune/pull/1226) so for example:
> 
> ```
>     let opam_prefix = Env.get env "OPAM_SWITCH_PREFIX" in
> ```
> 
> can be used for opam2, and also you allow the user to override that using a config step.

I considered that, but is the `lib` directory always fixed with respect to the prefix?

ejgallego: For all practical purposes yes.

rlepigre: > Or to hard-code paths like "/usr/local/lib/lambdapi/lib_root" which is probably OS dependent?

Well, that's a reasonable directory for POSIX systems. Any other systems can't even handle OCaml so I think we are good.

Anyway, that is not the most pressing kind of issue here. I'm trying to make things work on most systems that we are interested in, and I can't solve every problem in the world (especially not at once). So I'm making what I see as reasonable assumptions, and we will always be able to make things better later.

src/console.ml
```diff
@@ -35,15 +35,19 @@ let out_fmt = Pervasives.ref Format.std_formatter
 (** [err_fmt] warning/error output formatter. *)
 let err_fmt = Pervasives.ref Format.err_formatter
 
+(** [no_wrn] disables warnings when set to [true]. *)
+let no_wrn = Pervasives.ref false
+
 (** [wrn popt fmt] prints a yellow warning message with [Printf] format [fmt].
```
fblanqui: Isn't it [Format] instead of [Printf]?

src/files.ml
```diff
@@ -3,6 +3,211 @@
...
+
+    (** [pp oc mp] prints [mp] to channel [oc]. *)
+    let pp : module_path pp = fun oc mp ->
+      Format.pp_print_string oc (String.concat "." mp)
+  end
+
+(** Functional maps with module paths as keys. *)
+module PathMap = Map.Make(Path)
+
+(** Synonym of [string] for file paths. *)
```
fblanqui: Wouldn't be better before [Path] module definition?

src/files.ml
```diff
@@ -12,52 +217,47 @@ let obj_extension : string = ".lpo"
...
+  (* Finding the best mapping under the root. *)
+  let mapping = ref None in
+  let fn mp path =
+    if String.is_prefix path base then
+      match !mapping with
+      | None       -> mapping := Some(mp, path)
+      | Some(_, p) -> if String.length p < String.length path then
+                        mapping := Some(mp, p)
+  in
+  ModMap.iter fn (current_mappings ());
```
fblanqui: Can't you define a more efficient ModMap.find function that would not go through all mappings?

src/files.ml
fblanqui: Some times you use string, and sometimes file_path. Why not using string (or file_path) everywhere?

src/files.ml
fblanqui: It would be nice to add the example of https://github.com/Deducteam/lambdapi/pull/289#issuecomment-558197206 here.

fblanqui:
The update of doc/ is missing.

src/extra.ml
```diff
@@ -36,6 +36,11 @@ module String =
...
       is_sub 0
 
+    let is_prefix : string -> string -> bool = fun p s ->
+      let len_p = String.length p in
+      let len_s = String.length s in
+      len_p <= len_s && String.sub s 0 len_p = p
+
     let for_all : (char -> bool) -> string -> bool = fun p s ->
       let len_s = String.length s in
       let rec for_all i = i >= len_s || (p s.[i] && for_all (i+1)) in
```
fblanqui: Not related to this PR but could be slightly improved:
```
let rec for_all i = i < 0 || (p s.[i] && for_all (i-1)) in for_all (String.length s - 1)
```

src/files.ml
rlepigre: > Wouldn't be better before [Path] module definition?

No, I dont think so.

> Some times you use string, and sometimes file_path. Why not using string (or file_path) everywhere?

I'm only using `file_path` where an actual file path is expected. A module path member is definitely not a file path, so it is normal to not use `file_path` everywhere.

src/files.ml
rlepigre: No, you absolutely need to go through all mappings to find the "best" one.

src/extra.ml
rlepigre: It is in general expected that the elements are traversed from left to right (this is important in case of side effects), so I would rather not change that.

src/files.ml
```diff
@@ -3,6 +3,211 @@
...
+    new mapping is registered. In case of failure the program terminates and a
+    graceful error message is displayed. *)
+let new_lib_mapping : string -> unit = fun s ->
+  let fail s = exit_with ("Ill-formed argument to \"--map\": " ^^ s ^^ ".") in
+  let (module_path, file_path) =
+    match String.split_on_char ':' s with
+    | [mp; dir] -> (String.split_on_char '.' mp, dir)
+    | _         -> fail "bad syntax (use \"--help\")"
+  in
+  let valid_path_member s =
```
fblanqui: It would be better to have a common code with the parser to recognize regular idents.

src/files.ml
fblanqui: Sure but, later, you do not use file_path anymore. For instance, `lib_root : string option Pervasives.ref`.

src/files.ml
fblanqui: Ok but, then, why not moving this code in the  ModMap module?

rlepigre: The `ModMap` module is more of a library on which the function relies. Anyway, I might refactor this code at some point. For now, it is still a work in progress, that reviewing at that level of detail is not useful yet.

rlepigre:
We are now in a reasonable state here, we just need to check that the interface still works. But most importantly we need to change the command used to run the server. I'm not sure how to test all the LSP support, so maybe someone else can try to test that?

Note that when working with the server a package configuration file `lambdapi.pkg` should be placed at the root directory of the library being developed. This is similar to `_CoqProject` in Coq, but is mandatory for Lambdapi. It most importantly defines where under the library root the current library should be placed, without which the LSP server and Lambdapi itself cannot know where to look for files. I wrote a fair amount of doc, so please read that.

src/lambdapi.ml
```diff
@@ -104,6 +103,9 @@ let spec =
     ; ( "--verbose"
       , Arg.Int set_default_verbose
       , "<int> Set the default verbosity level" ^ verbose_values )
+    ; ( "--no-warnings"
+      , Arg.Unit (fun _ -> no_wrn := true)
```
gabrielhdt: Can't you use `Arg.Set`?

doc/sections/module.md
```diff
@@ -1,45 +1,84 @@
...
+the module path gets more members. File `<LIB_ROOT>/std/a/b/c/d.lp` has module
+path `std.a.b.c.d`.
+
+By default, every modules are looked up under the library root. However, there
+are cases where the files we want to work with are not (yet!) placed under the
+library root.  The typical case is when a library (or package) is still  under 
+development. In that case,  what we can do is map the development folder under
+the library root, similarly to what would happen when mounting a volume in our
+file system. There are two ways of doing that, the first one being to used the
+`--map <MOD_PATH>:<DIR>` command line option.  However, the best way is to use
```
gabrielhdt: `MOD_PATH` is the path as used in lambdapi while `DIR` is the filesystem path, is it?

src/core/files.ml
```diff
@@ -0,0 +1,280 @@
...
+  with Sys_error(_) | Invalid_argument(_) ->
+    exit_with "Invalid library root (no such file or directory [%s])." path
+
+(** [default_lib_root ()] returns the default library root curently configured
+    for the system. It depends on the current Opam switch (if any), and it may
+    or may not correspond to an existing directory. *)
+let default_lib_root : unit -> string = fun _ ->
+  match run_process "opam var lib" with
+  | Some([l]) -> Filename.concat l "lambdapi/lib_root"
+  | _         -> "/usr/local/lib/lambdapi/lib_root"
```
gabrielhdt: Shouldn't we allow the user to set an environment variable such as `LAMBDAPI_LIB_ROOT`?

src/lambdapi.ml
rlepigre: Indeed!

src/core/config.ml
```diff
@@ -0,0 +1,93 @@
...
+    in the same directory as [fname], then we look in the parent directory and
+    so on, until this is possible.  Note that [fname] is first normalized with
+    a call to [Filename.real_path]. *)
+let find_config : file_path -> file_path option = fun fname ->
+  let fname = Filename.realpath fname in
+  let rec find dir =
+    let file = Filename.concat dir config_file_name in
+    match Sys.file_exists file with
+    | true                   -> Some(file)
+    | false                  -> if dir = "/" then None else
```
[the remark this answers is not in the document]
rlepigre: No because that is not a separator in that case, and there is no similar variable for the "file system root".

doc/sections/module.md
rlepigre: Yes, exactly.

src/core/files.ml
rlepigre: I would say no, at least for now. There is already a command line option to change the library root, and that is all we need for development mode. Once installed, the choice if library root should always be consistent if we want to be able to have well-integrated tooling.

Of course, we will always be able to reconsider if we want to have several cohabiting (installed) versions of Lambdapi in the same Opam switch. But for now we do not need that.

rlepigre:
@gabrielhdt did you have a chance to test the LSP server with editors?

src/console.ml
rlepigre: Not really related to this PR, but OK.

src/files.ml
rlepigre: I don't think we want the full syntax of module paths here, the subset I am checking for seems reasonable to me. Anyway, we are changing our parser library so I will reconsider doing that when the PaComb stuff has been merged.

gabrielhdt:
Concerning the LSP server, I have just tried, and it does not work. However, it might come from my configuration since I use doom emacs, so you should not rely *too* much on me for this.

That said, the lsp server does not check the file (underlining in red ill typed terms) with this PR whereas it does so on the master branch.

rlepigre:
> Concerning the LSP server, I have just tried, and it does not work. However, it might come from my configuration since I use doom emacs, so you should not rely _too_ much on me for this.
> 
> That said, the lsp server does not check the file (underlining in red ill typed terms) with this PR whereas it does so on the master branch.
> 
> _Edit:_ attached the lsp log
> [lambdapi_lsp_log.txt](https://github.com/Deducteam/lambdapi/files/4237752/lambdapi_lsp_log.txt)

Thanks for trying. I think I fixed the problem, can you try again? (I'm really not great at using Emacs.)

gabrielhdt:
> can you try again?

It works now (I have not tested the proof mode though). I attach the log
[lambdapi_lsp_log.txt](https://github.com/Deducteam/lambdapi/files/4239374/lambdapi_lsp_log.txt)

rlepigre:
Since this PR has been open for a while, I would suggest that we merge it as soon as possible. Here is a summary of what has been done here:
- File paths can now be arbitrary, and they are normalized using `realpath` (together with the following three points this fixes #243).
- The module system has now a common root (the library root) that corresponds to a folder under which all libraries are intended to be installed.
- Arbitrary directories can be mounted under the library root to allow for a development package to behave as if it were already installed under the library root.
- A minimal package description file format is provided. For now it is only used as a way to specify the module path under which the files in the source tree are meant to be installed.
- A new option `--no-warnings` (or `-w`) is provided to disable all warnings (it fixes #296).
- The version of Lambdapi is now displayed correctly, even in development version (thus fixing #274, even though we considered it not to be a real issue due to dune limitations).
- The handling of command line options has been changed completely and now relies on `cmdliner`.
- The LSP server is now part of the Lambdapi binary. It is accessed through a sub-command (`lambdapi lsp`), while the main type-checking tool is accessed though another sub-command ( `lambdapi check`). Other commands are provided for parsing only and for "beautifying".
- The LSP server should now be able to work correctly according to dependencies. A configuration file should be placed in the working directory to ensure that the library mapping is made correctly.
- The documentation has been fixed according to the changes made here, and more.

I need help on the following points:
 - @gabrielhdt could you check again the Emacs mode? I had to change the server command again.
 - Could anyone make the necessary adaptations for vscode? This should not be too hard, I just don't have time to take care of that.
 - @fblanqui could you have a quick look and merge when proof modes have been confirmed to work or adapted?

Other that that, everything seems to work fine. And this PR put us in a good position to eventually provide tooling (installation tool, package manager) when it will become necessary.

gabrielhdt:
@rlepigre it still works.
[log-lp.txt](https://github.com/Deducteam/lambdapi/files/4302986/log-lp.txt)

fblanqui:
Hi @Rehan-MALAK . Could you please check the vscode interface?

README.md
```diff
@@ -8,30 +8,35 @@ in the [documentation](doc/DOCUMENTATION.md).
 Installation via [Opam](http://opam.ocaml.org/)
 ---------------------
 
+We will only publish a new version of `lambdapi` Opam pacakge when the
+development has reached a more stable test. For now, we advise you to
+pin the development repository to get the latest bug fixes.
+
 ```bash
-opam install lambdapi
+opam pin add lambdapi https://github.com/Deducteam/lambdapi.git
```
fblanqui: It doesn't look like a good idea to pin add the development version. Please revert.

README.md
```diff
@@ -41,7 +46,7 @@ Using Opam, a suitable OCaml environment can be setup as follows.
 ```bash
 opam switch 4.05.0
 eval `opam config env`
-opam install dune odoc menhir yojson cmdliner bindlib.5.0.0 timed.1.0 earley.2.0.0 ppx_inline_test why3.1.2.0
+opam install dune odoc menhir yojson cmdliner bindlib.5.0.0 timed.1.0 earley.2.0.0 why3.1.2.0
```
fblanqui: no need to give version numbers

doc/sections/install.md
```diff
@@ -1,15 +1,20 @@
-Installation via Opam
----------------------
+Installation via [Opam](http://opam.ocaml.org/)
+-----------------------------------------------
 
-See http://opam.ocaml.org/.
+The version of `lambdapi` that is directly available through Opam (with the
+command `opam install lambdapi`) is currently outdated. We will only publish
+a new version of `lambdapi` Opam pacakge when the development has reached a
```
fblanqui: package (spelling)

doc/sections/install.md
```diff
@@ -1,15 +1,20 @@
...
+more stable test. For now, we advise you to pin the development repository to
+get the latest bug fixes.
 
 ```bash
-opam install lambdapi
+opam pin add lambdapi https://github.com/Deducteam/lambdapi.git
```
fblanqui: please revert, just mention in comment that one can pin add the development version

doc/sections/module.md
```diff
@@ -1,45 +1,84 @@
...
+split our developments into several files, and even several folders.  The main
+thing to know is that a file holds a single module,  and accessing this module
+in other files requires using its *full* module path.
+
+#### Module path
+
+The module path that corresponds to a source file is defined using the name of
+the file, but also the position of the file under the *library root* (a folder
+under which the standard library and all packages are installed).
+
```
fblanqui: I would prefer not to speak about some standard library as we have none. And the day we will have one, if any, it should be in a distinct repo anyway, like it is done with Lean or Agda.

doc/sections/module.md
```diff
@@ -1,45 +1,84 @@
...
+is accessed with module path `std.bool`.  And if there are nested folders then
+the module path gets more members. File `<LIB_ROOT>/std/a/b/c/d.lp` has module
+path `std.a.b.c.d`.
+
+By default, every modules are looked up under the library root. However, there
+are cases where the files we want to work with are not (yet!) placed under the
+library root.  The typical case is when a library (or package) is still  under 
+development. In that case,  what we can do is map the development folder under
+the library root, similarly to what would happen when mounting a volume in our
+file system. There are two ways of doing that, the first one being to used the
```
fblanqui: to use (grammar)

doc/sections/module.md
```diff
@@ -1,45 +1,84 @@
...
+library (or package) under development. It should be named `lambdapi.pkg`, and
+contain the following fields (an example is given below for the syntax):
+ - `package_name` giving a globally unique name for the package being defined.
+   This is not yet used,  but will be necessary when we eventually publish our
+   packages online so that they can be automatically downloaded when our users
+   need them. (The idea is to come up with a system similar to Cargo in Rust.)
+ - `root_path` gives the module path under which the library is to be  placed.
+   Assuming that our configuration file is at `<REPO_ROOT>/lambdapi.pkg`, this
+   means that `<REPO_ROOT>/a/b/c.lp` will get module path `<ROOT_PATH>.a.b.c`.
+   In other words this is amounts to giving `--map-dir <ROOT_PATH>:<REPO_ROOT>`
```
fblanqui: this amounts to give

doc/sections/module.md
```diff
@@ -1,45 +1,84 @@
...
+
+A package configuration file can be placed at the root of the source tree of a
+library (or package) under development. It should be named `lambdapi.pkg`, and
+contain the following fields (an example is given below for the syntax):
+ - `package_name` giving a globally unique name for the package being defined.
+   This is not yet used,  but will be necessary when we eventually publish our
+   packages online so that they can be automatically downloaded when our users
+   need them. (The idea is to come up with a system similar to Cargo in Rust.)
+ - `root_path` gives the module path under which the library is to be  placed.
+   Assuming that our configuration file is at `<REPO_ROOT>/lambdapi.pkg`, this
```
fblanqui: can the configuration file be somewhere else? what's the semantics in this case?

doc/sections/module.md
```diff
@@ -1,45 +1,84 @@
...
+   means that `<REPO_ROOT>/a/b/c.lp` will get module path `<ROOT_PATH>.a.b.c`.
```
fblanqui: Are you sure? Don't you want <package_name>.a.b.c instead? With <package_name>.a.b.c maped to <ROOT_PATH>/package_name/a/b/c.lp?

doc/sections/module.md
```diff
@@ -1,45 +1,84 @@
...
+# Lines whose first non-whitespace charater is # are comments  
+# The end of a non-comment line cannot be commented.
+# The config file MUST be called "lambdapi.pkg".
+# The syntax for entries is:
+key = value
+# If the key is not known then it is ignored.
+# There are two used keys for now:
+package_name = my_package_name
+root_path    = a.b.c
+# We will use more entries lated (e.g., authors, version, ...)
```
fblanqui: later

README.md
rlepigre: It is worse for a user to install a completely outdated version. We will change that when we do a stable release, which I guess should be done soon.

README.md
rlepigre: No, version numbers are important.

rlepigre: It may not compile otherwise.

doc/sections/install.md
rlepigre: As I said somewhere else, this is better for now.

doc/sections/module.md
```diff
@@ -1,45 +1,84 @@
...
+
+A package should be installed under the library directory, under the specified
+root path. In other words a package whose root path is `x.y.z` should have its
+files installed under the directory `<LIB_ROOT>/x/y/z`. Moreover the directory
+structure relative to the package configuration file should be preserved.
+
+Of course, we should enforce that a root path is uniquely used by one package.
+This can be enforced using a central authority, but this can only be done when
+we actually decide to implement automatic package distribution. For now, it is
+assumed that every user uses reasonable root paths for their packages.
```
fblanqui: every user -> everyone
their -> her

doc/sections/module.md
rlepigre: Well, this is intended as an example of package creation. I don't think it is harmful, and we can actually use if for experimenting with things. It can always be made into a separate repository later, but I don't think this is a good idea for such a small tool.

doc/sections/options.md
```diff
@@ -1,134 +1,134 @@
...
+   and are up to date (they are regenerated otherwise).
+
+ - `--verbose NUM` sets the verbosity level to the given natural number (the
    default value is 1). A value of 0 should not print anything, and the higher
    values (up to 3) print more and more information.
 
-#### Confluence checking
+ - `--lib-root DIR` sets the library root, that is, the folder corresponding
+   to the entry point of the Lambdapi package system. This is the folder under
+   which every package is installed, and a default value is only only known if
```
fblanqui: only twice

doc/sections/options.md
```diff
@@ -1,134 +1,134 @@
...
    generated. For more information on trees (and how to read them), see
    `src/tree_graphviz.ml`.
- - `--keep-rule-order` forces the rewriting engine to use in priority
-   the topmost rules.
+
+##### Confluence checking
+
+Lambdapi provides an option `--confluence CMD` to check the confluence of the
+rewriting system by calling an external prover with the command `CMD`. The
+given command receives [TRS](http://project-coco.uibk.ac.at/problems/trs.php)
```
fblanqui: Isn't it the HRS format (http://project-coco.uibk.ac.at/problems/hrs.php) instead?

doc/sections/module.md
rlepigre: I believe what I wrote is correct (https://fr.wikipedia.org/wiki/They_singulier).

README.md
fblanqui: Right. My point was to anticipate this new release.

README.md
fblanqui: But that's the job of opam to figure out the version numbers so that it compiles!

doc/sections/options.md
fblanqui: @Rehan-MALAK : you checked the use of csiho recently. Could  you please check this paragraph?

doc/sections/options.md
```diff
@@ -1,134 +1,134 @@
...
+given command receives [XTC](http://cl2-informatik.uibk.ac.at/mercurial.cgi/TPDB/raw-file/tip/xml/xtc.xsd)
+formatted text on its standard input, and it is expected to output on the
+first line of its standard output either `YES`, `NO` or `MAYBE`.
+
+As for confluence, `echo MAYBE` is the simplest possible (valid) command for
+checking termination.
+
+To the best of our knowledge, the only termination checker that is compatible
+with all the features of Lambdapi is [`SizeChangeTool`](https://github.com/Deducteam/SizeChangeTool).
+It can be called using the flag `--termination "path/to/sct.native --no-color --stdin=xml"`
```
fblanqui: @GuillaumeGen : I don't remember whether SCT can handle LP files. Could you please check this paragraph?

doc/sections/module.md
rlepigre: I don't understand your question, what are you referring to?

doc/sections/module.md
rlepigre: Yes, I think this is the better choice. In any case, it subsumes what you propose.

doc/sections/options.md
rlepigre: Guillaume wrote this paragraph, I just reformulated it.

rlepigre: And it does not handle LP files, but XML files.

doc/sections/options.md
```diff
@@ -1,134 +1,134 @@
...
+
+If no type-level rewriting is used [`Wanda`](http://wandahot.sourceforge.net/)
+can also be used. However, it does not directly accept input on its standard
+input, so it is tricky to have Lambdapi call it directly. Alternatively, one
+can first generate a `.xml` file as described below.
```
fblanqui: below where?

rlepigre: Literally the next line.

doc/sections/structure.md
```diff
@@ -83,16 +85,23 @@ Structure of directories and files
...
+
+* `src/lsp/`: files related to the LSP server
+
+* `src/lambdapi.ml`: main program
 
 * `tests/`: unit tests
    - `OK/`: tests that should succeed
    - `KO/`: tests that should fail
 
+* `lib/`: a minimal standar library for Lambdapi.
```
fblanqui: standard

doc/sections/ui.md
```diff
@@ -1,4 +1,13 @@
-See the file [lp-lsp/README.md](../../lp-lsp/README.md) for more details.
+User interfaces
+---------------
+
+A prototype language server is provided with Lambdapi. It follows (an extension
+of) the Language Server Protocol (LSP), which is supported by most editors. See
```
fblanqui: URL: https://microsoft.github.io/language-server-protocol/

lib/std/bool.lp
```diff
@@ -0,0 +1,57 @@
+// Data type of booleans.
```
fblanqui: Could you please rename std into some thing else, say example? I don't want users to think that it could be the beginning of a standard library (even though it could be integrated in such a library later). Thank you.

doc/sections/module.md
fblanqui: I find the beginning of the sentence confusing. Could you please write:

A package configuration file  `<REPO_ROOT>/lambdapi.pkg` can be placed at the root of the source tree of a
library (or package) under development. It should contain the following fields (an example is given below for the syntax): ...

doc/sections/module.md
fblanqui: You do not get my point. Why <package_name> is not used? In addition, why do you say 'will get" and not "gets"?

fblanqui: Remark: I propose nothing. I just try to understand your doc.

doc/sections/options.md
fblanqui: in an extension of the XTC format I guess.

doc/sections/options.md
```diff
@@ -1,134 +1,134 @@
...
+the rewriting system by calling an external prover with the command `CMD`. The
+given command receives [XTC](http://cl2-informatik.uibk.ac.at/mercurial.cgi/TPDB/raw-file/tip/xml/xtc.xsd)
+formatted text on its standard input, and it is expected to output on the
+first line of its standard output either `YES`, `NO` or `MAYBE`.
+
+As for confluence, `echo MAYBE` is the simplest possible (valid) command for
+checking termination.
+
+To the best of our knowledge, the only termination checker that is compatible
+with all the features of Lambdapi is [`SizeChangeTool`](https://github.com/Deducteam/SizeChangeTool).
```
fblanqui: I propose: the only termination checker that can handle Lambdapi files is ....

fblanqui: Ignore this comment

doc/sections/options.md
fblanqui: Then, I propose:

If there is no type-level rewriting rules, then [`Wanda`](http://wandahot.sourceforge.net/) can also be used.

For provers like Wanda that do not read the standard input, one needs to generate a XTC file first as follows: `--termination "cat > output.trs; echo MAYBE"`.

doc/sections/module.md
rlepigre: Your version is not precise enough: the root of a package might not coincide with the root of a repository. For instance, it may be that several related but independent packages are developed in the same repository.

doc/sections/options.md
rlepigre: Doing that makes this paragraph less consistent with what I wrote for confluence.

Don't you think we could maybe improve on the doc later? I've clearly made a lot of improvements on the doc, actually a lot more than I should have since they were not directly related to what I did. Maybe you can see with @GuillaumeGen  later if he can clarify the doc. In particular, I have been meaning to try something like the following command for Wanda to see if it works: `--termination "cat > /tmp/output.xml; wanda output.xml"`. If it does, the doc can be made clearer by giving it directly.

lib/std/bool.lp
rlepigre: Well, the whole point was to have the stem of a minimal standard library. And I was hoping that someone would pick up on that to start building a collection of generally useful modules. I definitely makes sense to share very common stuff among all libraries, so why not start there?

Anyway, `std` is clearly documented as a very minimal standard library, I don't see how people could mistake that for anything else. And in all cases, I think that such a "standard library" will have to evolve with the tool. It is likely that developing it will make us realize that certain features are missing or not working well enough.

In any case, we can always decide to extend, rename or even remove `std` later. I dont think it is a problem to keep it as it is. Is it really a problem for you?

fblanqui: Yes, it is.

fblanqui:
Hi. Emacs doesn't seem to work with this PR. It would be good to check vscode as well.

rlepigre:
> Hi. Emacs doesn't seem to work with this PR. It would be good to check vscode as well.

It is supposed to work, @gabrielhdt tried it.

For vscode I have no idea how to make the necessary adaptation. There is no documentation and I don't want to spend my time on that, especially since I do not use vscode.

Rehan-MALAK:
> > Hi. Emacs doesn't seem to work with this PR. It would be good to check vscode as well.
> 
> It is supposed to work, @gabrielhdt tried it. Did you reinstall the Emacs mode?
> 
> For vscode I have no idea how to make the necessary adaptation. There is no documentation and I don't want to spend my time on that, especially since I do not use vscode.

It seems that pretty much every dedukti/lambdapi's users use emacs in one form or another (evil, dooms, spacemacs, vanilla, ...). I havn't use vscode for months (since I saw the lack of interest of the coq people for vscoq)
Both from a technical point of view and a strategical point of view, it's more interesting to concentrate our efforts on :
  - emacs > vscode
  - LSP serveur > LSP client

gabrielhdt:
I tested back the emacs mode, with a minimal configuration, and it seems that the proof mode does not work. The rest does.

To see it yourself,
```sh
wget https://sanemacs.com/sanemacs.el
echo '(use-package eglot)' >> sanemacs.el
emacs -q --no-site-file -l sanemacs.el -l <path/to/lambdapi.el>
```
and edit for instance `natproofs.lp`.

fblanqui:
I can now run emacs (I didn't install LP properly) but it doesn't print whether some thing is well type-checked or not.

fblanqui:
No, my fault. lambdapi.pkg is now mandatory even for a single file... Why not but, then, the error message is not very clear:
```
15:37 ~/lambdapi ((HEAD détachée sur rodolphe/filepath)) lambdapi check 303.lp 
[/home/blanqui/src/lambdapi/303.lp] cannot be placed under the library mapping.
```
Why not saying that there is no lambdapi.pkg file? Do you test this btw?

rlepigre:
> Why not saying that there is no lambdapi.pkg file? Do you test this btw?

Yes, the `lambdapi.pkg` file is mandatory now. At least when you use the LSP server. Otherwise there is no way for the server where files should be placed under the library root.

I have given some thoughts as to what we could do to make things simpler, but there is no easy solution. Indeed, the error message is not great, I'll improve it.

fblanqui:
It's not just when you use the LSP server. It's all the time, even for checking a single file with no dependencies. It's not very convenient. Couldn't we use a default mapping in this case?

rlepigre:
> It's not just when you use the LSP server. It's all the time, even for checking a single file with no dependencies. It's not very convenient. Couldn't we use a default mapping in this case?

I don't know, I don't have time to take care of that right now. Open an issue about that once this PR is merged and I'll try and find a way. However, I don't think this is going to be a big problem since the only thing you would need is to put a `lambdapi.pkg` file in your working directories. In fact, you could very well put one into your home directory, and this way all your "single files" would be type-checked no problem (assuming your directories have reasonable names, with no spaces for instance).

maximedenes:
> since I saw the lack of interest of the coq people for vscoq

FTR, I wonder on what kind of data this observation is based. Or maybe I misunderstood the claim.

Rehan-MALAK:
> FTR, I wonder on what kind of data this observation is based. Or maybe I misunderstood the claim.

Hi @maximedenes 
A bit of time spent on gitter vscoq some months ago and private conversations with coq intensive users (not especially coq developers).
But we could talk about this somewhere else. (I ping you privately on gitter vscoq)

Issue #290, opened 2019-11-25
tests on Mac OS


elhaddadyacine:
Hello,
make tests does not work on mac OS.
This is a fragment of what `make tests` outputs:
```
readlink: illegal option -- f
usage: readlink [-n] [file ...]
readlink: illegal option -- f
usage: readlink [-n] [file ...]
readlink: illegal option -- f
usage: readlink [-n] [file ...]
## OK tests ##
KO tests/OK/test.lp
/bin/sh: --verbose: command not found
KO tests/OK/244.lp
/bin/sh: --verbose: command not found
KO tests/OK/245.lp
/bin/sh: --verbose: command not found
KO tests/OK/245b.lp
/bin/sh: --verbose: command not found
KO tests/OK/262_legacy_parsing.dk
/bin/sh: --verbose: command not found
KO tests/OK/262_pair_ex_1.lp
/bin/sh: --verbose: command not found
```

rlepigre:
Yeah, you probably have the BSD version of `readlink`, which does not have the same interface as the GNU version. I think there is a way to install the GNU commands on mac OS.

Rehan-MALAK:
I have seen several projects including in the Makefile a "portable version" or readlink.

rlepigre:
Yeah, actually we can make it portable by just removing it.

elhaddadyacine:
The purpose is to setup CI for mac OS.

fblanqui:
Fixed in https://github.com/Deducteam/lambdapi/commit/ceba0a2383112aab85ec694be68ed4803316a27a.

PR #291, opened 2019-11-25
Avoid using the readlink command in the makefile.

rlepigre:
This solves #290.

Why did we use `readlink` anyway?


rlepigre:
> Another choice is to use `dune exec`

I think that's what we did at some point. No idea why this was changed.

Issue #292, opened 2019-11-26
Wrong warning


fblanqui:
With
```
symbol R : TYPE
constant symbol zero : R
symbol d2_1 : (R ⇒ R ⇒ R) ⇒ (R ⇒ R ⇒ R)
rule d2_1 (λ_ y,&F[y]) → λ_ _,zero
```
we get:
```
[tmp/wildcard.lp, 4:16-21] &F could be replaced by a wildcard.
```

rlepigre:
The warning is not wrong since `&F` is unused. Rather, the syntax is wrong because it does not allow `&_[y]`. Note that my initial proposal for the syntax did allow for a wildcard precise for that purpose. And if I remember @fblanqui insisted that this would never be useful. :) Is it now?

fblanqui:
I think that the warning is wrong because, for `d2_1(λxy,t)` to be reducible, we must check that x does not occur in `t`.

fblanqui:
`rule d2_1 (λx y,&F[y]) → λ_ _,zero` outputs no warning although the two declarations should be equivalent.

fblanqui:
The problem seems to come from the fact we check that that &F is applied to as many arguments as the number of variables in env:
https://github.com/Deducteam/lambdapi/blob/ceba0a2383112aab85ec694be68ed4803316a27a/src/scope.ml#L311
But in https://github.com/Deducteam/lambdapi/blob/ceba0a2383112aab85ec694be68ed4803316a27a/src/scope.ml#L241
and https://github.com/Deducteam/lambdapi/blob/ceba0a2383112aab85ec694be68ed4803316a27a/src/scope.ml#L246
we add no variable in env, and I think that it is important for metavariables to not get too many dependencies.

rlepigre:
> I think that the warning is wrong because, for d2_1(λxy,t) to be reducible, we must check that x does not occur in t.

Sure, that's exactly the meaning of `&_[y]` (versus a simple wildcard). However, there is no need to name the pattern variable since it is not used.

rlepigre:
> The problem seems to come from the fact we check that &F is applied to as many arguments as the number of variables in env:

Indeed, if an unused variable is applied to all variables in scope then it can be replaced by a wildcard since a wildcard automatically gets all the variables in its environment. That is not a very natural semantics for me, and I would rather force the user to give the list of variables that can appear in wildcards (if any). In any case, that only happens in higher-order patterns.

fblanqui:
[...]

PR #293, opened 2019-11-27
Fixing the behavior of wildcards + unused warning cleanup.

rlepigre:
This is aimed as a fix for #292.

A couple of things are fixed:
 - Syntax `&_[x,y]` can now be used in a LHS to match any term `t` with `FV(t) ⊆ {x,y}`.
 - Syntax `?_[t,u]` is now forbidden (it was mistakenly allowed before).
 - We now get a warning in case of unused λ-variable (except if its name is prefixed by `_`.
 - Warnings now seem to have the expected behavior regarding wildcards.
 - Warnings are printed unconditionally (they do not depend on the verbosity level). This is more consistent with the rest of the warnings in the code.

### DIFF
--- a/doc/syntax.bnf
+++ b/doc/syntax.bnf
@@ -83,7 +83,7 @@ reserved "why3"
   | "@"? <qident>
   | "_"
   | "?" - <ident> {"[" <term> {"," <term>}* "]"}?
-  | "&" - <ident> {"[" <term> {"," <term>}* "]"}?
+  | "&" - {<ident> | "_"} {"[" <term> {"," <term>}* "]"}?
   | "(" <term> ")"
   | "{" <term> "}"
   | <term> <term>
--- a/src/parser.ml
+++ b/src/parser.ml
@@ -303,15 +303,17 @@ let parser arg_ident =
 
 (** Metavariable identifier (regular or escaped, prefixed with ['?']). *)
 let parser meta =
-  | "?" - id:{regular_ident | escaped_ident} -> in_pos _loc id
+  | "?" - id:{regular_ident | escaped_ident} ->
+      if id = "_" then Earley.give_up (); in_pos _loc id
 
 (** Pattern variable identifier (regular or escaped, prefixed with ['&']). *)
 let parser patt =
-  | "&" - id:{regular_ident | escaped_ident} -> in_pos _loc id
+  | "&" - id:{regular_ident | escaped_ident} ->
+      if id = "_" then None else Some(in_pos _loc id)
 
 (** Any path member identifier (escaped idents are stripped). *)
 let parser path_elem =
-  | id:regular_ident -> KW.check id; (id, false)
+  | id:regular_ident          -> KW.check id; (id, false)
   | id:escaped_ident_no_delim -> (id, true)
 
 (** Module path (dot-separated identifiers. *)
--- a/src/scope.ml
+++ b/src/scope.ml
@@ -182,13 +182,11 @@ let get_args : p_term -> p_term * p_term list =
     state [ss] is used to hande module aliasing according to [find_qid]. *)
 let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
   (* Unique pattern variable generation for wildcards in a LHS. *)
-  let fresh_patt : env -> tbox =
+  let fresh_patt_name =
     let c = Pervasives.ref (-1) in
-    fun env ->
-      let open Pervasives in
-      let name = incr c; Printf.sprintf "#%i" !c in
-      _Patt None name (Env.to_tbox env)
+    fun _ -> Pervasives.(incr c; Printf.sprintf "#%i" !c)
   in
+  let fresh_patt env = _Patt None (fresh_patt_name ()) (Env.to_tbox env) in
   (* Toplevel scoping function, with handling of implicit arguments. *)
   let rec scope : env -> p_term -> tbox = fun env t ->
     (* Extract the spine. *)
@@ -262,8 +260,9 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
       | (Some x::l,d,i)::xs ->
           let v = Bindlib.new_var mkfree x.elt in
           let a = scope_domain env d in
-          let env = if x.elt = "_" then env else (x.elt,(v,a))::env in
-          let t = aux env ((l,d,i)::xs) in
+          let t = aux ((x.elt,(v,a)) :: env) ((l,d,i) :: xs) in
+          if x.elt.[0] <> '_' && not (Bindlib.occur v t) then
+            wrn x.pos "Variable [%s] could be replaced by [_]." x.elt;
           cons a (Bindlib.bind_var v t)
     in
     aux env xs
@@ -310,7 +309,7 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
           match unfold (Bindlib.unbox (scope env t)) with
           | Vari(x) -> x
           | _       -> fatal t.pos "Only bound variables are allowed in the \
-                                    environment of &%s." id.elt
+                                    environment of pattern variables."
         in
         let vs = Array.map scope_var ts in
         (* Check that [vs] are distinct variables. *)
@@ -319,23 +318,36 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
             if Bindlib.eq_vars vs.(i) vs.(j) then
               let name = Bindlib.name_of vs.(j) in
               fatal ts.(j).pos "Variable %s appears more than once in the \
-                                environment of &%s." name id.elt
+                                environment of a pattern variable." name
           done
         done;
         (* Find the reserved index, if any. *)
-        let i =
-          try Some(List.assoc id.elt m) with Not_found ->
-            (* If the pattern variable is applied to every bound variables and
-               if it has no reserved index then it is useless. *)
-            if !verbose > 1 && List.length env = Array.length vs then
-              wrn t.pos "&%s could be replaced by a wildcard." id.elt;
-            None
-        in
-        _Patt i id.elt (Array.map Bindlib.box_var vs)
+        begin
+          match id with
+          | None     ->
+              if List.length env = Array.length vs then
+                wrn t.pos "Pattern [%a] could be replaced by [_]."
+                  Pretty.pp_p_term t;
+              _Patt None (fresh_patt_name ()) (Array.map Bindlib.box_var vs)
+          | Some(id) ->
+              let i =
+                try Some(List.assoc id.elt m) with Not_found ->
+                  if List.length env = Array.length vs then
+                    wrn t.pos "Pattern variable [%a] can be replaced by a \
+                               wildcard [_]." Pretty.pp_p_term t
+                  else
+                    wrn t.pos "Pattern variable [&%s] does not need to be \
+                               named." id.elt;
+                  None
+              in
+              _Patt i id.elt (Array.map Bindlib.box_var vs)
+        end
     | (P_Patt(id,ts)   , M_RHS(m,_)       ) ->
         let x =
-          try List.assoc id.elt m with Not_found ->
-          fatal t.pos "Pattern variable not in scope." (* Cannot happen. *)
+          match id with
+          | None     -> fatal t.pos "Wildcard pattern not allowed in a RHS."
+          | Some(id) -> try List.assoc id.elt m with Not_found ->
+                          fatal t.pos "Pattern variable not in scope."
         in
         _TEnv (Bindlib.box_var x) (Array.map (scope env) ts)
     | (P_Patt(_,_)     , _                ) ->
@@ -410,16 +422,7 @@ let patt_vars : p_term -> (string * int) list * string list =
     | P_Iden(_)        -> acc
     | P_Wild           -> acc
     | P_Meta(_,ts)     -> Array.fold_left patt_vars acc ts
-    | P_Patt(id,ts)    ->
-        let (pvs, nl) as acc = Array.fold_left patt_vars acc ts in
-        let l = Array.length ts in
-        begin
-          try
-            if List.assoc id.elt pvs <> l then
-              fatal id.pos "Arity mismatch for pattern variable [%s]." id.elt;
-            if List.mem id.elt nl then acc else (pvs, id.elt::nl)
-          with Not_found -> ((id.elt,l)::pvs, nl)
-        end
+    | P_Patt(id,ts)    -> add_patt (Array.fold_left patt_vars acc ts) id ts
     | P_Appl(t,u)      -> patt_vars (patt_vars acc t) u
     | P_Impl(a,b)      -> patt_vars (patt_vars acc a) b
     | P_Abst(xs,t)     -> patt_vars (arg_patt_vars acc xs) t
@@ -430,12 +433,22 @@ let patt_vars : p_term -> (string * int) list * string list =
     | P_BinO(t,_,u)    -> patt_vars (patt_vars acc t) u
     | P_Wrap(t)        -> patt_vars acc t
     | P_Expl(t)        -> patt_vars acc t
+  and add_patt ((pvs, nl) as acc) id ts =
+    match id with
+    | None     -> acc
+    | Some(id) ->
+    try
+      if List.assoc id.elt pvs <> Array.length ts then
+        fatal id.pos "Arity mismatch for pattern variable [%s]." id.elt;
+      if List.mem id.elt nl then acc else (pvs, id.elt :: nl)
+    with Not_found -> ((id.elt, Array.length ts) :: pvs, nl)
   and arg_patt_vars acc xs =
     match xs with
     | []                       -> acc
     | (_, None,    _   ) :: xs -> arg_patt_vars acc xs
     | (_, Some(a), _) :: xs    -> arg_patt_vars (patt_vars acc a) xs
-  in patt_vars ([],[])
+  in
+  patt_vars ([],[])
 
 (** [scope_rule ss r] turns a parser-level rewriting rule [r] into a rewriting
     rule (and the associated head symbol). *)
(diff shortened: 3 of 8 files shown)


doc/syntax.bnf
```diff
@@ -83,7 +83,7 @@ reserved "why3"
   | "@"? <qident>
   | "_"
   | "?" - <ident> {"[" <term> {"," <term>}* "]"}?
-  | "&" - <ident> {"[" <term> {"," <term>}* "]"}?
+  | "&" - {<ident> | "_"} {"[" <term> {"," <term>}* "]"}?
```
fblanqui: Since <ident> includes "_", "&" - {<ident> | "_"} is equivalent to "&" - <ident>. Shouldn't you change the definition of <ident> so that it excludes "_"? Or any current ident starting with "_"?

src/parser.ml
```diff
@@ -303,15 +303,17 @@ let parser arg_ident =
 
 (** Metavariable identifier (regular or escaped, prefixed with ['?']). *)
 let parser meta =
-  | "?" - id:{regular_ident | escaped_ident} -> in_pos _loc id
+  | "?" - id:{regular_ident | escaped_ident} ->
+      if id = "_" then Earley.give_up (); in_pos _loc id
```
fblanqui: Will this generate a parse error? If so, couldn't you provide an informative error message? Perhaps there are other calls to Early.give_up() where we could do this as well.

src/scope.ml
```diff
@@ -243,8 +241,9 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
       | (Some x::l,d,i)::xs ->
           let v = Bindlib.new_var mkfree x.elt in
           let a = scope_domain env d in
-          let env = if x.elt = "_" then env else (x.elt,(v,a))::env in
-          let t = aux env ((l,d,i)::xs) in
+          let t = aux ((x.elt,(v,a)) :: env) ((l,d,i) :: xs) in
```
fblanqui: As I suggested in a previous comment (https://github.com/Deducteam/lambdapi/issues/292#issuecomment-558723937), I am not sure that this is a good idea: this impacts unification a lot. Variables that start with _ should be filtered out before generating new metavariables. Is it something that you do?

tests/OK/292.lp
```diff
@@ -0,0 +1,21 @@
...
+constant symbol zero : R
+symbol d2_1 : (R ⇒ R ⇒ R) ⇒ (R ⇒ R ⇒ R)
+
+// The following are semantically equivalent
+rule d2_1 (λ_ y, &F[y]) → λ_ _,zero // warning
+rule d2_1 (λ_ y, &_[y]) → λ_ _,zero // warning
+rule d2_1 (λ_ y, _    ) → λ_ _,zero // no warning
+
+// The following two are NOT semantically the same
+rule d2_1 (λx y, &_[y]) → λ_ _,zero // warning
```
fblanqui: Why would you get a warning here?

tests/OK/292.lp
```diff
@@ -0,0 +1,21 @@
...
+rule d2_1 (λ_ y, _    ) → λ_ _,zero // warning
```
fblanqui: Why do you get a warning here? Two lines above, you wrote that there is no warning...

doc/syntax.bnf
rlepigre: These are not equivalent, `<ident>` does not allow `"_"` since it is reserved.

src/parser.ml
rlepigre: Yes, but what you ask is not easily possible with Earley. I will probably replace Earley in the near future to solve that problem and the complexity problem as well.

src/scope.ml
rlepigre: Here I just removed dead code, the wildcard case is handled in the `None` branch.

tests/OK/292.lp
rlepigre: I think it is just a copy paste error, there is no warning.

tests/OK/292.lp
rlepigre: Same here.

doc/syntax.bnf
fblanqui: Ok. Could you please add a comment about that in syntax.bnf?

src/parser.ml
fblanqui: Perhaps we should exclude every ident starting with `_`. Perhaps we should introduce a new syntactic category for idents starting with `_`.

src/parser.ml
```diff
@@ -303,15 +303,17 @@ let parser arg_ident =
...
 
 (** Pattern variable identifier (regular or escaped, prefixed with ['&']). *)
 let parser patt =
-  | "&" - id:{regular_ident | escaped_ident} -> in_pos _loc id
+  | "&" - id:{regular_ident | escaped_ident} ->
+      if id = "_" then None else Some(in_pos _loc id)
```
fblanqui: Perhaps we should exclude every ident starting with _.

doc/syntax.bnf
rlepigre: I'm not sur it is a good idea, this file is an indication, I don't consider it as normative. Moreover, there are inconsistencies between the lexical conventions in the file and in the parser. For instance, `<ident>` can both include or exclude reserved words.

src/parser.ml
rlepigre: I don't know, maybe. For now prefixing a (non-empty) identifier with `_` is only used as a way to disable the warning. Which I think is common in many languages.

src/parser.ml
rlepigre: Again I'm not sure, there is no harm in having them. Anyway, we can always revise these syntax details later (they have little to do with this PR).

fblanqui: But isn't it the case that, currently in your PR, if I write `\ _x,t` then I cannot use `_x` in `t`, and if I write `&_x` in a LHS then I can use `_x` in the RHS? This is not consistent.

rlepigre: Currently, `_x` and `&_x` can be used. And there is no warning silencing for an unnecessary `&_x`.

src/parser.ml
fblanqui: Note that, here, it is about metavariables. It could be interesting to write _xx to generate a fresh metavariable with a name prefixed by xx.

tests/OK/292.lp
```diff
@@ -0,0 +1,21 @@
...
+rule d2_1 (λ_ y, &F[y]) → λ_ _,zero // warning
+rule d2_1 (λ_ y, &_[y]) → λ_ _,zero // warning
+rule d2_1 (λ_ y, _    ) → λ_ _,zero // no warning
+
+// The following two are NOT semantically the same
+rule d2_1 (λx y, &_[y]) → λ_ _,zero // no warning
+rule d2_1 (λ_ y, _    ) → λ_ _,zero // no warning
+// The difference is that in the former variable "x" may appear in the type
+// of the "y" variable. That is not the case in the second line.
+
```
fblanqui: and is `λ_x y, _` equivalent to `λ_ y, _`?

src/parser.ml
rlepigre: I'm not sure I see the interest in that. You can already do `?abc[x,y,z]` to generate a metavariable named `abc`.

tests/OK/292.lp
rlepigre: No, it would be equivalent to `λx y, _` in the current implementation.

fblanqui:
One potential problem I see is that you still have:
```
let parser arg_ident =
  | id:ident -> Some(id)
  | _wild_   -> None
```
So writing _xx is not equivalent to writing _. This will generate bigger environments to metavariables.

rlepigre:
> One potential problem I see is that you still have:
> 
> ```
> let parser arg_ident =
>   | id:ident -> Some(id)
>   | _wild_   -> None
> ```
> 
> So writing _xx is not equivalent to writing _. This will generate bigger environments to metavariables.

Yeah, I know. That is something we need to figure out. But I see that as being rather orthogonal to that PR. One idea to solve that though would be to stop using `None` to represent `_` and `Some(id)` to represent any other identifier in the parser. We could simply use an identifier, and reason by case on whether the first character is `'_'` or not.

src/parser.ml
fblanqui: But we will never do that. Imagine that you are in the middle of a proof with dozens of assumptions. You don't want to enumerate them all, but you may want to give an informative name to your metavariable. Moreover, currently, in proofs, goals, which are metavariables, are not referred to by their names but numbered instead (focus command), with a number that is not fixed. It could be useful at some point to be able to refer to some goal by its name. But this is not directly related to the current PR.

Meanwhile, and since _xx idents have a special meaning when used in an abstraction or a product, it would be good not to allow such idents for metas. So I propose to instead write `if id.[0] = "_" then Earley.give_up (); in_pos _loc id`.

tests/OK/292.lp
fblanqui: This is not the semantics I expect.

rlepigre: I have mixed feelings about that. This is the semantics in most programming languages I know (including OCaml, where you can write `fun _x -> _x`, so clearly identifiers starting with an underscore are in scope).

src/parser.ml
rlepigre: The thing is, they don't have a special meaning. Their only extra feature is to disable a warning. So I would just leave these things as they are no, until everything we want has been clarified. Remember that this is only a minor point of this PR, fixing a behavior that was not expected: we definitely do not want `?_` as a valid metavariable name. What we are talking about here has nothing to do with this PR. If you want to discuss that more, please open an issue and make a proposal for whatever semantics you think would be best.

fblanqui:
I agree. (This is partly what I was suggesting in https://github.com/Deducteam/lambdapi/issues/292#issuecomment-558723937.)

doc/syntax.bnf
fblanqui: Why not taking syntax.bnf as our reference? Then, a LP parser should implement it.

rlepigre: There are many details that you don't want to care about in a high-level document like `syntax.bnf`. It we want to be very precise then it will become less readable I fear.

src/parser.ml
fblanqui: It's related. This PR is about wildcard's. I propose that any ident starting with _ is equivalent to _ alone and thus cannot be referred to later, and thus used, so that we could write neither \_x,_x nor &_x in a LHS and &_x in a RHS. Since metavariables can be refered to, they must not start with _.

src/parser.ml
fblanqui: Why not changing this behavior?

tests/OK/292.lp
fblanqui: I don't like this. I prefer https://github.com/Deducteam/lambdapi/pull/293#discussion_r351349302.

src/parser.ml
rlepigre: That is a big change in our lexical conventions, and I don't want to change that on a whim. This needs to be discussed more, and here is not the place. There is also another, potentially bigger issue with the LHS of rewriting rules: we disallow writing type annotations for λ-abstractions, but yet we allow constraining the involved types by forbidding certain λ-variables (those that are introduced with `_` since they are not added to the environment).

fblanqui:
It's not orthogonal as this PR changes the previous behavior. Why not implementing this idea right now in this PR? In https://github.com/Deducteam/lambdapi/pull/293#discussion_r351349302, I propose that any ident starting with _ is equivalent to _ alone and thus cannot be referred to later, and thus used, so that we could write neither λ_x,_x nor &_x in a LHS and &_x in a RHS. Since metavariables can be refered to, they must not start with _.

src/parser.ml
fblanqui: What's the problem? And what's the relation with type annotations?

rlepigre: The problem is that currently we argue that types should not be part of matching, which makes sense. So the user does not annotate the lambdas in rules, which I understand as "there is no constraint on the types". On the other hand, using `_` in a λ-abstraction introduces a constraint in the types of subsequent abstractions. There is an example of that in the test file I added (in the middle of the file). But roughly, if you have `λ x y, ...` in a pattern, then `x` can appear in the type of `y`. But if you have `λ _ y, ...` then the first bound variable cannot appear in the type of `y` (it is not put in the environment of the corresponding metavariable).

tests/OK/292.lp
rlepigre: I don't particularly like it either, but taking different conventions as in every other language might not be the greatest idea.

src/parser.ml
fblanqui: Ok, I see. But I think that your comment: in tests/OK/292.lp is incorrect:
```
rule d2_1 (λx y, &_[y]) → λ_ _,zero // no warning
rule d2_1 (λ_ y, _    ) → λ_ _,zero // no warning
// The difference is that in the former variable "x" may appear in the type
// of the "y" variable. That is not the case in the second line.
```
From the point of view of matching, there should be no difference between the two rules: if a term is reducible by one rule, then it is reducible by the other. Changing the type of d2_1 to ∀x:R,Sx⇒T (or even ∀x:R,∀y:Sx,Txy) doesn't change any thing, as matching ignores types.
On the other, it may be important for checking subject-reduction. As there is no difference between these rules for the point of view of rewriting, when we translate pattern variables to metavariables, we need to take into account all variables, including the variables starting with underscore.
This means that we need to check that https://github.com/Deducteam/lambdapi/blob/ceba0a2383112aab85ec694be68ed4803316a27a/src/sr.ml#L72 is correct from this point of view, but this is completely orthogonal to this PR.

fblanqui: Currently, a pattern variable of arity l is replaced by a metavariable of arity l (k=0 as pattern variables will never be applied after https://github.com/Deducteam/lambdapi/pull/279).

fblanqui: Currently, if I understand well, `λx y, &_[y]` is equivalent to `λ(x:&A)(y:&B[x]),&C[y]` which in turn is translated to `λ(x:?A)(y:?B[x]),?C[y]`, while `(λ_ y, _    )` is equivalent to `λ(x:&A)(y:&B),&C[y]` which in turn is translated to `λ(x:?A)(y:?B),?C[y]`.

Do we need to fix to_m? An alternative is to fix https://github.com/Deducteam/lambdapi/blob/ceba0a2383112aab85ec694be68ed4803316a27a/src/scope.ml#L226 . Currently scope_domain generates a fresh pattern variable for domains (note by the way that we could put any dummy fixed term instead as domains are not matched). If in scoping, we keep track of all variables (those starting with underscore and those not starting with underscore) then, we can use the full environment in domains, and only the variables not starting with underscore else where.

So, again, it would be useful to replace string option by string, and filter out strings starting with underscore when necessary.

rlepigre: > From the point of view of matching, there should be no difference between the two rules: if a term is reducible by one rule, then it is reducible by the other.

Yes, of course. But it leads to a difference when we check SR!

rlepigre: > So, again, it would be useful to replace string option by string, and filter out strings starting with underscore when necessary.

I pretty much agree with all of what you said, but I think it is best to do these changes in a separate PR. The scoping module needs some refactoring and I'm going to get into conflicts with concurrent PRs if I go on now. Anyway, this PR already solves a couple of problems and it improves on the current state of the tool, so I don't see any reason not to merge it (after merging #288, and I rebase over it to cleanup the history).

src/syntax.ml
```diff
@@ -38,8 +38,8 @@ and p_term_aux =
   | P_Wild
   (** Wildcard (place-holder for terms). *)
   | P_Meta of strloc * p_term array
-  (** Meta-variable with the given environment. *)
-  | P_Patt of strloc * p_term array
+  (** Named or unnamed Meta-variable with the given environment. *)
```
fblanqui: Meta-variable -> Pattern-variable

tests/OK/292.lp
fblanqui: At least, could you please change your comment? The notion of semantical equivalence is undefined. And all these rules should really be equivalent. So, it is better to say so and open an issue for fixing the subject-reduction algorithm.

src/syntax.ml
rlepigre: Oups, I modified the wrong comment. Fixed.

tests/OK/292.lp
rlepigre: Done.

fblanqui: I opened https://github.com/Deducteam/lambdapi/issues/294 .

rlepigre:
@fblanqui when we merge PRs, it is better to use the **Squash and merge** feature, rather than the default one. It is accessed under the down arrow at the right of the merge button. This lets you modify the commit message to better reflect the changes, and the whole PR will only be seen as a single commit (which generally makes more sense). If there are several authors, I think they are all added as authors of the resulting commit. And I recommend to also erase the list of separate commits that is included in the comment section. It is generally not useful (many commits of the form "fixed this" or "typo..."), but it might make sense to keep a list in case there are many changes.

tests/OK/292.lp
fblanqui: I prefer to write:
```
// The following rules are equivalent
rule d2_1 (λ_ y, &F[y]) → λ_ _,zero // warning
rule d2_1 (λ_ y, &_[y]) → λ_ _,zero // warning
rule d2_1 (λ_ y, _    ) → λ_ _,zero // no warning
rule d2_1 (λx y, &_[y]) → λ_ _,zero // no warning
```
and not mention than there is currently a difference between the last two rules because this is a bug that will be fixed later, and we may forget to remove this comment when the problem will be fixed.

fblanqui:
Hello @rlepigre . I am happy with this PR. I just would like you to fix https://github.com/Deducteam/lambdapi/pull/293/files?file-filters%5B%5D=.bnf&file-filters%5B%5D=.lp&file-filters%5B%5D=.ml&file-filters%5B%5D=.mly#r351728567. Then you can squash and merge. :-)

tests/OK/292.lp
rlepigre: I just say that they are not the same and exactly explain that. I'm not gonna say that the last one is equivalent because that is simply not true. I can however add a reference to the issue.

rlepigre:
> Then you can squash and merge. :-)

I fixed what you asked, but not exactly in the way you wanted. However I referenced the issue in the test file.

I'll let you squash and merge, this way you can see how that process works (and the button will be the default one for you as well next time). :)

Note that in that case, there is no squash to do, but using the squash and merge feature avoids creating a merge commit.

Issue #267, opened 2019-11-07
doc: sections/structure.md should be updated


Rehan-MALAK:
structure.md is a nice entry for new contributors but currently mentions  :

1) non existing files :
- src/treecstr.ml
- src/dtree.ml
- src/confluence.ml
- src/termination.ml
- tools/bench.ml
2) non existing directories :
- examples/
- proofs/

and doesn't mention :
- src/tree.ml
- src/tree_graphviz.ml
- src/external.ml
- src/hrs.ml
- src/queries.ml
- src/xtc.ml
- src/why3_tactic.ml

fblanqui:
Right. This has been forgotten when we merged @gabrielhdt 's PR. Gabriel, could you please update this file?

fblanqui:
Non existing files and directories should be removed from structure.md.

fblanqui:
tree: compilation of rewriting rules to decision trees
tree_graphviz: representation of trees as graphviz files
external: call to external provers
hrs: translation of .lp files to .hrs files (format of the confluence competition)
queries: handling of queries to the system (type of a term, flags, assertions, etc.)
xtc: translation of .lp files to .xtc files (format of the termination competition)
why3_tactic.ml: why3 tactic (call to external provers)

fblanqui:
I'm going to do the change myself directly.

fblanqui:
Fixed in 1f6ec98.

fblanqui:
The structure of editors/vscode/ should be refined. @Rehan-MALAK could you please do that?

rlepigre:
Frankly, this section of the doc will never remain in sync. I think we should just remove it. Every module is documented and file names are quite self-explanatory.

fblanqui:
No, I think that it is useful to get an idea of the organization of the project for a newcomer. It's not a big deal to keep it in sync. It's just that we didn't do it in the latest PRs. Next time, we will have to make sure that every PR includes the necessary modifications of the documentation.

fblanqui:
Hello @Rehan-MALAK . Could you please update the content of structures.md to explain in one line what is the purpose of the files in editors/vscode/ so that we can close this issue?

PR #295, opened 2019-11-28
doc: explain the purpose of the VSCode files

Rehan-MALAK:
#267


doc/sections/structure.md
```diff
@@ -13,6 +13,17 @@ Structure of directories and files
    + `vim/`: code for vim
    + `vscode/`: code for vscode
 
+     - `.vscode/*.json`: config for launching, debugging, building the extension
+     - `src/*.ts`: the VSCode plugin source code written in the TypeScript language
+     - `package.json`: the manifest of the plugin (activation events, scripts, dependencies, ...)
+     - `tsconfig.json`: TypeScript configuration (directories, ...)
+     - `vscode.proposed.d.ts`: unstable VSCode API (likely to change)
+     - `lp.configuration.json`: specific characters
+     - `syntaxes/lp.tmLanguage.json`: grammar of lambdapi [WIP : currently only all the reserved keywords are highlighted in one unique color]
```
fblanqui: perhaps no need to add "[WIP : ...]"

doc/sections/structure.md
```diff
@@ -13,6 +13,17 @@ Structure of directories and files
    + `vim/`: code for vim
    + `vscode/`: code for vscode
 
+     - `.vscode/*.json`: config for launching, debugging, building the extension
+     - `src/*.ts`: the VSCode plugin source code written in the TypeScript language
+     - `package.json`: the manifest of the plugin (activation events, scripts, dependencies, ...)
+     - `tsconfig.json`: TypeScript configuration (directories, ...)
+     - `vscode.proposed.d.ts`: unstable VSCode API (likely to change)
```
fblanqui: no need to say "unstable" and "(likely to change)"

doc/sections/structure.md
```diff
@@ -13,6 +13,17 @@ Structure of directories and files
...
+     - `lp.configuration.json`: specific characters
+     - `syntaxes/lp.tmLanguage.json`: grammar of lambdapi [WIP : currently only all the reserved keywords are highlighted in one unique color]
+     - `Makefile`
+     - `.gitignore`
+     - `License-vscoq.txt` MIT License mentionning the original author of VSCoq, a similar plugin for Coq
```
fblanqui: Why do we need this License file? I would prefer that the content of this file is included in the file(s) taken from vscoq.

doc/sections/structure.md
Rehan-MALAK: it's not "unstable and likely to change" because of us ... it's in Microsoft side : they say that this API is unstable and likely to change

doc/sections/structure.md
Rehan-MALAK: I let @ejgallego and @rlepigre comment on this one
(I think that's pretty standard to include a License file like this)

doc/sections/structure.md
Rehan-MALAK: fixed in 59541b70

doc/sections/structure.md
fblanqui: I see. Perhaps you should mention here that this file is coming from Microsoft.

doc/sections/structure.md
fblanqui: That's normal indeed to include the License text in the files taken from Vscoq but putting a license file in this directory may suggest that all files are taken from Vscoq which is wrong. That's why it is important to identify that files that are coming from Vscoq.

doc/sections/structure.md
Rehan-MALAK: fixed in 609bd5c
(btw, I didn't mention the fact that it will be regularly updated with ''npx vscode-dts dev'' as this file is git tracked and we will do that ourselves)

fblanqui: It would be good to add this somewhere, in a editors/vscode/README.md file for instance.

doc/sections/structure.md
fblanqui: Hello @ejgallego . Could you please tell us what are the files that have been taken from VSCoq?

ejgallego: Originally the whole plugin was taken, but at this point not much remains.

fblanqui: @ejgallego Do we really need to keep this license file then? If you want, we can add a THANKS.md file where we say that you took some inspiration from VSCoq instead.

ejgallego: I guess you can remove it if you want and indeed place a note on the main `LICENSE` or`THANKS` file; licensing stuff is always complicated and has not definitive answer. Usually it doesn't bother me to keep extra stuff.

But after a review it is safe to say that no original code does remain tho.

PR #297, opened 2019-12-05
[ci] [build] Fix problem with LAMBDAPI invocation

ejgallego:
We fix the bug introduced in c7caa6f44af665ad567ae39b235dc85119b84764
indeed quoting of makefile variables is tricky.

I'd chosen to use `PATH` to locate lambdapi in the `libraries` tests,
however other choices are possible.

I took the opportunity to update the travis and opam files to recent
conventions and compilers.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -20,7 +20,7 @@ doc:
 
 #### Unit tests and sanity check #############################################
 
-LAMBDAPI     = 'dune exec -- lambdapi'
+LAMBDAPI     = dune exec -- lambdapi
 OK_TESTFILES = $(sort $(wildcard tests/OK/*.dk tests/OK/*.lp))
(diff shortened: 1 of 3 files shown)


GNUmakefile
```diff
@@ -20,7 +20,7 @@ doc:
 
 #### Unit tests and sanity check #############################################
 
-LAMBDAPI     = 'dune exec -- lambdapi'
+LAMBDAPI     = dune exec -- lambdapi
```
fblanqui: Is this variable declaration still useful?

fblanqui:
Thanks @ejgallego !

GNUmakefile
ejgallego: I dunno, I guess you can keep it until unless you want to port the tests to dune.

Issue #301, opened 2019-12-17
Bug on extra metavariable in rule RHS


Rehan-MALAK:
Dear all, 
There might be a bug with metavariables :

```dedukti
U : Type.
u : U.
f : U -> U.

(; h has a simple type ;)
h : U -> (U -> U) -> U.

(; hx has a dependent type ;)
def Ux : U -> Type.
[x] Ux x --> U.
hx : x:U -> (Ux x -> U) -> U.

(; rules with LHS type = RHS type = U ;)
def g : U -> U.
[i] g i --> h  u f.
[i] g i --> h  i f.
[i] g i --> hx u f.
[i] g i --> hx i f.
[i] g i --> h  u (dummy : U =>u).
[i] g i --> h  i (dummy : U =>u).
[i] g i --> hx u (dummy : U =>u).
[i] g i --> hx i (dummy : U =>u).
[i] g i --> h  u (dummy     =>u).
[i] g i --> h  i (dummy     =>u).
[i] g i --> hx u (dummy     =>u).
[i] g i --> hx i (dummy     =>u).
```
The last rule works with dedukti but not with lambdapi.
Here is the minimal reproduction of the problem in lambdapi syntax :
```lambdapi
constant symbol U : TYPE
constant symbol u : U
symbol Ux : U ⇒ TYPE
rule Ux _ → U
constant symbol hx : ∀ (x : U), (Ux x ⇒ U) ⇒ U
symbol g : U ⇒ U
// set debug +su
rule g &i → hx &i (λ_  , u)
```

fblanqui:
A temporary solution is to write `rule g &i → hx &i (λ_:U, u)`.

PR #302, opened 2019-12-21
fix checking of uninstantiated metavariables in rules #301

fblanqui:


fblanqui:
Thanks @Rehan-MALAK or @rlepigre to review my PR! :-)

src/extra.ml
```diff
@@ -61,11 +64,16 @@ module Option =
...
+    let get_default : 'a -> 'a option -> 'a = fun d o ->
       match o with
       | None    -> d
       | Some(e) -> e
 
+    let get : 'a option -> 'a = fun o ->
+      match o with
+      | None    -> assert false
+      | Some(e) -> e
+
```
rlepigre: I don't think this change is a good idea for two reasons:
 - `get` is a standard name for your `get_default`,
 - we probably don't want to have your `get` function in the interface. Such functions should be avoided since they can lead to tricky problems. It is by far better to inline it when used, which forces you to say why the `None` case is not accessible.

src/basics.ml
```diff
@@ -178,44 +178,45 @@ let iter : (term -> unit) -> term -> unit = fun action ->
...
-  | Symb(_)    -> ()
-  | Prod(a,b)
-  | Abst(a,b)  -> iter_meta f a; iter_meta f (Bindlib.subst b Kind)
-  | Appl(t,u)  -> iter_meta f t; iter_meta f u
-  | Meta(v,ts) -> f v; iter_meta f !(v.meta_type); Array.iter (iter_meta f) ts
-
-(** {b NOTE} that {!val:iter_meta} is not implemented using {!val:iter} due to
-    the fact this it is performance-critical. *)
+(** [iter_meta b f t] applies the function [f] to every metavariable of [t],
+   and of the type of every metavariable recursively if [b]. *)
```
rlepigre: Here "of" should be "to". Also the comment is not indented correctly.
The `fun b f ->` part can move to the end of previous line.

src/terms.ml
```diff
@@ -240,6 +240,10 @@ type term =
   ; meta_value : (term, term) Bindlib.mbinder option ref
   (** Definition of the metavariable, if known. *) }
 
+(** Comparison function on metavariables. *)
+let cmp_meta m1 m2 = m1.meta_key - m2.meta_key
+let eq_meta m1 m2 = m1.meta_key = m2.meta_key
+
```
rlepigre: Please, give a standard documentation comment to both functions.

src/extra.ml
fblanqui: Fixed in bc53195065b3db652ac59f271d627e03e9c7ab59

src/basics.ml
fblanqui: Fixed in a0113be89c8d0a93481b724398cc2c01ac0b4063. For indentation, I am not sure. Please do it yourself.

src/terms.ml
fblanqui: Fixed in dc44232e9d5f71ca6461469e84c26514bf9c9a32

Rehan-MALAK:
I tried your branch rebased on my "model" Rehan-MALAK/lambdapi@7d3c621c (the one I use to actually formalize stuff in lambdapi) and I'm ok with it. 
For the coding convention, it would be nice to have someday a git hook based on ocpindent or ocaml-format or something else because changing indentations messed up the git history.

src/basics.ml
```diff
@@ -178,44 +178,44 @@ let iter : (term -> unit) -> term -> unit = fun action ->
...
-  iter_meta (fun m -> l := m :: !l) t;
-  List.sort_uniq (fun m1 m2 -> m1.meta_key - m2.meta_key) !l
+  iter_meta b (fun m -> l := m :: !l) t;
+  List.sort_uniq cmp_meta !l
 
-(** [has_metas t] checks that there are metavariables in [t]. *)
-let has_metas : term -> bool =
-  let exception Found in fun t ->
-  try iter_meta (fun _ -> raise Found) t; false with Found -> true
+(** [has_metas t] checks whether there are metavariables in [t], and in the
```
Rehan-MALAK:
```ocaml
(** [has_metas t b] .... 
*)
```
the ```b``` is missing

fblanqui: Fixed in f30ef043af40bcbcd9bb3006bd131321d25e6fc5

fblanqui:
@Rehan-MALAK do you known how to install such a hook? If so, could you do this indeed?

Issue #215, opened 2019-04-18
"let" not supported in right hand side of rewrite rule


gabrielhdt:
A rule of the form
```
rule mult (S &x) &y → let larg = &y in plus larg (mult &x larg)
```
fails raising `Only allowed in terms` although the rhs is a valid term according to `syntax.bnf`.

rlepigre:
That's not a bug, but a feature. This is linked to the fact that annotations on abstractions are mandatory in the RHS (and `let` is compiled down to a redex).

Of course, we can discuss whether having `let` bindings in RHS is useful in practice. My personal feeling about that is: they are not very useful.

gabrielhdt:
They might be useful to introduce sharing in a non linear rhs of a rewrite rule. If I compute `mult x y` and `y` is, say, 54!, I'd like the 54! to be computed only once and not twice.  I thought that doing a let as mentioned above would allow one to compute the 54! only once.

rlepigre:
No, not at all. For now, `let x = t in u` is exactly the same thing as `(λx.u) t`. This might change in the future, but this has implications in the theory so we cannot just add a primitive `let`.

fblanqui:
I don't see your point Rodolphe. If t is 54!, it will be computed only once when reducing (λx.u) t.

Other remarks:
- Gabriel is right: the doc or error message is not clear enough.
- You are right that there is problem because let's have no type annotation and, in RHS, there cannot be metavariables. But this could be fixed. This is the object of https://github.com/Deducteam/lambdapi/issues/169.

rlepigre:
@fblanqui: no, it may be computed any number of times since we are not doing call-by-value evaluation (nor a fully lazy evaluation). There is only some sharing in the case where rewriting rules are involved, but again, not always. At some point Christophe experimented with lazy evaluation, but there was not a massive gain in performances. However, that is something we could consider again at some point.

rlepigre:
I just pushed some improvements to error messages (an positions) in scoping. Now the error should be clear. Now that we have fixed #169 we can choose to allow let-bindings in a RHS, but I don't think this is useful. Any thoughts on that @fblanqui? (Note that my comment above still holds.)

gabrielhdt:
Allowing `let` in rhs seems useful to me, see:
```
rule times (frac &a &b) (frac &c &d) →
  let bv ≔ fst &b in
  let bpr ≔ snd &b in
  let dv ≔ fst &d in
  let dpr ≔ snd &d in
[...]

PR #307, opened 2020-01-31
[WIP] Allowing `let` in rhs

gabrielhdt:
See #215


src/scope.ml
```diff
@@ -367,12 +367,13 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
     | (P_Prod(_,_)     , M_Patt   ) ->
         fatal t.pos "Dependent products are not allowed in a pattern."
     | (P_Prod(xs,b)    , _                ) -> scope_binder _Prod env xs b
-    | (P_LLet(x,xs,t,u), M_Term(_)        ) ->
+    | (P_LLet(x,xs,t,u), M_Term(_)        )
+    | (P_LLet(x,xs,t,u), M_RHS(_)         ) ->
         (* “let x = t in u” is desugared as “(λx.u) t” (for now). *)
         let t = scope env (if xs = [] then t else Pos.none (P_Abst(xs,t))) in
         _Appl (scope env (Pos.none (P_Abst([([Some x],None,false)], u)))) t
     | (P_LLet(_,_,_,_) , _                ) ->
```
fblanqui: please change to `     | (P_LLet(_,_,_,_) , M_LHS(_)                ) ->`

fblanqui:
Please add some test file.

gabrielhdt:
I face a strange behaviour, the following code typechecks
```
rule times (frac &a &b) (frac &c &d) →
  frac
   (N.times &a &c)
   (N.nznatc
    (N.times (fst N.not_zero &b) (fst N.not_zero &d))
    (N.prod_not_zero
     (fst N.not_zero &b)
     (fst N.not_zero &d)
     (snd N.not_zero &b)
     (snd N.not_zero &d)))
```
but this one does not,
```
rule times (frac &a &b) (frac &c &d) →
  let bv ≔ fst N.not_zero &b in
  frac
   (N.times &a &c)
   (N.nznatc
    (N.times bv (fst N.not_zero &d))
    (N.prod_not_zero
     bv
     (fst N.not_zero &d)
     (snd N.not_zero &b)
     (snd N.not_zero &d)))
```
I'll try to get a minimal example

src/basics.ml
```diff
@@ -100,6 +100,10 @@ let eq : term -> term -> bool = fun a b -> a == b ||
     | (Prod(a1,b1), Prod(a2,b2))
     | (Abst(a1,b1), Abst(a2,b2)) -> let (_, b1, b2) = Bindlib.unbind2 b1 b2 in
                                     eq ((a1,a2)::(b1,b2)::l)
+    | (LLet(t1,a1,u1), LLet(t2,a2,u2)) ->
+        let (_, u1, u2) = Bindlib.unbind2 u1 u2 in
+        (* FIXME or equality of substituted? *)
```
rlepigre: No, what you are doing is correct. This is not conversion, but syntactic equality (up to α-equivalence).

gabrielhdt:
This PR for now tries to implement `let` the light way, that is, during type inference, as soon as a `LLet(t,a,u)` is encountered, we substitute `t` in `u` and continue inference.

This process raises issues regarding meta variables, their free variables (e.g. `?0[x,y]`) can be substituted by the `let` rule, yielding non valid variables since they are not applied to distinct bound variables (but some term).

The other way
------------------
the branch [`heavy-let`](https://github.com/gabrielhdt/lambdapi/tree/heavy-let) implements what is done in other proof assistants, that is, contexts with values and unification with context.

rlepigre:
> This process raises issues regarding meta variables, their free variables (e.g. `?0[x,y]`) can be substituted by the `let` rule, yielding non valid variables since they are not applied to distinct bound variables (but some term).

I don't think that's a problem, we just need to use the equivalence `?0[x,y] = (λx.λy.?0[x,y]) x y`. (Inserting the minimal number of β-expansions to avoid as many higher-order unification problems as possible.)

PR #308, opened 2020-02-16
Better emacs mode

gabrielhdt:
A slight improvement on emacs mode
- entering greek letters [à la `cdlatex`](https://www.gnu.org/software/emacs/manual/html_node/org/CDLaTeX-mode.html)
- magic `autoloads` cookies for faster startup of emacs
- conventional emacs packaging
- abbrev mode for user-defined unicode input
- basic indentation engine with `smie`
- basic symbol completion.

This PR mainly sets up a correct emacs structure to allow easily further development, and integration of package into MELPA or alike.


editors/emacs/lambdapi.el
```diff
@@ -18,50 +18,47 @@
...
+     "open"
+     "print"
+     "private"
+     "proof"
+     "proofterm"
+     "protected"
+     "qed"
+     "refine"
+     "reflexivity"
+     "require"
```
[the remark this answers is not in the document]
gabrielhdt: Good idea! But do we want to have `require` highlighted like `type` and so on?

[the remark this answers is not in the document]
Rehan-MALAK: maybe only the queries... ```assert``` ```assertnot``` ```compute``` ```type```

and remove #REQUIRE and ```require```

editors/emacs/Makefile
```diff
@@ -0,0 +1,19 @@
+VERSION = 0.1
+PKG = lambdapi-mode-$(VERSION).tar
+
+$(PKG): $(wildcard *.el)
+	mkdir lambdapi-mode-$(VERSION)
+	cp *.el lambdapi-mode-$(VERSION)
+	tar --create --file $@ lambdapi-mode-$(VERSION)
+	$(RM) -rf lambdapi-mode-$(VERSION)
```
rlepigre: Note that instead of all that you can use (up to typo, I have not tried it):
```
tar --create --file $@ --transform 's,^,lambdapi-mode-$(VERSION)/,' $^
```

rlepigre: Or even:
```
tar --create --file $@ --transform 's,^,$(patsubst %.tar,%,$@)/,' $^
```

fblanqui:
Hi @gabrielhdt . I cannot compile your branch:
```
16:29 ~/src/lambdapi (better-emacs) make
make: command: Command not found
```

gabrielhdt:
Hi @fblanqui, it should be fixed now. You don't have to compile, `make install_emacs` should install the mode; and open emacs as a side effect.

Make sure you remove `(load "lambdapi")` from your emacs config.

fblanqui:
```
11:34 ~/src/lambdapi (better-emacs) make install_emacs
Will not install emacs mode.
```

fblanqui:
In addition, make install_emacs should not open emacs.

gabrielhdt:
> In addition, make install_emacs should not open emacs.

In fact, once this is merged or anything, I will put the package on [melpa](https://melpa.org/), so the installation shall be done with `M-x package-install RET lambdapi-mode`.

fblanqui:
Ok but, meanwhile, don't open emacs after install. Thanks. Or is it necessary for completing the installation?

gabrielhdt:
> Ok but, meanwhile, don't open emacs after install. Thanks. Or is it necessary for completing the installation?

Installing the package requires to load the user configuration file (or at least initialise the package manager), whereas the non interactive mode (i.e. option `--batch`) does not load the user configuration file. So it's either emacs is opened, or a minimal configuration file must be provided, but it might be incompatible with some user's weird configuration. And anyway, in the long run, the package shouldn't be installed with the makefile.

editors/emacs/lambdapi-mode.el
```diff
@@ -0,0 +1,143 @@
...
+  ("||" ?∨) ("\\vee"    ?∨) ("\\/" ?∨)
+  ("~~" ?¬) ("\\neg"    ?¬)
+  ("??" ?∃) ("\\exists" ?∃)
+  ("`a" ?α) ("`b" ?β) ("`c" ?γ) ("`d" ?δ)
+  ("`e" ?ε) ("`h" ?η) ("`i" ?ι) ("`k" ?κ)
+  ("`m" ?μ) ("`n" ?ν) ("`o" ?ω) ("`p" ?π)
+  ("`r" ?ρ) ("`s" ?σ) ("`t" ?τ) ("`u" ?υ)
+  ("`w" ?ω) ("`x" ?χ) ("`y" ?υ) ("`z" ?ζ)
+  ("`C" ?Γ) ("`D" ?Δ) ("`G" ?Γ) ("`L" ?Λ)
+  ("`O" ?Ω) ("`P" ?Π) ("`S" ?Σ) ("`W" ?Ω))
```
fblanqui, line 103: This doesn't seem very standard. Why not converting \alpha into α, etc. People are more used to Latex commands. Isn't there already a mode doing this?

gabrielhdt: Well, it's standard if you use `cdlatex`... but I see what you mean, it is done this way in the Agda mode,
<https://github.com/agda/agda/blob/323f58f9b8dad239142ed1dfa0c60338ea2cb157/src/data/emacs-mode/agda-input.el>

gabrielhdt: `cdlatex` is quite standard for emacs, and I personally find it convenient.

gabrielhdt: And no, it seems there are no mode doing it (at least Agda hasn't found any mode either). But there is the TeX input method (`C-x RET C-\` or `M-x set-input-method` to try it). One possible option would be to merge the TeX input method with the LambdaPi one

fblanqui: What do you think of the TeX input mode? See https://github.com/emacs-mirror/emacs/blob/master/lisp/leim/quail/latin-ltx.el . Could you activate it for LP files? I also found https://github.com/Christoph-D/LaTeX-unicode-math-mode.

gabrielhdt: I initially thought it would be too much, but after all, why not. Your second suggestion is made for TeX modes, and needs the strings to be in math mode (so inside `$`s)

fblanqui: This is not what says https://github.com/Christoph-D/LaTeX-unicode-math-mode

fblanqui: Better: https://github.com/vspinu/company-math
```
(add-hook 'lambdapi-mode-hook (lambda () (setq-local company-backends '((company-math-symbols-unicode)))))
```

gabrielhdt: For the company mode, I tried it, but without success (I can't figure out why); and it's more up to the user to configure this.
Indeed, I misread, but I found a satisfactory solution relying on
<https://elpa.gnu.org/packages/math-symbol-lists.html>.

As of c4a2470, all greek letters are available via their latex command.

fblanqui: With https://github.com/Deducteam/lambdapi/pull/308/commits/c4a24709b661882c810d164338783ea396e34684 my previous hook doesn't work anymore!

fblanqui: @gabrielhdt could you please make your new emacs mode use https://github.com/vspinu/company-math ?

gabrielhdt: @fblanqui, you could before? I never managed to get this company thing working. Can you get what you want using alpha input, though?

fblanqui: I don't understand. It doesn't work anymore even if I go back to master.
Question: with the new emacs mode, do we need to write (load "lambdapi") in our ~/.emacs?

gabrielhdt: What does not work anymore? The new emacs mode does not require any `(load "lamdbapi")` line, when the package is installed, some functions are auto (and lazy) loaded into emacs. To remove the package, `package-delete RET lambdapi-mode-0.1` should do the trick.

fblanqui: First, if I do not add (load "lambdapi") in ~/.emacs, I get no lambdapi mode when I open a LP file.

gabrielhdt: It's normal if you removed my package or haven't installed it

gabrielhdt:
To conclude on unicode input,
- I figured out why `company-math` did not work: `Eglot` messed up `company-backends`, but this should now be fixed;
- the behaviour is the following
  + if `company` is installed, then use `company-math` (I prefer not to force users to use company since it is non standard, and some might prefer other completion tools);
  + otherwise, enrich the input method with more latex translations (greek letters, binary operators, relations).

rlepigre:
From what I understand @gabrielhdt, you do not plan to have the Emacs mode be installed on `make install`. Is that right?

In any case, I think it would be nice to tell dune to install the package somewhere, probably under the `share` section, so that it can be installed locally. Is there any standard folder where Emacs looks for packages locally? If yes, I think that would be a better alternative than using some external package distribution. The reason is that the Emacs package is really linked to Lambdapi, so any change in syntax or new feature will need some modification, and it will be hard to keep these in sync. If you install some version of Lambdapi from Opam, you would like to get the corresponding version of the Emacs package for instance. Not a more recent or an older one. Does this make sense?

gabrielhdt:
@rlepigre, I see what you mean.

First, the MELPA thing contains a recipe to build the package, so if the recipe is "fetch the master branch, compile and install" (which is the usual, see for instance the recipe for company which is
`(company :repo "company-mode/company-mode" :fetcher github)`), then on the MELPA side, it's "setup and forget", and the user only has to update its emacs packages through `(package-refresh-content)`. Modern package manager like `straight` can do it automatically.

But it is true that emacs reads automatically in some places (see the [documentation](https://www.gnu.org/software/emacs/manual/html_node/emacs/Init-File.html)). So we can also provide a file `default.el` or `site-start.el` to be placed in, for instance, `/usr/local/share/emacs/site-lisp/`. I do not understand yet perfectly this aspect.

I think offering both method would be nice.

rlepigre:
I see, thanks @gabrielhdt. According to dune, you need to have a `dune` file that contains something like the following:
```
(install
 (section share_root)
 (files   (lambdapi.el as emacs/site-lisp/lambdapi.el)))
```
Placed in the same folder as the `lambdapi.el` file you want to install.

Of course, you can install several files if needed and so on. And I think Emacs should know about the `site-lisp` folder installed that is under your Opam prefix (where the above install target would place the file).

gabrielhdt:
@rlepigre: now the mode is automatically installed when executing `make install` (and `dune install`, so I think opam installs it as well). If the mode is installed this way, then, provided that `$(opam var share)/emacs/site-lisp` is in Emacs' load path, adding
```emacs-lisp
(load "lambdapi-site-file")
```
is enough. Otherwise, more instructions are in the [readme](editors/emacs/README.md).

The package can also be installed through Emacs' package manager (once it is uploaded on MELPA).

GNUmakefile
```diff
@@ -147,15 +147,10 @@ else
...
 .PHONY: install_emacs
 install_emacs:
-ifeq ($(EMACS),)
-	@printf "\e[36mWill not install emacs mode.\e[39m\n"
-else
-	@cd editors/emacs && $(MAKE) install
-	@printf "\e[36mEmacs mode installed.\e[39m\n"
-endif
+	@printf "\e[36mAdd (load \"$(EMACSDIR)/lambdapi-site-file\") to \
+your Emacs configuration file.\e[39m\n"
```
rlepigre: In that case, I guess that the `install_emacs` target should depend on the `install target` right? Otherwise it only shows the message.

On the other end, I'm wondering if this is even useful now. We could very well remove every mention to emacs in the `Makefile`.

rlepigre:
@gabrielhdt thanks, that's great!

gabrielhdt:
The development is finished, so it can be tested (@fblanqui, @rlepigre, @amelieled) and merged if you are happy.

The instructions are [here](https://github.com/gabrielhdt/lambdapi/blob/better-emacs/doc/sections/editors.md).

doc/DOCUMENTATION.md
```diff
@@ -25,6 +25,8 @@ Table of contents
 
  - [Tactics](sections/tactics.md)
 
+ - [Editing Lambdapi source code](sections/editors.md)
```
fblanqui, line 28: Shouldn't this be in ui.md instead?

fblanqui: And you could rename editors.md into emacs.md as it is specific to emacs.

doc/sections/editors.md
```diff
@@ -0,0 +1,138 @@
...
+----------------------------
+
+Although lambdapi source code can be edited with any text editors, users are
+encouraged to use [Emacs](https://www.gnu.org/software/emacs/),
+[Vim](https://www.vim.org/) or [VsCode](https://code.visualstudio.com/).
+
+### Emacs
+
+`lambdapi-mode` is a major mode for editing lambdapi source code. It requires
+Emacs 26.1 and provides:
```
fblanqui, line 11: 26.1 or higher?

fblanqui:
Currently, if I want to test your emacs mode, I need to do `make install` which will replace my local LP installation. Wouldn't be better to keep a Makefile target `make install_emacs` to just install emacs mode files, while adding `make install_emacs` to the opam installation script?

gabrielhdt:
> Currently, if I want to test your emacs mode, I need to do `make install` which will replace my local LP installation. Wouldn't be better to keep a Makefile target `make install_emacs` to just install emacs mode files, while adding `make install_emacs` to the opam installation script?

Its an advantage to have only on install target, it makes things a lot simpler and clearer. You must be able to revert to any lambdapi using `make install` from another branch, emacs lisp files shouldn't be erased from `$(opam var share)/emacs/site-lisp`. Or are they?

fblanqui:
I'm trying to use your mode but get no feedback on whether there is an error or not.

gabrielhdt:
> I'm trying to use your mode but get no feedback on whether there is an error or not.

Regarding LSP? If that's the case, call `M-x eglot-reconnect`

fblanqui:
Error messages:
```
Loading lambdapi-site-file...done
For information about GNU Emacs and the GNU system, type C-h C-a.
File mode specification error: (void-variable eglot-stay-out-of)
scroll-up-command: End of buffer
scroll-down-command: Beginning of buffer
jsonrpc-error: jsonrpc-error: "No current JSON-RPC connection", (jsonrpc-error-code . 32603), (jsonrpc-error-message . "No current JSON-RPC connection")
```

gabrielhdt:
> Error messages:
> 
> ```
> Loading lambdapi-site-file...done
> For information about GNU Emacs and the GNU system, type C-h C-a.
> File mode specification error: (void-variable eglot-stay-out-of)
> jsonrpc-error: jsonrpc-error: "No current JSON-RPC connection", (jsonrpc-error-code . 32603), (jsonrpc-error-message . "No current JSON-RPC connection")
> ```

Can you give me the version of `eglot` you're using? (with `M-x pkg-info-pkg-version RET eglot`)

fblanqui:
eglot 1.4

fblanqui:
pardon 1.1. je vais mettre à jour.

fblanqui:
How to make emacs update packages as much as possible (like opam upgrade)?

gabrielhdt:
`M-x package-list-packages RET U x`

fblanqui:
Well, I have a problem:
```
Failed to verify signature flymake-1.0.8.el.sig:
No public key for 066DAFCB81E42C40 created at 2019-09-21T19:57:24+0200 using RSA
Command output:
gpg: Signature faite le sam. 21 sept. 2019 19:57:24 CEST
gpg:                avec la clef RSA C433554766D3DDC64221BFAA066DAFCB81E42C40
gpg: Impossible de vérifier la signature : Pas de clef publique
```

gabrielhdt:
> Well, I have a problem:
> 
> ```
> Failed to verify signature flymake-1.0.8.el.sig:
> No public key for 066DAFCB81E42C40 created at 2019-09-21T19:57:24+0200 using RSA
> Command output:
> gpg: Signature faite le sam. 21 sept. 2019 19:57:24 CEST
> gpg:                avec la clef RSA C433554766D3DDC64221BFAA066DAFCB81E42C40
> gpg: Impossible de vérifier la signature : Pas de clef publique
> ```

According to [stackoverflow](https://stackoverflow.com/questions/58202993/emacs-failed-to-verify-signature-archive-contents-sig), you have to `M-x package-install RET gnu-elpa-keyring-update RET`

fblanqui:
I get the same error for this too:
```
Failed to verify signature gnu-elpa-keyring-update-2019.3.tar.sig:
No public key for 066DAFCB81E42C40 created at 2019-09-21T19:55:14+0200 using RSA
Command output:
gpg: Signature faite le sam. 21 sept. 2019 19:55:14 CEST
gpg:                avec la clef RSA C433554766D3DDC64221BFAA066DAFCB81E42C40
gpg: Impossible de vérifier la signature : Pas de clef publique
```

gabrielhdt:
> I get the same error for this too:
> 
> ```
> Failed to verify signature gnu-elpa-keyring-update-2019.3.tar.sig:
> No public key for 066DAFCB81E42C40 created at 2019-09-21T19:55:14+0200 using RSA
> Command output:
> gpg: Signature faite le sam. 21 sept. 2019 19:55:14 CEST
> gpg:                avec la clef RSA C433554766D3DDC64221BFAA066DAFCB81E42C40
> gpg: Impossible de vérifier la signature : Pas de clef publique
> ```

It looks like the elpa page of the package had foreseen this, see the instructions on
<http://elpa.gnu.org/packages/gnu-elpa-keyring-update.html>

fblanqui:
I found the good page to solve this problem of key dates:
https://emacs.stackexchange.com/questions/233/how-to-proceed-on-package-el-signature-check-failure

fblanqui:
I now have eglot 1.5 but still the same problem: no feedback. Yet eglot-reconnect works.

fblanqui:
Wait, yes, it works now!

fblanqui:
Small remarks: the keywords "open", "constant", "injective" are written in red. I would prefer to use the same color as for "infix" and reserve red for errors. Where are the colors set?

gabrielhdt:
> Small remarks: the keywords "open", "constant", "injective" are written in red. I would prefer to use the same color as for "infix" and reserve red for errors. Where are the colors set?

The categories of words are in `lambdapi-vars.el` and the colours are in `lambdapi-mode.el`. Let me commit the change.

amelieled:
And `admit` in yellow ?
It's very confusing with Qed !

fblanqui:
I cannot enter unicode symbols by \alpha, and cannot get company-math running either.

fblanqui:
Yes, admit, you could put it in red! :-)

fblanqui:
Do we have a lambdapi-mode-hook?

gabrielhdt:
Yes there is a `lambdapi-mode-hook`. Can you give me the content of the variable `company-backends`?

fblanqui:
How do I get this?

gabrielhdt:
`M-x describe-variable RET company-backends RET`

fblanqui:
2 remarks on abbrev mode:
- I cannot get it working
- abbrevs are recorded in a file $DIR/.dir-locals.el where $DIR is where emacs has been launched (it may not be related to the file that has been opened)

fblanqui:
```
company-backend is a variable defined in ‘company.el’.
Its value is nil
Local in buffer List.lp; global value is the same.

  Automatically becomes buffer-local when set.

Documentation:
Not documented as a variable.
```

fblanqui:
```
company-backends is a variable defined in ‘company.el’.
Its value is
(company-math-symbols-unicode company-bbdb company-nxml company-css company-eclim company-semantic company-clang company-xcode company-cmake company-capf
                              (company-dabbrev-code company-gtags company-etags company-keywords)
                              company-oddmuse company-files company-dabbrev)
Original value was 
(company-bbdb company-nxml company-css company-eclim company-semantic company-clang company-xcode company-cmake company-capf
              (company-dabbrev-code company-gtags company-etags company-keywords)
              company-oddmuse company-files company-dabbrev)

  This variable is safe as a file local variable if its value
  satisfies the predicate ‘company-safe-backends-p’.
```

gabrielhdt:
Your last message suggests that `company-math` work. Does it? If you hit `M-TAB` after a word?

fblanqui:
Concerning the abbrev mode: I understand that there is a number of limitations on the characters that can be used as abbrevations. For instance, \ and : cannot be used inside an abbreviation.

gabrielhdt:
> 2 remarks on abbrev mode:
> 
>     * I cannot get it working
> 
>     * abbrevs are recorded in a file $DIR/.dir-locals.el where $DIR is where emacs has been launched (it may not be related to the file that has been opened)

For the second remark, I don't think there's an easy solution. Perhaps I'll just remove this thing. Can you try creating abbrevs  with `add-global-abbrev`?

fblanqui:
M-TAB does nothing

gabrielhdt:
For company, two things to check
* type `\al` in the buffer and then `M-x company-math-symbols-unicode`. If company-math is properly set, you'll see the tooltip and all
* type `\al` and `M-x company-complete`

fblanqui:
The 2 things work. But completion should be done automatically, right?

gabrielhdt:
> The 2 things work. But completion should be done automatically, right?

Then I suspect that it depends on your configuration of company, like the timeout before the popup appear &c. (I thought that `M-TAB` would call `company-complete` but apparently not).

fblanqui:
I don't kow how to activate company. For information, what I have in my status bar at the bottom is:
```
LambdaPiU:**-   List.lp  Top (17,0)   Git-master  (LambdaPi Flymake[0 0 48] ElDoc Abbrev)  [eglot: mylib]
```
Nothing about company.

fblanqui:
How to add a hook to company from the lambdapi-mode?

gabrielhdt:
> I don't kow how to activate company. For information, what I have in my status bar at the bottom is:
> 
> ```
> LambdaPiU:**-   List.lp  Top (17,0)   Git-master  (LambdaPi Flymake[0 0 48] ElDoc Abbrev)  [eglot: mylib]
> ```
> 
> Nothing about company.

Ah, add `(global-company-mode)` to your emacs conf file to enable it everywhere. To enable it in the buffer, use `M-x company-mode`.

gabrielhdt:
> How to add a hook to company from the lambdapi-mode?

```emacs-lisp
(add-hook 'lambdapi-mode-hook #'company-mode)
```

fblanqui:
When I do this, "company" appears in the status bar but completion does not work.

fblanqui:
Probably not related but I have the following warning:
```
Warning [flymake List.lp]: Disabling backend flymake-proc-legacy-flymake because (error Can’t find a suitable init function)
```

fblanqui:
When I create a new file from emacs foo.lp, or launche emacs with a non-existing file foo.lp, nothing seems to be checked.

amelieled:
Yes ! Use `M-x eglot` ou `M-x eglot-reconnect` to reconnect (correctly) eglot

gabrielhdt:
Indeed it looks like eglot is not automatically launched or something of the kind, I'll see what I can do.

gabrielhdt:
> When I create a new file from emacs foo.lp, or launch emacs with a non-existing file foo.lp, nothing seems to be checked.

@fblanqui, do you know how the old mode behaved?

fblanqui:
Hi @gabrielhdt . Could you please improve the way indentation works? In particular, I would prefer to come back to the beginning of the line after a newline than having a big tabulation.

gabrielhdt:
> Hi @gabrielhdt . Could you please improve the way indentation works? In particular, I would prefer to come back to the beginning of the line after a newline than having a big tabulation.

Unfortunately, this is linked to how the indentation engine works (the emacs smie for simple minded indentation engine), you can see the same behaviour in the caml mode. You can indent correctly by hitting `TAB` after having written the keyword (such as `symbol`)

gabrielhdt:
@fblanqui, as a workaround, with `electric-indent-mode`, the line is automatically re-indented when typing ':'.

fblanqui:
What should I add in my ~/.emacs to do this automatically?

gabrielhdt:
> What should I add in my ~/.emacs to do this automatically?

You should only ensure that `electric-indent-mode` is activated with lambdapi, either by calling `M-x describe-mode` when editing a lambdapi file, or by trying indenting with `:` (or `RET` in fact). If it is not activated by default, add a hook.

fblanqui:
Can I merge @gabrielhdt ?

gabrielhdt:
@fblanqui, you can now.

rlepigre:
There are broken links in the README, do not merge before this is fixed please.

README.md
```diff
@@ -65,6 +65,13 @@ make install_emacs # Install emacs (>= 26.1) support (needs the eglot package)
 
 **Note:** you can run `lambdapi` without installing with `dune exec -- lambdapi`.
 
+**Note on Emacs:** `make install` installs an Emacs major mode for editing
+`lambdapi` files in `$(opam var share)/emacs/site-lisp`. To load the
+`lambdapi-mode` automatically when editing `*.lp` files, add
+`(load "lambdapi-site-file")` to your `~/.emacs.d/init.el` or `~/.emacs`. If
+`lambdapi-mode` is not activated when editing a file ending in `.lp`, read the
+installation section of [`lambdapi-mode` readme](editors/emacs/README.md).
```
rlepigre, line 73: Broken link.

gabrielhdt: Well spot

fblanqui:
idem

Le 09/04/2020 à 08:32, Gabriel Hondet a écrit :

README.md
```diff
@@ -65,6 +65,13 @@ make install_emacs # Install emacs (>= 26.1) support (needs the eglot package)
```
rlepigre, line 64: And there is still a reference to `make install_emacs` above, please grep for that and remove!

amelieled:
And just one point: is it possible to have 2 commands C-c r and C-c v?
I mean, C-c r is equivalent to "lambdapi check" (so you just have the current goal) and C-c v is the verbose version (for declaration, type and computation information).

So, you can just add these lines in ~/.emacs:
```
(quickrun-add-command "lambdapi"
  '((:command . "lambdapi check")
    (:exec    . ("%c --verbose 1 %s")))
     :mode 'lambdapi-mode)
(add-hook 'lambdapi-mode-hook
	  (lambda () (define-key lambdapi-mode-map (kbd "C-c v") #'quickrun)))

(quickrun-add-command "lambdapi"
  '((:command . "lambdapi check")
    (:exec    . ("%c %s")))
     :mode 'lambdapi-mode)
(add-hook 'lambdapi-mode-hook
	  (lambda () (define-key lambdapi-mode-map (kbd "C-c r") #'quickrun)))
```

README.md
rlepigre: (Weirdly, GitHub does not allow you to comment on code that was not directly modified.)

gabrielhdt:
@amelieled, I prefer not to add these `quickrun` shenanigans (although I like them) to the mode, to keep it simple and avoid too many dependencies; and because it will be replaced by a true interactive mode (some day). I thus prefer to let the user add these things to their configuration file. I can add it to the documentation, though.

gabrielhdt:
If people think it should be a *must*, I can add it. But let's conclude this PR.

README.md
gabrielhdt: It's been removed in 5c5ba3195d5a99

fblanqui:
Thank you @gabrielhdt . Great job!

PR #317, opened 2020-03-25
Fix type inference and unification

fblanqui:


src/core/infer.ml
```diff
@@ -150,18 +144,13 @@ and check : Ctxt.t -> term -> term -> unit = fun ctx t c ->
       check ctx a Type;
       (* We (hopefully) evaluate [c] to a product, and get its body. *)
       let b =
-        let c = Eval.whnf c in
+        let c = unfold (Eval.whnf c) in
```
gabrielhdt: Isn't the `unfold` useless here?

fblanqui: fixed in 225de7aae00d7d62bb41700d48d0a348b7a70034

rlepigre:
@fblanqui it would be nice if you could add a description of what you are doing in this PR. What is the problem that you are solving?

fblanqui:
1) The lines https://github.com/Deducteam/lambdapi/blob/19c53e64d4e0ffb54a9f01307cd6e6d1a5b8a52a/src/core/infer.ml#L101-L109
and
https://github.com/Deducteam/lambdapi/blob/19c53e64d4e0ffb54a9f01307cd6e6d1a5b8a52a/src/core/infer.ml#L156-L163
are simply wrong because they extend the arity of a metavariable in a local context instead of extending it in the context of the metavariable type (metavariables can be instantiated by closed terms only).

2) Unification was not dealing with the cases Meta-Prod and Prod-Meta.

fblanqui:
@rlepigre ok for merging?

src/core/env.ml
```diff
@@ -77,3 +77,34 @@ let of_prod_vars : tvar array -> term -> env * term = fun vars t ->
...
+(** Given a metavariable [m] of arity [n] and type [∀x1:A1,..,∀xn:An,B] (with
+   [B] being a sort normally), [extend_meta_type m] returns
+   [m[x1,..,xn],(∀y:p,q),bp,bq] where p=m1[x1,..,xn], q=m2[x1,..,xn,y], bp is
+   a mbinder of [x1,..,xn] over p, and bq is a mbinder of [x1,..,xn] over q,
+   where [y] is a fresh variable, and [m1] and [m2] are fresh metavariables of
+   arity [n] and [n+1], and type [∀x1:A1,..,∀xn:An,TYPE] and
+   [∀x1:A1,..,∀xn:An,∀y:m1[x1,..,xn],B] respectively. *)
+let extend_meta_type : meta -> term * term
+    * (term, term) Bindlib.mbinder * (term, tbinder) Bindlib.mbinder
+  = fun m ->
```
rlepigre: It would seem that only the two first elements of the result tuple are ever used. I would suggest simplifying this function to only return these two, this way the type will be easier to understand and the doc can be simplified accordingly.

(I'll need to read the doc and code of this function more carefully tomorrow, I'm too tired for that kind of thing now.)

src/core/unif.ml
```diff
@@ -61,19 +60,24 @@ let nl_distinct_vars : term array -> tvar array option =
...
   with Not_a_var -> None
 
-(** [instantiate m ts u] check whether [m] can be instantiated for solving the
-    unification problem “m[ts] = u”. The returned boolean tells whether [m]
-    was instantiated or not. *)
+(** [instantiation m ts u] checks whether, in a problem [m[ts]=u], [m] can be
+   instantiated and returns the corresponding instantiation. It does not check
+   whether the instantiation is closed though. *)
+let instantiation : meta -> term array -> term ->
+                    (term, term) Bindlib.mbinder Bindlib.box option
```
rlepigre: You can use type `tbinder` (exported by `Terms`).

src/core/unif.ml
```diff
@@ -61,19 +60,24 @@ let nl_distinct_vars : term array -> tvar array option =
...
-  && match nl_distinct_vars ts with
-     | None -> false
-     | Some vs ->
-        let bu = Bindlib.bind_mvar vs (lift u) in
-        Bindlib.is_closed bu (* To make sure that there is no non-linear
-                                variable of [vs] occurring in [u]. *)
-        && (set_meta m (Bindlib.unbox bu); true)
+  match instantiation m ts u with
+  | Some bu when Bindlib.is_closed bu -> set_meta m (Bindlib.unbox bu); true
+  | _ -> false
```
rlepigre: I like the idea of unboxing as late as possible, and only when closed. In that case this is actually a no-op.

src/core/env.ml
fblanqui: No, the two last elements are also used in infer.ml. That's why I introduced this function.

src/core/unif.ml
fblanqui: No, its a mbinder I need.

src/core/env.ml
rlepigre: Oh I missed that when I used Ctrl+F for some reason. :)

src/core/unif.ml
```diff
@@ -91,25 +95,22 @@ let rec solve : problems -> unif_constrs = fun p ->
 and solve_aux : term -> term -> problems -> unif_constrs = fun t1 t2 p ->
   let (h1, ts1) = Eval.whnf_stk t1 [] in
   let (h2, ts2) = Eval.whnf_stk t2 [] in
-  if !log_enabled then
-    log_unif "solve %a ≡ %a" pp (add_args h1 ts1) pp (add_args h2 ts2);
+  let t1 = add_args h1 ts1 in
+  let t2 = add_args h2 ts2 in
+  if !log_enabled then log_unif "solve %a ≡ %a" pp t1 pp t2;
```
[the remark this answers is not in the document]
fblanqui: To this respect, the current code is not optimal. See next commit.

src/core/eval.ml
```diff
@@ -45,9 +45,8 @@ type stack = term list
 let rec whnf_beta : term -> term = fun t ->
   if !log_enabled then log_eval "evaluating [%a]" pp t;
   let s = Stdlib.(!steps) in
-  let t = unfold t in
   let (u, stk) = whnf_beta_stk t [] in
-  if Stdlib.(!steps) <> s then add_args u stk else t
+  if Stdlib.(!steps) <> s then add_args u stk else unfold t
```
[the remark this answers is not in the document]
fblanqui: I checked. There is no loss in efficiency. In fact, it can only improve as there are less redundant calls to unfold.

fblanqui:
Hi @rlepigre . I took into account your comments. We should be able to merge now.

rlepigre:
@fblanqui let me have a quick look first.

PR #309, opened 2020-02-20
local let

gabrielhdt:
Introduce local let in the `term` datatype.
Unlike in #307, when inferring the type of a `let x := t in u`, there is no substitution of `t` in `u`. Instead, the following have changed
- contexts contain type declarations as well as definitions, they thus are of the form
  ```
  C ::= x : A :: C | x := t : A :: C | []
  ```
- a unification problem contains equalities as well as a context, and is of the form `C |- t =? u`.

**Note** commits up to 7aefcae are shared with #307.

Issue #298, opened 2019-12-06
Add environments to the assert and assertnot commands


fblanqui:
Allow `assert x1:A1 .. xn:An ⊢ t : A` and  `assert x1:A1 .. xn:An ⊢ t ≡ u`.

After a suggestion by @Rehan-MALAK  in https://github.com/Rehan-MALAK/lambdapi/tree/syntax-lambdaification .

fblanqui:
Hi @Rehan-MALAK . You already implemented this. Could you please make a PR? It would be useful to other people as well. @elhaddadyacine for instance. #309 could be useful (it should be merged very soon first).

fblanqui:
Fixed in #320

Issue #315, opened 2020-03-22
Compatibility between Dedukti 2.7 and Lambdapi


francoisthire:
I raise one more time this issue but I think it would be really nice to actually do something about it. 

So I recall that today, there are two main implementations of Dedukti which are:
- Dedukti (aka Dedukti 2.7) : https://github.com/Deducteam/Dedukti
- Lambdapi (aka Dedukti 3.0) : https://github.com/Deducteam/lambdapi

The problem is: If the development of Dedukti 2.7 is dropped, what will happen of tools using Dedukti 2.7?

Fortunately, for many tools, it is only a matter of syntax and except for one feature (brackets), it seems that the syntax of Dedukti 2.7 is a subset of the syntax of Dedukti 3.0 (old or new). 

The real problem is for tools that uses Dedukti 2.7 as an OCaml library. IIRC, they are five:
- Dkmeta
- Logipedia (the backend part that produces JSON files and proof files that are exported)
- SCT
- Sukerujo (used by Focalize)
- Universo

From what I know, a way to get around this issue on the short term is to have a full compatibility of the syntax between Dedukti 2.7 and Dedukti 3.0 . To achieve this part, I think the easiest way is to have a printer of Dedukti 3.0 to Dedukti 2.7 . This printer could be itself a tool around Dedukti 3.0 maybe? 

On the long term though, this solution may be not well suited because the development of Dedukti 2.7 will be dropped. On the long term, we can rewrite the code from scratch of these tools for Lambdapi, but given the experience of the past two years, I suspect this will not happen any time soon. 

Some effort has been made to have a proper API for Dedukti 2.7 . Maybe an intermediate step before translating the code of these tools from Dedukti 2.7 to Dedukti 3.0 would be to see whether the API we have proposed for Dedukti 2.7 can be implemented (again, as a tool) for Dedukti 3.0 ?

To do so, the three main interfaces that need to be implemented are:

- The top-level commands: https://github.com/Deducteam/Dedukti/blob/another-api/parsing/entry.mli
- The environment (and interface over the kernel of Dedukti 2.7) : https://github.com/Deducteam/Dedukti/blob/another-api/api/env.mli
- The processors (a generic way to process top-level commands using the environment) : https://github.com/Deducteam/Dedukti/blob/another-api/api/processor.mli

This API is not finished but almost. Only the first two interfaces are really important, I think that the last one can be in 90% built on the other two.

[...]

PR #320, opened 2020-03-25
assert convertibility and typing with context

Rehan-MALAK:
@fblanqui @elhaddadyacine please make comments about the syntax you are expecting here  #298 and @fblanqui maybe what you had in mind concerning @gabrielhdt #309

please note that currently, it allows to do

```
assert {(n m:N) (o:O)} f1 n m o ≡ f2 n m o
```
but not
```
assert {n m o} f1 n m o ≡ f2 n m o
```


fblanqui:
For the syntax, I propose:
(assert|assertnot) decl1, .., decln |- a = b
decl := x [: A] [:= t]
(using unicode symbols for |-, = and := )
types and definitions should be optional

src/core/legacy_lexer.mll
```diff
@@ -24,7 +24,7 @@ let locate_lexbuf : Lexing.lexbuf -> Pos.pos = fun lexbuf ->
 
 type token =
   (* Special characters and symbols. *)
-  | L_SQB | R_SQB | L_PAR | R_PAR | ARROW | LARROW | FARROW | DEFEQ | COMMA
+  | L_SQB | R_SQB | L_PAR | R_PAR | LEFTBRA | RIGHTBRA | ARROW | LARROW | FARROW | DEFEQ | COMMA
```
fblanqui: Why do you change the legacy parser? The legacy parser is a parser for the current version of Dedukti. It should not include things that are not in Dedukti.

fblanqui:
Thanks to update syntax.bnf as well.

src/core/legacy_lexer.mll
fblanqui: Well, in this case, this is harmless. So, that's fine.

rlepigre:
> For the syntax, I propose:
> (assert|assertnot) decl1, .., decln |- a = b
> decl := x [: A] [:= t]
> (using unicode symbols for |-, = and := )
> types and definitions should be optional

Yes, I agree with @fblanqui that is a much better syntax.

I thing that I don't agree with in what you have done is the extension of the legacy syntax. Only the standard syntax should be extended, the legacy syntax is only here for compatibility.

Rehan-MALAK:
I added initially the extension of the legacy syntax because this feature was actually used by @barras for some time. It's not that important now.

But if we really care more about compatibility than trying to limit divergences between old and new syntax, there are several things to change in legacy parser. I will add a quick test in #315.

Rehan-MALAK:
> use the argument syntax in the context

yes, the idea was to use what's already done : the fact that arg can be ``x`` or ``(x : X)`` or ``x y (z:X)``
with minimal modifications in core/src/parser.ml as this parser is going to go away anyway in N days

> more general contexts introduced by #309

yes, it's not finished, you guys can merge it now or consider it as a WIP and wait for #309

Rehan-MALAK:
sorry  , just corrected a typo and rebased on current master

fblanqui:
Thanks Rehan!

Rehan-MALAK:
It seems that without write access, it's not possible to trigger a CI build (unless I close/reopen PR or push force another time)
https://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit

Rehan-MALAK:
the commit message is horrible ... can we do something about it, rewrite the history now before it's too late ?

rlepigre:
I think we should not try and extend the mechanism introduced by this PR to contexts containing definitions yet (i.e., in #309). As @Rehan-MALAK noted, this will interact with the parser.

fblanqui:
Don't worry. It is not so important.

rlepigre:
> the commit message is horrible ... can we do something about it, rewrite the history now before it's too late ?

No, that's fine. We have had worse commit messages before...

fblanqui:
But I don't understand the right access you are talking about.

Rehan-MALAK:
well it's really confusing... 

    assert convertibility and typing with context (#320)
    
    * assert convertibility and typing with context
    
      * old syntax :
    
        #ASSERT {(n:N) (m:N) (o:O)} f1 n m o == f2 n m o.
        #ASSERT {(n:N) (m:N) (o:O)} f n m o : P.
    
      * new syntax :
    
        assert {(n m:N) (o:O)} f1 n m o ≡ f2 n m o
        assert {(n m:N) (o:O)} f n m o : P
    
    * no old syntax after all
    
    * introduce unicode ⊢
    
      assert n m (o:N) ⊢ f n m o ≡ f n m o
      assert n (m:N) ⊢ f n m : N ⇒ N

fblanqui:
Ok, we can wait.

Le 26/03/2020 à 14:47, Rodolphe Lepigre a écrit :

PR #321, opened 2020-03-26
add examples in module doc

fblanqui:


fblanqui:
See the rendering in https://github.com/fblanqui/lambdapi/blob/doc/doc/sections/module.md
and https://github.com/fblanqui/lambdapi/blob/doc/doc/sections/options.md

doc/sections/options.md
```diff
@@ -49,21 +49,21 @@ command line arguments and flags.
...
+   Lambdapi has been installed with Opam or `make install` as, in this
+   case, the library root is already known.
+
+ - `--map-dir MOD:DIR` maps an arbitrary directory `DIR` under a
+   module path `MOD` relative to the library root. This option is
+   useful only during the development of a package, before its
+   installation under the library root. It can also be set by using a
+   package configuration file `lambdapi.pkg` (this is necessary for
+   Emacs to know where to find the files). More information on that is
+   given in the section about the module system.
```
rlepigre: The changes you propose here make thing a lot less precise, and in some cases too specific. For instance, there is nothing specific to Emacs here. And there will only ever be ONE library root, so THE library root is the way to go. And there is absolutely no reason to mention possible installation methods, that is not relevant.

Also, there was a reason why I put the `--keep-rule-order` at the end: it is not used very often.

Overall, I'd rollback these changes completely (my English and use of parentheses is fine). And only keep the one piece of useful information that you added: the fact that a package configuration file is mandatory for using the LSP server (this is not specific to Emacs). I would add it as a parenthesis at the end of the sentence.

fblanqui: I strongly disagree Rodolphe. Several other people already faced this problem: Yacine, Catherine, François, etc. You seem the only one to understand how to write, compile and install a project with several files. This is essential that newcomers can do this in a few seconds. Otherwise people will simply stay away from this system.

rlepigre: I'm not saying that the documentation is perfect as it is, only that these particular changes do not improve anything and only make things less precise. It is important to be precise, this is not a tutorial, this is the technical documentation.

Things like "A library root" are a no go, I don't get how you could disagree with me on that.

Also, since I introduced these changes, maybe it would be better to give me the time to fix the documentation and the few bugs that remain myself. Following the discussion about Yacine's work I was thinking of doing that...

fblanqui: It's important to have feedbacks from users and especially newcomers for doing the doc. And we have various newcomers with internships starting now or soon.
So, @elhaddadyacine , @francoisthire , @amelieled , @Rehan-MALAK and anyone else, it would be nice to have your feedback on the doc. Especially, do the proposed examples help you figure out how to write files in LP? Thanks for your comments.

fblanqui: @rlepigre thanks to fix this PR as you want but keep the examples. They are very useful.

rlepigre: I'll take care of that soon.

rlepigre:
See a0c953f59ad59fa70086d4d2a2357cb6b56cba15 for a better documentation and tooling. I created a "getting started" that should be enough to understand how to use the system. Feel free to open issues if you feel there is something missing.

PR #326, opened 2020-04-02
Cpu problem

amelieled:
I've got a problem between the CPU and the lambdapi. I don't know why, but when I opened the file &quot;tp2. lp";, automatically, this process uses 100% of my CPU. After a backup, this process stops and if I try to kill it, it'll duplicate itself. . . So the only way to kill the emacs window is to shut down my computer. 
(See the file in tests/preuves_Ledein/ )


fblanqui:
HI @amelieled . This should not be a PR but an issue. Thanks to create an issue about this problem. See also my mail about our politics concerning branch creation.

PR #328, opened 2020-04-02
Work on inference, unification and subject-reduction

### LINKED ISSUE #303: Bug in rule RHS type inference.
Dear all, 

The following example ``test.dk`` typechecks with Deducteam/Dedukti@e6310f84
```
U : Type.
u : U.
f : (U -> U) -> Type.
def G : U -> Type.
H : U -> Type.
[v] G v --> a : H v -> f (b => u).
```
```bash
$ dkcheck test.dk 
[SUCCESS] File 'test.dk' was successfully checked.
```
while rule RHS type inference fails with lambdapi
```bash
$ lambdapi test.dk 
[test.dk, 6:26-27] Variable [b] could be replaced by [_].
[test.dk, 6:0-34] Uncaught exception [File "src/basics.ml", line 113, characters 36-42: Assertion failed].
```
At some point, Lambdapi is testing equality between metavariable ``v`` and "term environment" ``v`` and then fails.

fblanqui:
fix type inference, unification, handling of implicit arguments in rules, subject-reduction
fix issues #303 #330 #334


src/core/extra.ml
```diff
@@ -239,6 +239,15 @@ module List =
             | _ -> false
       in in_sorted
 
+    (** [insert cmp x l] inserts [x] in the list [l] assuming that [l] is
+       sirted wrt [cmp]. *)
```
gabrielhdt, line 243: sirted -> sorted

src/core/terms.ml
```diff
@@ -143,8 +143,10 @@ type term =
   (** Right hand side (or RHS). *)
   ; arity : int
   (** Required number of arguments to be applicable. *)
-  ; vars  : (string * int) array
-  (** Name and arity of the pattern variables bound in the RHS. *) }
+  ; pvs   : (string * int) array
+  (** Name and arity of the pattern variables bound in the RHS. *)
```
rlepigre, line 147: Having the name here is redundant, it is already available twice in a rule `r`:
 - `Bindlib.mbinder_names r.lhs` (constant time),
 - `Array.map Bindlib.name_of r.vars`.
I would suggest only having an array of integers to indicate the intended arity of each variable.

src/core/terms.ml
```diff
@@ -242,11 +244,12 @@ type term =
...
 
-(** Comparison function on metavariables. *)
-let cmp_meta : meta -> meta -> int = fun m1 m2 -> m1.meta_key - m2.meta_key
+module Meta = struct
+  type t = meta
+  let compare m1 m2 = m1.meta_key - m2.meta_key
+end
 
-(** Equality on metavariables. *)
-let eq_meta : meta -> meta -> bool = fun m1 m2 -> m1.meta_key = m2.meta_key
```
rlepigre, line 249: I would still keep this definition. The `compare` function that you expose in the `Meta` module does not make much sense outside of building sets (there is no useful order on metavariables). I would hence suggest defining the following to avoid exporting your `Meta` module:
```ocaml
module MetaSet = Set.Make(
  struct
    type t = meta
    let compare m1 m2 = m1.meta_key - m2.meta_key
  end)
```

src/core/terms.ml
fblanqui: Why not. But this is not part of this PR. I didn't change anything here (just renamed the old vars into pvs so as to reuse vars for something else). In fact, pvs is only used in xtc (and perhaps hrs too). I think xtc should be rewritten. Perhaps pvs could be completely removed then.

src/core/terms.ml
fblanqui: But eq_meta is used nowhere at the moment...

src/core/terms.ml
rlepigre: No you did change something: the old `vars` is still here.

src/core/terms.ml
rlepigre: It does not hurt to have it, but whatever. However the second point is important.

fblanqui: Meta.compare is used in proof.ml

rlepigre: That was not the point, I still think such a function should not be exposed. And it should also not be exposed by `Proofs.Goal`. What I would propose is to expose a set interface as part of the `Proofs.Goal` module as well. I will make suggestion in the relevant part of the code.

src/core/proof.ml
```diff
@@ -22,6 +22,9 @@ module Goal :
 
     (** [simpl g] simplifies the given goal, evaluating its type to SNF. *)
     val simpl : t -> t
+
+    (** Comparison function. *)
+    val compare : t -> t -> int
```
rlepigre, line 27: I would replace that by:
```ocaml
module Set : Set.S with type elt = t
```
And the corresponding implementation:
```ocaml
module Set = Set.Make(
  type struct
    type t = t (* you might need to do things differently here, I am not sure if this is valid. *)
    let compare {goal_meta = m1; _} {goal_meta = m2; _} = m1.key - m2.key
  end)
```

You can then use sets of goals instead of sorted lists of goals.

rlepigre: And again, there is no order on goals, so we really do not want a `compare` function to be exposed.

fblanqui: In the end, we need lists, not sets, as the order is important (new goals should come first).

fblanqui:
There is a bug I need to fix. I am working on it.

fblanqui:
I finished this PR. Thanks to review it. @Rehan-MALAK @rlepigre

fblanqui:
This PR fixes a number of problems in SR and unification. For instance:
https://github.com/Deducteam/lambdapi/blob/920205298f980bd91a261ac9a67644831a05b104/src/core/terms.ml#L312
https://github.com/Deducteam/lambdapi/blob/920205298f980bd91a261ac9a67644831a05b104/src/core/infer.ml#L155-L172

fblanqui:
And the performances seem a little bit better, at least on `make tests` (28.8s instead of 32.5s on my machine). I didn't check the libraries yet.

fblanqui:
Note that, in sr,
https://github.com/Deducteam/lambdapi/blob/920205298f980bd91a261ac9a67644831a05b104/src/core/sr.ml#L20
was useless

fblanqui:
The code for SR is now ready for being improved by closed Knuth-Bendix completion (see #200).

src/core/terms.ml
```diff
@@ -242,11 +244,13 @@ type term =
...
+module MetaSet = Set.Make(Meta)
+module MetaMap = Map.Make(Meta)
```
rlepigre, line 253: This is unused.

src/core/terms.ml
rlepigre: My point still stands here.

src/core/tactics.ml
```diff
@@ -56,9 +56,12 @@ let handle_tactic : sig_state -> Proof.t -> p_tactic -> Proof.t =
     set_meta m (Bindlib.unbox (Bindlib.bind_mvar (Env.vars env) (lift t)));
     (* New subgoals and focus. *)
     let metas = Basics.get_metas true t in
-    let new_goals = List.map Proof.Goal.of_meta metas in
+    let add_goal m = List.insert Proof.Goal.compare (Proof.Goal.of_meta m) in
+    let new_goals = MetaSet.fold add_goal metas [] in
+    (* New goals must appear first. *)
     Proof.({ps with proof_goals = new_goals @ gs})
   in
+
```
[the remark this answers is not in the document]
fblanqui: 1) Meta keys increase. So metas are ordered in the order of generation. Indeed, using sets didn't break any proof script.
2) Your code snippet is not well typed-checked: a goal is not just a meta.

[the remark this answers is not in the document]
rlepigre: 1. Order of generation is probably not the order we want in presence of instantiation. I don't have any precise example in mind, but the fact that this does not break any proof script is not so surprising since we have basically none.
2. Some `List.map` should do the trick.

fblanqui:
I still have a bug to fix.

src/core/basics.ml
```diff
@@ -75,48 +75,6 @@ let add_args : term -> term list -> term = fun t args ->
...
-    | (TRef(r)    , b          ) -> r := Some(b); eq l
-    | (a          , TRef(r)    ) -> r := Some(a); eq l
-    | (Patt(_,_,_), _          )
-    | (_          , Patt(_,_,_))
-    | (TEnv(_,_)  , _          )
-    | (_          , TEnv(_,_)  ) -> assert false
-    | (_          , _          ) -> raise Not_equal
-  in
-  try eq [(a,b)]; true with Not_equal -> false
-
```
rlepigre, line 115: Why move this to `rewrite.ml`??? This is a general purpose function, it useful independently of the `rewrite` tactic.

rlepigre:
This PR is becoming unreviewable because there are too many unrelated things going on. This is not the right way to do things. Plus CI is broken because now the local library root does not exist...

src/core/basics.ml
fblanqui: It is used in rewrite.ml only.

fblanqui: For testing equality, one should use Eval.eq_modulo.

fblanqui:
This PR fixes many bugs that were in unification and subject-reduction. It affects many things indeed but there is no other solution. It's a difficult topic and I am spending a lot of time solving them.

src/core/terms.ml
```diff
@@ -309,7 +313,7 @@ let fresh_meta : ?name:string -> term -> int -> meta =
 (** [set_meta m v] sets the value of the metavariable [m] to [v]. Note that no
     specific check is performed, so this function may lead to cyclic terms. *)
 let set_meta : meta -> (term, term) Bindlib.mbinder -> unit = fun m v ->
-  m.meta_type := Kind (* to save memory *); m.meta_value := Some(v)
+  m.meta_value := Some(v)
```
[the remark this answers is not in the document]
fblanqui: It was needed at some point. It is back again.

rlepigre:
> This PR fixes many bugs that were in unification and subject-reduction. It affects many things indeed but there is no other solution. It's a difficult topic and I am spending a lot of time solving them.

It is a difficult topic indeed, and this is why careful review is extremely important. This is only possible if useless modifications like moving functions around or removing harmless folders are not intermixed.

src/core/basics.ml
rlepigre: This is not true, the function is also used in the `Why3_tactic` module. This function is used for many things, testing convertibility and testing equality are not the same thing. Equality may be useful for implementing certain tactics, and for optimisation for example. I still disagree.

fblanqui: This is not used in why3_tactic anymore for some good reason.

fblanqui:
@rlepigre, @gabrielhdt  If you have no other comments, I will merge this PR.

src/core/scope.ml
```diff
@@ -477,10 +476,11 @@ let patt_vars : p_term -> (string * int) list * string list =
 let scope_rule : sig_state -> p_rule -> sym * pp_hint * rule loc = fun ss r ->
   let (p_lhs, p_rhs) = r.elt in
   (* Compute the set of pattern variables on both sides. *)
-  let (pvs_lhs, nl) = patt_vars p_lhs in
+  let (pvs_lhs, _nl) = patt_vars p_lhs in
```
gabrielhdt, line 479: `_nl` can be replaced by `_` (less confusing)

src/core/scope.ml
```diff
@@ -328,10 +326,11 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
         begin
           match id with
           | None     ->
-              if List.length env = Array.length vs then
+              let l = List.length env in
+              if l = Array.length vs then
```
gabrielhdt, line 330: It seems that `l` isn't used anywhere else, the two lines can be merged as one

src/core/infer.ml
```diff
@@ -37,7 +37,7 @@ let make_meta_codomain : ctxt -> term -> tbinder = fun ctx a ->
    constraints are satisfied. [ctx] must be well-formed. This function
    never fails (but constraints may be unsatisfiable). *)
 let rec infer : ctxt -> term -> term = fun ctx t ->
-  if !log_enabled then log_infr "infer [%a]" pp t;
+  if !log_enabled then log_infr "infer %a" pp t;
```
gabrielhdt, line 40: Shouldn't the context be printed as well for debug?

src/core/rewrite.ml
```diff
@@ -13,6 +13,51 @@ open Print
...
+    | (TRef(r)    , b          ) -> r := Some(b); eq l
+    | (a          , TRef(r)    ) -> r := Some(a); eq l
+    | (Patt(_,_,_), _          )
+    | (_          , Patt(_,_,_))
+    | (TEnv(_,_)  , _          )
+    | (_          , TEnv(_,_)  ) -> assert false
+    | (_          , _          ) -> raise Not_equal
+    end
+  in
+  try eq [(a,b)]; true with Not_equal -> false
```
rlepigre, line 59: I still think this has nothing to do in the `Rewrite` module. The previous discussion was not over. Again, the point of reviews is to convince the reviewer that they are wrong, or that you are right.

rlepigre: And frankly, I should really not have to review that kind of changes as part of this PR.

src/core/eval.ml
```diff
@@ -49,7 +49,7 @@ type stack = term list
 
 (** [whnf_beta t] computes a weak head beta normal form of the term [t]. *)
 let rec whnf_beta : term -> term = fun t ->
-  if !log_enabled then log_eval "evaluating [%a]" pp t;
+  if !log_enabled then log_eval "evaluating %a" pp t;
```
rlepigre, line 52: Again, I would keep the brackets for readability.

src/core/handle.ml
```diff
@@ -117,7 +117,7 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
         if !(s.sym_def) <> None then
           fatal pr.pos "Rewriting rules cannot be given for defined \
                         symbol [%s]." s.sym_name;
-        Sr.check_rule ss.builtins r; r
+        Sr.check_rule ss.builtins r
```
rlepigre, line 120: It feels weird that SR checking should modify the rule. Why is that? I can understand that it could have an effect on the rule (e.g., instantiate some meta variables), but why should it be allowed to make arbitrary modifications by its interface?

src/core/proof.ml
rlepigre: I still don't like to expose a compare function here.

src/core/basics.ml
rlepigre: Well, someone making a mistake and using the wrong function is not much of a justification for moving `eq`.

src/core/basics.ml
```diff
@@ -4,15 +4,9 @@ open Extra
...
-  struct
-    type t = term Bindlib.var
-    let compare = Bindlib.compare_vars
-  end
-
-module VarSet = Set.Make(Var)
-module VarMap = Map.Make(Var)
+(** Create an array of fresh variables [|x1;..;xn|]. *)
+let fresh_vars : int -> tvar array = fun n ->
+  Bindlib.new_mvar mkfree (Array.init n (Printf.sprintf "x%i"))
```
rlepigre, line 9: I'm not sure how I feel about this function. It is generally a good idea to preserve some kind of name, so that kind of thing should only be done as a last resort. Also, generic names should probably be avoided here. I would rather implement it as:
```ocaml
  Bindlib.new_mvar mkfree (Array.make n "_generated_var_" (* or "_var_" *)) 
```
Note that `Bindlib` should take care of numbering to avoid name clashes. This will happen at the time of binding.

src/core/basics.ml
```diff
@@ -178,13 +182,18 @@ let occurs : meta -> term -> bool =
...
-(** [get_metas b t] returns the list of all the metavariables in [t], and in
-    the types of metavariables recursively if [b], sorted wrt [cmp_meta]. *)
-let get_metas : bool -> term -> meta list = fun b t ->
+(** [add_metas b t ms] extends [ms] with all the metavariables of [t], and
+   those in the types of these metavariables recursively if [b]. *)
+let add_metas : bool -> term -> MetaSet.t -> MetaSet.t = fun b t ms ->
   let open Stdlib in
-  let l = ref [] in
-  iter_meta b (fun m -> l := m :: !l) t;
-  List.sort_uniq cmp_meta !l
```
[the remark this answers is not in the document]
rlepigre: I still think this would be a better idea.

src/core/infer.ml
```diff
@@ -127,54 +135,16 @@ let rec infer : ctxt -> term -> term = fun ctx t ->
...
-      (* We type-check the body with the codomain. *)
-      let (x,t,ctx') = Ctxt.unbind ctx a None t in
-      check ctx' t (Bindlib.subst b (Vari(x)))
-  | t           ->
-      (*  ctx ⊢ t ⇒ a
-         -------------
-          ctx ⊢ t ⇐ a  *)
-      conv ctx (infer ctx t) c
+  if !log_enabled then log_infr "check %a : %a" pp t pp c;
+  conv ctx (infer ctx t) c
```
rlepigre, line 147: If I remember well, this had a significant performance impact. Did you run benchmarks?

src/core/infer.ml
```diff
@@ -14,7 +14,7 @@ let constraints = Stdlib.ref []
 
 (** Function adding a constraint. *)
 let conv ctx a b =
-  if not (Basics.eq ctx a b) then
+  if not (Eval.eq_modulo ctx a b) then
```
rlepigre, line 17: Using equality here had a significant positive impact on performances if I remember well. Conversion on the other end can be very costly, and if it fails it is gonna be tested again partially by unification.

src/core/infer.ml
```diff
@@ -204,8 +174,8 @@ let check : ctxt -> term -> term -> unif_constrs = fun ctx t c ->
   let constrs = Stdlib.(!constraints) in
   if !log_enabled then
     begin
-      log_infr (gre "check [%a] [%a]") pp t pp c;
-      List.iter (log_infr "  assuming %a" pp_constr) constrs;
+      log_infr (gre "check %a : %a") pp t pp c;
```
rlepigre, line 177: I feel that brackets were useful for readability here.

src/core/infer.ml
```diff
@@ -187,8 +157,8 @@ let infer : ctxt -> term -> term * unif_constrs = fun ctx t ->
   let constrs = Stdlib.(!constraints) in
   if !log_enabled then
     begin
-      log_infr (gre "infer [%a] yields [%a]") pp t pp a;
-      List.iter (log_infr "  assuming %a" pp_constr) constrs;
+      log_infr (gre "infer %a : %a") pp t pp a;
```
rlepigre, line 160: Same here, I think having brackets was useful.

src/core/sr.ml
```diff
@@ -102,50 +87,109 @@ let check_rule : sym StrMap.t -> sym * pp_hint * rule Pos.loc -> unit =
...
+    Array.map (function Some m -> m | None -> assert false) lhs_metas
   in
-  let te_envs = Array.map fn metas in
-  let rhs = Bindlib.msubst r.elt.rhs te_envs in
-  (* Infer the type of the LHS and the constraints. *)
+
+  (* Build a map LHS pattern variable -> term_env variable. *)
+  let sym_name m = "&" ^ match m.meta_name with
+                         | Some(n) -> n
+                         | None    -> string_of_int m.meta_key
```
rlepigre, line 99: Please do not use this style, move the concatenation into both branches and have the match on the next line.

src/core/sr.ml
```diff
@@ -102,50 +87,109 @@ let check_rule : sym StrMap.t -> sym * pp_hint * rule Pos.loc -> unit =
...
+  in
+  let map =
+    let map = ref StrMap.empty in
+    let fn i m =
+      map := StrMap.add (sym_name m) (rule.vars.(i), m.meta_arity) !map
+    in
+    Array.iteri fn lhs_metas;
+    !map
```
rlepigre, line 107: I would probably do:
```ocaml
let map = ref StrMap.empty in
let fn x m = map := StrMap.add (sym_name m) (x, m.meta_arity) !map in
Array.iter2 fn rule.vars lhs_metas; !map
```
or
```ocaml
let fn (i, map) m =
  let elt = (rule.vars.(i), m.meta_arity) in
  (i + 1,  StrMap.add (sym_name m) elt map)
in
snd (Array.fold_left fn (0, StrMap.empty) lhs_metas)
```
or I would define an `Array.foldi` in `Extra` and do:
```ocaml
let fn map i m =
  StrMap.add (sym_name m) (rule.vars.(i), m.meta_arity) map
in
Array.foldi fn StrMap.empty lhs_metas
```

src/core/sr.ml
```diff
@@ -102,50 +87,109 @@ let check_rule : sym StrMap.t -> sym * pp_hint * rule Pos.loc -> unit =
...
+    let s = { sym_name  = sym_name m
+            ; sym_type  = m.meta_type
+            ; sym_path  = []
+            ; sym_def   = ref None
+            ; sym_impl  = []
+            ; sym_rules = ref []
+            ; sym_tree  = ref Tree_types.empty_dtree
+            ; sym_prop  = Defin
+            ; sym_expo  = Public }
+    in
```
rlepigre, line 166: This is not the correct indentation, please start the record on its own line.

src/core/scope.ml
```diff
@@ -135,22 +132,22 @@ let get_root : p_term -> sig_state -> sym * pp_hint = fun t ss ->
...
 type mode =
-  | M_Term of metamap Stdlib.ref * expo
+  | M_Term of meta StrMap.t Stdlib.ref * expo
   (** Standard scoping mode for terms, holding a map of metavariables that can
       be updated with new metavariables on scoping and the exposition of the
       scoped term. *)
   | M_Patt
   (** Scoping mode for patterns in the rewrite tactic. *)
-  | M_LHS  of (string * int) list * bool
+  | M_LHS  of (string * int) list * bool * int Stdlib.ref
```
rlepigre, line 141: I don't understand the role of the new reference here. Can you explain? What problem does that solve?
In particular, I don't see what implicit arguments have to do with anything here.

rlepigre: I can only review the rest of the file once I understand what you want to do here.

rlepigre:
> @rlepigre, @gabrielhdt If you have no other comments, I will merge this PR.

I will take care of merging when this is ready. CI seems to not work on the PR, so it needs to be checked directly.

GNUmakefile
```diff
@@ -16,7 +16,7 @@ doc:
 
 #### Unit tests and sanity check #############################################
 
-LAMBDAPI     = dune exec -- lambdapi check --lib-root lib
+LAMBDAPI     = dune exec -- lambdapi check --lib-root /tmp
```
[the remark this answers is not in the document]
fblanqui: The problem is that we need to provide an existing directory for --lib-root even if the files are never going to be installed in this directory. This is annoying.

src/core/sr.ml
fblanqui: No, in this case, I find it better as it clearly shows that the resulting string always starts with '&'.

src/core/scope.ml
fblanqui: Now, you can also have implicit arguments in LHS too! :-)
For instance, with polymorphic lists, you can simply write concat (cons x l) m -> cons x (concat l m)!

GNUmakefile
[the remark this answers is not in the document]
rlepigre: Yeah, but we have to be prepared for the long run here. Moreover, this change also breaks installation of lambdapi. This means that now even an installed version needs to use `--lib-root`. I would advise you to revert that and to reinstate the `example` library.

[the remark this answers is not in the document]
fblanqui: I don't understand. This variable is only used for compiling tests.

src/core/scope.ml
gabrielhdt: Wasn't it already the case?? For instance, I can pick the following rewriting rule
```
symbol lif {U: Set} {V: Set} (S: Set)
: ε (U ⊑ S) ⇒ ε (V ⊑ S) ⇒ Bool ⇒ Deferred.t U ⇒ Deferred.t V ⇒ η S

// lazy if
rule lif _ &pr _ true &t _ → ↑ &pr (Deferred.unquote &t)
```

fblanqui: In your example, the RHS does not depend on the implicit variables of the LHS. In general, implicit variables of the LHS needs to be used in the RHS as well.

GNUmakefile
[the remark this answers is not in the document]
rlepigre: There must be a library root. There is no question there. By removing the example you have also remove their installation under the system library root, and hence the creation of the system library root. That is why you do not just do arbitrary changes in the middle of a PR, this problem could have easily been overlooked: in the current state, installation is broken.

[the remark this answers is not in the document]
fblanqui: Yes, there is a question. In some cases, I simply don't care about the library root. What is the "system library root"? The one under opam? But I don't want to pollute this directory with our tests, and I don't want `make` to compile the files in the old lib/ directory either.

src/core/scope.ml
gabrielhdt: I think it does, if I call with `--verbose 3` and flag implicits on, I obtain
```
lif &#0 &#1 &#2 &pr &#3 true &t &#4 → ↑ ?#0 ?#2 &pr (Deferred.unquote ?#0 &t)
```
the upcast ↑ has implicits as well. Here, the first implicit of the upcast is the first implicit of the `lif`

GNUmakefile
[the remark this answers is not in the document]
rlepigre: The library root is the directory under which packages must be installed. It is essential if we want to be able to provide a notion of installable packages at some further point. I already explained that in my previous PR on the file management stuff. This is why I first intended to distribute a standard library with Lambdapi, but for some reason you seem to think that is stupid.

For the record, I intended the standard library to be developed under the `lib` directory for many good reasons, one being the ability for tests to use this library. This is important for testing features such as the module and file system stuff that are not covered by the current tests. Developing a standard library anywhere else makes no sense to me because it definitely has to remain in sync with the tool.

To be clear, you should never have to care about the library root as a user. I don't understand why you are making such a fuss about it. The only point is that the library root cannot be assumed to exist prior to installation, that is why a path needs to be provided when running Lambdapi as a developer, prior to installation.

In any case, this is not the place to complain about the design I spent hours coming up with when working in a previous PR. This is completely __unrelated__ to the issues at hand, and this PR should simply __not break unrelated stuff__.

src/core/scope.ml
rlepigre: Yeah, implicits were already handled correctly I think. I agree with @gabrielhdt.

fblanqui: No! It was completely buggy. As you can see in your example: the RHS of the rule added in the signature contains uninstantiated metavariables (identifiers prefixed by '?')!!! The master branch doesn't complain but it is wrong. ?#0 should be replaced by say &#0, etc.

GNUmakefile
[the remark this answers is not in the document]
fblanqui: Rodolphe, I never said it was stupid! I was thinking that it could be better to separate the development of the tool from the one of the library. This is what Agda and Lean people do. Testing external libraries when one makes changes is no problem. We already do it for generated libraries. This is also what Coq developers do.

Why did I remove the lib/ directory in my PR? Because make was failing while I was modifying the LP files. It was annoying. So, yes, it is unrelated but I don't think that make should compile any library or examples.

Now, what is broken? It works perfectly well on my machine.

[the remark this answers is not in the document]
rlepigre: Agda and Lean have a big standard library, so this kind of makes sense. For now it would make more sense for us to have everything in one place I think.

You do not have to run `make` to compile Lambdapi since `dune exec - - lambdapi ...` takes care of that already before running. Basically, you never have to run `make` when you modify Lambdapi. And `make tests` should work as well.

What is broken is exactly what I said: the system library root is not created on install now. The only reason you cannot observe this is because you have an installed version of Lambdapi.

src/core/scope.ml
gabrielhdt: So in your version, we have
```
lif &#0 &#1 &#2 &pr &#3 true &t &#4 → ↑ &#0 &#2 &pr (Deferred.unquote &#0 &t)
```
?

rlepigre: But why should it be any different to meta-variables in the domains of lambdas? I think that it should be the role of SR checking to instantiate such meta variables with the proper thing. For now what do you generate exactly instead of meta variables?

rlepigre: Morally, this could be any term depending on the higher-order variables bound in the LHS.

src/core/sr.ml
rlepigre: In that case, at least change it to:
```ocaml
let sym_name m =
  let name =
    match ...
  in
  "&" ^ name
in
```
or
```ocaml
let sym_name m =
  "&" ^
  match ..
```

src/core/scope.ml
fblanqui: @gabrielhdt if you give me a standalone file, I can tell you. You can also check the output by yourself on my branch.

fblanqui: Here is an example;
```
set flag "print_implicits" on
set verbose 3

// Type of set codes

constant symbol Set : TYPE

// Interpretation of set codes in TYPE

set declared "τ"

injective symbol τ : Set ⇒ TYPE

// Type of natural numbers

constant symbol nat : Set

definition N ≔ τ nat

constant symbol zero : N
constant symbol succ : N ⇒ N

// Type of polymorphic lists

constant symbol list : Set ⇒ Set

definition L a ≔ τ (list a)

set declared "□"

constant symbol □ {a} : L a
constant symbol cons {a} : τa ⇒ L a ⇒ L a

set infix 4 "⸬" ≔ cons

// Length of a list

symbol length {a} : L a ⇒ N

rule length □ → zero
and length (_ ⸬ &l) → succ (length &l)

// Concatenation of two lists

symbol concat {a} : L a ⇒ L a ⇒ L a

set infix right 5 "⋅" ≔ concat

rule □ ⋅ &m → &m
and (&x ⸬ &l) ⋅ &m → &x ⸬ (&l ⋅ &m)
```
master branch output:
```
(flag) verbose → 3     
(symb) Set
(conf) declared identifier [τ]
(symb) τ
(symb) nat
(symb) N ≔ τ nat
(symb) zero
(symb) succ
(symb) list
(symb) L ≔ λa, τ (list a)
(conf) declared identifier [□]
(symb) □
(symb) cons
(conf) new infix [⸬]
(symb) length
(rule) length &#0 (□ &#1) → zero
(rule) length &#0 ((⸬) &#1 &#2 &l) → succ (length ?#0 &l)
(symb) concat
(conf) new infix [⋅]
(rule) (⋅) &#0 (□ &#1) &m → &m
(rule) (⋅) &#0 ((⸬) &#1 &x &l) &m → (⸬) ?#0 &x ((⋅) ?#0 &l &m)
Checked [/home/blanqui/src/lambdapi/tests/OK/328.lp]
```
rhs_metas branch output:
```
(flag) verbose → 3     
(symb) Set
(conf) declared identifier [τ]
(symb) τ
(symb) nat
(symb) N ≔ τ nat
(symb) zero
(symb) succ
(symb) list
(symb) L ≔ λa, τ (list a)
(conf) declared identifier [□]
(symb) □
(symb) cons
(conf) new infix [⸬]
(symb) length
(rule) length &#0 (□ &#1) → zero
(rule) length &#1 ((⸬) &#2 &#3 &l) → succ (length &#1 &l)
(symb) concat
(conf) new infix [⋅]
(rule) (⋅) &#1 (□ &#2) &m → &m
(rule) (⋅) &#3 ((⸬) &#4 &x &l) &m → (⸬) &#3 &x ((⋅) &#3 &l &m)
```

GNUmakefile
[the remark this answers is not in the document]
fblanqui: Ok. I have put the lib/ dir back.

src/core/scope.ml
gabrielhdt: All right, my example is getting more and more involved (I'm working on it), and actually it requires your branch (the master raises `assert false`).

rlepigre: @fblanqui I really do not get it, the LHS is fine on the master branch. The problem is with the RHS. However one thing that seems like a regression with your version is that the numbering of pattern variables seems somewhat broken. There is no `#0` or `#1` in the last rule for example.

The existing problem had to do with the RHS, not the LHS. And it indeed seems to have been solved by your PR. However the changes for the LHS are most certainly not what we want. Am I wrong there?

fblanqui: They are just names. It should not be too difficult to fix. This is not a reason to reject this PR.

rlepigre: OK so I did some tests, and I am coming back to something I said long ago when we started to work on implicit arguments: I think that the LHS of rules should always have their implicit arguments explicitly given.

I am starting to understand what you tried to do here @fblanqui. Any pattern variable of the LHS (including those that stand for a wildcard) can actually appear in the RHS through unification of meta-variables corresponding to implicit arguments or wildcards in the RHS. So the optimisation that consisted in only binding the pattern variables that at least appear explicitly in the RHS or are non-left-linear is wrong. We should really bind all the variables, or at least wait until after SR-checking to do the optimisation rightly.

fblanqui: The reason why it starts at 3 is because the slots 0 to 2 are taken by &x, &l and &m.

fblanqui: Exactly. But I don't think that this optim is really useful.

fblanqui:
I have a question: do we assume somewhere in he code that a meta of arity n has a type of the form \forall x1:A1,..,\forall xn:An,B?

src/core/scope.ml
rlepigre: > They are just names. It should not be too difficult to fix. This is not a reason to reject this PR.

Well, nobody is talking about rejecting the PR. This is a regression, it should probably be taken care of while we are at it.

> The reason why it starts at 3 is because the slots 0 to 2 are taken by &x, &l and &m.

I see. It is still weird that the generation of numbers does not go from left to right in the LHS. In any case, I would change printing to get "&%i" when no name, and "&%i_%s" when there is a name, that would be clearer and useful for debugging.

> Exactly. But I don't think that this optim is really useful.

OK, I kind of agree with that. In that case, you should turn the `int option` argument of the `Patt` constructor into an `int`. And the code of rule scoping can probably be simplified a lot. I can help with that if you want.

rlepigre: > I have a question: do we assume somewhere in he code that a meta of arity n has a type of the form \forall x1:A1,..,\forall xn:An,B?

I'm not sure, but most probably.

rlepigre:
> I have a question: do we assume somewhere in he code that a meta of arity n has a type of the form \forall x1:A1,..,\forall xn:An,B?

I'm not sure, but most probably.

src/core/scope.ml
fblanqui: It would affect tree.ml. In particular, None is used in tree.ml line 529. I propose that you or Gabriel (this is his code) do this modification in another PR.

rlepigre: I am looking into it at the moment actually.

fblanqui: As you want but I prefer to merge the PR as it is now for not mixing these modifications with mine. Modifying tree.ml will require more extensive testing. And this PR is waited for by various people. So, if you have no other comment on the current code, then I will merge.

rlepigre: So here are my conclusions after working late on the problem:
 - Removing the option on `Patt` naively does not work, and in fact we should better keep it.
 - I want to keep it so that we can optimise rules after SR-checking. At that point we know what are the pattern variables that are useful and those that are not, so we can apply a transformation there. What we did before was incorrect and would definitely lead to bugs.
 - What I did is define a notion of pre-rule, that contains some extra useful information. SR-checking takes a pre-rule and returns a rule. This way we have all the freedom to optimise or not. And we can also get some useful informations that simplify the implementation of SR-checking.
 - I have basically finished the work, but I want to make sure that I now use Bindlib correctly. The current PR as it stands has quite a few bugs that I exposed during refactoring (missing cases, wrong bindings). I now know how to solve the problem completely, and I will finish my work tomorrow.
 - Roughly, the problem is again that the types of metavariables in a RHS can contain free term environment variables, and lifting assumes that metavariables are closed (including their type). The solution is to do more lifting in some cases.

rlepigre: Here is what I have done so far: https://github.com/rlepigre/lambdapi/commit/d112faaef8e9602235c866eb1859b8a79f6816f8.

rlepigre: I am done @fblanqui. If you agree to the changes in https://github.com/rlepigre/lambdapi/commit/790c0d294735df4ecd4ffb7adca386c71948896d then I will push them here and merge your PR. The performances are almost identical with my changes (although I observed a small, not very significant speedup on both `make tests` and `make focalide`). I will implement the optimisation of rules in another PR, for now there is only a `TODO` in `src/core/sr.ml`.

rlepigre: Note that there is quite a bit more code, but this is necessary to handle variable binding correctly and with less chance of having bugs. The thing you did updating the types of metavariables was certainly not correct in general. Now we have more general scoping for rules, and sanity checks to ensure that metavariables outside of the checking of a RHS cannot contain variables other than those of the environment.

fblanqui: Thanks to check your commit on https://github.com/fblanqui/lib/.

fblanqui: Thanks to merge the last commits I did on rhs_metas too.

rlepigre: All merged, and the library still works fine.

rlepigre:
So can I merge @fblanqui? (I want to do it myself because I want to fix the commit message.)

src/core/extra.ml
```diff
@@ -239,6 +239,15 @@ module List =
...
 
+    (** [insert cmp x l] inserts [x] in the list [l] assuming that [l] is
+       sorted wrt [cmp]. *)
+    let insert : 'a cmp -> 'a -> 'a list -> 'a list = fun cmp x ->
+      let rec insert acc l =
+        match l with
+        | y :: m when cmp x y > 0 -> insert (y::acc) m
+        | _ -> List.rev_append acc (x::l)
+      in insert []
+
```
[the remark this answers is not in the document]
fblanqui: thanks to do it

src/core/infer.ml
fblanqui: This is not clear. We should compare the performances.

src/core/unif.ml
```diff
@@ -270,8 +335,7 @@ and solve_aux : ctxt -> term -> term -> problems -> unif_constrs =
      [t = s^{-1}(v)] when [s] is injective and [ts=[t]]. *)
   let solve_inj s ts v =
     if !log_enabled then
-      log_unif "solve_inj [%a] [%a]" pp (add_args (symb s) ts) pp v;
-    if is_constant s then error ();
+      log_unif "solve_inj %a ≡ %a" pp (add_args (symb s) ts) pp v;
```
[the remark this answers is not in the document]
fblanqui: Isn't \equiv reserved?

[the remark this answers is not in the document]
rlepigre: Of course not! You would complain if it was! ;)

src/core/unif.ml
```diff
@@ -340,16 +404,23 @@ and solve_aux : ctxt -> term -> term -> problems -> unif_constrs =
...
+
+  | (_          , _          ) -> add_to_unsolved ()
+
+(** [solve builtins flag problems] attempts to solve [problems]. If there is
+   no solution, the value [None] is returned. Otherwise [Some(cs)] is
+   returned, where the list [cs] is a list of unsolved convertibility
+   constraints. *)
+let solve : sym StrMap.t -> problems -> unif_constrs option =
+  fun _builtins p ->
   try Some (solve p) with Unsolvable -> None
```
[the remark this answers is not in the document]
fblanqui: I didn't change that part. Indeed, currently, they are not used. But they could be so later.

src/core/sr.ml
```diff
@@ -102,50 +61,184 @@ let check_rule : sym StrMap.t -> sym * pp_hint * rule Pos.loc -> unit =
...
+  let rec symb_to_tenv t =
+    log_subj "symb_to_tenv %a" pp_term t;
+    let (h, ts) = Basics.get_args t in
+    let ts = List.map symb_to_tenv ts in
+    let (h, ts) =
+      match h with
+      | Symb(f,_) when List.memq f syms ->
+          let i =
+            try Hashtbl.find htbl f.sym_name with Not_found ->
+            fatal pos "Introduced symbol [%s] cannot be removed." f.sym_name
```
fblanqui, line 99: I don't understand how this could happen. Please explain. Do you have an example?

src/core/unif.ml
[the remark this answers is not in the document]
rlepigre: In any case, I think we could distinguish the parts of the terms from the rest using colors. I might do that in a further PR. It is true that we are wasting space with brackets, but it is for a good reason.

src/core/sr.ml
rlepigre: Do you at least agree that not all the symbols we introduce may correspond to a term environment variable? If yes, it seems obvious to be that such substituted symbol could escape. If no, can you explain why?

src/core/unif.ml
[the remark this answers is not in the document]
rlepigre: Will they ever be useful here @fblanqui?

src/core/infer.ml
rlepigre: We can't compare anymore because `eq` is in the wrong module now (cyclic depencency).

src/core/sr.ml
fblanqui: Do you mean a symbol representing a metavariable occurring in the type of a pattern variable?

rlepigre: Yes, that is what I have in mind.

fblanqui:
I have a question: do you keep the LHS unchanged after scoping?

src/core/unif.ml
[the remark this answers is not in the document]
fblanqui: I think so but I don't mind if you want to remove them now.

rlepigre:
> I have a question: do you keep the LHS unchanged after scoping?

For now yes, but it will have to change once I implement the optimisation.

src/core/sr.ml
fblanqui: Ok. Then, for the record, thanks to add a comment.

fblanqui:
Ok. The optimisation does not change the matching behavior. It is important to not change this indeed. Only the RHS metavariables should be instantiated by something.

fblanqui:
Thanks to merge @rlepigre . And thanks for your improvements on this PR!

rlepigre:
> Ok. The optimisation does not change the matching behavior. It is important to not change this indeed. Only the RHS metavariables should be instantiated by something.

Yeah, the only thing that the optimisation will do is just not record the term matching a pattern variable in the case where it is not useful in the RHS (and does not appear non-linearly). This means that we will have wildcards again (i.e., terms of the form `Patt(None,...,...)`) when reducing. For now these do not appear anymore I think. This wastes space since we allocate arrays that are bigger than they need to be.

PR #350, opened 2020-04-10
fix #341: no warning if a variable does not occur in the type of a definition

### LINKED ISSUE #341: Wrong warning in scoping
The following code
```
symbol A:TYPE
definition f x : A ≔ x
```
generates the following wrong warning:
```
[/home/blanqui/src/lambdapi/tmp/zo.lp, 2:13-14] Variable [x] could be replaced by [_].
```
The problem comes from the fact that the output type of f, A, indeed does not depend on x, but we should not care in this case. scope_binder should take an additional boolean argument for knowing whether a warning can be output or not.
https://github.com/Deducteam/lambdapi/blob/83ea99311b628742fa78372fb0cd13dfca5099d3/src/core/scope.ml#L251
https://github.com/Deducteam/lambdapi/blob/83ea99311b628742fa78372fb0cd13dfca5099d3/src/core/scope.ml#L267-L268

**Edit 17 March 2021:**
Here is another problematic example with a parametrized inductive type:
```
symbol Set:TYPE;
symbol τ:Set → TYPE;
symbol Prop:TYPE;
builtin "Prop" ≔ Prop;
symbol π:Prop → TYPE;
builtin "P" ≔ π;
(a:Set) inductive 𝕃:TYPE ≔
| □ : 𝕃 a
| ⸬ : τ a → 𝕃 a → 𝕃 a;
// wrong warning: [7:1-2] Variable [a] could be replaced by [_].
```

fblanqui:


rlepigre:
The fix you are implementing now won't do the trick I think. Note that neither of the two definitions in the following example should trigger a warning:
```
symbol A : TYPE
symbol e : A
definition f x : A ≔ x
definition g (x : A) : ((λ _, A) x) ≔ e
```
To scope `definition f <args> : <type>  ≔ <term>` we want a new function in scoping that takes as input:
 - the list of arguments `<args>` with their types,
 - the type `<type>`,
 - the term `<term>`.
The idea is to build an environment using the elements of args, and then to call normal scoping for `<type>` and `<term>` in this environment. Then, only at the time of creating the binders we need to do something special for warnings: for each variable in the environment, check if it is bound in the term and type under construction. **Only give a warning if it is bound in neither.** And in any case, create an abstraction and a product. Repeat for all the other variables of the environment.

PR #352, opened 2020-04-11
Change path of gv files

### LINKED ISSUE #346: Module names and relative paths for symbols tree graphs
There is a utf8 error when someone tries to write the tree of a utf8 symbol:
```
Writing file [personoj/encodings/lhol.ϕ.gv]
[Sys_error("personoj/encodings/lhol.\207\149.gv: No such file or directory")]
```

Do you know how to handle this, @rlepigre?

gabrielhdt:
With the old module system, module hierarchy was replicating the architecture of the filesystem. Writing trees to the module path replacing `.` with `/` was enough.

With the new module system, the replication isn't maintained, so the path of the graph is likely to be wrong. This PR writes the trees to the current working directory, replacing `.` of modules with `%`.

It fixes #346.

**Question:** could we create a command such as `print_tree <symbol>` rather than a cli? In the spirit of CafeOBJ, <https://cafeobj.org/intro/en/parsing.html>.


rlepigre:
Actually, I think we should maybe provide a command `lambdapi decision-tree <MOD_PATH>.<NAME>` that would print on the standard output the decision tree for the symbol. That would be quite easy to do, and we could then remove the old CLI argument which is indeed broken.

src/lambdapi.ml
```diff
@@ -333,6 +355,11 @@ let check_cmd =
   Term.(const check_cmd $ global_config $ timeout $ recompile $ files),
   Term.info "check" ~doc ~man:man_pkg_file
 
+let decision_tree_cmd =
+  let doc = "Prints decision tree of a symbol to standard output." in
```
rlepigre, line 359: Maybe give a bit more details about what format? Also give some help for visualising the result.

src/core/tree_graphviz.ml
```diff
@@ -10,10 +10,11 @@
...
 
-    To create dot files, use the [--write-trees] flag on command line: given a
-    module [M], a symbol [s] of the signature of [M], a file [M.s.gv]
-    containing the decision tree will be created. If [M] is [a.b.c], the file
-    will be [a/b/c.s.gv]. *)
+    To obtain the dot representation of a decision tree, use the
+    [decision-tree] command with a fully qualified symbol name to print the
+    graphviz code on standard output. To display symbol [sym] of module
+    [mod.ule] using [imagemagick],
+    [lambdapi decision-tree mod.ule.sym | dot -Tpng | display]. *)
```
rlepigre, line 17: I think this can be removed completely, provided you put the information in the doc of the command (see my comment there).

src/lambdapi.ml
```diff
@@ -279,6 +294,13 @@ let lsp_log_file : string Term.t =
   in
   Arg.(value & opt string default & info ["log-file"] ~docv:"FILE" ~doc)
 
+(** Specific to the ["decision-tree"] command. *)
+
+let qsym : string Term.t =
+  let doc = "Fully qualified symbol name with dot separated identifiers." in
+  Arg.(required & pos 0 (some string) None
+       & info [] ~docv:"MOD_PATH.SYM" ~doc)
```
rlepigre, line 302: Please, rather use the same thing as everywhere else:
```ocaml
let i = Arg.(info [] ~docv:"MOD_PATH.SYM" ~doc) in
Arg.(required & pos 0 (some string) None i)
```

rlepigre: Actually, you can even do the splitting of the module path here using `Cmcliner.Arg.list` (maybe combined with `Cmdliner.Arg.non_empty`).

rlepigre: And the nicest thing would be that this function takes care of returning a pair of a module path and a symbol name. You can do that easily I think.

src/lambdapi.ml
```diff
@@ -111,6 +109,30 @@ let lsp_server_cmd : config -> bool -> string -> unit =
...
+      let rec get red mp =
+        match red with
+        | [sym] -> (mp, sym)
+        | m::ms -> get ms (m::mp)
+        | []    -> assert false
+      in
+      let (mp, sym) = get (String.split_on_char '.' qsym) [] in
+      (List.rev mp, sym)
+    in
+    Timed.(verbose := 0); (* To avoid printing the "Checked ..." line *)
```
rlepigre, line 126: Please move this before `init cfg;` so that the user can override. Otherwise the user's input for the `--verbose` option is ignored completely. You just want to modify the default behaviour here.

src/lambdapi.ml
```diff
@@ -111,6 +109,30 @@ let lsp_server_cmd : config -> bool -> string -> unit =
...
+    let (mp, sym) =
+      let rec get red mp =
+        match red with
+        | [sym] -> (mp, sym)
+        | m::ms -> get ms (m::mp)
+        | []    -> assert false
+      in
+      let (mp, sym) = get (String.split_on_char '.' qsym) [] in
+      (List.rev mp, sym)
+    in
```
rlepigre, line 125: This should move to argument parsing (see other comment), but if you have the guarantee that the list of items is non-empty I would just do something like:
```ocaml
let (mp,s) = match List.rev path with [] -> assert false | s::mp -> (List.rev mp, s) in
```

src/lambdapi.ml
gabrielhdt: This is intentional, you don't want the "Checked ..." to pipe to `dot`

rlepigre: Yeah I know, but in that case it is weird to allow the `--verbose` option here. I would say, apply my comment for now (we might still want to use `--verbose` together with that command, who knows?). I can always come back later and improve this by changing the available options for this command. (I am planning to work on other commands soon, so I will need to do that anyway. No need to do that now.)

gabrielhdt: Ok. It's a bit weird as the option *must* be used with `--verbose 0` to work properly. Perhaps I can default it to 0 in that command?

rlepigre: That is exactly what I was suggesting by moving the update.

rlepigre: Oh but `--verbose` has the default option... So actually you need to make it so that verbose only carries on option.

src/lambdapi.ml
gabrielhdt: To perform your last suggestion, I'd need a `map` function of type
`('a -> 'b) -> 'a conv -> 'b conv`, right?

rlepigre: No, the full thing is a `string Term.t`, so assuming the current thing is called `arg`, you need to do something like `Term.(pure f $ arg)` where `f` is your function from `string list` to `Path.t * sym`.

gabrielhdt: Nice.

src/lambdapi.ml
```diff
@@ -290,10 +287,15 @@ let lsp_log_file : string Term.t =
...
 
-let qsym : string list Term.t =
+let qsym : (string list * string) Term.t =
   let doc = "Fully qualified symbol name with dot separated identifiers." in
   let i = Arg.(info [] ~docv:"MOD_PATH.SYM" ~doc) in
-  Arg.(non_empty & pos 0 (list ~sep:'.' string) [] & i)
+  let separate lst =
+    let rv = List.rev lst in
+    ((List.rev (List.tl rv)), List.hd rv)
+  in
```
rlepigre: I would prefer
```ocaml
let separate l =
  match List.rev l with
  | []   -> assert false (* Unreachable. *)
  | s::l -> (List.rev l, s)
in
```
since `List.hd` and `List.tl` can raise exceptions.

Also, please use `Path.t` instead of `string list` for the type.

rlepigre: And rather use:
```ocaml
let arg = Arg.(non_empty & pos 0 (list ~sep:'.' string) [] & i) in
Term.(const separate $ arg)
```

src/lambdapi.ml
```diff
@@ -175,14 +175,14 @@ let keep_order : bool Term.t =
...
-let verbose : int Term.t =
+let verbose : int option Term.t =
   let doc =
     "Set the verbosity level to $(docv). A value smaller or equal to 0 will \
      disable all printing (on standard output). Greater numbers lead to more \
      and more informations being written to standard output. There is no \
      difference between the values of 3 and more."
   in
-  Arg.(value & opt int 1 & info ["verbose"; "v"] ~docv:"NUM" ~doc)
+  Arg.(value & opt (some int) None & info ["verbose"; "v"] ~docv:"NUM" ~doc)
```
rlepigre, line 185: I think that just `opt int` would suffice here. But I am not exactly sure.

gabrielhdt: Is the important thing to perform an `Option.iter` in `init`? If so, then the `some` is required to have on `option` type (I discovered cmdliner this morning, so I may be wrong).

src/lambdapi.ml
gabrielhdt: what about the one-liner?

src/lambdapi.ml
rlepigre: Oh yeah, you are right I think. I am still not completely used to it myself. I only had a serious look at it for my PR.

src/lambdapi.ml
```diff
@@ -333,6 +348,13 @@ let check_cmd =
   Term.(const check_cmd $ global_config $ timeout $ recompile $ files),
   Term.info "check" ~doc ~man:man_pkg_file
 
+let decision_tree_cmd =
+  let doc = "Prints decision tree of a symbol to standard output using the \
+             Dot language. Piping to `dot -Tpng | display' displays the tree."
```
rlepigre, line 353: Please start a new line as in other doc strings.

src/lambdapi.ml
rlepigre: Yeah that is not too bad (I was actually considering asking you to do exactly that!), but I don't like to have functions that can raise exceptions outside of an exception handler. I would much rather have the `assert false` this way if a further modification leads to the error happening we will know where it comes from.

src/lambdapi.ml
rlepigre: Actually, I just remembered: if you don't use the `some` thingy` then you need to give the default verbosity value in the command line argument, which is not what we want since it is what was done before...

I checked, the default value is indeed set to `1` in `src/core/console.ml` so what you have here is perfect.

src/lambdapi.ml
```diff
@@ -111,6 +109,21 @@ let lsp_server_cmd : config -> bool -> string -> unit =
...
+  fun cfg (mp, sym) ->
+  let run _ =
+    Timed.(verbose := 0); (* To avoid printing the "Checked ..." line *)
+    init cfg;
+    let sign = Compile.compile false mp in
+    let (sym, _) =
+      try StrMap.find sym Timed.(!(sign.sign_symbols))
+      with Not_found -> fatal_no_pos "Symbol not found."
+    in
+    Tree_graphviz.to_dot !out_fmt sym
```
rlepigre, line 123: Here you should rather use `out 0 Tree.graphviz.to_dot sym`.

src/lambdapi.ml
```diff
@@ -111,6 +109,21 @@ let lsp_server_cmd : config -> bool -> string -> unit =
...
+(** Printing a decision tree. *)
+let decision_tree_cmd : config -> (Path.t * string) -> unit =
+  fun cfg (mp, sym) ->
+  let run _ =
+    Timed.(verbose := 0); (* To avoid printing the "Checked ..." line *)
+    init cfg;
+    let sign = Compile.compile false mp in
+    let (sym, _) =
+      try StrMap.find sym Timed.(!(sign.sign_symbols))
+      with Not_found -> fatal_no_pos "Symbol not found."
```
rlepigre, line 121: Last thing, here I would prefer:
```ocaml
    let sym =
      try fst (StrMap.find sym Timed.(!(sign.sign_symbols)))
      with Not_found -> fatal_no_pos "Symbol not found."
    in
```
This way the logic of the function is easier to identify (even though in that case it is pretty easy already, but I want to encourage that style).

PR #354, opened 2020-04-12
new syntax

fblanqui:


src/core/parser.ml
```diff
@@ -107,8 +107,8 @@ let declared_id = Prefix.grammar declared_ids
 (** The following should not appear as substrings of binary operators, as they
     would introduce ambiguity in the parsing. *)
 let forbidden_in_ops =
-  [ "("; ")"; "."; "λ"; "∀"; "&"; "["; "]"; "{"; "}"; "?"; ":"; "→"; "⊢"
-  ; "⇒"; "@"; ","; ";"; "\""; "\'"; "≔"; "//"; " "; "\r"; "\n"; "\t"; "\b" ]
```
fblanqui, line 111: Why // is forbidden?

francoisthire:
I think it would be nice to modify the `doc` too such as the [syntax](https://github.com/Deducteam/lambdapi/blob/master/doc/syntax.bnf).

src/core/parser.ml
gabrielhdt: Comments?

rlepigre:
One big problem I have here is that the new arrow symbol for rewriting rule is not displayed correctly in the terminal. Is it supposed to be a wide arrow? As I said, we need to pick a symbol that is widely supported. It might look OK in emacs, but it definitely doesn't in my terminal or editor (at least with the fonts I have). For the rest, whatever.

Did you write a sed script for the translation? Your changes do not look very systematic to me, and there are several typos I could spot at a first glance. Plus, it will be a pain if some changes are forgotten in the doc or in some random files.

amelieled:
I agree with the update notation concerning : \Pi, $ and -> for type.
But I think, it's not easy to see the difference between -> for type and -> for rewriting.
Maybe, -> for type is perfect because is like in Coq, and \hookrightarrow for rewriting, because it's readable in emacs and in console (mate-terminale).

fblanqui:
Ok, I replaced the long arrow by the hook arrow. We should be able to merge now.

rlepigre:
@fblanqui you did not answer about using a script for the translation. I cannot review that kind of work directly because we can only see the diff, not the places where something should have been changed and wasn't. So what I did is come up with my own script:
```bash
LP_FILES=`find . -type f -and -not -path "./.git*" -and -name "*.lp"`
DOC_FILES=`find *.md doc -type f -not -name AUTHORS.md -not -name biblio.md`
SRC_FILES=`find src -type f -not -name why3_tactic.ml -not -name lambdapi.ml -and -not -name queries.ml`

grep "Π" -q ${LP_FILES} ${DOC_FILES} ${SRC_FILES} || (echo "Problem with [Π]." && exit 1)
grep "↪" -q ${LP_FILES} ${DOC_FILES} ${SRC_FILES} || (echo "Problem with [↪]." && exit 1)
grep "$" -q ${LP_FILES} ${DOC_FILES} ${SRC_FILES} || (echo "Problem with [$]." && exit 1)

sed -i 's/→/↪/g' ${LP_FILES} ${DOC_FILES} ${SRC_FILES}
sed -i 's/⇒/→/g' ${LP_FILES} ${DOC_FILES} ${SRC_FILES}
sed -i 's/∀/Π/g' ${LP_FILES} ${DOC_FILES} ${SRC_FILES}
sed -i 's/&/$/g' ${LP_FILES} ${DOC_FILES}

sed -i 's/"&/"$/g' ${SRC_FILES}
sed -i "s/'&/'$/g" ${SRC_FILES}
```
It is far from perfect, so I had to do quite a few things by hand afterwards (using sed commands and checking the diffs). In the end, I obtained this: https://github.com/rlepigre/lambdapi/tree/new_syntax.

It contains basically all the relevant changes made here (I did not reproduced unrelated changes like capitalising module names), plus many missing ones. As I mentioned, this PR introduces typos by wrongly removing space before the word "and" in some places. And there are also missing translations from the long arrow to the hook arrow. If you want to see what they are use `git diff --word-diff rodolphe/new_syntax` assuming `rodolphe` points to my fork on your system.

fblanqui:
Thanks to update tools/listings.tex as well.
Then, thanks to merge your branch and close this issue.

rlepigre:
OK, but I'm not sure what you mean with `tools/listings.tex`, I think it is up-to-date in my branch. In yours there is no squiggly arrow. Is there anything missing there?

fblanqui:
Right. Thank you to merge your branch then.

PR #356, opened 2020-04-16
Change some of the reserved symbols (improves over #354)

rlepigre:
The syntax changes introduced are the following:
 - `→` is replaced by `↪` in rewriting rules,
 - `&` is replaced by `$` for pattern variables in rewriting rules,
 - the syntax `rule ... and ...` becomes `rule ... with ...`,
 - `⇒` is replaced by `→` for implication, and
 - `∀` is replaced by `Π` for the dependent product type.


rlepigre:
@gabrielhdt can you quickly review the changes to the Emacs mode and approve the PR if that is alright?

PR #361, opened 2020-04-18
Unification rules

gabrielhdt:
Based on <http://www.cs.unibo.it/~asperti/PAPERS/tphol09.pdf>

Unifications hints are seen as rewrite rules on a special symbol [=?]. If unification [T =? U] fails, the problem is rewritten using hints.

Implementation wise,
* A unification hint of the form `hint t1 ≡ u1 ↪ t2 ≡ u2, t3 ≡ u3 ` is parsed as a rewrite rule on the (special) symbol `≡`
* Scoping is similar to scoping a rule, except that SR is replaced with another set of checks (which are all in `scope.ml` for now)
* Symbols `≡` and `,` are defined in `terms.ml` in sub-module `Unif_hints`
* Hints are used in `unif.ml`: before adding a unification problem to "unsolved", we look for a hint that matches the problem. If so, the problem is rewritten to the RHS of the hint, and the sub-problems composing the RHS are added as new to-be-solved problems.

There are some design issues to discuss here, particularly regarding management of "special" symbols.


src/core/eval.ml
```diff
@@ -47,6 +47,11 @@ let appl_to_tref : term -> term = fun t ->
 (** Abstract machine stack. *)
 type stack = term list
 
+(** FIXME: unification hints must be allowed on types, which may involve
```
fblanqui, line 50: must -> should ... on types too

gabrielhdt: This is outdated wrt yesterday's PR isn't it?

doc/syntax.bnf
```diff
@@ -112,6 +113,10 @@ reserved "why3"
 
 // TODO rule naming, positive / negative
 
+// Unification hint
+<hint> ::=
+  | <patt> "≡" <patt> "↪" <pvar> "≡" <term> {"," <term> "≡" <term>}*
```
fblanqui, line 118: Why do you restrict to <patt> and <pvar>? Why not using <term> every where?

src/core/handle.ml
```diff
@@ -126,6 +126,14 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
       let syms = List.remove_phys_dups (List.map (fun (s, _, _) -> s) rs) in
       List.iter Tree.update_dtree syms;
       (ss, None)
+  | P_hint(h)                    ->
```
fblanqui, line 129: Thanks to call it P_unif_hint because we may have other sorts of hints.

doc/syntax.bnf
gabrielhdt: I'm afraid I don't understand the question.

fblanqui: Where <hint> is used is not defined. It should be part of <command>:
```
| set "unif" <term> "≡" <term> "↪" <term> "≡" <term> {"," <term> "≡" <term>}*
```

fblanqui:
Because I forgot to put some words between backquotes. This is fixed now.

Le 23/04/2020 à 09:41, Gabriel Hondet a écrit :

doc/syntax.bnf
fblanqui: You can reject unification rules that cannot be encoded into a rewrite rule at scoping time.

fblanqui: I prefer using set "unif_rule" for not adding yet another reserved word in the grammar, and also because it is just a way to modify the unification algorithm.

fblanqui: Ah, I see now why you want to have a `<patt>`. Note however that, in parser.ml, a rule is parsed as term*term. LHS that are not patterns are rejected at scoping time. But I don't understand why you have `<pvar>`.

src/core/terms.ml
```diff
@@ -263,6 +263,34 @@ type ctxt = (term Bindlib.var * term * term option) list
...
+
+  (** Symbol representing an atomic unification problem. The term [atom t u]
+      represents [t =? u]. *)
+  let atom : term =
+    let sym =
+      { sym_name="unif_atom"; sym_type=ref Kind; sym_path=[]
+      ; sym_def=ref None; sym_impl=[];sym_tree=ref Tree_types.empty_dtree
+      ; sym_prop=Defin; sym_expo=Public; sym_rules=ref [] }
+    in
+    Symb(sym, Binary("≡"))
```
fblanqui, line 277: It would be better that atom and list return symbols instead. Then, you won't need to_sym anymore. pp_hints are going to be removed soon anyway (see #363).

src/core/terms.ml
```diff
@@ -263,6 +263,34 @@ type ctxt = (term Bindlib.var * term * term option) list
...
+
+  (** Symbol for a list of atoms. *)
```
fblanqui, line 279: More explanations would be welcome here.

src/core/terms.ml
```diff
@@ -263,6 +263,34 @@ type ctxt = (term Bindlib.var * term * term option) list
 (** Type of a list of unification constraints. *)
 type unif_constrs = (ctxt * term * term) list
 
+(** Some definitions for unification hints. *)
```
fblanqui, line 266: More explanations would be welcome here. Mainly, you define symbols that will be used to rewrite unification constraints. Why not moving this module in unif.ml?

src/core/unif.ml
```diff
@@ -142,8 +173,10 @@ and solve_aux : ctxt -> term -> term -> problems -> unif_constrs =
 
   let add_to_unsolved () =
     if Eval.eq_modulo ctx t1 t2 then solve p else
-    (* Keep the context *)
-    solve {p with unsolved = (ctx,t1,t2) :: p.unsolved}
+    match try_hints ctx t1 t2 with
+    | None     -> solve {p with unsolved = (ctx,t1,t2) :: p.unsolved}
+    | Some([]) -> assert false (* Hint shouldn't have been necessary *)
```
fblanqui, line 178: I don't understand the comment. Why not simply saying that this cannot happen since, by definition, a unification rule always generates at least one unification constraint.

src/core/scope.ml
```diff
@@ -603,6 +608,157 @@ let scope_rule : sig_state -> p_rule -> pre_rule loc = fun ss r ->
...
+    rewriting rule that rewrite a unification problem into one or more
+    sub-unification problems. *)
+let scope_hint : sig_state -> p_hint -> rule loc = fun ss h ->
+  let (p_l, p_rs) = h.elt in
+  (* NOTE: in the following comments, we consider a unification hint of the
+     form P ≡ Q → x1 ≡ H1, ..., xn ≡ Hn. *)
+  (* Compute the pattern variables. *)
+  let data =
+    (* Get pattern variables of [l] and [r] and verify that they are
+       linear. *)
```
fblanqui, line 622: Why enforcing linearity?

fblanqui: In the following, why duplicating the code of scope_rule and not trying to reuse it instead? Couldn't you see l=r --> t1=u1,..,tn=un as just syntactic sugar for the p_rule (\#unif l r) --> \#list (\#unif t1 u1) .. (\#unif tn un) and modify find_sym function to return the symbols mapped to \#unif and \#list?

gabrielhdt: I think linearity is required to have desirable properties. I should check the paper and proofs, though.

fblanqui: Perhaps. But I prefer let this to the user for the moment.

doc/syntax.bnf
gabrielhdt: The form of a hint is always `?x == term`. There is an error in my line, there shouldn't be `<term> == <term>`.

src/core/scope.ml
gabrielhdt: And we have some other checks to perform. I think reusability of the code is a real question here: on one hand, duplicating involves more code, but it can be better to duplicate code and specialise it rather than generalise and twist the already existing code for unification hints.

fblanqui: Right but I see no reason here. This feature is very experimental. I propose you to do no check at all on unification rules for the moment. Later, if needed, we could add a function for checking some conditions on unification rules, like we do with Sr.check_rule for rewriting rules.

fblanqui: No need to modify find_sym after all. And it's better not to do so anyway.

fblanqui:
We could use an internal symbol \#prod to have unification rules on products as well. At scoping time once for unification rules, and before application of unification rules, a product \Pi x:a,b can be replaced by \#prod a (\lambda x:a,b), and \#prod can be given the rule \#prod $a $b --> \Pi x:$a, $b x, to get back a product.

src/core/scope.ml
gabrielhdt: OK, in the end we'll have to make a `Sr` module counter part for unification hints.

fblanqui:
@gabrielhdt you also need to register unification rules in the signature, and add them in the current sig_state when a module is open.

gabrielhdt:
> @gabrielhdt you also need to register unification rules in the signature, and add them in the current sig_state when a module is open.

@fblanqui, registering amounts to using `add_rule`, doesn'it? And shouldn't we import unification rules when a module is opened *or* required?

src/core/sign.ml
```diff
@@ -273,7 +273,9 @@ let read : string -> t = fun fname ->
     its dependencies. *)
 let add_rule : t -> sym -> rule -> unit = fun sign sym r ->
   sym.sym_rules := !(sym.sym_rules) @ [r];
-  if sym.sym_path <> sign.sign_path then
+  if sym.sym_path <> sign.sign_path &&
+     not (sym == Unif.Hint.atom || sym == Unif.Hint.list)
+  then
```
rlepigre, line 278: I think this change should not be necessary if things are done correctly with your special symbols. I think it is better to give them impossible paths (e.g., the empty path), and not to rely on specific code here.

rlepigre:
> > @gabrielhdt you also need to register unification rules in the signature, and add them in the current sig_state when a module is open.
> 
> @fblanqui, registering amounts to using `add_rule`, doesn'it? And shouldn't we import unification rules when a module is opened _or_ required?

Yes, this is indeed useless. The signatures can declare rewriting rules for external symbols, and this is the mechanism @gabrielhdt needs to rely on here. Together with the fact that the special symbols that Gabriel is considering are part of a kind of "ghost signature".

However, one thing that must be done is to make sure that `Sign.link` works as expected for the symbols of the "ghost signature". To do that, I would actually create a signature for them and add it to the loading path. This way, the corresponding symbols will be looked up normally, without any modification to the `Sign` module.

gabrielhdt:
If I recap, the two things that must be done in priority on this PR is,

* process unification rules as standard rewriting rules on the `#unif_atom` symbol (*temporarily* discarding all checks performed on these unification rules)
* manage signature properly.

Sounds good, @fblanqui, @rlepigre?

fblanqui:
It's fine to consider unification rules as rewrite rules in order to reuse the rewriting engine. But do we want them to be treated in the same way wrt require? Currently, when you require a module, rewrite rules are automatically included in the environment. Do we want the same behavior for unification rules?

src/core/unif.ml
```diff
@@ -33,6 +33,38 @@ let pp_problem oc p =
 let no_problems : problems =
   {to_solve  = []; unsolved = []; recompute = false}
 
+(** Definition of symbols used to encode unification rules. *)
+module Hint = struct
+
```
rlepigre, line 38: Here you need to create a signature with `Sign.create`, and immediately register it into the `Sign.loaded`. So you need to pick a unique module path for your ghost signature. Maybe we should reserve module paths starting with a an empty member? So maybe use `[""; "hint"]`?

Once you have done that you need to register your symbols using `Sign.add_symbol` (or to manually insert them if the function is not suitable for some reason). Doing all this will ensure correct linking by `Sign.link`.

I think you might need to make sure that the "ghost signature" is always open in the case where its symbols can be entered by the user. But you may also go around that in some other way. I am not sure. Is the user intended to have access to these symbols actually?

rlepigre:
> If I recap, the two things that must be done in priority on this PR are,
> 
> * process unification rules as standard rewriting rules on the `#unif_atom` symbol (_temporarily_ discarding all checks performed on these unification rules)
> * manage signature properly.
> 
> Sounds good, @fblanqui, @rlepigre?

Yeah, for the second point see my comments.

> It's fine to consider unification rules as rewrite rules in order to reuse the rewriting engine. But do we want them to be treated in the same way wrt require? Currently, when you require a module, rewrite rules are automatically included in the environment. Do we want the same behavior for unification rules?

I think we do: when you give hints in a file, you expect them to apply in every file that depends on it.

gabrielhdt:
> It's fine to consider unification rules as rewrite rules in order to reuse the rewriting engine. But do we want them to be treated in the same way wrt require? Currently, when you require a module, rewrite rules are automatically included in the environment. Do we want the same behavior for unification rules?

I think we want this behaviour. Otherwise, you would have weird behaviours, such as having terms that type check in some modules and not in some others. For instance, considering the problem of inversing rule, you would have `eq true false` that type checks where the hint is defined, but if you only require the file as, say `F`, `F.eq F.true F.false` wouldn't type check.

rlepigre:
I agree with @fblanqui that this approach is OK, even though it is kind of a big hack. But we need to make sure that this is not too invasive in the end by only having limited changes to the core components. For now it seems that this is going in the right direction @gabrielhdt, I an quite confident we can make this work in a way that is not too ugly! ;)

fblanqui:
Well, we can start implementing this behavior for the moment and see later if we need to change it. It would be useful then to have a command for printing all the rules for a given symbol (#375). This makes me think that this is not the current behavior for builtins, which is strange, isn't it?

rlepigre:
> Well, we can start implementing this behavior for the moment and see later if we need to change it. It would be useful then to have a command for printing all the rules for a given symbol (#375). This makes me think that this is not the current behavior for builtins, which is strange, isn't it?

Builtins are quite different, they are a syntactic component, and it is very natural to want to override them (e.g., you might have a different notion of natural numbers), or to simply not have them. On the other hand, there is no good reason not to want to import the unification hints. In a way they are very similar to rewriting rules, and I cannot think of any situation where you would want to disable them. Once you have a signature in your scope, unification problems can involve any of its symbols, and so they can benefit from any unification hint available.

fblanqui:
Gabriel, it could be the case that some unification rules are incompatible and lead to non-termination. Perhaps, in the future, we may need a more fine grained control of what is imported in the environment when we do require [open]. But let's keep this for the future. FYI, concerning unification, @Rehan-MALAK is currently working on this topic. He will develop a tactic language for solving unification constraints interactively or automatically. Unification rules will be seen as basic tactics then.

rlepigre:
> It could be the case that some unification rules are incompatible and lead to non-termination. Perhaps, in the future, we may need a more fine grained control of what is imported in the environment when we do require [open]. But let's keep this for the future. FYI, concerning unification, @Rehan-MALAK is currently working on this topic. He will develop a tactic language for solving unification constraints interactively or automatically. Unification rules will be seen as basic tactics then.

As for rewriting rules, this can be controlled by registering unification hints in an auxiliary module and only requiring it when you want to have these hints. This use case is already supported by the approach that is implemented here.

src/core/unif.ml
gabrielhdt: No, I don't think the user should have access to these symbols. Or I don't see the need.

rlepigre: Then I am confused about this `Syntax.Unif_hints` module. What is it for?

gabrielhdt: Ah it must be completely useless with my last commit.

gabrielhdt:
Now,
* unification rules are much closer to rewriting rule, the only difference is that SR is not checked
* their syntax is lightened for now, a unification rule is declared with
    `set unif_rule u_atom t u --> u_list (u_atom v w) ...`
* `sig_state` always have the symbols of the "ghost" signature in scope

src/core/handle.ml
```diff
@@ -263,6 +263,21 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+            (* Approximately same processing as rules without SR checking. *)
+            let pur = (scope_rule ss h).elt in
+            let urule =
+              { lhs = pur.pr_lhs
+              ; rhs = Bindlib.unbox (Bindlib.bind_mvar pur.pr_vars pur.pr_rhs)
+              ; arity = List.length pur.pr_lhs
+              ; arities = pur.pr_arities
+              ; vars = pur.pr_vars }
+            in
+            Sign.add_rule Unif.Hint.sign Unif.Hint.atom urule;
```
rlepigre, line 276: The rule should be rather added to the current signature here, right?

gabrielhdt: Ah yes. It seems to me that it does not make a difference, but it keeps things tidier.

src/core/parser.ml
```diff
@@ -564,6 +564,7 @@ let parser config =
       sanity_check _loc_id id;
       Prefix.add declared_ids id id;
       P_config_ident(id)
+  | "unif_rule" h:rule -> P_config_unif_rule(h)
```
rlepigre, line 567: I actually preferred the previous syntax, even though it required more code. The problem here is that you would need to declare infix symbols to get what you had before. That is annoying. Better have some specific scoping handling I think.

src/core/handle.ml
gabrielhdt: Except that I must add the hint signature to the dependencies of all signatures.

rlepigre: Yeah, there is no avoiding that I think.

src/core/parser.ml
gabrielhdt: Me too, I'll come back to it when we are settled.

src/core/pretty.ml
```diff
@@ -227,6 +227,7 @@ let pp_command : p_command pp = fun oc cmd ->
         | Assoc_right -> " right"
       in
       out "set infix%s %f \"%s\" ≔ %a" a p s pp_qident qid
+  | P_set(P_config_unif_rule(ur))   -> pp_p_rule true oc ur
```
fblanqui, line 230: You need to print "set unif_rule " first.

src/core/sign.ml
```diff
@@ -26,7 +26,8 @@ type t =
 
 (** [create path] creates an empty signature with module path [path]. *)
 let create : Path.t -> t = fun sign_path ->
-  { sign_path; sign_symbols = ref StrMap.empty; sign_deps = ref PathMap.empty
+  { sign_path; sign_symbols = ref StrMap.empty
+  ; sign_deps = ref (PathMap.singleton [""; "unif_rule"] [])
```
fblanqui, line 30: It would be better to introduce a function Path.ghost s = [""; s].

src/core/unif.ml
```diff
@@ -33,6 +33,30 @@ let pp_problem oc p =
 let no_problems : problems =
   {to_solve  = []; unsolved = []; recompute = false}
 
+
+(** Definition of symbols used to encode unification rules. *)
+module Hint = struct
+
+  let sign =
+    let pth = [""; "unif_rule"] in
```
fblanqui, line 41: Use Path.ghost instead

src/core/unif.ml
```diff
@@ -33,6 +33,30 @@ let pp_problem oc p =
...
+    let s = Sign.create pth in
+    s.sign_deps := Files.PathMap.empty;
+    Sign.loaded := Files.PathMap.add pth s !(Sign.loaded);
+    s
+
+  (** Symbol representing an atomic unification problem. The term [atom t u]
+      represents [t ≡ u]. The left-hand side of a unification rule is made
+      of only one atom. *)
+  let atom : sym =
```
fblanqui, line 50: I would prefer to call it, say equiv. It would be clearer as atom is very generic.

src/core/unif.ml
```diff
@@ -33,6 +33,30 @@ let pp_problem oc p =
...
+    Sign.add_symbol sign Public Defin (Pos.none "u_atom") Kind []
```
fblanqui, line 51: Why do you want to add a prefix "u_"?

src/core/unif.ml
```diff
@@ -120,6 +144,37 @@ let instantiate : ctxt -> meta -> term array -> term -> bool =
       set_meta m (Bindlib.unbox bu); true
   | _ -> false
 
+(** [try_hints ctx s t] tries to solve unification problem [ctx ⊢ s ≡ t] using
+    declared unification hints. *)
+let try_hints : ctxt -> term -> term -> unif_constrs option =
```
fblanqui, line 149: Thanks to rename hint into unif_rule everywhere.

src/core/unif.ml
gabrielhdt: Otherwise we miss some information when symbols are printed out. Except if we print them fully qualified

gabrielhdt: I mean, getting for instance `atom symbol not found` may be confusing

src/core/unif.ml
```diff
@@ -33,6 +33,30 @@ let pp_problem oc p =
 let no_problems : problems =
   {to_solve  = []; unsolved = []; recompute = false}
 
+
+(** Definition of symbols used to encode unification rules. *)
+module Hint = struct
```
fblanqui, line 38: I would rather call this module Sym instead. But is it really useful to have a module here?

gabrielhdt: Well, here the module is used only to have an "elegant" name scheme in the caml code, to have `Unif.Hint.atom` rather that `Unif.atom_hint` and `Unif.sign_hint` and alike. So it is purely a question of taste.

src/core/unif.ml
fblanqui: In this case, I would prefer a prefix that is not allowed by the parser, say "#" for instance. Here, "#equiv" and "#list". What do you think?

src/core/unif.ml
fblanqui: I see. Ok for Sym then.

src/core/urs.ml
```diff
@@ -0,0 +1,95 @@
+(** Soundness of unification rules. *)
```
fblanqui, line 1: Let's keep the verification of some properties of unification rules for later. (Btw, "urs" is too short.)

gabrielhdt: Yeah, it was just to have the code at hand. Do you want it away from the PR?

fblanqui: Yes, I think it is better to do this in another PR.

src/core/unif.ml
gabrielhdt: Good idea.

src/core/sig_state.ml
```diff
@@ -24,11 +24,39 @@ type sig_state =
...
+        signature is an automatic dependency of all other signatures, and is
+        automatically loaded. *)
+    let sign : Sign.t =
+      let open Files in
+      let pth = Path.ghost "unif_rule" in
+      let s = Sign.create pth in
+      (* Remove the dependency on itself. *)
+      s.sign_deps := PathMap.remove pth !(s.sign_deps);
+      Sign.loaded := Files.PathMap.add pth s !(Sign.loaded);
+      s
```
fblanqui, line 41: To have \#equiv and \#list printed as "≡" and "," respectively, it would be better to declare them as binops too, with "," associative. Btw, it would then be better to rename \#list into \#comma. Then, there would be no need to have a special parser rule. Finally, you should call open_sign at some point.

fblanqui: Ah but, currently, "," cannot be used in an identifier... So, indeed, we need a special parser rule.

src/core/unif.ml
```diff
@@ -58,6 +58,36 @@ let copy_prod_env : tvar array -> term -> env = fun xs t ->
   in
   build_env 0 [] t
 
+(** [try_rules ctx s t] tries to solve unification problem [ctx ⊢ s ≡ t] using
+    declared unification rules. *)
+let try_rules : ctxt -> term -> term -> constr list option = fun ctx s t ->
+  if !log_enabled then log_unif "hint [%a]" pp_constr (ctx,s,t);
```
fblanqui, line 64: hint -> try_rules

src/core/parser.ml
```diff
@@ -467,6 +467,18 @@ let term = term PFunc
...
+(** [unif_rule] is a parser for unification rules. *)
+let parser unif_rule =
+  | l:{term "≡" term} "↪" r:{term "≡" term} rs:{"," term "≡" term}* ->
+      let equiv = Pos.none (P_Iden(Pos.none ([], "#equiv"), true)) in
+      let eqlist = Pos.none (P_Iden(Pos.none([], "#list"), true)) in
+      let mkequiv (l, r) =
+        Pos.none (P_Appl(Pos.none (P_Appl(equiv, l)), r))
+      in
+      let lhs = mkequiv l in
+      let rhs = p_add_args eqlist (List.map mkequiv (r::rs)) in
```
fblanqui, line 479: Here, for better printing later, I would have 2 cases:
- rs = []: rhs = mkequiv r
- r=r1, rs=r2,..,rn<>[]: rhs = mkcomma s1 (mkcomma s2 ... (mkcomma s(n-1) sn) ...) where si = mkequiv ri

src/core/sig_state.ml
gabrielhdt: OK, so it would rather be a cons symbol than a list symbol.

fblanqui: Right. Otherwise you won't have a nice printing (unless you use a specific printing function).

gabrielhdt: I don't think we need to call `open_sign` since `of_sign` initialises "correctly" the fields.

fblanqui: No, in_scope and pp_hints are not set correctly.

gabrielhdt: OK, it should be now that I `open_sign` when a `sig_state` is created.

src/core/sig_state.ml
```diff
@@ -24,17 +24,62 @@ type sig_state =
...
+
+    (** Holds a list of equivalences. The right-hand side of a unification
+        rule is made of such a list, [list (equiv t u) (equiv v w) ...]. *)
+    let comma : sym =
+      Sign.add_symbol sign Public Defin (Pos.none "#comma") Kind []
+
+    (** Pretty printing hints for symbols [equiv] and [comma]. *)
+    let pp_hints : pp_hint SymMap.t =
+      let open Stdlib in
+      let res = ref SymMap.empty in
```
fblanqui, line 57: Why do you use a ref to do that ?

src/core/sig_state.ml
```diff
@@ -24,17 +24,62 @@ type sig_state =
...
+      res := begin
+        let h = Infix("≡", Assoc_none, 1.1, Pos.none ([], "#equiv")) in
```
fblanqui, line 59: [] should be replaced by pth.

src/core/sig_state.ml
```diff
@@ -24,17 +24,62 @@ type sig_state =
...
+        SymMap.add equiv h !res
+      end;
+      res := begin
+        let h = Infix(",", Assoc_right, 1.0, Pos.none([], "#comma")) in
```
fblanqui, line 63: [] should be replaced by pth.

src/core/sig_state.ml
fblanqui: You need to move pth to make it an element of the module (and rename it into "path")

src/core/sig_state.ml
```diff
@@ -24,17 +24,62 @@ type sig_state =
...
+
+    (** [unpack eqs] transforms a term of the form [t =? u, v =? w, ...]
+        into a list [[t =? u; v =? w; ...]]. *)
+    let rec unpack : term -> (term * term) list = fun eqs ->
+      match Basics.get_args eqs with
+      | (Symb(s), [v; w]) ->
+          if s == comma then
+            match Basics.get_args v with
+            | (Symb(e), [t; u]) ->
+                assert (e == equiv);
```
fblanqui, line 76: use when clause instead

src/core/sign.ml
```diff
@@ -29,7 +29,8 @@ type t =
 
 (** [create path] creates an empty signature with module path [path]. *)
 let create : Path.t -> t = fun sign_path ->
-  { sign_path; sign_symbols = ref StrMap.empty; sign_deps = ref PathMap.empty
+  { sign_path; sign_symbols = ref StrMap.empty
+  ; sign_deps = ref (PathMap.singleton (Path.ghost "unif_rule") [])
```
fblanqui, line 33: use Unif_rule.path instead

fblanqui:
@gabrielhdt you also need to update the doc and CHANGES.md

src/core/sig_state.ml
fblanqui: Where do you do this? I cannot find it.

src/core/sign.ml
gabrielhdt: Ah but I can't because `Unif_rule` is in `sig_state`, or I move create to `sig_state`.

fblanqui: Ah right.

src/core/sig_state.ml
gabrielhdt: in `of_sign`

src/core/sign.ml
gabrielhdt: A solution would be to have a `Sign.create` that creates an *empty* signature, and a `Sig_state.create` that creates a signature with the pervasive dependencies set. Is this OK?

src/core/sig_state.ml
```diff
@@ -44,6 +86,53 @@ let remove_pp_hint :
...
+    are also handled in a similar way. *)
+let open_sign : sig_state -> Sign.t -> sig_state = fun ss sign ->
+  let f _key _v1 v2 = Some(v2) in (* open signature hides previous symbols *)
+  let in_scope = StrMap.union f ss.in_scope Sign.(!(sign.sign_symbols)) in
+  let builtins = StrMap.union f ss.builtins Sign.(!(sign.sign_builtins)) in
+  let pp_hints = update_pp_hints ss sign in
+  {ss with in_scope; builtins; pp_hints}
+
+(** [of_sign sign] creates a state from the signature [sign] with only
+    pervasive signatures imported. *)
```
fblanqui, line 122: perhaps add a comment about added ghost signatures here

src/core/sig_state.ml
```diff
@@ -44,6 +86,53 @@ let remove_pp_hint :
...
+let of_sign : Sign.t -> sig_state = fun sign ->
+  let s =
+    { signature = sign
+    ; in_scope  = !(Unif_rule.sign.sign_symbols)
```
fblanqui, line 126: not needed if you open Unif_rule.sign

src/core/sig_state.ml
```diff
@@ -44,6 +86,53 @@ let remove_pp_hint :
...
+    ; aliases   = StrMap.empty
+    ; builtins  = StrMap.empty
+    ; pp_hints  = Unif_rule.pp_hints }
```
fblanqui, line 129: not needed if you open Unif_rule.sign

src/core/sig_state.ml
```diff
@@ -44,6 +86,53 @@ let remove_pp_hint :
...
+  in
+  open_sign s Unif_rule.sign
```
fblanqui, line 131: so, you could just do `open_sign {empty with signature = s} Unif_rule.sign`

src/core/sign.ml
fblanqui: Yes, why not. But it seems better in this case to rename Sign.create into Sign.empty.

src/core/sig_state.ml
gabrielhdt: Not with the current definition of `empty`.

fblanqui:
Do you plan to allow rules on products in this PR? (see https://github.com/Deducteam/lambdapi/pull/361#issuecomment-618282333)

gabrielhdt:
That's a good question. The current PR already allows quite much, so if you're happy with what's done here, perhaps we can make rule on products in another PR. This will possibly require a non negligible amount of code, whereas this PR is quit light.

src/core/sig_state.ml
```diff
@@ -50,15 +50,68 @@ type sig_state =
...
+      s
+
+    (** Symbol representing an atomic unification problem. The term [equiv t
+        u] represents [t ≡ u]. The left-hand side of a unification rule is
+        made of only one unification. *)
+    let equiv : sym =
+      Sign.add_symbol sign Public Defin (Pos.none "#equiv") Kind []
+
+    (** Holds a list of equivalences. The right-hand side of a unification
+        rule is made of such a list, [list (equiv t u) (equiv v w) ...]. *)
```
fblanqui, line 81: comment to be updated

src/core/sig_state.ml
```diff
@@ -50,15 +50,68 @@ type sig_state =
...
 type t = sig_state
 
-(** [init_with sign] creates a state from the signature [sign] with no symbols
-   in scope, module aliases, builtins or printing hints. *)
-let of_sign : Sign.t -> sig_state = fun sign ->
-  { signature = sign ; in_scope  = StrMap.empty ; aliases = StrMap.empty
-  ; path_map = PathMap.empty ; builtins  = StrMap.empty
-  ; unops = StrMap.empty ; binops = StrMap.empty ; pp_hints = SymMap.empty }
+(** Symbols and signature used for unification rules. *)
+module Unif_rule =
```
fblanqui, line 54: cant' this module be moved in a separate file?

src/core/sig_state.ml
```diff
@@ -50,15 +50,68 @@ type sig_state =
...
+            | _          (* Ill-formed term. *) -> assert false
+          else if s == equiv then [(v, w)] else
+          assert false (* Ill-formed term. *)
+      | _                 -> assert false (* Ill-formed term. *)
+end
+
+(** [create_sign path] creates a signature with pervasive modules as
+    dependencies. *)
+let create_sign : Path.t -> Sign.t = fun sign_path ->
+  { sign_path; sign_symbols = ref StrMap.empty
```
fblanqui, line 111: Why not introducing a Sign.empty (or Sign.void or Sign.dummy if you prefer) for not rewriting all the components each time?

gabrielhdt: The idea was to force using `create_sign` to obtain a `Sign.t`. But I understand that it can become tedious. It's as you wish.

fblanqui: Yes, it will simplify the reading of the code.

fblanqui: It's already the case that, in handle, we do not call functions from Sign but those of Sig_state.

doc/sections/commands.md
```diff
@@ -383,3 +383,13 @@ set builtin "eq"    ≔ eq    // : Π {a}, T a → T a → Prop
...
 set builtin "eqind" ≔ eqind // : Π {a} x y, P (x = y) → Π (p:T a→Prop), P (p y) → P (p x)
 ```
+
+**unification rules** The unification engine can be guided using 
+*unification rules*. Given a unification problem `t ≡ u`, if the engine cannot
+find a solution, it will try to match the pattern `t ≡ u` against the defined
+rules and rewrite the problem to the right-hand side of the matched  rule.
+For instance, given the unification rule
+```
+set unif_rule Bool ≡ T $t ↪ $t ≡ bool
```
fblanqui, line 393: Better provide an example with a comma:  f $x1 $y1 = f $x2 $y2 --> $x1 = $x2, $y1 = $y2

fblanqui: You should also add that this is experimental and no sanity check is done, so it may lead to invalid deductions on unification or non-termination.

src/core/parser.ml
```diff
@@ -564,6 +582,7 @@ let parser config =
       sanity_check _loc_id id;
       Prefix.add declared_ids id id;
       P_config_ident(id)
+  | "unif_rule" r:unif_rule -> P_config_unif_rule(r)
```
fblanqui, line 585: Why not using parser rule instead finally?

src/core/sig_state.ml
```diff
@@ -196,6 +194,19 @@ let open_sign : sig_state -> Sign.t -> sig_state = fun ss sign ->
...
+let of_sign : Sign.t -> sig_state = fun sign ->
+  let empty =
+    { signature = sign; in_scope = StrMap.empty; aliases = StrMap.empty
+    ; path_map = PathMap.empty; builtins = StrMap.empty; unops = StrMap.empty
+    ; binops = StrMap.empty; pp_hints = SymMap.empty }
+  in
+  open_sign empty Unif_rule.sign
+
+(** [empty] state. *)
+let empty : sig_state = of_sign (create_sign [])
```
fblanqui, line 208: You can now replace create_sign [] by dummy. Perhaps the empty state should be rename into dummy too.

src/core/sig_state.ml
```diff
@@ -50,15 +50,13 @@ type sig_state =
...
-  { signature = sign ; in_scope  = StrMap.empty ; aliases = StrMap.empty
-  ; path_map = PathMap.empty ; builtins  = StrMap.empty
-  ; unops = StrMap.empty ; binops = StrMap.empty ; pp_hints = SymMap.empty }
-
-(** [empty] state. *)
-let empty = of_sign (Sign.create [])
+(** [create_sign path] creates a signature with pervasive modules as
+    dependencies. *)
+let create_sign : Path.t -> Sign.t = fun sign_path ->
+  { sign_path; sign_symbols = ref StrMap.empty
```
fblanqui, line 56: You can use dummy now.

src/core/sig_state.ml
```diff
@@ -196,6 +194,19 @@ let open_sign : sig_state -> Sign.t -> sig_state = fun ss sign ->
   in
   {ss with in_scope; builtins; unops; binops; pp_hints}
 
+(** [of_sign sign] creates a state from the signature [sign] with ghost
+    signatures opened. *)
+let of_sign : Sign.t -> sig_state = fun sign ->
+  let empty =
+    { signature = sign; in_scope = StrMap.empty; aliases = StrMap.empty
```
fblanqui, line 201: You can use dummy now.

src/core/parser.ml
fblanqui: To do this, you need to add the idents "\equiv" et "," in Unif_rule.sign.

fblanqui: Ah but I guess it won't work because "," cannot be used as identifier...

src/core/sig_state.ml
gabrielhdt: Ah no because `create_sign` sets the dependencies on `Unif_rule`

gabrielhdt: And I prefer leaving `empty` as a name for "ready-to-use" elements and `dummy` for not usable elements: the dummy signature needs to be initialised with ghost signatures and path before being used but the empty `sig_state` can be nominally used.

src/core/sig_state.ml
gabrielhdt: Where? This function acts as the constructor of `sig_state` doesn't it? We don't have a dummy `sig_state` elsewhere.

src/core/sig_state.ml
fblanqui: But this empty/dummy sig_state is not usable ([] is not a valid path). It is used in print only. That's why you can rename it into dummy and use it in of_sign.

src/core/sig_state.ml
fblanqui: Yes, see comment above.

src/core/sig_state.ml
gabrielhdt: If it's used in `Print` only, don't we want to define it there?

src/core/pretty.ml
```diff
@@ -227,6 +227,8 @@ let pp_command : p_command pp = fun oc cmd ->
         | Assoc_right -> " right"
       in
       out "set infix%s %f \"%s\" ≔ %a" a p s pp_qident qid
+  | P_set(P_config_unif_rule(ur))   ->
+      out "set unif_rule %a" (pp_p_rule true) ur
```
fblanqui, line 231: You should use a specific function for printing unif_rule's. Otherwise it will print some \#equiv's and \#comma's...

src/core/sig_state.ml
```diff
@@ -196,6 +192,21 @@ let open_sign : sig_state -> Sign.t -> sig_state = fun ss sign ->
...
+(** Dummy [sig_state] made from the dummy signature. *)
+let dummy : sig_state =
+  { signature = Sign.dummy (); in_scope = StrMap.empty; aliases = StrMap.empty
+  ; path_map = PathMap.empty; builtins = StrMap.empty; unops = StrMap.empty
+  ; binops = StrMap.empty; pp_hints = SymMap.empty }
+
+(** [of_sign sign] creates a state from the signature [sign] with ghost
+    signatures opened. *)
+let of_sign : Sign.t -> sig_state = fun sign ->
+  open_sign {dummy with signature = sign} Unif_rule.sign
```
fblanqui, line 204: avoid = sign by renaming sign into signature?

src/core/sig_state.ml
```diff
@@ -196,6 +192,21 @@ let open_sign : sig_state -> Sign.t -> sig_state = fun ss sign ->
...
+
+(** Dummy state made from the empty signature but with ghost modules
+    opened. *)
+let dummy : sig_state = of_sign (create_sign [])
```
fblanqui, line 208: You do not remove it finally? (if not you should replace create_sign [] by Sign.dummy())

src/core/pretty.ml
gabrielhdt: Perhaps we can more simply modify `pp_p_ident` to handle the "#..." identifiers

fblanqui: It looks very hacky. As you have a specific parsing rule, it is not surprising to have a specific pp rule.

src/core/sig_state.ml
gabrielhdt: We have to make a choice: `dummy` depends on `of_sign`, so I cannot use it in `of_sign`... or I can make a truly dummy signature which does not open the ghost modules (is that what you meant?)

fblanqui: Yes, this is what I meant. There are Sign.dummy and Sig_state.dummy.

PR #363, opened 2020-04-18
Adapt code to improve term printing

fblanqui:
In this PR, pp_hints are not attached to every Symb constructor. Instead, a map from sym to pp_hints is built in every sig_state, and many functions now take a sig_state as argument in order to print terms in a nice way (except in Eval, at least for the moment).


rlepigre:
I think it is a good idea to try and improve printing, and I agree that using the `sig_state` is the right way to go. Nonetheless, I think it would have been beneficial to discuss design before going straight into implementation. I don't like to always have to be the annoying guy that says "I think you're not doing things the right way"... (This would easily be avoiding by having a quick design discussion in an issue.) But here I go: I think that it is totally unnecessary to carry around a `sig_state` everywhere.

Indeed, the scope of symbols and builtins only change at the granularity of commands, so it is enough to register the data needed for the printer (in fact, exactly the `sig_state`) at the beginning of `Handle.handle_cmd` using a reference in the `Print` module. This would be a lot simpler and would avoid having to carry an extra value around in all functions. After all, this argument is never directly useful, and it is only here for debug printing. So it has no real reason to be there.

Note also that I don't like what you have done with the `Builtin` module I created. It would have been a lot better to solve the dependency cycle by extracting the definition of `sig_state` from `scope.ml` to a new module `sig_state.ml` that would contain the related functions (basically all the beginning of `scope.ml` up to `find_sym` included). The scoping module is already huge, it is better not to dump even more things in there. Doing this is also necessary for what I propose.

If you agree with my analysis, then I think there is no choice but to start over this PR. In that case, I think that a good plan would be the following (each point being either a separate commit or a separate PR):
 1. Remove the printing hints on `Symb`, and make it so that printing always displays a fully qualified symbol. Here a separate review is desirable because despite the fact that the change can be quite systematic it is easy to forget something.
 2. Create the `sig_state.ml` module and set up the reference in `print.ml` and its update in `handle.ml`.
 3. Modify the printing function to do a kind of reverse lookup in the signature state in the case of symbols.

Some thoughts about the last point:
 - If a symbol has a corresponding unary or binary operator then print it with the right priority and associativity (management of priority / associativity can be done later as a fourth point. As a first version it is good enough to insert parentheses systematically around binary operator applications). If there are implicit arguments and their printing is turned on then you can decide to give up using the unary or binary syntax.
 - Otherwise, if the symbol corresponds to an available `"+1"` or `"0"` builtin, then try and print the term in decimal notation.
 - Otherwise check whether the symbol is in scope (looking up its name in the `in_scope` member of the `sig_state`), and if yes display only the symbol name with no qualification.
 - Otherwise check for aliases for the module path of the symbol, and if one exists use the short qualification.
 - Otherwise fall back to the fully qualified syntax.

fblanqui:
It doesn't look like a good idea to multiply global references. It will prevent us from having several threads in the future (@ejgallego and @01mf02 already mentioned this problem).

fblanqui:
It is not efficient to do a reverse lookup each time we want to print something. It is better to compute the reverse table each time we update the sig_state as I do in the PR.

fblanqui:
For printing, scope checking must be done first I think.

rlepigre:
> It doesn't look like a good idea to multiply global references. It will prevent us from having several threads in the future (@ejgallego and @01mf02 already mentioned this problem).

Well, OCaml does not support concurrency well, so it is not like we are gonna go multi-threaded any time soon. And in any case, none of our code is thread safe in the current state. So doing what you propose actually makes the code worst for no reason.

> It is not efficient to do a reverse lookup each time we want to print something. It is better to compute the reverse table each time we update the sig_state as I do in the PR.

Obviously, but then efficiency does not matter at all for printing. In any case, having the reverse table is a good approach, but if you do that it makes even more sense to put `sig_state` in its own module and to provide functions to work on it (that would take care of updating two tables).

> For printing, scope checking must be done first I think.

No it does not. The `sig_state` is only extended by specific commands (those that introduce a symbol or a built-in, or opening a signature). For all the others, the state is the same before and after the command, so in any case, the signature state at the beginning of command handling is valid throughout the command.

fblanqui:
For multi threads, who knows. Michael already has a multi threaded fork of Dedukti.

You didn't check the last commits. sig_state is already in its own file.

I don't understand your last point. There must be a misunderstanding. What I am saying is: we should not use a notation if it is defined in a non-opened module.

rlepigre:
> I don't understand your last point. There must be a misunderstanding. What I am saying is: we should not use a notation if it is defined in a non-opened module.

Well, that is exactly why you are using a `sig_state`, it only contains the notations that are in scope before the handling of a command. This initial scope (before handling the command) is the one you need both for scoping and for printing. None of the interesting commands modify the `sig_state` in a way that is meaningful for printing (only some queries are). For instance, when a symbol is defined it is not gonna be in scope in its own type or definition for example.

fblanqui:
Back to ref. PR ready to merge.

rlepigre:
Please @fblanqui, do not modify while I am reviewing! You said this was ready...

src/core/sign.ml
```diff
@@ -64,6 +67,13 @@ let current_sign () =
   in
   PathMap.find mp !loaded
 
+(** [new_sym ()] creates a new (private definable) symbol. *)
+let new_sym : string -> term -> sym = fun name typ ->
+  let path = (current_sign()).sign_path in
+  { sym_name = name; sym_type = ref typ; sym_path = path; sym_def = ref None
+    ; sym_impl = []; sym_rules = ref []; sym_prop = Defin; sym_expo = Privat
+    ; sym_tree = ref Tree_types.empty_dtree }
```
rlepigre, line 75: Indentation is not correct, please align the leading `;` with the `{`.

src/core/syntax.ml
```diff
@@ -95,6 +95,22 @@ type p_assertion =
...
+  | SNF
+  (** Reduce to strong normal form. *)
+  | NONE
+  (** Do nothing. *)
+
+(** Configuration for evaluation. *)
+type eval_config =
+  { strategy : strategy   (** Evaluation strategy.          *)
+  ; steps    : int option (** Max number of steps if given. *) }
+
```
rlepigre, line 113: As I said, I think this should be reverted.

src/core/terms.ml
```diff
@@ -451,7 +416,7 @@ let rec lift : term -> tbox = fun t ->
   | Vari(x)     -> _Vari x
   | Type        -> _Type
   | Kind        -> _Kind
-  | Symb(s,h)   -> _Symb s h
+  | Symb(s)     -> _Symb s
```
rlepigre, line 419: You can replace that by:
```ocaml
  | Symb(_) as t -> Bindlib.box t
```
This will save an allocation.

src/core/terms.ml
```diff
@@ -497,16 +462,49 @@ end
...
+  { to_solve  : constr list
+  (** List of unification problems to solve. *)
+  ; unsolved  : constr list
+  (** List of unification problems that could not be solved. *)
+  ; recompute : bool
+  (** Indicates whether unsolved problems should be rechecked. *) }
+
+(** Empty problem. *)
+let empty_problem : problem =
+  {to_solve  = []; unsolved = []; recompute = false}
```
rlepigre, line 510: I don't understand why these things should be here. This module is for the AST and closely related things.

src/core/why3_tactic.ml
```diff
@@ -4,7 +4,9 @@ open Console
 open Terms
 open Extra
 open Timed
-open Scope
+open Print
+
+module Builtin = Sig_state.Builtin
```
rlepigre, line 9: Please avoid module aliases, rather open the module locally using `let open Sig_state.Builtin`.

src/core/rewrite.ml
```diff
@@ -58,18 +59,6 @@ let eq : ctxt -> term -> term -> bool = fun ctx a b -> a == b ||
...
-    Assia Mahboubi and Enrico Tassi, INRIA Research Report 6455, 2016,
-    http://hal.inria.fr/inria-00258384, section 8, p. 48. *)
-type rw_patt =
-  | RW_Term           of term
-  | RW_InTerm         of term
-  | RW_InIdInTerm     of (term, term) Bindlib.binder
-  | RW_IdInTerm       of (term, term) Bindlib.binder
-  | RW_TermInIdInTerm of term * (term, term) Bindlib.binder
-  | RW_TermAsIdInTerm of term * (term, term) Bindlib.binder
-
```
rlepigre, line 72: As I already said, I think this should stay here. This is only relevant for `rewrite` and related future tactics.

src/core/rewrite.ml
```diff
@@ -161,22 +153,22 @@ let _ =
...
-    triple containing the type in which equality is used and the equated terms
-    (LHS and RHS). *)
-let get_eq_data : popt -> eq_config -> term -> term * term * term =
-  fun pos cfg t ->
-  match get_args t with
+  register_builtin "eqind" expected_eqind_type
+
+(** [get_eq_data pos cfg a] extra data from an equality type [a]. It
+   consists of a triple containing the type in which equality is used and the
+   equated terms (LHS and RHS). *)
```
rlepigre, line 160: Please, indent comments correctly.

src/core/rewrite.ml
```diff
@@ -161,22 +153,22 @@ let _ =
...
+let get_eq_data : popt -> eq_config -> term -> term * term * term
+  = fun pos cfg t ->
```
rlepigre, line 162: Please, leave `=` on the previous line,

src/core/rewrite.ml
```diff
@@ -202,20 +194,21 @@ let break_prod : term -> term * tvar array = fun a ->
...
-(** [match_pattern (xs,p) t] attempts to match the pattern [p] (containing the
-    “pattern variables” of [xs]) with the term [t]. If successful,  it returns
-    [Some(ts)] where [ts] is an array of terms such that substituting elements
-    of [xs] by the corresponding elements of [ts] in [p] yields a term that is
-    equal to [t] (in terms of [eq]). *)
+(** [match_pattern (xs,p) t] attempts to match the pattern [p] (containing
+   the “pattern variables” of [xs]) with the term [t]. If successful, it
+   returns [Some(ts)] where [ts] is an array of terms such that substituting
+   elements of [xs] by the corresponding elements of [ts] in [p] yields a term
+   that is equal to [t] (in terms of [eq]). *)
```
rlepigre, line 201: Why do you mess up the comment alignment? You did not even modify anything here!

src/core/rewrite.ml
```diff
@@ -202,20 +194,21 @@ let break_prod : term -> term * tvar array = fun a ->
...
 
-(** [find_subst t (xs,p)] is given a term [t] and a pattern [p] (with “pattern
-    variables” of [xs]),  and it finds the first instance of (a term matching)
-    [p] in [t] (if there is any). If successful, the function returns an array
-    of terms corresponding to the substitution (see [match_pattern]). *)
+(** [find_subst t (xs,p)] is given a term [t] and a pattern [p] (with
+   “pattern variables” of [xs]), and it finds the first instance of (a term
+   matching) [p] in [t] (if there is any). If successful, the function returns
+   an array of terms corresponding to the substitution (see
+   [match_pattern]). *)
```
rlepigre, line 211: Same here. Why?

src/core/rewrite.ml
```diff
@@ -235,11 +228,11 @@ let find_subst : term -> to_subst -> term array option = fun t (xs,p) ->
...
-(** [make_pat t p] is given a term [t], and a pattern [p] containing reference
-    cells (that are not instantiated) and wildcards.  It then tries to find  a
-    subterm of [t] that matches [p], using (instantiating) syntactic equality.
-    In case of success, the function returns [true],  and the matching term is
-    [p] itself (through instantiation). *)
+(** [make_pat t p] is given a term [t], and a pattern [p] containing
+   reference cells (that are not instantiated) and wildcards.  It then tries
+   to find a subterm of [t] that matches [p], using (instantiating) syntactic
+   equality.  In case of success, the function returns [true], and the
+   matching term is [p] itself (through instantiation). *)
```
rlepigre, line 235: And again...

src/core/rewrite.ml
```diff
@@ -257,9 +250,9 @@ let make_pat : term -> term -> bool = fun t p ->
       end
   in make_pat_aux t
 
-(** [bind_match p t] binds every occurence of the pattern [p] in the term [t].
-    We require [t] not to contain products, abstractions, metavariables or any
-    other awkward term constructor. *)
+(** [bind_match p t] binds every occurence of the pattern [p] in the term
+   [t].  We require [t] not to contain products, abstractions, metavariables
+   or any other awkward term constructor. *)
```
rlepigre, line 255: And again...

src/core/rewrite.ml
```diff
@@ -285,25 +278,26 @@ let bind_match : term -> term -> tbinder =  fun p t ->
...
-    equality. Every occurrence of the first instance of the left-hand side  is
-    replaced by the right-hand side of the obtained proof. It also handles the
-    full set of SSReflect patterns. *)
-let rewrite : popt -> Proof.t -> rw_patt option -> term -> term =
-  fun pos ps p t ->
+(** [rewrite ss ps po t] rewrites according to the equality proved by [t] in
+   the current goal of [ps]. The term [t] should have a type corresponding to
+   an equality. Every occurrence of the first instance of the left-hand side
+   is replaced by the right-hand side of the obtained proof. It also handles
+   the full set of SSReflect patterns. *)
```
rlepigre, line 285: And again...

src/core/rewrite.ml
```diff
@@ -618,80 +618,81 @@ let rewrite : popt -> Proof.t -> rw_patt option -> term -> term =
...
     end;
 
   (* Return the proof-term. *)
   term
 
-(** [reflexivity ps] applies the reflexivity of equality on the focused  goal.
-    If successful, the corresponding proof term is returned. *)
-let reflexivity : popt -> Proof.t -> term = fun pos ps ->
+(** [reflexivity ss pos ps] applies the reflexivity of equality on the focused
+   goal. If successful, the corresponding proof term is returned. *)
```
rlepigre, line 649: Comment not indented correctly.

src/core/rewrite.ml
```diff
@@ -618,80 +618,81 @@ let rewrite : popt -> Proof.t -> rw_patt option -> term -> term =
...
+  add_args (Symb cfg.symb_refl) [a; l]
 
-(** [symmetry ps] attempts to use symmetry of equality on the focused goal. If
-    successful,  a new goal is generated,  and the corresponding proof term is
-    returned. The proof of symmetry is built from the axioms of equality. *)
-let symmetry : popt -> Proof.t -> term = fun pos ps ->
+(** [symmetry ss pos ps] attempts to use symmetry of equality on the focused
+   goal. If successful, a new goal is generated, and the corresponding proof
+   term is returned. The proof of symmetry is built from the axioms of
+   equality. *)
```
rlepigre, line 664: Comment not indented correctly.

src/core/rewrite.ml
```diff
@@ -618,80 +618,81 @@ let rewrite : popt -> Proof.t -> rw_patt option -> term -> term =
...
+  let cfg = get_eq_config ss pos in
   (* Get the type of the focused goal. *)
   let (g_env, g_type) = Proof.focus_goal pos ps in
   (* Check that the type of [g] is of the form “P (eq a l r)”. *)
   let (a, l, r) = get_eq_data pos cfg g_type in
   (* NOTE The proofterm is “eqind a r l M (λx,eq a l x) (refl a l)”. *)
   (* We create a new metavariable (“M” in the above). *)
   let meta_type =
-    Appl(symb cfg.symb_P, (add_args (symb cfg.symb_eq) [a; r; l])) in
+    Appl(Symb cfg.symb_P, (add_args (Symb cfg.symb_eq) [a; r; l])) in
```
rlepigre, line 675: Please, put the `in` on its own line (I know it was here before).

src/core/rewrite.ml
```diff
@@ -618,80 +618,81 @@ let rewrite : popt -> Proof.t -> rw_patt option -> term -> term =
...
     let pred = Bindlib.unbox (Bindlib.bind_var x (lift pred)) in
-    Abst(Appl(symb cfg.symb_T, a), pred)
+    Abst(Appl(Symb cfg.symb_T, a), pred)
   in
   (* We build the proof term. *)
-  let refl_a_l = add_args (symb cfg.symb_refl) [a; l] in
+  let refl_a_l = add_args (Symb cfg.symb_refl) [a; l] in
   let term =
-    add_args (symb cfg.symb_eqind) [a; r; l; meta_term; pred; refl_a_l] in
+    add_args (Symb cfg.symb_eqind) [a; r; l; meta_term; pred; refl_a_l] in
```
rlepigre, line 687: Same here, please put the `in` on its own line.

src/core/tree.ml
```diff
@@ -279,7 +279,7 @@ module CM = struct
 
   (** [pp ?(pp_cond=false) o m] prints matrix [m] to out channel [o]. The flag
       [pp_cond] indicates whether the condition pull should be printed. *)
-  let pp : ?pp_cond:bool -> t pp = fun ?(pp_cond=false) oc m ->
+  (*let pp : ?pp_cond:bool -> t pp = fun ?(pp_cond=false) oc m ->
```
rlepigre, line 282: Please, either uncomment or remove the code.

src/core/typing.ml
```diff
@@ -3,49 +3,47 @@
...
-(** [sort_type builtins ctx t] checks that the type of the term [t] in context
+(** [sort_type ctx t] checks that the type of the term [t] in context
    [ctx] is a sort. If that is not the case, the exception [Fatal] is
    raised. *)
-let sort_type : sym StrMap.t -> ctxt -> term -> unit =
-  fun builtins ctx t ->
-  match infer builtins ctx t with
-  | None    -> fatal None "Unable to infer a sort for %a." pp t
+let sort_type : ctxt -> term -> unit = fun ctx t ->
+  let pp_term = Print.pp_term in
```
rlepigre, line 42: I would prefer to leave the `open Print` at the top instead of doing that.

src/core/typing.ml
```diff
@@ -3,49 +3,47 @@
...
+  match infer ctx t with
+  | None    -> fatal None "Unable to infer a sort for %a." pp_term t
   | Some(a) ->
   match unfold a with
   | Type | Kind -> ()
-  | a           -> fatal None "%a has type %a (not a sort)." pp t pp a
+  | a           ->
+      fatal None "%a has type %a (not a sort)." pp_term t pp_term a
```
rlepigre, line 49: And we see here that the short `pp` was useful to save us some lines. I think it makes sense to have it because it is such a common function (and there can be no error due to type-checking).

src/core/typing.ml
```diff
@@ -3,49 +3,47 @@
 open Extra
 open Console
 open Terms
-open Print
 open Unif
 
-(** [check builtins ctx t a] tells whether [t] has type [a] in context
+(** [check ctx t a] tells whether [t] has type [a] in context
    [ctx]. *)
```
rlepigre, line 9: This should fit on one line.

src/core/unif.ml
```diff
@@ -109,30 +113,27 @@ let instantiation : ctxt -> meta -> term array -> term ->
         Some (Bindlib.bind_mvar vs (lift u))
   else None
 
-(** [instantiate ctx m ts u] check whether, in a problem [m[ts]=u], [m] can be
-    instantiated and, if so, instantiate it. *)
+(** [instantiate ctx m ts u] check whether, in a problem [m[ts]=u], [m] can
+   be instantiated and, if so, instantiate it. *)
```
rlepigre, line 117: Comment indentation is incorrect, please be careful with this (or set up your editor).

src/core/unif.ml
```diff
@@ -417,10 +422,16 @@ and solve_aux : ctxt -> term -> term -> problems -> unif_constrs =
...
    no solution, the value [None] is returned. Otherwise [Some(cs)] is
    returned, where the list [cs] is a list of unsolved convertibility
    constraints. *)
-let solve : sym StrMap.t -> problems -> unif_constrs option =
-  fun _builtins p ->
+let solve : problem -> constr list option = fun p ->
   try Some (solve p) with Unsolvable -> None
+
+(** [eq c t u] tries to unify the terms [t] and [u] in context [c], by
+   instantiating their metavariables. *)
```
rlepigre, line 433: Comment not indented correctly.

src/core/unif.ml
```diff
@@ -417,10 +422,16 @@ and solve_aux : ctxt -> term -> term -> problems -> unif_constrs =
...
+let eq c t u =
```
rlepigre, line 434: Please give the type.

src/core/unif.ml
```diff
@@ -417,10 +422,16 @@ and solve_aux : ctxt -> term -> term -> problems -> unif_constrs =
...
+  match solve {empty_problem with to_solve=[c,t,u]} with
+  | Some [] -> true
+  | _ -> false
```
rlepigre, line 437: This can be written as:
```ocaml
solve {empty_problem with to_solve=[c,t,u]} = Some([])

File: src/core/handle.ml [hunk 26]
```diff
@@ -4,20 +4,33 @@ open Extra
...
 open Sign
 open Pos
 open Files
 open Syntax
+open Sig_state
 open Scope
+open Print
 
-(* Register a check for the type of builtin successor symbol ["+1"]. *)
+(* Register a check for the type of the builtin symbol "0" and "+1". *)
```
rlepigre, line 15: Typo: symbolS.

src/core/handle.ml
```diff
@@ -4,20 +4,33 @@ open Extra
...
+  register "0" expected_zero_type;
+  let expected_succ_type ss _pos =
+    let typ_0 =
+      try lift !((StrMap.find "0" ss.builtins).sym_type)
+      with Not_found -> _Meta (fresh_meta Type 0) [||]
+    in
     Bindlib.unbox (_Impl typ_0 typ_0)
   in
-  Builtin.Check.register_expected_type "+1" expected_successor
+  register "+1" expected_succ_type
```
rlepigre, line 33: What was wrong with the previous approach? This seemed like a good fit for this pattern.

src/core/handle.ml
```diff
@@ -148,27 +160,28 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
       let a =
         match ao with
         | Some(a) ->
-            Typing.sort_type ss.builtins [] a;
-            if Typing.check ss.builtins [] t a then a else
-            fatal cmd.pos "The term [%a] does not have type [%a]." pp t pp a
+            Typing.sort_type [] a;
+            if Typing.check [] t a then a else
+              fatal cmd.pos "The term [%a] does not have type [%a]."
+                pp_term t pp_term a
```
rlepigre, line 166: Using `pp` saved lines here as well.

src/core/handle.ml
```diff
@@ -268,12 +278,13 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
     indicate commands that take too long to execute. *)
 let too_long = Stdlib.ref infinity
 
-(** [handle_cmd ss cmd] is similar to [handle_cmd_aux ss cmd] but it adds some
+(** [handle_cmd ss cmd] is similar to [handle_cmd ss cmd] but it adds some
```
rlepigre, line 281: Surely `handle_cmd ss cmd` is similar to `handle_cmd ss cmd`. Did the comment make sense before or was that not up to date already?

src/core/sr.ml
```diff
@@ -77,23 +76,24 @@ let new_function_symbol name a =
...
-let symb_to_tenv : pre_rule Pos.loc -> sym list -> index_tbl -> term -> tbox =
-    fun {elt={pr_vars=vars;pr_arities=arities;_};pos} syms htbl t ->
+(** [symb_to_tenv ss pr syms htbl t] builds a RHS for the pre-rule [pr]. The
+   term [t] is expected to be a version of the RHS of [pr] whose term
+   environments have been replaced by function symbols of [syms]. This
+   function builds the reverse transformation, replacing symbols by the term
+   environment variable they stand for.  Here, [htbl] should give the index in
+   the RHS environment for the symbols of [syms] that have corresponding
+   [term_env] variable. The pre-rule [pr] is provided to give access to these
+   variables and also their expected arities. *)
```
rlepigre, line 86: Comment not indented properly.

src/core/sr.ml
```diff
@@ -77,23 +76,24 @@ let new_function_symbol name a =
...
+let symb_to_tenv
+    : Scope.pre_rule Pos.loc -> sym list -> index_tbl -> term -> tbox
+  = fun {elt={pr_vars=vars;pr_arities=arities;_};pos} syms htbl t ->
```
rlepigre, line 89: Please do not start lines with `:` or `=`, rather use:
```ocaml
let symb_to_tenv : Scope.pre_rule Pos.loc -> sym list -> index_tbl ->
    -> term -> tbox =
  fun {elt={pr_vars=vars;pr_arities=arities;_};pos} syms htbl t ->
```

src/core/sr.ml
```diff
@@ -131,13 +131,14 @@ let symb_to_tenv : pre_rule Pos.loc -> sym list -> index_tbl -> term -> tbox =
   in
   symb_to_tenv t
 
-(** [check_rule bmap r] checks whether the pre-rule [r] is well-typed and then
-    construct the corresponding rule. The “bultin map” [bmap] is passed to the
-    unification engine. Note that [Fatal] is raised in case of error. *)
-let check_rule : sym StrMap.t -> pre_rule Pos.loc -> rule = fun bmap pr ->
+(** [check_rule r] checks whether the pre-rule [r] is well-typed in
+   signature state [ss] and then construct the corresponding rule. Note that
+   [Fatal] is raised in case of error. *)
```
rlepigre, line 136: Comment not indented properly.

src/core/sr.ml
```diff
@@ -131,13 +131,14 @@ let symb_to_tenv : pre_rule Pos.loc -> sym list -> index_tbl -> term -> tbox =
...
+let check_rule : Scope.pre_rule Pos.loc -> rule = fun pr ->
   (* Unwrap the contents of the pre-rule. *)
-  let (pos, (s, h), lhs, vars, rhs_vars, arities) =
-    let Pos.{elt={pr_sym;pr_lhs;pr_vars;pr_rhs;pr_arities;_}; pos} = pr in
+  let (pos, s, lhs, vars, rhs_vars, arities) =
+    let Pos.{elt=Scope.{pr_sym;pr_lhs;pr_vars;pr_rhs;pr_arities;_}; pos} = pr
+    in
```
rlepigre, line 141: Please, no, that is horrible style. Rather start the function with `let open Scope in`. Or reinstate the `open Scope`, which was not a problem at all.

src/core/sr.ml
```diff
@@ -229,21 +230,21 @@ let check_rule : sym StrMap.t -> pre_rule Pos.loc -> rule = fun bmap pr ->
...
   (* Solving the typing constraints of the RHS. *)
-  match Unif.(solve bmap {no_problems with to_solve}) with
+  match Unif.(solve {empty_problem with to_solve}) with
   | None     -> fatal pos "The rewriting rule does not preserve typing."
   | Some(cs) ->
   let is_constr c =
     let eq_comm (_,t1,u1) (_,t2,u2) =
       (* Contexts ignored: [Infer.check] is called with an empty context and
-         neither [Infer.check] nor [Unif.solve] generate contexts with defined
+         neither [Infer.check] nor [solve] generate contexts with defined
```
rlepigre, line 247: Is `solve` local to this function now?

src/core/infer.ml
```diff
@@ -31,13 +70,25 @@ let make_meta_codomain : ctxt -> term -> tbinder = fun ctx a ->
...
+      if !log_enabled then
+        log_infr (yel "add %a") pp_constr (ctx,a,b);
+      let open Stdlib in constraints := (ctx,a,b) :: !constraints
+    end
+
+(** [infer ctx t] infers a type for the term [t] in context [ctx] and
+   signature state [ss], possibly under some constraints recorded in
+   [constraints] using [conv]. The returned type is well-sorted if recorded
+   unification constraints are satisfied. [ctx] must be well-formed. This
+   function never fails (but constraints may be unsatisfiable). *)
```
rlepigre, line 89: Comment indentation is not correct.

src/core/infer.ml
```diff
@@ -132,49 +181,51 @@ let rec infer : ctxt -> term -> term = fun ctx t ->
...
-  if !log_enabled then log_infr "check %a : %a" pp t pp c;
-  conv ctx (infer ctx t) c
-
-(** [infer ctx t] returns a pair [(a,cs)] where [a] is a type for the
-   term [t] in the context [ctx], under unification constraints [cs].
-   In other words, the constraints of [cs] must be satisfied for [t]
-   to have type [a]. [ctx] must be well-formed. This function never
+(** [check ctx t a] checks that the term [t] has type [a] in context [ctx],
+   possibly under some constraints recorded in [constraints] using
+   [conv]. [ctx] must be well-formed and [a] well-sorted. This function never
```
rlepigre, line 195: Comment not indented correctly.

src/core/infer.ml
```diff
@@ -132,49 +181,51 @@ let rec infer : ctxt -> term -> term = fun ctx t ->
...
-let infer : ctxt -> term -> term * unif_constrs = fun ctx t ->
+and check : ctxt -> term -> term -> unit = fun ctx t a ->
+  if !log_enabled then log_infr "check %a : %a" pp_term t pp_term a;
+  conv ctx (infer ctx t) a
+
+(** [infer ctx t] returns a pair [(a,cs)] where [a] is a type for the term [t]
+   in the context [ctx] under unification constraints [cs].  In other words,
+   the constraints of [cs] must be satisfied for [t] to have type [a]. [ctx]
+   must be well-formed. This function never fails (but constraints may be
+   unsatisfiable). *)
```
rlepigre, line 205: Comment not indented correclty.

src/core/infer.ml
```diff
@@ -132,49 +181,51 @@ let rec infer : ctxt -> term -> term = fun ctx t ->
...
-(** [check ctx t c] checks returns a list [cs] of unification
-   constraints for [t] to be of type [c] in the context [ctx]. The
-   context [ctx] must be well-typed, and the type [c]
-   well-sorted. This function never fails (but constraints may be
-   unsatisfiable). *)
-let check : ctxt -> term -> term -> unif_constrs = fun ctx t c ->
+(** [check ctx t c] checks returns a list [cs] of unification constraints for
+   [t] to be of type [c] in the context [ctx]. The context [ctx] must be
+   well-typed, and the type [c] well-sorted. This function never fails (but
+   constraints may be unsatisfiable). *)
```
rlepigre, line 221: Comment not indented correctly.

src/core/sig_state.ml
```diff
@@ -0,0 +1,227 @@
...
+        let expected = fn ss pos in
+        if not (eq !(sym.sym_type) expected) then
+          fatal pos "The type of %s is not of the form %a."
+            sym.sym_name pp expected
+      in
+      register name check
+  end
+
+(** [remove_hint ss s hints] removes from [hints] the mapping for [s] if
+   [s.sym_name] is mapped in [ss.in_scope]. *)
```
rlepigre, line 95: Comment not indented correctly.

src/core/sig_state.ml
```diff
@@ -0,0 +1,227 @@
...
+
+    let register_expected_type eq pp name fn =
+      let check ss pos sym =
+        let expected = fn ss pos in
+        if not (eq !(sym.sym_type) expected) then
+          fatal pos "The type of %s is not of the form %a."
+            sym.sym_name pp expected
+      in
+      register name check
+  end
```
rlepigre, line 92: Please, revert deletion of `builtin.ml`.

src/core/sig_state.ml
```diff
@@ -0,0 +1,227 @@
...
+
+(** [remove_hint ss s hints] removes from [hints] the mapping for [s] if
+   [s.sym_name] is mapped in [ss.in_scope]. *)
+let remove_hint : sig_state -> sym -> hint SymMap.t -> hint SymMap.t
+  = fun ss s hints ->
```
rlepigre, line 97: Please put `=` on the previous line.

src/core/sig_state.ml
```diff
@@ -0,0 +1,227 @@
...
+  if StrMap.mem s.sym_name ss.in_scope then SymMap.remove s hints
+  else hints
+
+(** [add_symbol ss e p x a impl] adds a symbol in [ss]. *)
+let add_symbol
+    : sig_state -> expo -> prop -> strloc -> term -> bool list
+      -> term option -> sig_state
+  = fun ss e p x a impl t ->
```
rlepigre, line 105: This is not the style we use, use rather:
```ocaml
let add_symbol : sig_state -> expo -> prop -> strloc -> term -> bool list
    -> term option -> sig_state =
  fun ss e p x a impl t ->
```

src/core/sig_state.ml
```diff
@@ -0,0 +1,227 @@
...
+    | Some t -> s.sym_def := Some(t)
+    | None -> ()
+  end;
+  let hints = SymMap.add s No_hint (remove_hint ss s ss.hints) in
+  let in_scope = StrMap.add x.elt (s, x.pos) ss.in_scope in
+  {ss with in_scope; hints}
+
+(** [add_unop ss n (s,unop)] declares [n] as prefix and maps it to [s]. *)
+let add_unop : sig_state -> string -> (sym * unop) -> sig_state
+  = fun ss n (sym, unop) ->
```
rlepigre, line 118: Please, `=` on previous line.

src/core/sig_state.ml
```diff
@@ -0,0 +1,227 @@
...
+  Sign.add_unop ss.signature n (sym, unop);
+  let hints = SymMap.add sym (Prefix unop) (remove_hint ss sym ss.hints) in
+  {ss with hints}
+
+(** [add_binop ss n (s,binop)] declares [n] as infix and maps it to [s]. *)
+let add_binop : sig_state -> string -> (sym * binop) -> sig_state
+  = fun ss n (sym, binop) ->
```
rlepigre, line 125: Please, `=` on previous line.

src/core/sig_state.ml
```diff
@@ -0,0 +1,227 @@
...
+    empty, we search for the name [snd qid.elt] in the opened modules of [st].
+    The boolean [b] only indicates if the error message should mention
+    variables, in the case where the module path is empty and the symbol is
+    unbound. This is reported using the [Fatal] exception.
+    {!constructor:Terms.expo.Protec} symbols from other modules
+    are allowed in left-hand side of rewrite rules (only) iff [~prt] is true.
+    {!constructor:Terms.expo.Privat} symbols are allowed iff [~prv]
+    is [true]. *)
+let find_sym : prt:bool -> prv:bool -> bool -> sig_state -> qident -> sym
+  = fun ~prt ~prv b st qid ->
```
rlepigre, line 179: Please, `=` on previous line.

src/core/sig_state.ml
```diff
@@ -0,0 +1,227 @@
...
+    match (prt, prv, s.sym_expo) with
+    | (false, _    , Protec) ->
+        if s.sym_path <> st.signature.sign_path then
+          (* Fail if symbol is not defined in the current module. *)
+          fatal pos "Protected symbol not allowed here."
+    | (_    , false, Privat) ->
+        fatal pos "Private symbol not allowed here."
+    | _                      -> ()
+  end;
+  s
```
rlepigre, line 227: This can be simplified into:
```ocaml
  match (prt, prv, s.sym_expo) with
  | (false, _    , Protec) ->
      if s.sym_path = st.signature.sign_path then s else
      (* Protected symbol not defined in the current module, fail. *)
      fatal pos "Protected symbol not allowed here."
  | (_    , false, Privat) ->
      fatal pos "Private symbol not allowed here."
  | (_    , _    , _     ) -> s
```

src/core/print.ml
```diff
@@ -94,83 +134,75 @@ and pp_term : term pp = fun oc t ->
...
-          | Abst(a,t) -> let (x,t) = Bindlib.unbind t in
-                         out oc " %a%a" pp_arg (x,a) pp_absts t
-          | _         -> out oc ", %a" (pp `Func) t
-        in
-        pp_absts oc t;
+        out oc "λ";
+        let (x,t) = Bindlib.unbind b in
+        if Bindlib.binder_occur b then out oc "%a" pp_tvar x else out oc "_";
+        if !print_domains then out oc ": %a" (pp `Func) a;
+        out oc ", %a" (pp `Func) t;
```
rlepigre, line 148: So here you are not printing several abstractions at once anymore. Is that intended?

src/core/print.ml
```diff
@@ -94,83 +134,75 @@ and pp_term : term pp = fun oc t ->
...
-            let rec pp_prods oc c =
-              match unfold c with
-              | Prod(a,b) when Bindlib.binder_occur b ->
-                  let (x,b) = Bindlib.unbind b in
-                  out oc " %a%a" pp_arg (x,a) pp_prods b
-              | _                                      ->
-                  out oc ", %a" (pp `Func) c
-            in
-            pp_prods oc c
-          end
```
rlepigre, line 134: Same here, why did you remove the more compact printing?

src/core/print.ml
```diff
@@ -8,6 +8,16 @@
...
 open Extra
 open Terms
+open Console
+open Syntax
+open Sig_state
+
+(** Logging function for printing. *)
+let log_prnt = new_logger 'p' "prnt" "pretty-printing"
+let log_prnt = log_prnt.logger
+
```
rlepigre, line 18: This does not seem to be used.

src/core/eval.ml
```diff
@@ -377,25 +377,9 @@ let rec hnf : ctxt -> term -> term = fun ctx t ->
...
-  | SNF
-  (** Reduce to strong normal form. *)
-  | NONE
-  (** Do nothing. *)
-
-(** Configuration for evaluation. *)
-type config =
-  { strategy : strategy   (** Evaluation strategy.          *)
-  ; steps    : int option (** Max number of steps if given. *) }
-
```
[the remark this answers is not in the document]
rlepigre: I know it is not completely independent due to the `binary` and `unary` types as well, but I think we should still try to keep a separation here.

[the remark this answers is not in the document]
fblanqui: I would prefer to do the other way around. If we also move expo and prop from terms to syntax, then syntax does not depend on terms anymore, which is good. Then, we can also move the type definitions that are shared between syntax and terms into another file. Proposition for a name?

src/core/scope.ml
```diff
@@ -426,24 +339,24 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
...
         let rec unsugar_nat_lit acc n =
           if n <= 0 then acc else unsugar_nat_lit (_Appl sym_s acc) (n-1)
         in
         unsugar_nat_lit sym_z n
     | (P_UnaO(u,t)    , _                   ) ->
         let (s, impl) =
-          let (op,_,qid) = u in
-          let (s, _) = find_sym ~prt:true ~prv:true true ss qid in
-          (_Symb s (Unary(op)), s.sym_impl)
+          let (_op,_,qid) = u in
```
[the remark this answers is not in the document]
fblanqui: Why??? I find this feature very useful.

src/core/sig_state.ml
fblanqui: ok (but this was not my code...)

src/core/print.ml
fblanqui: I prefer this one. It has less parentheses. Too much parentheses is bad for readability.

src/core/print.ml
fblanqui: It has been used and may be useful later as well.

src/core/print.ml
fblanqui: Yes, by symmetry with products.

src/core/handle.ml
fblanqui: It was depending on the order the builtins where defined. See added examples. LP was failing if +1 was defined before 0.

src/core/handle.ml
fblanqui: I prefer not to use pp. It is not explicit enough.

src/core/infer.ml
fblanqui: We don't have the same environment and editor. Next time, it's better that you fix the indentation yourself.

src/core/rewrite.ml
fblanqui: See my previous comment. Thanks to fix indentation yourself. I don't have the time to do this.

src/core/rewrite.ml
fblanqui: My editor...

src/core/rewrite.ml
fblanqui: A problem with emacs I guess...

src/core/rewrite.ml
fblanqui: I will ignore these comments in the future.

src/core/eval.ml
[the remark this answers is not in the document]
rlepigre: The whole point here is that the "kernel", that is the important part for the theory, should be as small as possible. With what you are saying we would be making it bigger. It does not matter if the front-end (i.e., there parser level syntax) depends on types defined by the "kernel". However the other way around is more problematic.

fblanqui:
@rlepigre please do not do comments on indentation anymore. Just fix indentations you don't like by yourself. I don't have the time to take care of all these comments.

src/core/scope.ml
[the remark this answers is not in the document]
rlepigre: Because that make harder to parse when you look over the code quickly. We don't need it, there is no reason to name it. The underscore disables the warning but does not prevent someone from using the value in the code. This is just bad style.

src/core/tree.ml
fblanqui: This function is unused. @gabrielhdt do you want to keep it?

src/core/print.ml
rlepigre: It has maybe less parentheses, but it introduces several dependent product symbols.
Do you really prefer `Π a : A, Π b : B, Π c : C, D` over `Π (a : A) (b : B) (c : C), D`? The choice does not seem obvious to me. And we can modify the later to factor consecutive variables of the same type so this would become `Π a : A, Π b : A, Π c : A, B` against `Π (a b c : A), B`, which is a clear win. This would have been a trivial modification from the factored one.

src/core/print.ml
rlepigre: Yeah, but when you have no domain you end up with `λ x, λ y, λ z, t` instead of `λ x y z, t`.

src/core/print.ml
fblanqui: Mathematicians do not write parentheses everywhere.  Π a b c : A, B is good as well and can be done in another PR.

rlepigre:
> @rlepigre please do not do comments on indentation anymore. Just fix indentations you don't like by yourself. I don't have the time to take care of all these comments. In the future, I will ignore these comments.

That is unacceptable @fblanqui. I have spent more than an hour reviewing, you cannot just dismiss comments as you please. This is your PR, so it is your responsibility to answer comments and to ensure that the code is uniform with the rest of the implementation. Especially if problems are pointed out. Now you are just wasting my time twice: the first time when I reviewed code that did not follow our usual guidelines in the first place, and now because you dismiss the comments I took the time to write. As I already told you, code quality is important. We cannot allow PRs to make the code worst. Especially code that you should not have touched.

src/core/print.ml
fblanqui: Done in c1dd49e.

fblanqui:
Concerning this kind of comments, yes, I can and will do. Rodolphe, I try to follow the style you imposed with the tools I have. But they are not perfect. I accepted this style because I don't care. But you should not exaggerate and make me lose my time with such comments. If you are not happy with the style, fix it by yourself. Otherwise, I will impose ocamlformat as suggested by @francoisthire in https://github.com/Deducteam/lambdapi/pull/367#issuecomment-616091916. Like this, there will be no such problem.

rlepigre:
> Concerning this kind of comments, yes, I can and will do. Rodolphe, I try to follow the style you imposed with the tools I have. But they are not perfect. I accepted this style because I don't care. But you should not exaggerate and make me lose my time with such comments. If you are not happy with the style, fix it by yourself. Otherwise, I will impose ocamlformat as suggested by @francoisthire in [#367 (comment)](https://github.com/Deducteam/lambdapi/pull/367#issuecomment-616091916). Like this, there will be no such problem.

Well, I am not your engineer. And it is not like I have time to go cleanup after everyone's mess. Since the beginning of the project **I am forcing myself** to adhere to a strict and uniform coding style that, I am certain, makes it easier for everyone to read, understand, review, and work on the code. And I think it is not too much to ask that everyone does the same. The point of commenting on style is that people are intended to get into the habit of respecting the conventions. This happens in most software projects at some level, and this is no exaggeration. It would be a big mistake to do otherwise here.

What makes us lose time is you not following the style of the project in the first place. Me having to comment on that constantly is only a consequence. I don't understand why @gabrielhdt can follow the style close to perfection, while you don't even seem to try. This is not a tool problem, I know for a fact that Gabriel and I use very different setups. What is a big part of the problem though, it that you keep doing unnecessary modifications throughout the code all the time. There would be a lot fewer comments if you only touched the code you need to in a given PR.

src/core/tree.ml
gabrielhdt: Well, it sure is useful for debugging, but one can always reimplement it. It can be removed.

gabrielhdt: I mean it is more up to you to decide, rather we want easily debuggable code at the expense of some dead code; or we want to avoid forgotten piece of code. I don't have a clear stand on this question.

fblanqui: Well, then I am going to keep just in case.

ejgallego:
> It doesn't look like a good idea to multiply global references. It will prevent us from having several threads in the future (@ejgallego and @01mf02 already mentioned this problem).

Printers having imperative state have been a constant source of problems for example in Coq, keep in mind that printers are often called asynchronously (for example to print errors), so indeed once the codebase is large it becomes hard to keep things in sync.

There is consensus among Coq devs in making the printing entry API referentially transparent, it is just a lot or work for a system such as Coq.

People doing data-mining with Coq SerAPI for example are often users of async printing; imagine a machine learning application, they may want to print random parts of the document so it is hard to manage unless you can indeed encapsulate the state. SerAPI does try to workaround that but even so bugs happen here and there from time to time.

ejgallego:
Regarding OCaml multicore I don't know, but I have been running it for a while and indeed it works. Coq doesn't yet have multicore support because of all the bad API choices made there; cleanup is gonna take months or even more :/

PR #367, opened 2020-04-19
Setup git hook

### LINKED ISSUE #366: Setting up a git hook to check sanity
We should force the sanity check to run before each commit.

gabrielhdt:
This PR adds a command in the makefile to install a pre-commit hook running sanity check. Any commit that make `make sanity_check` return some string will be refused **before** comitting.

Closes #366.


GNUmakefile
```diff
@@ -55,6 +55,12 @@ real_tests: bin
 sanity_check: tools/sanity_check.sh
 	@./$<
 
+.PHONY: git-setup
+git-setup:
+	@printf '#!/bin/sh\n[ -z "$$(make sanity_check)" ]\n' > \
+.git/hooks/pre-commit
```
fblanqui, line 61: What if someone has its own pre-commit file already?

fblanqui:
What not simply have make start by `make sanity_check`?

GNUmakefile
gabrielhdt: Well, perhaps we can expect that person not to run `make git-setup`.

gabrielhdt: Or that person shall know how to call sanity check from his/her own hook.

fblanqui:
I would be in favor of simply having make start by make sanity_check.

gabrielhdt:
> What not simply have make start by `make sanity_check`?

Why not. It's just that git hooks are tailored for this. I can't see a clear advantage of a method on another, except that when calling `make`, IMO, the user expects the code to be compiled, not linted or checked this way; and we want to avoid surprising anyone.

gabrielhdt:
> What not simply have make start by `make sanity_check`?

And more simply, someone who commit without compiling won't have his/her code checked.

fblanqui:
Right. By then, why not having a pre-commit that runs make and make tests?

gabrielhdt:
> Right. But then, why not having a pre-commit that runs make and make tests?

Yeah, I thought of that, it can be done. The only drawback is that it forces people to install dune and everything. But perhaps we want that as well. We might even be able to select some tests and run them (we can't run them all, having to wait even 10s before committing would be annoying).

GNUmakefile
francoisthire: I agree with @fblanqui , as it is you erase the file if it exists. You should add a check before and print a message if it already exists.

gabrielhdt: OK. In first place I wanted to have a short instruction, but you're both right. I'll fix it.

francoisthire:
Proably @rlepigre knows it already but there is a package called `ocamlformat` that formats code according to some pre-defined rules. `ocamlformat` can be hooked to your editor when you save a file. The advantage of `ocamlformat` is that it can do more than the sanity check script and can be plugged easily in a CI. However, it means that everyone agrees to use the same style. `ocamlformat` can be parameterized with a single file in every sub-directories.

rlepigre:
> Proably @rlepigre knows it already but there is a package called `ocamlformat` that formats code according to some pre-defined rules. `ocamlformat` can be hooked to your editor when you save a file. The advantage of `ocamlformat` is that it can do more than the sanity check script and can be plugged easily in a CI. However, it means that everyone agrees to use the same style. `ocamlformat` can be parameterized with a single file in every sub-directories.

I have not tried this one, but my usual complain with such tools is that they are both two rigid and two flexible. There are many situations where I want to not follow the global conventions locally. For example, we often use the following scheme:
```
match ... with
| ... -> ..
| ... -> ...
| ... ->
match ... with
| ... -> ..
| ... -> ...
```
and this is typically rejected by formatting tools. However it is very natural to do that in some cases.

Then the tools are too flexible because I would like to reject stuff like
```
let x, y = ... in
```
to force parentheses around tuples. But that cannot be done it seems. Similarly, I would like to reject uses of
```
let x = ... and y = ... in
```
but that cannot be done either. In the end, I always end up enforcing conventions by hand.

rlepigre:
@gabrielhdt I am not sure that having a `Makefile` target is the way to go here. This setup step is only be performed once, and there is no reason to do it again later. I think it would be better to extend our guidelines for developers to explain that they can use the file you propose as a pre-commit hook.

gabrielhdt:
> @gabrielhdt I am not sure that having a `Makefile` target is the way to go here. This setup step is only be performed once, and there is no reason to do it again later. I think it would be better to extend our guidelines for developers to explain that they can use the file you propose as a pre-commit hook.

It's fine by me. We can as well set a receive hook which would make the github server reject any "invalid" file. It's a bit brutal but why not, even more if we provide easy ways to check locally. What do you think?

rlepigre:
> It's fine by me. We can as well set a receive hook which would make the github server reject any "invalid" file. It's a bit brutal but why not, even more if we provide easy ways to check locally. What do you think?

I'm not sure I understand what you are proposing here. Are you saying that we can set up GitHub so that it does the rejection? Can we do that?

gabrielhdt:
> > It's fine by me. We can as well set a receive hook which would make the github server reject any "invalid" file. It's a bit brutal but why not, even more if we provide easy ways to check locally. What do you think?
> 
> I'm not sure I understand what you are proposing here. Are you saying that we can set up GitHub so that it does the rejection? Can we do that?

At least git allows this, see https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks.
I guess if we use the github interface to create a file `.git/hooks/pre-receive`, it'll run and reject any invalid push.

gabrielhdt:
Ah, it cannot be done, we can't touch the `.git` folder of github.

rlepigre:
> Ah, it cannot be done, we can't touch the `.git` folder of github.

Yeah, I guess you need the enterprise version of GitHub to get such a feature. But no matter, developers should just configure their local hooks correctly then.

gabrielhdt:
So here is my proposition.

* The script `tools/git_hook_helper.sh` creates a `pre-commit` hook that run sanity check and `make bin` *if* there is no hook already.
* If there is already a hook, it only displays a message with the command that should be run.
* The script can take a `-s` option to only check for sanity. This can be useful when we are working on PR that do not compile yet.

rlepigre:
I don't like that the `all` target depends on `sanity_check`, because that is too restrictive while you are developing (and typing `make` all the time).

Installing should be the way to go, preventing a commit is the perfect way to catch the problem.

gabrielhdt:
> I don't like that the `all` target depends on `sanity_check`, because that is too restrictive while you are developing (and typing `make` all the time).
> 
> Installing should be the way to go, preventing a commit is the perfect way to catch the problem.

I see what you mean. You're suggesting to run sanity check on the `install` rule?

rlepigre:
No, I am suggesting to never run sanity checks from the Makefile (except with `make sanity_check`). However, we should never accept a commit that is not "sane", hence the Git hook.

rlepigre:
To rephrase what I am saying: we should not care in what crappy state your own repository may be, as long as you do not push crap. Running the hook exactly prevents that, and it is a good idea to check for compilation as well.

CONTRIBUTING.md
```diff
@@ -18,6 +18,10 @@ You should at the very least run `make sanity_check` before committing anything.
 Please check you have GNU awk (gawk) installed or another UTF-8 compatible
 implementation of the AWK programming language interpreter.
 
+The script `tools/git_hook_helper.sh` helps setting up a git hook to run
+`make sanity_check` and `make bin` automatically *before* each commit.
+The option `-s` makes the hook only check for sanity (if you are in early stage
+of development).
```
rlepigre, line 24: I don't understand, do you expect people to use this script more than once? (The one time being after they cloned the repository, or whenever they decide to set up the hook.)

gabrielhdt: Not really, but some may want a hook that builds before commit (normal behaviour when doing rather limited modifications), and other might want to push commits that do not compile (in certain cases).

In short, you may want to change your hook when you're starting a big PR and you want to push content that don't compile (to your branch, so it does not involve the upstream).

gabrielhdt: Or we can expect the user to disable the hook by him/herself if needed.

rlepigre: Well, it is always a good idea to keep the code compiling, even if it is on your own branch. In general it is also a good idea to keep your history clean to make review easier, which is something that most of us do not do. So I would say that we should advise everyone to set up hooks that run sanity checks only all the time (of course there is no way to impose that), but everything should be tested by CI.

Ideally, we would like all individual commits to go through CI.

gabrielhdt: Anyway, everything *is* tested by CI, as long as we make pull requests.

gabrielhdt: Does it look clearer @rlepigre ?

rlepigre: Yeah, that looks fine to me.

Issue #209, opened 2019-04-03
Natural numbers builtin are not printed as such


fblanqui:
When defining the builtins "0" and "+1", we expect goals to be printed using natural numbers, at least for "0" which cannot be used as identifier. This is not the case currently.

fblanqui:
This means that printing should depend on the signature, or at least the builtins.

rlepigre:
No, not really. This particular example only concerns syntactic sugar like zero and successor. Other builtins do not suffer from the same problem, since they only give a "standard" name to some constructor. In their case, printing the actual constructor (not the builtin name) is the expected result.

For natural number literals, we could arguably recognize literals when the builtins are set. This would not be difficult to do.

fblanqui:
Ok, but how to know that those builtins are set? By using some reference?

rlepigre:
Good point, it is not completely trivial.

PR #377, opened 2020-04-27
fix operator printing + print nats using builtins (#209)

fblanqui:


src/core/sig_state.ml
```diff
@@ -7,112 +7,180 @@ open Files
...
-  ; builtins  : Sign.builtin_map          (** Builtin symbols.     *)
-  ; pp_hints  : pp_hint SymMap.t          (** Printing hints.      *) }
+  { signature : Sign.t          (** Current signature.   *)
+  ; in_scope  : in_scope_map    (** Symbols in scope.    *)
+  ; aliases   : Path.t StrMap.t (** Established aliases. *)
+  ; path_map  : path_map        (** Reverse map of [aliases]. *)
+  ; builtins  : builtin_map     (** Builtin symbols.     *)
+  ; unops     : sym StrMap.t    (** Unary operators.     *)
+  ; binops    : sym StrMap.t    (** Binary operators.    *)
+  ; pp_hints  : pp_hint_map     (** Printing hints.      *) }
```
gabrielhdt, line 44: Closing comments may be aligned using a rectangular selection (`C-x SPC` or `M-x rectangle-mark-mode`) then `M-x align-regexp RET \*)`.

src/core/sig_state.ml
```diff
@@ -7,112 +7,180 @@ open Files
...
+  fun map name pp_hints ->
+  try SymMap.remove (StrMap.find name map) pp_hints
+  with Not_found -> pp_hints
+
+(** [remove_pp_hint_eq map name h pp_hints] removes from [pp_hints] the
+   mapping for [s] if [s] is mapped to [(name,h')] in [map], and [eq_pp_hint h
+   h' = true]. *)
+let remove_pp_hint_eq :
+      in_scope_map -> string -> pp_hint -> pp_hint_map -> pp_hint_map =
+  fun in_scope name h pp_hints ->
```
gabrielhdt, line 70: The name of the arguments and the comment don't match.

src/core/sig_state.ml
```diff
@@ -7,112 +7,180 @@ open Files
...
-  let pp_hints = remove_pp_hint ss sym ss.pp_hints in
+  fun ss name ((sym, binop) as x) ->
+  Sign.add_binop ss.signature name x;
+  let binops = StrMap.add name sym ss.binops in
+  let pp_hints = remove_pp_hint ss.binops name ss.pp_hints in
   let pp_hints = SymMap.add sym (Infix binop) pp_hints in
-  {ss with pp_hints}
+  {ss with binops; pp_hints}
 
 (** [add_builtin ss n s] adds builtin [n] and maps it to [s]. *)
```
gabrielhdt, line 110: `ss` should be mentioned in the comments.

src/core/print.ml
```diff
@@ -36,24 +36,27 @@ let pp_assoc : assoc pp = fun oc assoc ->
...
+     | Assoc_right -> " right")
 
 (** [hint oc a] prints hint [h] to channel [oc]. *)
 let pp_hint : pp_hint pp = fun oc pp_hint ->
   match pp_hint with
-  | No_hint -> ()
+  | Unqual -> ()
   | Prefix(n,p,_) -> Format.fprintf oc "prefix %s %f" n p
-  | Infix(n,a,p,_) -> Format.fprintf oc "infix %s %a %f" n pp_assoc a p
+  | Infix(n,a,p,_) -> Format.fprintf oc "infix %s%a %f" n pp_assoc a p
```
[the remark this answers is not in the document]
gabrielhdt: And we might need to set some convention to quote terms of the language, we used `[f]` for some time but I understood you preferred `"f"`, we can use any of the two, but it would be good to remain consistent and always mark the elements that are from the language while printing.

src/core/sig_state.ml
gabrielhdt: Arrows `->` can be aligned as well with this.

src/core/print.ml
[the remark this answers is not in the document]
fblanqui: About pp_hint printing: it's mainly used for debugging but, yes, we can have more verbose messages.

About uniformity: you are right, it would be nice to be more uniform. But I also want to have something light. [] may be useful when to clearly distinguish multiple terms in a message. But, for a function symbol, I do not see the need, unless there may be some confusion with other elements of the message. I use "" for builtins as this is currently the syntax used to input them. If you spot places that could be fixed, let me know.

src/core/print.ml
```diff
@@ -65,6 +68,25 @@ let pp_symbol : sym pp = fun oc s ->
 let pp_tvar : tvar pp = fun oc x ->
   Format.pp_print_string oc (Bindlib.name_of x)
 
+(** Exception raised when trying to convert a term into a nat. *)
+exception Not_a_nat
+
+(** [nat_of_term t] converts a term into a natural number. Raises [Not_a_nat]
+   if this is not possible. *)
```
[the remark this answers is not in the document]
fblanqui: Good to know. We could do this in many more places.

src/core/sig_state.ml
fblanqui: Ok. But all these alignment constraints artificially increase diff's...

gabrielhdt: True, but github allows to hide whitespace change (although you have to reset the option each time...)

src/core/sig_state.ml
```diff
@@ -1,4 +1,13 @@
-(** Signature state. *)
+(** Signature state.
+
+   This module provides a record type [sig_state] containing all the
+   informations needed for scoping p_terms and printing terms, and functions
+   on this type for manipulating it. In particular, it provides functions
+   [open_sign], [add_symbol], [add_binop], etc. taking a [sig_state] as
+   argument and returning a new [sig_state] as result. These functions call
+   the corresponding functions of [Sign] which should not be called directly
+   but though the current module only, in order to setup the [sig_state]
```
gabrielhdt, line 9: ~~though~~ through?

PR #382, opened 2020-04-30
WIP induction

amelieled:
My first step about the tactic induction (without my previous proofs).


src/core/handle.ml
```diff
@@ -187,6 +187,7 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
       let d = if op then None else Some(t) in
       let ss = Sig_state.add_symbol ss e Defin x a impl d in
       (ss, None)
+  | P_inductive(_, _, _, _) -> failwith "to be implemented" (* P_inductive(e, s, t, tl) @Moi *)
```
fblanqui, line 190: Please, to be use failwith. Instead, just make a warning using wrn (see example at the end of handle.ml).

src/core/pretty.ml
```diff
@@ -205,6 +205,10 @@ let pp_command : p_command pp = fun oc cmd ->
       List.iter (out " %a" pp_p_arg) args;
       Option.iter (out " :@ @[<hov>%a@]" pp_p_term) ao;
       out " ≔@ @[<hov>%a@]@]" pp_p_term t
+  | P_inductive(e,s,t,tl)   ->                         (* @Moi *)
```
fblanqui, line 208: Thanks Amélie to avoid comments like (*@Moi*).

src/core/syntax.ml
```diff
@@ -198,6 +198,8 @@ type p_command_aux =
   | P_definition of Terms.expo * bool * ident * p_arg list * p_type option
                   * p_term
   (** Definition of a symbol (unfoldable). *)
+  | P_inductive of Terms.expo * ident * p_type * p_term list   (* @Moi *)
```
fblanqui, line 201: Shouldn't you have (ident * p_term) list instead of p_term list? (You need the names of constructors.) And p_type should optional (p_term option).

src/core/pretty.ml
amelieled: It's OK now.

src/core/syntax.ml
amelieled: You right but I have a problem to define the good pretty in pretty.ml.

src/core/handle.ml
amelieled: It's OK now.

src/core/syntax.ml
amelieled: Now it's OK.

src/core/handle.ml
```diff
@@ -187,6 +187,9 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
       let d = if op then None else Some(t) in
       let ss = Sig_state.add_symbol ss e Defin x a impl d in
       (ss, None)
+  | P_inductive(_, _, _, _) -> wrn cmd.pos "to be implemented";
+                               (Sig_state.empty ,None)
```
fblanqui, line 191: you should return the state unchanged: (ss, None)

src/core/parser.ml
```diff
@@ -624,6 +625,12 @@ let parser cmd =
       -> P_rules(r::rs)
   | e:exposition? _definition_ s:ident al:arg* ao:{":" term}? "≔" t:term
       -> P_definition(Option.get Terms.Public e,false,s,al,ao,t)
+  | e:exposition? _inductive_ i:ident t:{":" term}? "≔"
+                   "|"? ifirst:ident ":" tfirst:term
+                 c:{ "|"  ilist :ident ":" tlist :term }*
+      -> P_inductive(Option.get Terms.Public e, i, t,
+                     (*(ifirst, tfirst)::(List.combine c.ilist tlist))*)
```
fblanqui, line 632: comment to be removed

src/core/pretty.ml
```diff
@@ -182,6 +182,9 @@ let pp_p_tactic : p_tactic pp = fun oc t ->
   | P_tac_why3(Some(s))      -> out "why3 %s" s
   | P_tac_query(q)           -> pp_p_query oc q
 
+let pp_pair pp_a pp_b oc fmt (a,b) =
+  Format.fprintf oc fmt pp_a a pp_b b
```
fblanqui, line 186: This function doesn't seem very useful as it could easily be inlined.

src/core/pretty.ml
```diff
@@ -205,6 +208,15 @@ let pp_command : p_command pp = fun oc cmd ->
       List.iter (out " %a" pp_p_arg) args;
       Option.iter (out " :@ @[<hov>%a@]" pp_p_term) ao;
       out " ≔@ @[<hov>%a@]@]" pp_p_term t
+  | P_inductive(e,s,t,tl)   ->
+      out "@[<hov 2>%ainductive %a" pp_expo e pp_ident s;
+      Option.iter (out " :@ @[<hov>%a@]" pp_p_term) t;
+      List.iter (pp_pair pp_ident pp_p_term oc (" ≔@ @[<hov>%a@] :@ @[<hov>%a@]")) tl
```
fblanqui, line 214: ":=" should be printed before List.iter. And List.iter should print "|" before each case, except perhaps the first one.

src/core/syntax.ml
```diff
@@ -338,6 +340,13 @@ let eq_p_command : p_command eq = fun c1 c2 ->
   | (P_definition(e1,b1,s1,l1,a1,t1), P_definition(e2,b2,s2,l2,a2,t2)) ->
       e1 = e2 && b1 = b2 && eq_ident s1 s2 && List.equal eq_p_arg l1 l2
       && Option.equal eq_p_term a1 a2 && eq_p_term t1 t2
+  | (P_inductive(e1,s1,t1,tl1)   , P_inductive(e2,s2,t2,tl2)    ) ->
+      e1 = e2 && eq_ident s1 s2 && Option.equal eq_p_term t1 t2
+      &&
+        let (t1_ident, t1_term) = List.split tl1 in
+        let (t2_ident, t2_term) = List.split tl2 in
+        List.equal eq_ident t1_ident t2_ident &&
+          List.equal eq_p_term t1_term t2_term
```
fblanqui, line 349: There is no need to split: just give the right function to List.equal.

doc/syntax.bnf
```diff
@@ -133,7 +133,7 @@ reserved "why3"
   | "focus" <nat_lit>
   | "print"
   | "proofterm"
-  | "why3" <string_lit>?
+  | "why3" <string_lit>?
```
fblanqui, line 136: Why do you change anything here? On the other hand, you should add something about "inductive". Idem in doc/sections/commands.md (just saying that it generates new symbols; details will be given later).

amelieled: I think it's just a bug from github. As you can see the two lines are the same.
OK I won't forget that.

tests/OK/inductive_type.lp
```diff
@@ -0,0 +1,11 @@
+inductive nat : TYPE ≔
+ | z    : nat
+ | succ : nat → nat
+
+inductive list ≔
+ | nil  : list
+ | cons : a → list → list
```
fblanqui, line 7: You probably don't want this example to be accepted as it contains an undefined symbol `a`. Perhaps replace `a` by `nat` for the moment.

tests/OK/inductive_type.lp
```diff
@@ -0,0 +1,11 @@
...
+
+symbol nat : TYPE
```
fblanqui, line 9: If `inductive nat` generates a symbol `nat`, this command will fail as one cannot define the same object twice in the same file.

tests/OK/inductive_type.lp
```diff
@@ -0,0 +1,11 @@
...
+
+definition test (x:nat) ≔ x
```
fblanqui, line 11: Once inductive is implemented, more interesting tests should be given instead like buillding a nat or a list to check that constructors and types are indeed correctly added in the signature: `assert succ z : nat`.

doc/syntax.bnf
fblanqui: They look the same but perhaps they are not (a different space character).

tests/OK/inductive_type.lp
amelieled: It's a mistake. I deleted it. 
Now, the file pretty.ml is OK (it wasn't the case before).

doc/syntax.bnf
amelieled: Oh I know!
Yes, before this space character wasn't it. Probably someone used Ctrl + space.
Now it's OK.

fblanqui:
@amelieled you also need to update the emacs mode (lambdapi-smie.el and perhaps other files).

src/core/handle.ml
```diff
@@ -89,19 +90,20 @@ let handle_require_as : popt -> sig_state -> Path.t -> ident -> sig_state =
     This structure contains the list of the tactics to be executed, as well as
     the initial state of the proof.  The checking of the proof is then handled
     separately. Note that [Fatal] is raised in case of an error. *)
-let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
+let rec handle_cmd : sig_state -> p_command
+                     -> sig_state * proof_data option * inductive option =
```
fblanqui, line 94: A less disruptive solution would be to factorize out the code for the case P_symbol by adding a function handle_symbol.

src/core/handle.ml
```diff
@@ -186,7 +188,48 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
-      (ss, None)
+      (ss, None, None)
+  | P_inductive(e, s, t, tl) ->
+      (* Obtaining of the global type *)
+      let type_symbol_head = match t with
+        | None -> { elt = P_Type ; pos = None }
+        | Some a -> a
+      in
+      (* Add the first symbol in the signature *)
+      let (ss, _, _) = handle_cmd ss
```
fblanqui, line 199: Thanks to follow the style currently used. See CONTRIBUTING.md.

src/core/handle.ml
```diff
@@ -186,7 +188,48 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+                      ({elt = P_symbol(e, Defin, s, [], type_symbol_head);
+                        pos = cmd.pos})
```
fblanqui, line 201: Use t.pos instead.

src/core/handle.ml
```diff
@@ -186,7 +188,48 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
       out 3 "(symb) %s ≔ %a\n" x.elt pp_term t;
       let d = if op then None else Some(t) in
       let ss = Sig_state.add_symbol ss e Defin x a impl d in
-      (ss, None)
+      (ss, None, None)
+  | P_inductive(e, s, t, tl) ->
+      (* Obtaining of the global type *)
+      let type_symbol_head = match t with
+        | None -> { elt = P_Type ; pos = None }
```
fblanqui, line 195: This is `none P_Type`.

src/core/handle.ml
```diff
@@ -186,7 +188,48 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+        | Some a -> a
+      in
+      (* Add the first symbol in the signature *)
+      let (ss, _, _) = handle_cmd ss
+                      ({elt = P_symbol(e, Defin, s, [], type_symbol_head);
+                        pos = cmd.pos})
+      in
+      let create_symbol : string -> term -> term option -> sym  =
```
fblanqui, line 203: This function doesn't seem necessary as this is done in handle_symbol.

fblanqui: You just need to modify Sig_state.add_symbol and handle_symbol so that they return the added symbol as well.

src/core/handle.ml
```diff
@@ -186,7 +188,48 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+                                -> sig_state * sym list =
+        fun ss_cur syml l -> match l with
+                   | [] -> (ss_cur, syml)
+                   | (ident, pterm)::q ->
+                       let (ss_new, _, _) =
+                         handle_cmd ss_cur
+                           ({elt = P_symbol(e, Injec, ident, [], pterm);
+                             pos = cmd.pos})
+                       in
+                       let term = scope_term e ss_cur Env.empty pterm in
```
fblanqui, line 223: Scoping is done in handle_symbol.

src/core/inductive.ml
```diff
@@ -0,0 +1,64 @@
...
+  | "and"
+  | "imp"
+*)
+
+open Timed
+open Terms
+open Syntax
+
+type inductive =
+  { name  : ident      ; (* the name of the inductive type     *)
```
fblanqui, line 28: Prefix all field names by "ind_" to avoid name clashes.

src/core/inductive.ml
```diff
@@ -0,0 +1,64 @@
...
+    sort  : p_type     ; (* the sort of the inductive type     *)
```
fblanqui, line 29: sort -> type

src/core/inductive.ml
```diff
@@ -0,0 +1,64 @@
...
+    rule  : sym list   ; (* the list of constructors           *)
+    _ind  : sym option ; (* one inductive principle (Prop one) *)
+    _rec  : sym option ; (* one recursive principle (Set one)  *)
+    _rect : sym option ; (* one recursive principle (Type one) *)
+    _inv  : sym option   (* one inversion principle            *) }
```
fblanqui, line 34: Focus on induction principle in Prop only for the moment.

src/core/inductive.ml
```diff
@@ -0,0 +1,64 @@
...
+    | []   -> assert false
+    | [a]  -> constructor_to_term a
+    | t::q -> Appl(constructor_to_term t, create_premises q)
+  in
+  let premises = create_premises i.rule in
+  (* ending principal *)
+  let ending = Wild (* Prod *) in
+  let t = Appl(head, Appl(premises, ending)) in
+  let induc_principle =
+    { sym_name  = (i.name.elt)^"_ind" ;
```
fblanqui, line 54: You shouldn't create symbols directly but call Sign.add_symbol instead.

src/core/syntax.ml
```diff
@@ -198,6 +198,8 @@ type p_command_aux =
   | P_definition of Terms.expo * bool * ident * p_arg list * p_type option
                   * p_term
   (** Definition of a symbol (unfoldable). *)
+  | P_inductive of Terms.expo * ident * p_type option * (ident * p_term) list
```
fblanqui, line 201: p_type -> p_term, as it may not be actually a LP type later

tests/OK/inductive_type.lp
```diff
@@ -0,0 +1,9 @@
+inductive nat : TYPE ≔
+ | z    : nat
+          | succ : nat → nat
```
fblanqui, line 3: indentation

src/core/inductive.ml
fblanqui: Parentheses are useless here.

src/core/handle.ml
amelieled: Yes, I can do that but the main problem is the new type for the function "handle_cmd", not the fact that it becomes recursive, no?
But I must do this last modification.

src/core/inductive.ml
amelieled: I'm not sure to understand your first remark.
The job about the second is done.

src/core/inductive.ml
amelieled: Yes, don't worry :)

src/core/handle.ml
fblanqui: I think that you do not need to change the type and make it recursive. It is sufficient to add handle_symbol.

src/core/inductive.ml
fblanqui: Don't create a sym data structure directly, call Sign.add_symbol instead.

src/core/handle.ml
amelieled: I don't see the problem here. I'm not in this case:
```
| A(x) ->
>>>>let y = ... in
>>>>...
```

fblanqui: I am talking of the following lines. As I won't be available the next 2 days, thanks to check this with @gabrielhdt .

src/core/handle.ml
amelieled: OK, but how can I create my structure, I mean my record, to store enough data to build the inductive principle?
In fact, I followed the same process that with the constructor `P_theorem` where you need to store some information.

src/core/handle.ml
amelieled: Maybe the name of the function is a bit confusing.
The function `create_symbol` is useful to build my record, and now I use `handle_symbol` (and not `handle_cmd`) to add the symbols in the signature.

src/core/handle.ml
amelieled: Yes but I need again this information when I would like to build my record.
I can change the type of my function `handle_symbol` to obtain a pair with the scoping term, but I'm not sure it's better.

src/core/handle.ml
gabrielhdt: Regarding style, if a `let` binding is using several lines, the line is broken after the `=`,  so
```ocaml
let (ss,_,_) =
  handle_cmd ...
in
```

By the way, when defining records, use
```ocaml
t = { f  : t 
    (** Ocamldoc comment *)
    ; f2 : u
    (** Ocamldoc comment *) }
```
and we tend to put parentheses around arguments of constructors.

src/core/handle.ml
fblanqui: Hi @amelieled . Could you please make handle_symbol return a symbol. Then, you do not need to change the type of handle_cmd and make it recursive? You just have to call handle_symbol from handle_inductive. Thanks.

src/core/handle.ml
```diff
@@ -74,35 +75,22 @@ let handle_require : bool -> popt -> sig_state -> Path.t -> sig_state =
   if b then handle_open pos ss p else ss
 
 (** [handle_require_as pos ss p id] handles the command [require p as id] with
-   [ss] as the signature state. On success, an updated signature state is
-   returned. *)
+    [ss] as the signature state. On success, an updated signature state is
```
fblanqui, line 78: Thanks Amélie to not change parts not related with your PR.

src/core/handle.ml
```diff
@@ -187,6 +197,35 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
       let d = if op then None else Some(t) in
       let ss = Sig_state.add_symbol ss e Defin x a impl d in
       (ss, None)
+  | P_inductive(e, s, t, tl)     ->
+      (* Obtaining of the global type *)
+      let type_symbol_head = match t with
+        | None   -> none P_Type
```
fblanqui, line 203: We should probably make the term t mandatory at the parsing level because there is no reason to set this type to TYPE.

src/core/handle.ml
```diff
@@ -122,8 +110,30 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
         end;
       (* Actually add the symbol to the signature and the state. *)
       out 3 "(symb) %s : %a\n" x.elt pp_term a;
-      (Sig_state.add_symbol ss e p x a impl None, None)
-  | P_rules(rs)                ->
+      Sig_state.add_symbol ss e p x a impl None
```
fblanqui, line 113: It would be good to have handle_symbol, and thus Sig_state.add_symbol, return the created symbol as well.

src/core/parser.ml
```diff
@@ -624,6 +625,11 @@ let parser cmd =
       -> P_rules(r::rs)
   | e:exposition? _definition_ s:ident al:arg* ao:{":" term}? "≔" t:term
       -> P_definition(Option.get Terms.Public e,false,s,al,ao,t)
+  | e:exposition? _inductive_ i:ident t:{":" term}? "≔"
```
fblanqui, line 628: Please make t mandatory.

src/core/syntax.ml
```diff
@@ -198,6 +198,8 @@ type p_command_aux =
   | P_definition of Terms.expo * bool * ident * p_arg list * p_type option
                   * p_term
   (** Definition of a symbol (unfoldable). *)
+  | P_inductive of Terms.expo * ident * p_term option * (ident * p_term) list
```
fblanqui, line 201: Thanks to replace p_term option by p_term.

doc/syntax.bnf
fblanqui, line 136: Thanks to update syntax.bnf and commands.md.

src/core/handle.ml
```diff
@@ -187,6 +197,35 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+      (* Create the body of the inductive type *)
+      let rec add_constructor : sig_state -> sym list -> (ident * p_term) list
+                                -> sig_state * sym list = fun ss_cur syml l ->
+        match l with
+        | []                -> (ss_cur, syml)
+        | (ident, pterm)::q ->
+            let ss = handle_symbol ss_cur e Injec ident [] pterm in
+            let term = scope_term e ss Env.empty pterm in (*empty env ?*)
+            let new_sym = create_sym (ident.elt) term Injec e in
+            add_constructor ss ((new_sym)::syml) q
```
fblanqui, line 218: create_sym is useless and wrong here as the right symbol is created in handle_symbol. You can get it back by changing the code of Sig_state.add_symbol and handle_symbol so that they return a sig_state * sym.
Then, you could write something like:
```
let (ss,typ) = handle_symbol ss e Const s [] t in
let rec add_cons : sig_state * sym list -> ident * p_term -> sig_state * sym list
= fun (ss, cons) (id, a) ->
  let (ss, c) = handle_symbol ss e Const id [] a in
  (ss, c::cons)
in
let (ss, cons) = List.fold_left add_cons (ss, []) tl in
let pr = Inductive.principle typ cons in
Sign.add_inductive ss.signature typ cons pr;
(ss, None)
```

src/core/handle.ml
```diff
@@ -286,7 +325,8 @@ let too_long = Stdlib.ref infinity
     exception handling. In particular, the position of [cmd] is used on errors
     that lack a specific position. All exceptions except [Timeout] and [Fatal]
     are captured, although they should not occur. *)
-let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
+let handle_cmd : sig_state -> p_command
+                 -> sig_state * proof_data option =
```
fblanqui, line 329: Thanks to revert this change.

src/core/pretty.ml
```diff
@@ -205,6 +205,13 @@ let pp_command : p_command pp = fun oc cmd ->
       List.iter (out " %a" pp_p_arg) args;
       Option.iter (out " :@ @[<hov>%a@]" pp_p_term) ao;
       out " ≔@ @[<hov>%a@]@]" pp_p_term t
+  | P_inductive(e,s,t,tl)   ->
+      out "@[<hov 2>%ainductive %a" pp_expo e pp_ident s;
+      Option.iter (out " :@ @[<hov>%a@] ≔@ " pp_p_term) t;
+      let pp_pair pp_a pp_b oc fmt (a,b) =
+        Format.fprintf oc fmt pp_a a pp_b b in
+      List.iter (pp_pair pp_ident pp_p_term oc
+                   ("\n |@ @[<hov>%a@] :@ @[<hov>%a@]")) tl
```
fblanqui, line 214: Could be simplified as follows:
```
let pp_cons oc (id,a) = Format.fprintf "@[<hov>%a@] :@ @[<hov>%a@]" pp_ident id pp_p_term a in
List.pp pp_cons "\n| "
```

src/core/sig_state.ml
```diff
@@ -43,7 +43,7 @@ type sig_state =
   ; in_scope  : (sym * Pos.popt) StrMap.t (** Symbols in scope.         *)
   ; aliases   : Path.t StrMap.t           (** Established aliases.      *)
   ; path_map  : string PathMap.t          (** Reverse map of [aliases]. *)
-  ; builtins  : sym StrMap.t               (** Builtin symbols.          *)
+  ; builtins  : sym StrMap.t              (** Builtin symbols.          *)
```
fblanqui, line 46: Thanks to revert.

src/core/sign.ml
```diff
@@ -64,12 +66,15 @@ let current_sign () =
   in
   PathMap.find mp !loaded
 
+let create_sym : string -> term -> prop -> expo -> sym = fun name typ p e ->
```
fblanqui, line 69: Thanks to remove.

fblanqui: On the other hand, it would be useful to add a function similar to add_symbol but for inductive:
```
add_inductive : t -> sym -> sym list -> sym -> unit
= fun sign typ ind_cons ind_prop ->
let ind = { ind_cons; ind_prop } in
sign.sign_ind := SymMap.add typ ind sign.sign_ind
```

amelieled: This creates a "Dependency cycle" due to t in the signature.

src/core/handle.ml
amelieled: Why is it wrong? The function `create_sym` is only used to create my record.
And it is more general than `new_sym`..
I can change the signature, but this change a lot of things in the code, no? (just for my function)

src/core/pretty.ml
amelieled: Where is used the variable "oc" ?

src/core/handle.ml
amelieled: It's like in Coq, no?

src/core/pretty.ml
fblanqui: The code I write here is not type check; you should be able to fix by yourself. oc is the first argument of fprintf...

src/core/handle.ml
fblanqui: It's wrong because handle_symbol already creates a symbol, and create_sym creates a new one as well which is different. That's why I say that it is sufficient that handle_symbol returns the symbol that it created. Thanks Amélie to read my comments carefully and try to take them into account.

src/core/handle.ml
fblanqui: LP is not Coq.

src/core/sign.ml
amelieled: Done

doc/syntax.bnf
amelieled: done

src/core/parser.ml
amelieled: done

src/core/pretty.ml
fblanqui:
```
let pp_cons oc (id,a) = Format.fprintf oc "@[<hov>%a@] :@ @[<hov>%a@]" pp_ident id pp_p_term a in
List.iter pp_cons "\n| " oc tl
```

doc/sections/commands.md
```diff
@@ -207,6 +207,20 @@ arguments on the left side of the `≔` symbol (similarly to a value declaration
 in OCaml). Some arguments can be declared as implicit by enclosing them in
 curly brackets.
 
+<!---------------------------------------------------------------------------->
+### `inductive`
+The `inductive` command is used to define an inductive type, as in Coq.
```
fblanqui, line 212: No need to speak of Coq here ;-)

Thanks to explain what is the effect of the command.

doc/sections/commands.md
```diff
@@ -207,6 +207,20 @@ arguments on the left side of the `≔` symbol (similarly to a value declaration
...
+
+```
+inductive nat : TYPE ≔
+ | z    : nat
+ | succ : nat → nat
+
+inductive list : TYPE ≔
```
fblanqui, line 219: One example is enough but give the code the command `inductive` is equivalent to:
```
constant symbol nat : TYPE
constant symbol z : nat
constant symbol succ : nat → nat
```

src/core/handle.ml
amelieled: done

src/core/handle.ml
amelieled: Done

doc/sections/commands.md
amelieled: done

doc/sections/commands.md
amelieled: Done

src/core/pretty.ml
amelieled: Done, thanks

fblanqui:
Hello @amelieled . For generating the induction principle, I suggest you to use a code for inductive.ml similar to (I didn't type-check it):
```
(** Builtin configuration for induction. *)
type config =
  { symb_Prop : sym (** : TYPE. Type of propositions. *)
  ; symb_prf  : sym (** : Prop → TYPE.
                        Interpretation of propositions as types. *) }

(** [get_config ss pos] build the configuration using [ss]. *)
let get_config : Sig_state.t -> Pos.popt -> config = fun ss pos ->
  let builtin = Builtin.get ss pos in
  { symb_Prop = builtin "Prop"
  ; symb_prf  = builtin "π" }

let principle : Sig_state.t -> popt -> sym -> sym list -> term =
  fun pos sind scons_list ->
  let c = get_config ss pos in
  let ind = _Symb sind in
  let prop = _Symb c.symb_Prop in
  let p = Bindlib.new_var mkfree "p" in
  let prf_of_p t = _Appl (_Symb c.symb_prf) (_Appl (_Vari p) t) in
  let app = List.fold_left _Appl in

  let case_of scons =
    let rec case xs a =
      match unfold a with
      | Symb(s) ->
         if s == sind then prf_of_p (app (_Symb scons) (List.rev xs))
         else fatal pos "%a is not a constructor of %a"
                pp_symbol scons pp_symbol sind
      | Prod(a,b) ->
         let (x,b) = Bindlib.unbind b in
         let b = case (x:xs) b in
         begin
           match unfold a with
           | Symb(s) ->
              let b = if s == sind then _Impl (prf_of_p x) b else b in
              _Prod a (Bindlib.bind_var x b)
           | _ -> fatal pos "The type of %a is not supported"
                    pp_symbol scons
         end
      | _ ->
         fatal pos "The type of %a is not supported"
           pp_symbol scons
    in
    case [] scons.sym_type
  in

  let t =
    let x = Bindlib.new_var mkfree "x" in
    _Prod ind (Bindlib.bind_var x (prf_of_p (_Vari x)))
  in
  let add_case t scons = _Impl (case_of scons) t in
  let t = List.fold_left add_case t scons_list in
  let t = _Prod (_Impl ind prop) (Bindlib.bind_var p t) in
  Bindlib.unbox t
```
This code does not handle polymorphic or dependent types but just monomorphic non-dependent first-order data types.

src/core/handle.ml
```diff
@@ -185,7 +194,35 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+        = fun (ss, cons) (id, a) ->
+        let (ss, c) = handle_symbol ss e Injec id [] a in
+        (ss, c::cons)
+      in
+      (* Add the inductive type in the signature *)
+      let (ss, cons) = List.fold_left add_cons (ss, []) tl in
+      (* Compute the inductive principle *)
+      let pr = Inductive.principle typ cons in
+      let path = (current_sign()).sign_path in
+      let pr =  { sym_name = ((typ.sym_name)^"_ind")
```
fblanqui, line 214: You can simply write `typ.sym_name ^ "_ind"`.

fblanqui: Try to not create symbol structures directly. Instead you can write:
```
let ind_sym = Sign.add_symbol ss.signature e cmd.pos pr [] in
```

amelieled: OK, but this code it's just a "glue code".
I'm deleting this record.

amelieled:
Thanks !
This is the previous code with type-checking:

src/core/handle.ml
```diff
@@ -185,7 +194,25 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
       (* Actually add the symbol to the signature. *)
       out 3 "(symb) %s ≔ %a\n" x.elt pp_term t;
       let d = if op then None else Some(t) in
-      let ss = Sig_state.add_symbol ss e Defin x a impl d in
+      let (ss, _) = Sig_state.add_symbol ss e Defin x a impl d in
+      (ss, None)
+  | P_inductive(e, s, t, tl)     ->
+      (* Add the head symbol in the signature *)
```
fblanqui, line 200: head symbol -> inductive type

src/core/handle.ml
```diff
@@ -185,7 +194,25 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+      let (ss, typ) = handle_symbol ss e Const s [] t in
+      (* To create the body of the inductive type *)
```
fblanqui, line 202: `(* Add the constructors in the signature. *)`

src/core/handle.ml
```diff
@@ -185,7 +194,25 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+      let add_cons :
+                sig_state * sym list -> ident * p_term -> sig_state * sym list
+        = fun (ss, cons) (id, a) ->
+        let (ss, c) = handle_symbol ss e Injec id [] a in
+        (ss, c::cons)
+      in
+      (* Add the inductive type in the signature *)
```
fblanqui, line 209: Thanks to remove this comment.

src/core/handle.ml
```diff
@@ -185,7 +194,25 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+      let (ss, cons) = List.fold_left add_cons (ss, []) tl in
+      (* Compute the inductive principle *)
```
fblanqui, line 211: inductive -> induction

src/core/handle.ml
```diff
@@ -223,7 +250,8 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
             (* Add a symbol corresponding to the proof, with a warning. *)
             out 3 "(symb) %s (admit)\n" x.elt;
             wrn cmd.pos "Proof admitted.";
-            Sig_state.add_symbol ss e Const x a impl None
+            let (ss, _) = Sig_state.add_symbol ss e Const x a impl None
+            in ss
```
fblanqui, line 254: Here and below, you can simply write `fst (Sig_state.add_symbol ss e Const x a impl None)`.

src/core/inductive.ml
```diff
@@ -0,0 +1,9 @@
+open Timed
+open Terms
+
+type inductive =
+  { ind_constructors : sym list   ; (** the list of constructors           *)
+    ind_prop         : sym option ; (** one inductive principle (Prop one) *) }
```
fblanqui, line 6: No option type here I think. If we cannot generate the induction principle then the `inductive` command is useless and should fail, don't you think?

src/core/inductive.ml
```diff
@@ -0,0 +1,9 @@
+open Timed
+open Terms
+
+type inductive =
```
fblanqui, line 4: This type should be define in sign.ml I think because you need to access some builtins in inductive.ml.
Thanks to rename ind_constructors into ind_cons (this is shorter and clear enough).

src/core/sign.ml
```diff
@@ -307,6 +312,15 @@ let add_binop : t -> string -> (sym * binop) -> unit = fun sign s sym ->
 let add_ident : t -> string -> unit = fun sign id ->
   sign.sign_idents := StrSet.add id !(sign.sign_idents)
 
+(** [add_inductive sign typ ind_cons ind_prop] add the inductive type which
+    consists of a type [typ], constructors [ind_cons] and an induction
+    principle [ind_prop] to [sign]. *)
+let add_inductive : t -> sym -> sym list -> sym -> unit =
+  fun sign typ ind_cons ind_prop ->
+  let ind = { Inductive.ind_constructors=ind_cons
```
fblanqui, line 320: You can simply write `let ind = { ind_cons; ind_prop } in`

src/core/syntax.ml
```diff
@@ -338,6 +340,12 @@ let eq_p_command : p_command eq = fun c1 c2 ->
   | (P_definition(e1,b1,s1,l1,a1,t1), P_definition(e2,b2,s2,l2,a2,t2)) ->
       e1 = e2 && b1 = b2 && eq_ident s1 s2 && List.equal eq_p_arg l1 l2
       && Option.equal eq_p_term a1 a2 && eq_p_term t1 t2
+  | (P_inductive(e1,s1,t1,tl1)   , P_inductive(e2,s2,t2,tl2)    ) ->
+      e1 = e2 && eq_ident s1 s2 && eq_p_term t1 t2
+      &&
+        let eq_pair eq_a eq_b (a1,b1) (a2, b2) =
+          eq_a a1 a2 && eq_b b1 b2 in
+        List.equal (eq_pair eq_ident eq_p_term) tl1 tl2
```
fblanqui, line 348: You can simplify to:
```
let eq_cons (s1,t1) (s2,t2) = eq_ident s1 s2 && eq_p_term t1 t2 in
List.equal eq_cons tl1 tl2
```

fblanqui:
As explained by @gabrielhdt , you should not write
```
let b = if s == sind
                       then _Impl (prf_of_p (Bindlib.box_var x)) b
                       else b in
```
but
```
let b =
  if s == sind then
    _Impl (prf_of_p (Bindlib.box_var x)) b
  else
    b
in
```
when it cannot be put on one line (which is not the case here).

src/core/inductive.ml
amelieled: Yes, I think too.

src/core/handle.ml
amelieled: At the end of the case `P_theorem`, the couple returned is really (ss, ...)?
It should be the result of the function `finalize`, no?
I think there is something I don't understand here!

src/core/inductive.ml
amelieled: Yes, I think too!

src/core/syntax.ml
```diff
@@ -338,6 +340,11 @@ let eq_p_command : p_command eq = fun c1 c2 ->
   | (P_definition(e1,b1,s1,l1,a1,t1), P_definition(e2,b2,s2,l2,a2,t2)) ->
       e1 = e2 && b1 = b2 && eq_ident s1 s2 && List.equal eq_p_arg l1 l2
       && Option.equal eq_p_term a1 a2 && eq_p_term t1 t2
+  | (P_inductive(e1,s1,t1,tl1)   , P_inductive(e2,s2,t2,tl2)    ) ->
+      e1 = e2 && eq_ident s1 s2 && eq_p_term t1 t2
+      &&
+        let eq_cons (s1,t1) (s2,t2) = eq_ident s1 s2 && eq_p_term t1 t2 in
+        List.equal eq_cons tl1 tl2
```
fblanqui, line 347:
```
let eq_id_p_term (s1,t1) (s2,t2) = eq_ident s1 s2 && eq_p_term t1 t2 in
e1 = e2 && List.equal eq_id_p_term ((s1,t1)::tl1) ((s2,t2)::tl2)
```

src/core/inductive.ml
```diff
@@ -0,0 +1,5 @@
+open Timed
+open Terms
+
+let principle : sym -> sym list -> term = fun head _ -> (* @WIP *)
+  !(head.sym_type)
```
fblanqui, line 5: Thanks to complete ;-)

src/core/handle.ml
```diff
@@ -233,10 +257,10 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
               end;
             (* Add a symbol corresponding to the proof. *)
             out 3 "(symb) %s (qed)\n" x.elt;
-            Sig_state.add_symbol ss e Const x a impl None
+            fst (Sig_state.add_symbol ss e Const x a impl None)
       in
       let data =
-        { pdata_stmt_pos = stmt.pos ; pdata_p_state = st ; pdata_tactics = ts
+        { pdata_stmt_pos = stmt.pos ; pdata_p_state  = st ; pdata_tactics = ts
```
fblanqui, line 263: useless change

src/core/inductive.ml
amelieled: done

src/core/handle.ml
```diff
@@ -185,7 +193,23 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+        let (ss, c) = handle_symbol ss e Injec id [] a in
+        (ss, c::cons)
+      in
+      let (ss, cons) = List.fold_left add_cons (ss, []) tl in
+      (* Compute the induction principle *)
+      let pr = Inductive.principle typ cons in
+      let ind_sym = Sign.add_symbol ss.signature e Defin
+                      ({ elt = "ind"; pos = cmd.pos }) pr [] in
+      Sign.add_inductive ss.signature typ cons ind_sym;
       (ss, None)
```
fblanqui, line 213: I propose some renamings and some small changes (constructors are Const and the induction principle name):
```
  | P_inductive(e, id, a, l)     ->
      (* Add the inductive type in the signature *)
      let (ss, sym_typ) = handle_symbol ss e Const id [] a in
      (* Add the constructors in the signature. *)
      let add_cons (ss, cons_list) (id, a) ->
        let (ss, sym_cons) = handle_symbol ss e Const id [] a in
        (ss, sym_cons::cons_list)
      in
      let (ss, cons_list) = List.fold_left add_cons (ss, []) l in
      (* Compute the induction principle *)
      let ind_typ = Inductive.principle sym_typ cons_list in
      let ind_name = in_pos cmd.pos (id.elt ^ "_ind") in
      let sym_ind = Sign.add_symbol ss.signature e Defin sl ind_name [] in
      Sign.add_inductive ss.signature sym_typ cons_list sym_ind;
      (ss, None)
```

amelieled: Yes, I agree but:
* remove the signature of the function `add_cons` is maybe less readable. But as you want.
* why constructors will be constant ? One property of constructors is to be injective, no?

fblanqui: Const is stronger than Injec: constant => injective.

amelieled: Ah OK perfect! :+1:

src/core/sign.ml
```diff
@@ -8,6 +8,11 @@ open Terms
 open Syntax
 open Pos
 
+(** Representation of an inductive type *)
+type inductive =
+  { ind_cons : sym list (** the list of constructors           *)
+  ; ind_prop : sym      (** one inductive principle (Prop one) *) }
```
fblanqui, line 14: `(** Induction principle on propositions. *)`

src/core/sign.ml
```diff
@@ -8,6 +8,11 @@ open Terms
 open Syntax
 open Pos
 
+(** Representation of an inductive type *)
+type inductive =
+  { ind_cons : sym list (** the list of constructors           *)
```
fblanqui, line 13: `(** List of constructors. *)`

src/core/sign.ml
```diff
@@ -64,12 +71,15 @@ let current_sign () =
   in
   PathMap.find mp !loaded
 
+let create_sym : string -> term -> prop -> expo -> sym = fun name typ p e ->
```
fblanqui, line 74: Thanks to use `create_sym : expo -> prop -> ...` (as in Sign.add_symbol).

src/core/sign.ml
```diff
@@ -64,12 +71,15 @@ let current_sign () =
...
+  let path = (current_sign()).sign_path in (*file_to_module (current_path ())?*)
+  { sym_name = name ; sym_type = ref typ ; sym_path = path
+    ; sym_def = ref None ; sym_impl = []; sym_rules = ref []
+    ; sym_prop = p ; sym_expo = e ; sym_tree = ref Tree_types.empty_dtree }
+
 (** [new_sym ()] creates a new (private definable) symbol. *)
 let new_sym : string -> term -> sym = fun name typ ->
```
fblanqui, line 81: Thanks to remove this function and replace it by its definition where it is used.

src/core/sign.ml
```diff
@@ -64,12 +71,15 @@ let current_sign () =
   in
   PathMap.find mp !loaded
 
+let create_sym : string -> term -> prop -> expo -> sym = fun name typ p e ->
+  let path = (current_sign()).sign_path in (*file_to_module (current_path ())?*)
```
fblanqui, line 75: Thanks to remove the comment.

tests/OK/inductive_type.lp
```diff
@@ -0,0 +1,16 @@
+inductive nat : TYPE ≔
+ | z    : nat
+ | succ : nat → nat
+
+assert z : nat
+assert succ z : nat
+assert succ (succ (succ z)) : nat
+
```
fblanqui, line 8: Thanks to add a test to check that the induction principle is generated with the correct type. Idem for list below.

src/core/inductive.ml
```diff
@@ -0,0 +1,63 @@
...
+type config =
+  { symb_Prop : sym (** : TYPE. Type of propositions. *)
+  ; symb_prf  : sym (** : Prop → TYPE.
+                        Interpretation of propositions as types. *) }
+
+(** [get_config ss pos] build the configuration using [ss]. *)
+let get_config : Sig_state.t -> Pos.popt -> config = fun ss pos ->
+  let builtin = Builtin.get ss pos in
+  { symb_Prop = builtin "Prop"
+  ; symb_prf  = builtin "π" }
```
fblanqui, line 18: Please use `"P"` for the moment (as for the rewrite and why3 tactics), we will change this later.

tests/OK/inductive_type.lp
amelieled: Les symboles "nat_ind" et "list_ind" ne sont pas connus.

fblanqui: They should be...

doc/commands.md
```diff
@@ -202,6 +202,24 @@ arguments on the left side of the `≔` symbol (similarly to a value declaration
...
+ | succ : nat → nat
+```
+is equivalent to:
+
+```
+constant symbol nat  : TYPE
+constant symbol z    : nat
+constant symbol succ : nat → nat
+```
+The main goal is to generate automatically an inductive principle (work in progress).
```
fblanqui, line 221: More details needed here (name of the induction principle).

src/core/handle.ml
fblanqui: This doesn't seem to be fixed yet.

src/core/handle.ml
```diff
@@ -185,7 +193,23 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+      let add_cons :
+            sig_state * sym list -> ident * p_term -> sig_state * sym list
+        = fun(ss, cons_list) (id, a) ->
+        let (ss, sym_cons) = handle_symbol ss e Const id [] a in
+        (ss, sym_cons::cons_list)
+      in
+      let (ss, cons_list) = List.fold_left add_cons (ss, []) l in
+      (* Compute the induction principle *)
+      let ind_typ = Inductive.principle ss cmd.pos sym_typ cons_list in
+      let ind_name = make cmd.pos (id.elt ^ "_ind") in
```
fblanqui, line 210: make -> Pos.make

src/core/inductive.ml
```diff
@@ -0,0 +1,63 @@
+(** Builtin configuration for induction. *)
```
fblanqui, line 1: This is a comment for the type `config` not for the whole file. Thanks to move this comment below and add one for the file.

src/core/inductive.ml
```diff
@@ -0,0 +1,63 @@
...
+                pp_symbol scons pp_symbol sind
+      | Prod(a,b) ->
+         let (x,b) = Bindlib.unbind b in
+         let b = case ((Bindlib.box_var x)::xs) b in
+         begin
+           match unfold a with
+           | Symb(s) ->
+               let b =
+                 if s == sind then _Impl (prf_of_p (Bindlib.box_var x)) b
+                 else b in
```
fblanqui, line 44: Thanks to put `in` on a distinct line.

amelieled:
Hi @gabrielhdt ! Could you check the files `lambdapi-smie.el` and `lambdapi-var.el`, please?
Thanks :)

gabrielhdt:
> Hi @gabrielhdt ! Could you check the files `lambdapi-smie.el` and `lambdapi-var.el`, please?
> Thanks :)

Hi Amélie! I've made a PR to propose a somewhat simpler `smie` grammar than your suggestion, if you don't mind.

The PR includes a merge with last commit of master regarding emacs (I hope it will only make things simpler, but I'm not sure...)

fblanqui:
@amelieled . I don't understand: why did you close this PR???

fblanqui:
Is it possible to revert this? Otherwise we are going to lose the comments.

gabrielhdt:
It's because of me, I suggested to create a new branch containing all her commits (instead of her master branch), and to make the PR based on this new branch. It will allow her to keep her master branch in sync with Deducteam master, and provide a cleaner PR for `contributing.md`.

fblanqui:
But she could have renamed her master branch into say induction and recreate a new master branch in sync with origin/master, couldn't she?

gabrielhdt:
In a sense yes, she could, but in the end it would have become error prone to work on a master branch that is not truly the master branch... but it's true that it would have allowed to not touch this PR. I did not think that the comments were important (given that they are all resolved and still accessible through closed PRs).

fblanqui:
They are not lost on github but it is not convenient!

amelieled:
Yes I tried to do what Gabriel told me and I think it's the right thing to do. Now this PR was always pointing to the master branch and it was while trying other commands that I caused this problem. 
I'm sorry.
I'm going to start my fork from scratch, then create an "induction" and "contributing_doc" branch to be able to recreate the PRs. I think it's not so bad that I lost the comments about induction because most of them were details, and I've grouped the rules together to code in CONTRIBUTING.md for the other PR. In addition, the important points were discussed by e-mail.

Issue #154, opened 2019-02-13
Dependency graph of the project


fblanqui:
It would be nice to provide a dependency graph of the source files to help newcomers to enter the code.

fblanqui:
After http://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html, a .dot graph can be generated using ocamldoc -dot:

> Generate a dependency graph for the toplevel modules, in a format suitable for displaying and processing by dot. The dot tool is available from http://www.research.att.com/sw/tools/graphviz/. The textual representation of the graph is written to the file ocamldoc.out, or to the file specified with the -o option. Use dot ocamldoc.out to display it.

fblanqui:
It would be very useful to me to have this right now. As expert on dot, is there any chance @gabrielhdt that you can do it this morning?

francoisthire:
Actually, dune can do it too.

gabrielhdt:
> Actually, dune can do it too.

Good news, we need to do it through dune.

gabrielhdt:
https://github.com/mjambon/dune-deps seems to be the way to go.

gabrielhdt:
Apparently, `dune-deps` does not provide dependencies between files. So we have three options
* @francoisthire has a solution with dune,
* someone finds the right setup and command line to make `ocamldoc -dot` work on the dune project (I tries `dune exec -- ocamldoc -dot src/core/*.ml` without success),
* or we implement something using the output of `ocamldep`.

fblanqui:
Any solution is fine. If @francoisthire has a solution already, that's great.

rlepigre:
I am just wondering: how is such a thing useful to you @fblanqui?

francoisthire:
Sorry, I had `dune-deps` in mind indeed.

ejgallego:
Dune already knows module deps so it would be easy to add a patch for it to generate this information, tho in general module dependencies in OCaml can be quite tricky due to virtual libs, etc... There is a dune tool that allows to query all that Dune knows using GraphQL I think.

On the other hand I would maybe take advantage of `odoc` and build an overview page for beginners `dune build @doc` , API doc + a bit of introduction to the modules could be very helpful.

amelieled:
It's not possible with `ocamldoc -dot` ?

gabrielhdt:
At least not trivially (all the setup is drowned in dune, so one has to explicit paths and dependencies and libraries for `ocamldoc`...)

[...]

PR #383, opened 2020-04-30
depgraph.sh generates the dependency graph of the project (#154)

Rehan-MALAK:
#154


gabrielhdt:
Just a side note, on developing shell scripts, I couldn't recommend enough [ShellCheck](https://www.shellcheck.net/) (can be installed and used by flycheck).

Rehan-MALAK:
Nice tool !
I corrected most of the errors/warnings (mostly due to double quoting missing around arrays and not checking if a directory doesn't exist)
But I took the liberty to add a .shellcheckrc disabling 
''SC2086 Double quote array expansions to avoid re-splitting elements'' because most of the time, if the string is splitted, the script wouldn't be working properly anyway.

Rehan-MALAK:
sorry, correcting with shellcheck suggestions actually broke something

Rehan-MALAK:
Ok, now this should be ok.

``git clean -dffx && rm -vf graph.png graph.dot && make install && cd tools/depgraph/ && ./depgraph.sh``

graph

tools/depgraph/.shellcheckrc
```diff
@@ -0,0 +1 @@
+disable=SC2086
```
gabrielhdt, line 1: I would avoid creating files using
```bash
#!/bin/bash
# shellcheck disable=SC2086
```
in the main script

tools/depgraph/depgraph.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+
+set -o nounset
```
gabrielhdt, line 3: It's also a good habit to use
```bash
set -eu -o pipefail
```
see <https://sipb.mit.edu/doc/safe-shell/>

tools/depgraph/depgraph.sh
```diff
@@ -0,0 +1,79 @@
...
+errordirectory="error : non-existing directory"
+workingdir=$(pwd)
+scriptdir=$(dirname "${BASH_SOURCE[0]}")
+lambdapidir=$(cd ${scriptdir}/../.. || { echo $errordirectory ; exit 1; } ; pwd)
+graph=graph
+graphdot=$workingdir/$graph.dot
+graphpng=$workingdir/$graph.png
+
+# CHOOSE THE INTERESTING_MODULES MODULES
+interesting_modules=( compile handle infer proof sr tactics typing unif )
```
gabrielhdt, line 24: Perhaps we could specify the modules on command line? The simplest way seems to use `"$@"` as `interesting_modules` (perhaps you'll have to use `modules=( "$@" )` to convert to a proper array).

fblanqui:
@Rehan-MALAK @gabrielhdt what is the status of this PR? Thanks to tell me when it will be ready to merge.

PR #386, opened 2020-05-04
[WIP] Unification rule on products

gabrielhdt:
Objectify products to be able to define unification rules on them, such as
```
set unif_rule T $a → T $b ≡ T $c ↪ $c ≡ arrow $a $b
```

### DIFF
--- a/src/core/handle.ml
+++ b/src/core/handle.ml
@@ -276,9 +276,10 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
         | P_config_unif_rule(h)   ->
             (* Approximately same processing as rules without SR checking. *)
             let pur = (scope_rule ss h).elt in
+            let rhs = Bindlib.(unbox (bind_mvar pur.pr_vars pur.pr_rhs)) in
+            let lhs = List.map Unif_rule.prod_obj pur.pr_lhs in
             let urule =
-              { lhs = pur.pr_lhs
-              ; rhs = Bindlib.unbox (Bindlib.bind_mvar pur.pr_vars pur.pr_rhs)
+              { lhs ; rhs
               ; arity = List.length pur.pr_lhs
               ; arities = pur.pr_arities
               ; vars = pur.pr_vars }
--- a/src/core/unif_rule.ml
+++ b/src/core/unif_rule.ml
@@ -45,7 +45,7 @@ let comma : sym =
   sym
 
 (** [unpack eqs] transforms a term of the form [t =? u, v =? w, ...]
-    into a list [[t =? u; v =? w; ...]]. *)
+    into a list [[...; v =? w; t =? u]]. *)
 let rec unpack : term -> (term * term) list = fun eqs ->
   match Basics.get_args eqs with
   | (Symb(s), [v; w]) ->
@@ -56,6 +56,9 @@ let rec unpack : term -> (term * term) list = fun eqs ->
       else if s == equiv then [(v, w)] else
       assert false (* Ill-formed term. *)
   | _                 -> assert false (* Ill-formed term. *)
+
+(* TODO: clean *)
+let unpack eqs = List.rev (unpack eqs)
 
 (** [p_unpack eqs] is [unpack eqs] on syntax-level equivalences [eqs]. *)
 let rec p_unpack : p_term -> (p_term * p_term) list = fun eqs ->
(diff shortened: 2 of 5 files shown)


src/core/unif_rule.ml
```diff
@@ -71,3 +71,37 @@ let rec p_unpack : p_term -> (p_term * p_term) list = fun eqs ->
       else if id s = "#equiv" then [(v, w)] else
       assert false (* Ill-formed term. *)
   | _                               -> assert false (* Ill-formed term. *)
+
+let prod : sym =
+  Sign.add_symbol sign Public Defin (Pos.none "#prod") Kind []
+
+let rec prod_obj : term -> term = fun t ->
```
fblanqui, line 78: Comments missing for explaining the transformation.

src/core/unif_rule.ml
```diff
@@ -71,3 +71,37 @@ let rec p_unpack : p_term -> (p_term * p_term) list = fun eqs ->
...
+      let b = Bindlib.unbox (Bindlib.bind_var x b) in
+      Prod(prod_unobj a, b)
+  | Appl(t, u)                              ->
+      Appl(prod_unobj t, prod_unobj u)
+  | Abst(a, b)                              ->
+      let (x, b) = Bindlib.unbind b in
+      let b = lift (prod_unobj b) in
+      let b = Bindlib.unbox (Bindlib.bind_var x b) in
+      Abst(prod_unobj a, b)
+  | _                                       -> t
```
fblanqui, line 107: The LLet case is missing (may occur in a RHS).

src/core/unif_rule.ml
```diff
@@ -71,3 +71,37 @@ let rec p_unpack : p_term -> (p_term * p_term) list = fun eqs ->
...
+
+let rec prod_obj : term -> term = fun t ->
+  match unfold t with
+  | Appl(t,u) -> Appl(prod_obj t, prod_obj u)
+  | Prod(a,b) ->
+      let (x, b) = Bindlib.unbind b in
+      let b = lift (prod_obj b) in
+      let b = Bindlib.unbox (Bindlib.bind_var x b) in
+      Appl(Symb(prod), Abst(prod_obj a,b))
+  | Abst(a,b)         ->
```
fblanqui, line 86: Since we need to transform top produtcs only (right?), the cases Appl and Abst are useless.

tests/OK/unif_hint.lp
```diff
@@ -44,7 +44,10 @@ set unif_rule (plus $x $y) ≡ z ↪ $x ≡ z, $y ≡ z
 compute H (plus _ _) iz
 
 // The arrow problem
+set flag "print_implicits" on
+set flag "print_domains" on
 symbol arrow : U → U → U
 rule T (arrow $t $u) ↪ T $t → T $u
-// hint (T $a → T $b) ≡ T (arrow $c $d) ↪ $a ≡ $c, $b ≡ $d
-// FIXME: implications not allowed in LHS (scoping level)
+set unif_rule (T $a → T $b) ≡ T $c ↪ $c ≡ arrow $a $b
```
fblanqui, line 51: Perhaps write `arrow $a $b ≡ $c` to keep the order.

fblanqui:
@gabrielhdt Would it be difficult to modify the trees and the matching algorithm to do matching on products as we already do matching on abstractions? Then, the functions obj/unobj could be implemented by simply adding/removing \#prod without doing any lifting/unboxing: Prod(a,b) would simply be transformed into Appl(Symb(\#prod),Abst(a,b)), or even Appl(Symb(\#prod),Prod(a,b)). Perhaps we could even get rid of obj/unobj by applying he tree of \#prod in the case Prod.

src/core/unif_rule.ml
fblanqui: Since we rewrite step by step, do we really need prod_unobj? Only the top product(s?) of the LHS and of the term to rewrite needs to be objectified.

src/core/handle.ml
```diff
@@ -276,9 +276,15 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
         | P_config_unif_rule(h)   ->
             (* Approximately same processing as rules without SR checking. *)
             let pur = (scope_rule ss h).elt in
+            let rhs =
+              Bindlib.unbox (Bindlib.bind_mvar pur.pr_vars pur.pr_rhs)
+            in
+            let (xs, rhs) = Bindlib.unmbind rhs in
```
fblanqui, line 282: bind followed by unbind could be avoided by just unboxing pur.pr_rhs (binders can be unboxed too).

src/core/handle.ml
```diff
@@ -276,9 +276,15 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+            let rhs = Unif_rule.prod_obj rhs in
```
fblanqui, line 283: Is it necessary to apply prod_obj to the RHS?

gabrielhdt: > @gabrielhdt Would it be difficult to modify the trees and the matching algorithm to do matching on products as we already do matching on abstractions? Then, the functions obj/unobj could be implemented by simply adding/removing #prod without doing any lifting/unboxing: Prod(a,b) would simply be transformed into Appl(Symb(#prod),Abst(a,b)), or even Appl(Symb(#prod),Prod(a,b)). Perhaps we could even get rid of obj/unobj by applying he tree of #prod in the case Prod.

If I understand well, you want to be able to consider `Prod` as a tree symbol to match on it and use a `#prod` symbol to hold a tree. But if we allow matching on products, a `#prod` symbol should be useless since the tree is hold by `#equiv`.

So I think matching on products can be implemented, and it would even allow to not use any `#prod` symbol.

gabrielhdt:
> @gabrielhdt Would it be difficult to modify the trees and the matching algorithm to do matching on products as we already do matching on abstractions? Then, the functions obj/unobj could be implemented by simply adding/removing #prod without doing any lifting/unboxing: Prod(a,b) would simply be transformed into Appl(Symb(#prod),Abst(a,b)), or even Appl(Symb(#prod),Prod(a,b)). Perhaps we could even get rid of obj/unobj by applying he tree of #prod in the case Prod.

If I understand well, you want to use a `#prod` symbol to hold a tree and be able to match on `Prod` (just like `Abst`). But I think that if we match on products, the `#prod` symbol becomes useless since the tree is held by the `#equiv` symbol. So yes, it is possible to modify the trees to match on products and it would allow to get rid of `prod_obj/unobj` and `#prod`.

src/core/handle.ml
gabrielhdt: No indeed

PR #390, opened 2020-05-07
Allow extra variables in RHS

gabrielhdt:
This PR is in preparation of unification on products. It solves two issues with the previous implementation of unification rules:
- the order of sub problems (`x = y, t = u, ...`) was poorly managed,
- we allow the creation of metavariables in the right hand side of unification rules.

The creation of metavariables is needed for instance for the arrow rule,
```
set unif_rule $a → $b ≡ T $c ↪ $a ≡ T ?aa, $b ≡ T ?bb, $c ≡ arrow ?aa ?bb
```

### DIFF
--- a/src/core/eval.ml
+++ b/src/core/eval.ml
@@ -192,9 +192,12 @@ and tree_walk : dtree -> ctxt -> stack -> (term * stack) option =
     let open Tree_types in
     match tree with
     | Fail                                                -> None
-    | Leaf(env_builder, act)                              ->
-        (* Allocate an environment for the action. *)
-        let env = Array.make (Bindlib.mbinder_arity act) TE_None in
+    | Leaf(env_builder, (act, xvars))                     ->
+        (* Allocate an environment where to place terms coming from the
+           pattern variables for the action. *)
+        let env_len = Bindlib.mbinder_arity act in
+        (* We have [length env_builder = env_len - xvars] *)
+        let env = Array.make env_len TE_None in
         (* Retrieve terms needed in the action from the [vars] array. *)
         let fn (pos, (slot, xs)) =
           match bound.(pos) with
@@ -211,8 +214,13 @@ and tree_walk : dtree -> ctxt -> stack -> (term * stack) option =
                 env.(slot) <- TE_Some(Bindlib.unbox (Bindlib.bind_mvar xs b))
         in
         List.iter fn env_builder;
-        (* Actually perform the action. *)
-        Some(Bindlib.msubst act env, stk)
+        (* Complete the array with fresh meta-variables if needed. *)
+        for i = env_len - xvars to env_len - 1 do
+          let t = make_meta [] Kind in
+          let b = Bindlib.raw_mbinder [||] [||] 0 mkfree (fun _ -> t) in
+          env.(i) <- TE_Some(b)
+        done;
+        Some (Bindlib.msubst act env, stk)
     | Cond({ok; cond; fail})                              ->
         let next =
           match cond with
--- a/src/core/print.ml
+++ b/src/core/print.ml
@@ -66,8 +66,8 @@ let pp_symbol : sym pp = fun oc s ->
     Format.pp_print_string oc s.sym_name
   else pp_qualified oc s
 
-(** [pp_tvar oc x] prints the term variable [x] to the channel [oc]. *)
-let pp_tvar : tvar pp = fun oc x ->
+(** [pp_var oc x] prints the Bindlib variable [x] to the channel [oc]. *)
+let pp_var : 'a Bindlib.var pp = fun oc x ->
   Format.pp_print_string oc (Bindlib.name_of x)
 
 (** Exception raised when trying to convert a term into a nat. *)
@@ -166,7 +166,7 @@ and pp_term : term pp = fun oc t ->
           match unfold b with
           | Abst(a,b) ->
               let (x,p) = Bindlib.unbind b in
-              out oc "%a%a" pp_symbol s pp_tvar x;
+              out oc "%a%a" pp_symbol s pp_var x;
               if !print_implicits then out oc ": %a" (pp `Func) a;
               out oc ", %a" (pp `Func) p
           | _ -> assert false
@@ -189,7 +189,7 @@ and pp_term : term pp = fun oc t ->
     | Wild        -> assert false
     | TRef(_)     -> assert false
     (* Atoms are printed inconditonally. *)
-    | Vari(x)     -> pp_tvar oc x
+    | Vari(x)     -> pp_var oc x
     | Type        -> out oc "TYPE"
     | Kind        -> out oc "KIND"
     | Symb(s)     -> pp_symbol oc s
@@ -200,7 +200,7 @@ and pp_term : term pp = fun oc t ->
     | Abst(a,b)   ->
         if wrap then out oc "(";
         let (x,t) = Bindlib.unbind b in
-        out oc "λ%a" pp_var (b,x);
+        out oc "λ%a" pp_bvar (b,x);
         if !print_domains then out oc ": %a, %a" (pp `Func) a (pp `Func) t
         else pp_abstractions oc t;
         if wrap then out oc ")"
@@ -208,23 +208,23 @@ and pp_term : term pp = fun oc t ->
         if wrap then out oc "(";
         let (x,t) = Bindlib.unbind b in
         if Bindlib.binder_occur b then
-          out oc "Π%a: %a, %a" pp_tvar x (pp `Func) a (pp `Func) t
+          out oc "Π%a: %a, %a" pp_var x (pp `Func) a (pp `Func) t
         else out oc "%a → %a" (pp `Appl) a (pp `Func) t;
         if wrap then out oc ")"
     | LLet(a,t,b) ->
         if wrap then out oc "(";
         let (x,u) = Bindlib.unbind b in
-        pp_var oc (b,x);
+        pp_bvar oc (b,x);
         if !print_domains then out oc ": %a" (pp `Atom) a;
         out oc " ≔ %a in %a" (pp `Atom) t (pp `Atom) u;
         if wrap then out oc ")"
-  and pp_var oc (b,x) =
-    if Bindlib.binder_occur b then out oc "%a" pp_tvar x else out oc "_"
+  and pp_bvar oc (b,x) =
+    if Bindlib.binder_occur b then out oc "%a" pp_var x else out oc "_"
   and pp_abstractions oc t =
     match unfold t with
     | Abst(_,b) ->
         let (x,t) = Bindlib.unbind b in
-        out oc " %a" pp_var (b,x); pp_abstractions oc t
+        out oc " %a" pp_bvar (b,x); pp_abstractions oc t
     | t -> out oc ", %a" (pp `Func) t
   in
   pp `Func oc (cleanup t)
@@ -243,9 +243,9 @@ let pp_ctxt : ctxt pp = fun oc ctx ->
     let pp_ctxt : ctxt pp = fun oc ctx ->
       let pp_e oc (x,a,t) =
         match t with
-        | None    -> Format.fprintf oc "%a: %a" pp_tvar x pp_term a
+        | None    -> Format.fprintf oc "%a: %a" pp_var x pp_term a
         | Some(t) ->
-            Format.fprintf oc "%a: %a ≔ %a" pp_tvar x pp_term a pp_term t
+            Format.fprintf oc "%a: %a ≔ %a" pp_var x pp_term a pp_term t
       in
       if ctx = [] then Format.pp_print_string oc "∅"
       else List.pp pp_e ", " oc (List.rev ctx)
--- a/src/core/scope.ml
+++ b/src/core/scope.ml
@@ -73,11 +73,24 @@ type mode =
   (** Scoping mode for rewriting rule left-hand sides. The constructor carries
       a flag that is set to [true] if {!constructor:Terms.expo.Privat} symbols
       are allowed, and also additional data. *)
-  | M_RHS  of bool * (string, tevar) Hashtbl.t
-  (** Scoping mode for rewriting rule righ-hand sides. The constructor carries
-      a flag that is set to [true] if {!constructor:Terms.expo.Privat} symbols
-      are allowed, and the environment for variables that we known to be bound
-      in the RHS. *)
+  | M_RHS  of
+      { m_rhs_prv             : bool
+      (** True if {!constructor:Terms.expo.Privat} symbols are allowed. *)
+      ; m_rhs_data            : (string, tevar) Hashtbl.t
+      (** Environment for variables that we know to be bound in the RHS. *)
+      ; mutable m_rhs_vars_nb : int
+      (** Number  of  distinct  variables  in  the rewriting  rule,  including
+          variables  only in  the  RHS. It  is initialised  to  the number  of
+          (distinct) variables in the LHS and incremented each time a variable
+          of the RHS that was not in the LHS is scoped. *)
+      ; mutable m_rhs_xvars   : (string * tevar) list
+      (** Variables scoped that  were not in the LHS. This  field is only used
+          in  unification  rules and  is  updated  imperatively for  each  new
+          variable scoped. A couple [(n, v)]  is the name of the variable with
+          the variable itself. The name is needed to ensure that two variables
+          with the same name are scoped as the same variable. *) }
+  (** Scoping  mode for  rewriting rule right-hand  sides. During  scoping, we
+      always have [m_rhs_vars_nb = m_lhs_size + length m_rhs_xvars]. *)
 
 (** [get_implicitness t] gives the specified implicitness of the parameters of
     a symbol having the (parser-level) type [t]. *)
@@ -191,8 +204,8 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
         let m = _Meta_full (fresh_meta_box a (Array.length vs)) vs in
         (* Sanity check: only variables of [env] free in [m] if not in RHS. *)
         match md with
-        | M_RHS(_,_) -> m
-        | _          ->
+        | M_RHS(_) -> m
+        | _        ->
         assert (Bindlib.is_closed (Bindlib.bind_mvar (Env.vars env) m)); m
   (* Scoping of a binder (abstraction or product). The environment made of the
      variables is also returned. *)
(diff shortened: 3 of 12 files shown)


fblanqui:
@gabrielhdt you also need to define what is the effect of applying a rule with metavariables in the RHS. Shouldn't these metavariables be replaced by fresh ones each time a rule is applied?

fblanqui:
Instead why not writing `set unif_rule fst $p ≡ $f ↪ $p ≡ pair $f $s` where $s is an extra pattern variable? Then, each time a rule is applied we can easily replace $s by a fresh metavariable.

gabrielhdt:
> Instead why not writing `set unif_rule fst $p ≡ $f ↪ $p ≡ pair $f $s` where $s is an extra pattern variable? Then, each time a rule is applied we can easily replace $s by a fresh metavariable.

I agree that in a way it would be more coherent. But in rewriting we want all the variables of the RHS to be in the LHS, so it is not absurd to distinguish them. Implementation wise, isn't it substantially simpler with metavariables?

gabrielhdt:
And please restrain from merging at will the master branch, it botches my commit history (I wanted to keep it clean, for once...)

gabrielhdt:
> @gabrielhdt you also need to define what is the effect of applying a rule with metavariables in the RHS. Shouldn't these metavariables be replaced by fresh ones each time a rule is applied?

So the RHS of unification rules should be binders, right? And each time a unification rule is applied, new meta variables are created and substituted.

gabrielhdt:
So the plan seems to be:
- we can allow pattern variables that are not in the RHS in unification rules,
- such pattern variables are replaced at scoping by free variables,
- the variables are collected so the RHS of unification rules is a binder on those new variables,
- in the unification algorithm, when a rule matches, we create new metavar and substitute them in the RHS.

What do you think @fblanqui ?

fblanqui:
In your first item, do you mean RHS or LHS?

gabrielhdt:
> In your first item, do you mean RHS or LHS?

Ah yes indeed.

fblanqui:
So, you agree with using extra pattern variables instead of meta variables?

gabrielhdt:
Yes, I tend to agree because we're not exactly doing rewriting, we're doing unification indications (via rewriting)

fblanqui:
Note however there I believe exists a notion of conditional rewriting with extra variables in RHS and conditions that are instantiated when checking conditions, probably using narrowing.

fblanqui:
I agree with your plan. It seems to me that we do not need to change many things. When binding pattern variables in the RHS, we should put extra pattern variables at the end and record their number in the type rule. And checking whether RHS pattern variables occur in the LHS should be optional and postponed.

gabrielhdt:
Don't we have to create a 2 layers binders to be able to substitute separately?

fblanqui:
Why do you want to substitute separately?

gabrielhdt:
Ah you suggest to modify a bit the rewriting engine to automatically create new metavariables?

gabrielhdt:
To sum up what have changed, in brief I introduced the notion of _extra variables_ which are the variables of the RHS which are not in the LHS. These variables are stored at the end of the `vars` array of the rules.
- The type `rule` has a `xvars` integer which is the number of variables.
- Same for the `pre_rule` type.
- This `xvars` is propagated to the leaves of the trees.
- The rewriting engine creates the remaining meta variables (`xvars` meta-variables).

src/core/eval.ml
```diff
@@ -192,9 +192,11 @@ and tree_walk : dtree -> ctxt -> stack -> (term * stack) option =
     let open Tree_types in
     match tree with
     | Fail                                                -> None
-    | Leaf(env_builder, act)                              ->
-        (* Allocate an environment for the action. *)
-        let env = Array.make (Bindlib.mbinder_arity act) TE_None in
+    | Leaf(env_builder, act, xvars)                       ->
```
fblanqui, line 195: @gabrielhdt Could you please remind me and update tree_types.ml of what is the meaning of the elements of the first argument of Leaf?

src/core/eval.ml
```diff
@@ -212,7 +214,18 @@ and tree_walk : dtree -> ctxt -> stack -> (term * stack) option =
         in
         List.iter fn env_builder;
         (* Actually perform the action. *)
-        Some(Bindlib.msubst act env, stk)
+        begin
+          try Some(Bindlib.msubst act env, stk)
+          with Invalid_argument(_) ->
```
fblanqui, line 219: It looks kind of hackish to rely on an Invalid_argument exception... Isn't there another way to do this? Anyway, a comment is necessary.

src/core/print.ml
```diff
@@ -70,6 +70,11 @@ let pp_symbol : sym pp = fun oc s ->
 let pp_tvar : tvar pp = fun oc x ->
   Format.pp_print_string oc (Bindlib.name_of x)
 
+(** [pp_tevar oc x] prints the term with environment variable [x] to the
+    channel [oc]. *)
+let pp_tevar : tevar pp = fun oc x ->
```
fblanqui, line 75: The functions pp_tvar and pp_tevar are in fact the same. They should be replaced by a single polymorphic function ('a Bindlib.var pp) named pp_var.

src/core/scope.ml
```diff
@@ -73,11 +73,16 @@ type mode =
   (** Scoping mode for rewriting rule left-hand sides. The constructor carries
       a flag that is set to [true] if {!constructor:Terms.expo.Privat} symbols
       are allowed, and also additional data. *)
-  | M_RHS  of bool * (string, tevar) Hashtbl.t
-  (** Scoping mode for rewriting rule righ-hand sides. The constructor carries
-      a flag that is set to [true] if {!constructor:Terms.expo.Privat} symbols
-      are allowed, and the environment for variables that we known to be bound
-      in the RHS. *)
+  | M_RHS  of
+      { m_r_p             : bool
```
fblanqui, line 77: Following what is done for M_LHS, introduce a type m_rhs_data with fields prefixed by m_rhs. By the way, you could include the boolean argument of M_LHS in m_lhs_data as well.

src/core/scope.ml
```diff
@@ -73,11 +73,16 @@ type mode =
...
+      (** True if {!constructor:Terms.expo.Privat} symbols are allowed. *)
+      ; m_r_data          : (string, tevar) Hashtbl.t
+      (** Environment for variables that we know to be bound in the RHS. *)
+      ; mutable m_r_xvars : (int * tevar list)
+      (** Slot of the next extra variable in the [pr_vars] array and collected
```
fblanqui, line 82: What is pr_vars?

src/core/scope.ml
```diff
@@ -73,11 +73,16 @@ type mode =
...
-  (** Scoping mode for rewriting rule righ-hand sides. The constructor carries
-      a flag that is set to [true] if {!constructor:Terms.expo.Privat} symbols
-      are allowed, and the environment for variables that we known to be bound
-      in the RHS. *)
+  | M_RHS  of
+      { m_r_p             : bool
+      (** True if {!constructor:Terms.expo.Privat} symbols are allowed. *)
+      ; m_r_data          : (string, tevar) Hashtbl.t
+      (** Environment for variables that we know to be bound in the RHS. *)
+      ; mutable m_r_xvars : (int * tevar list)
```
fblanqui, line 81: Why not splitting in two fields? This would avoid to have to do match's and fst/snd's.

src/core/scope.ml
```diff
@@ -433,13 +442,16 @@ type pre_rule =
...
-  (** Pattern variables that can appear in the RHS. *)
+  (** Pattern variables that can appear in the RHS. The [pr_xvars] last
+      variables do not appear in the LHS. *)
   ; pr_rhs     : tbox
   (** Body of the RHS, should only be unboxed once. *)
   ; pr_names   : (int, string) Hashtbl.t
   (** Gives the original name (if any) of pattern variable at given index. *)
   ; pr_arities : int array
-  (** Gives the arity of all the pattern varialbes in field [pr_vars]. *) }
+  (** Gives the arity of all the pattern varialbes in field [pr_vars]. *)
```
fblanqui, line 452: Typo: varialbes -> variables

src/core/scope.ml
```diff
@@ -433,13 +442,16 @@ type pre_rule =
...
+  ; pr_xvars: int
+  (** Number of extra variable in [pr_vars]. *) }
```
fblanqui, line 454: variable+s. Comment not clear: number of variables of the RHS not occurring in the LHS.

src/core/sr.ml
```diff
@@ -136,19 +136,27 @@ let symb_to_tenv
    [Fatal] is raised in case of error. *)
 let check_rule : Scope.pre_rule Pos.loc -> rule = fun pr ->
   (* Unwrap the contents of the pre-rule. *)
-  let (pos, s, lhs, vars, rhs_vars, arities) =
-    let Pos.{elt=Scope.{pr_sym;pr_lhs;pr_vars;pr_rhs;pr_arities;_}; pos} = pr
+  let (pos, s, lhs, vars, rhs_vars, arities, xvars) =
+    let Pos.{elt=Scope.{pr_sym;pr_lhs;pr_vars;pr_rhs;pr_arities;
```
fblanqui, line 140: Perhaps we should use a let here to make things more readable.

src/core/tree_types.ml
```diff
@@ -61,12 +61,13 @@ type tree_cond =
 type 'rhs tree =
   | Fail
   (** Empty decision tree, used when there are no rewriting rules. *)
-  | Leaf of (int * (int * int array)) list * 'rhs
-  (** The value [Leaf(m, rhs)] stores the RHS [rhs] of the rewriting rule that
+  | Leaf of (int * (int * int array)) list * 'rhs * int
+  (** Triple [Leaf(m, rhs, x)] stores the RHS [rhs] of the rewriting rule that
```
fblanqui, line 65: As mentioned previously, more details on the meaning of the elements of [m] would be very welcome here.

src/core/eval.ml
gabrielhdt: Yes, we can as well check whether the `xvars` is equal to 0. Performance wise, I thought that relying on the exception would avoid a comparison in the case of regular rewriting rules. I'll check if there's an actual difference.

src/core/scope.ml
gabrielhdt: I wanted to pack the mutable elements, as you wish.

src/core/sr.ml
gabrielhdt: Agreed.

src/core/scope.ml
gabrielhdt: What's the advantage of doing so? It's a guess but it might be more efficient to avoid creating a new type (and we don't need that type, do we?).

gabrielhdt: I'd even be in favour of inlining the record of `M_LHS`.

src/core/tree_types.ml
gabrielhdt: Tell me if the documentation isn't clear enough.

src/core/scope.ml
gabrielhdt: I'm using two fields now.

src/core/tree_types.ml
```diff
@@ -61,12 +61,17 @@ type tree_cond =
...
       to construct the environment of the RHS. Note that we do not need to use
-      a map here since we only need to insert at the head, and iterate over
-      the elements of the structure. *)
+      a map here  since we only need  to insert at the head,  and iterate over
+      the elements of the structure. Triplet  [(p, (v, xs))] of [m] means that
+      when a rule matches,  the term to be used as the [v]th variable of the
+      RHS is found in position [p] in the array containing all the terms
+      gathered during matching. The pattern may have an environment made of
+      variables [xs]. The integer [x] indicates the number of meta variables
+      to generate to populate the RHS. *)
```
fblanqui, line 74: to populate the RHS -> for the extra variables of the RHS

src/core/scope.ml
gabrielhdt: I don't know what to think of 4be4073, what's your opinion @fblanqui?

src/core/scope.ml
```diff
@@ -69,15 +56,33 @@ type mode =
...
+      (** Stores the arity of the pattern variable at the given index. *)
+      ; m_lhs_names   : (int   , string) Hashtbl.t
+      (** Stores the name of the pattern variable at given index (if any). *)
+      ; mutable m_lhs_size : int
+      (** Stores the current known size of the environment of the RHS. *)
+      ; m_lhs_in_env  : string list
+      (** Pattern variables definitely needed in the RHS environment. *) }
+  (** Scoping mode for rewriting rule left-hand sides. *)
+  | M_RHS  of
+      { m_rhs_p             : bool
```
fblanqui, line 74: Please use a more explicit name like m_rhs_prv. Idem for M_LHS.

src/core/scope.ml
```diff
@@ -69,15 +56,33 @@ type mode =
...
+      (** True if {!constructor:Terms.expo.Privat} symbols are allowed. *)
+      ; m_rhs_data          : (string, tevar) Hashtbl.t
+      (** Environment for variables that we know to be bound in the RHS. *)
+      ; mutable m_rhs_xvi   : int
+      (** Number of the next variable not in the LHS scoped. Incremented each
```
fblanqui, line 79: Why not simply "Number of variables not in the LHS. Incr..." ?

src/core/scope.ml
```diff
@@ -69,15 +56,33 @@ type mode =
...
+          time such a variable is scoped. *)
+      ; mutable m_rhs_xvars : tevar list
```
fblanqui, line 81: Do we have the invariant List.length m_rhs_xvars = m_rhs_xvi? If so, it would be good to mention it and rename m_rhs_xvi into m_rhs_xvars_nb.

src/core/scope.ml
```diff
@@ -296,14 +309,19 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
...
+    | (P_Patt(id,ts)   , M_RHS(r)         ) ->
         let x =
           match id with
           | None     -> fatal t.pos "Wildcard pattern not allowed in a RHS."
           | Some(id) ->
-          try Hashtbl.find h id.elt with Not_found ->
-            fatal t.pos "Pattern variable not in scope."
+              try Hashtbl.find r.m_rhs_data id.elt
+              with Not_found ->
+                let name = Printf.sprintf "%i_%s" r.m_rhs_xvi id.elt in
```
fblanqui, line 320: prefix variable names with "v"

src/core/scope.ml
```diff
@@ -433,13 +451,16 @@ type pre_rule =
   ; pr_lhs     : term list
   (** Arguments of the LHS. *)
   ; pr_vars    : term_env Bindlib.mvar
-  (** Pattern variables that can appear in the RHS. *)
+  (** Pattern variables that can appear in the RHS. The [pr_xvars] last
```
fblanqui, line 454: -> last [pr_xvars]

src/core/scope.ml
```diff
@@ -433,13 +451,16 @@ type pre_rule =
...
+      variables do not appear in the LHS. *)
   ; pr_rhs     : tbox
   (** Body of the RHS, should only be unboxed once. *)
   ; pr_names   : (int, string) Hashtbl.t
   (** Gives the original name (if any) of pattern variable at given index. *)
   ; pr_arities : int array
-  (** Gives the arity of all the pattern varialbes in field [pr_vars]. *) }
+  (** Gives the arity of all the pattern variables in field [pr_vars]. *)
+  ; pr_xvars: int
```
fblanqui, line 462: rename into pr_xvars_nb ?

src/core/scope.ml
fblanqui: Why not simply writing "Pattern variables occurring in the LHS or in the RHS" ?

src/core/scope.ml
```diff
@@ -490,34 +515,47 @@ let scope_rule : sig_state -> p_rule -> pre_rule loc = fun ss r ->
...
         with Not_found -> Printf.sprintf "%i" i
       in
       Bindlib.new_var te_mkfree name
     in
-    Array.init data.m_lhs_size fn
+    Array.init lhs_size fn
   in
-  (* We scope the RHS. *)
-  let pr_rhs =
+  (* We scope the RHS and retrieve the RHS only variables. *)
```
fblanqui, line 525: and retrieve the variables not occurring in the LHS

src/core/terms.ml
```diff
@@ -130,7 +130,9 @@ type term =
   ; arities : int array
   (** Arities of the pattern variables bound in the RHS. *)
   ; vars    : term_env Bindlib.var array
-  (** Bindlib variables used to build [rhs]. *) }
+  (** Bindlib variables used to build [rhs]. *)
+  ; xvars   : int
```
fblanqui, line 134: I would rename it into xvars_nb.

src/core/scope.ml
gabrielhdt: We rather have `m_rhs_xvi = length m_rhs_xvars + length m_lhs_vars`

src/core/tree.ml
```diff
@@ -242,20 +242,22 @@ module CM = struct
...
-    { c_lhs : term array
+    { c_lhs       : term array
     (** Left hand side of a rule. *)
-    ; c_rhs : rhs
+    ; c_rhs       : rhs
     (** Right hand side of a rule. *)
     ; env_builder : env_builder
     (** Data required to apply the rule. *)
-    ; cond_pool : CP.t
+    ; xvars       : int
```
fblanqui, line 251: I would rename it into xvars_nb.

src/core/eval.ml
```diff
@@ -192,9 +192,11 @@ and tree_walk : dtree -> ctxt -> stack -> (term * stack) option =
...
     match tree with
     | Fail                                                -> None
-    | Leaf(env_builder, act)                              ->
-        (* Allocate an environment for the action. *)
-        let env = Array.make (Bindlib.mbinder_arity act) TE_None in
+    | Leaf(env_builder, act, xvars)                       ->
+        (* Allocate an environment where to place terms coming from the
+           pattern variables for the action. *)
+        (* We have [length env_builder = mbinder_arity act - xvars] *)
+        let env = Array.make (Bindlib.mbinder_arity act - xvars) TE_None in
```
fblanqui, line 199: Why not taking env of length mbinder_arity and filling the last xvars_nb variables by fresh metavariables (if xvars_nb <> 0 only for efficiency)? Then, msubst will not fail.

src/core/scope.ml
fblanqui: I see. It would be good to mention this in a comment. Then, m_rhs_xvi is not a number but an index and would be more appropriately named m_rhs_xvars_ind. I know that this is rather long but this will help understanding and maintenance by other people later.

gabrielhdt: Or it is the total number of variables and could be named `m_rhs_vars_nb`, which might be even a bit clearer (since we don't invoke some index in some array)

gabrielhdt: See my proposition, and tell me if that seems clear

src/core/tree_types.ml
```diff
@@ -61,12 +61,17 @@ type tree_cond =
 type 'rhs tree =
   | Fail
   (** Empty decision tree, used when there are no rewriting rules. *)
-  | Leaf of (int * (int * int array)) list * 'rhs
-  (** The value [Leaf(m, rhs)] stores the RHS [rhs] of the rewriting rule that
+  | Leaf of (int * (int * int array)) list * 'rhs * int
+  (** Triple [Leaf(m, rhs, x)] stores the RHS [rhs] of the rewriting rule that
```
fblanqui, line 65: Do you really need x in this type? In eval.ml, can't you use act.xvars_nb instead?

gabrielhdt: But `act` does not have such an attribute, it's a really the rhs as a binder, so we have to ship this `xvars_nb` in the `Leaf`

gabrielhdt: Ah yes it's a better idea to modify the instantiated `'rhs` type to carry this `act.xvars_nb`.

fblanqui:
Hello @gabrielhdt . Let me know if we can merge now.

gabrielhdt:
> Hello @gabrielhdt . Let me know if we can merge now.

Yeah, if you're happy with what it's done, we can.

PR #391, opened 2020-05-07
[WIP] Product matching

gabrielhdt:
This PR introduces rewriting on products, for unification rules. Is based off #390.

Products are treated just like abstractions.

### DIFF
--- a/src/core/tree.ml
+++ b/src/core/tree.ml
@@ -313,6 +313,7 @@ module CM = struct
     | Patt(_) -> false
     | Vari(_)
     | Abst(_)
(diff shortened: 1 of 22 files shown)


src/core/tree.ml
```diff
@@ -594,6 +597,32 @@ module CM = struct
     in
     (pos, List.filter_map transf cls)
 
+  (** [product col v pos cls] is like [abstract col v pos cls] for
+      products.
+      TODO: code factorisation? *)
```
fblanqui, line 602: If that is possible, that would be better.

PR #394, opened 2020-05-14
Matching on product (and domain of binders)

gabrielhdt:
This PR introduces rewriting on products, for unification rules. Is based off #390.

Products are treated just like abstractions.

The ability to match against the domain of binders has been added, it is required to write a unification rule since `$a → $b` is syntactic sugar for `Π_:$a, $b`
```
set unif_rule $a → $b ≡ T $c ↪ $a ≡ T $ea; $b ≡ T $eb; $c ≡ arrow $ea $eb
```

Commit 57c43d2ce904a0e1bc78c8a573aac1080f783c89 is an optimisation of the rewriting engine (not really related): it ensures non linear variables are not recorded twice in the environment builder, which may reduce its size in some cases, and thus should speed up rewriting (less iterations in`List.iter fn env_builder`).

Commit d868c95 allows to print decision trees of symbols from ghost signatures (for debug).

### DIFF
--- a/src/cli/lambdapi.ml
+++ b/src/cli/lambdapi.ml
@@ -78,7 +78,9 @@ let decision_tree_cmd : Config.t -> (Path.t * string) -> unit =
     Config.init cfg;
     let sym =
       let sign = Compile.compile false mp in
-      try fst (StrMap.find sym Timed.(!(sign.sign_symbols)))
+      let ss = Sig_state.of_sign sign in
+      let (prt, prv) = (true, true) in
+      try Sig_state.find_sym false ss (Pos.none ([], sym)) ~prt ~prv
       with Not_found -> fatal_no_pos "Symbol not found."
     in
     out 0 "%a" Tree_graphviz.to_dot sym
--- a/src/core/eval.ml
+++ b/src/core/eval.ml
@@ -253,10 +253,10 @@ and tree_walk : dtree -> ctxt -> stack -> (term * stack) option =
               else fail
         in
         walk next stk cursor vars_id id_vars
-    | Eos(l, r)                                           ->
+    | Eos(l, r)                                                    ->
         let next = if stk = [] then l else r in
         walk next stk cursor vars_id id_vars
-    | Node({swap; children; store; abstraction; default}) ->
+    | Node({swap; children; store; abstraction; default; product}) ->
         match List.destruct stk swap with
         | exception Not_found     -> None
         | (left, examined, right) ->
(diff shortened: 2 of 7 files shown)


src/cli/lambdapi.ml
```diff
@@ -79,6 +79,8 @@ let decision_tree_cmd : Config.t -> (Path.t * string) -> unit =
     let sym =
       let sign = Compile.compile false mp in
       try fst (StrMap.find sym Timed.(!(sign.sign_symbols)))
+      with Not_found ->
+      try fst (StrMap.find sym Timed.(Unif_rule.(!(sign.sign_symbols))))
```
fblanqui, line 83: Why do we need this if ghost modules are open?

src/core/eval.ml
```diff
@@ -285,19 +289,34 @@ and tree_walk : dtree -> ctxt -> stack -> (term * stack) option =
...
+          (* [default ()] carries on the matching on the default branch of the
+             tree. Nothing is added to the stack. *)
           let default () =
             let fn d =
               let stk = List.reconstruct left [] right in
               walk d stk cursor vars_id id_vars
             in
             Option.map_default fn None default
           in
+          (* [walk_binder a  b  id tr]  matches  on  binder  [b]  of type  [a]
```
fblanqui, line 301: I don't like very much the spacing in `[walk_binder a  b  id tr]`. In fact, I don't like very much justified text. I prefer flush left text, as in hand writing.

gabrielhdt: Ok, I'm fine with both.

src/cli/lambdapi.ml
gabrielhdt: Good remark, I'll look more into details.

src/core/tree.ml
```diff
@@ -493,9 +494,19 @@ module CM = struct
...
+             the  variables  collected.  Or  the  variable  is  subject  to  a
+             non-linearity  constraint. In  this  case we  do  not enrich  the
+             environment builder. *)
           match i with
           | Some(i) ->
-              (slot, (i, Array.map (index_var vi) e)) :: r.env_builder
+              if List.exists (fun (_,(j,_)) -> i = j) r.env_builder then
+                r.env_builder
+              else
+                (slot, (i, Array.map (index_var vi) e)) :: r.env_builder
```
fblanqui, line 509: You could simplify this by writing:
```
match i with
| Some(i) when not (List.exists (fun (_,(j,_)) -> i = j) r.env_builder) ->
  (slot, (i, Array.map (index_var vi) e)) :: r.env_builder
| _ -> r.env_builder
```

src/cli/lambdapi.ml
gabrielhdt: It's ok now that `find_sym` is used on a `sig_state`.

PR #398, opened 2020-05-16
Emacs fixes

gabrielhdt:
Here is a new round of emacs fixes. Users are incited to request changes in this PR.

For the moment,
- disable `electric-indent-mode` by default, so the mode does not automatically indent new lines (no more far away cursor for no reason)
- added some keywords for colours
- some indentation fixes,
  + queries indented according to the previous line (should be more consistent)
  + indent of 2 after a top colon


fblanqui:
Hi @gabrielhdt . Thanks for this PR. But it seems to mix two different things: the emacs mode for LP, and some formatting tool for OCaml developers. It would be better to split it in two PRs. And should we use your formatting tool?

gabrielhdt:
> Hi @gabrielhdt . Thanks for this PR. But it seems to mix two different things: the emacs mode for LP, and some formatting tool for OCaml developers. It would be better to split it in two PRs. And should we use your formatting tool?

Ok but anyway, if you prefer left-flushed comments, the two first commits should be discarded.

fblanqui:
@gabrielhdt Thanks to remove the two first commits.

gabrielhdt:
> @gabrielhdt Thanks to remove the two first commits.

It's done.

editors/emacs/lambdapi-smie.el
```diff
@@ -121,15 +106,15 @@ Indent by `lambdapi-indent-basic' in proofs, and 0 otherwise."
...
-               ("set" "infix" "right" ident "≔" ident)
-               ("set" "quantifier" ident)
-               ("set" "declared" ident)))
+               ("set" "builtin" ident "≔" sterm)
+               ("set" "prefix" ident "≔" sterm)
+               ("set" "infix" ident "≔" sterm)
+               ("set" "infix" "left" ident "≔" sterm)
+               ("set" "infix" "right" ident "≔" sterm)
+               ("set" "quantifier")
+               ("set" "declared")))
```
fblanqui, line 117: Perhaps we should put things in alphabetical order?

editors/emacs/GNUmakefile
```diff
@@ -0,0 +1,18 @@
+CASK = ~/.cask/bin/cask
```
fblanqui, line 1: What is cask? Is it a standard path? Do you document that it is necessary somewhere? (I don't have such a command on my machine.)

gabrielhdt: It's standard for emacs development (see any emacs package, for instance, https://github.com/emacs-lsp/lsp-mode/blob/master/Makefile), but I shouldn't set this path

fblanqui:
Thanks @gabrielhdt to update your branch and fix the path so that I can merge.

fblanqui:
But shouldn't they be already available with company? It's explicitly mentioned on https://github.com/vspinu/company-math .

gabrielhdt:
> But shouldn't they be already available with company? It's explicitly mentioned on https://github.com/vspinu/company-math .

Indeed, they are available with company, so if `lambdapi-unicode-prefer-company` is set to 1, you'll have it

fblanqui:
Hi @gabrielhdt . What's the status of this PR?

gabrielhdt:
Well I'm not sure. I think we can merge it since subscripts are ensured by company?

fblanqui:
What about adding also math-symbol-list-packages from https://github.com/vspinu/math-symbol-lists/blob/master/math-symbol-lists.el ?

gabrielhdt:
> What about adding also math-symbol-list-packages from https://github.com/vspinu/math-symbol-lists/blob/master/math-symbol-lists.el ?

Which elements from these lists would you like me to add?

fblanqui:
Why not adding all the elements of this list like you did for the other lists? Unless it is too much work.

gabrielhdt:
> Why not adding all the elements of this list like you did for the other lists? Unless it is too much work.

It's not a matter of work (because it's easy), but some symbols bindings are duplicated, so we can't dump the lists. For instance, the italic capital lambda is bound to `\Lambda`, as the upright capital lambda. They differ by their unicode representation.

fblanqui:
I see. Forget about this then. Let me know when your PR will be ready to merge.

gabrielhdt:
All right, I'll settle this today.

gabrielhdt:
It should be good.

fblanqui:
Thanks @gabrielhdt !

Issue #380, opened 2020-04-29
Admitted and admit


amelieled:
Is it possible to change "admit" to "admitted" (like in Coq) to add the tactic "admit" (like in Coq) in the future?

fblanqui:
Why would you like admit as a *tactic*?

amelieled:
It a very useful tactic when you tried to do a proof.
I know that LP isn't a proof assistant but maybe it's a good feature.

fblanqui:
LP *is* a proof assistant. It is just missing more tactics and a better interface.

Why not a tactic "fail"? The advantage is that you could put "fail" in the middle of a proof script.

amelieled:
Yes but it's before a logical framework, no?
Yes, your tactic "fail" is exactly the tactic Coq "admit".
Maybe "break" or "skip" are better then "fail", I don't know.

amelieled:
I tested it today, but it seems to be not correct.
This tactic just raises an error, but does nothing?

fblanqui:
Right.

amelieled:
Here, there is a problem again.

PR #399, opened 2020-05-18
add tactic fail

fblanqui:


fblanqui:
Should close #380 .

src/core/syntax.ml
```diff
@@ -164,9 +164,9 @@ type p_tactic_aux =
   (** Print the current proof term (possibly containing open goals). *)
   | P_tac_why3 of string option
   (** Try to solve the current goal with why3. *)
-
   | P_tac_query   of p_query
   (** Query. *)
+  | P_tac_fail
```
gabrielhdt, line 169: A short documentation could be added here.

src/core/tactics.ml
```diff
@@ -109,6 +109,8 @@ let handle_tactic : Sig_state.t -> Proof.t -> p_tactic -> Proof.t =
       handle_refine (Rewrite.symmetry ss tac.pos ps)
   | P_tac_why3(config)  ->
       handle_refine (Why3_tactic.handle ss tac.pos config g)
+  | P_tac_fail          ->
+      fatal tac.pos "Fail tactic"
```
gabrielhdt, line 113: As @francoisthire suggested, it would be a good idea to take time to rethink our error messages at some point. Here, I'd rather say `Intended failure`, which, IMO seems a bit friendlier and less debug-ish, but it's really subjective. What do you think?

PR #402, opened 2020-05-20
[vscode] Fix vscode after command line changes.

ejgallego:
The PR #289 broke the vscode plugin.

We do a hotfix for now, but the mode still remains broken if using
from the dev tree [for example with `dune exec -- code`].

This is quite inconvenient, for now lambdapi must be fully installed
as to be used from vscode.

### DIFF
--- a/doc/sections/ui.md
+++ b/doc/sections/ui.md
@@ -26,6 +26,9 @@ This requires to have `npm` and `node-typescript` installed:
 sudo apt install npm node-typescript
 ```
 
+_note_: as of today the vscode mode requires a full lambdapi install,
+it won't unfortunately run from a developer tree.
+
 ### Emacs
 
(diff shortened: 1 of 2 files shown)


ejgallego:
Note @fblanqui @vycastor I have confirmed the mode works in full fine.

fblanqui:
@gabrielhdt, the problem that @ejgallego mentions (running vscode with the dev version), isn't also the case with emacs?

gabrielhdt:
> @gabrielhdt, the problem that @ejgallego mentions (running vscode with the dev version), isn't also the case with emacs?

As things are now, it is not easily possible to run the emacs mode with a development version of emacs, because the lsp server fetches the binary in the usual path. So there are no "breaking" problem, but the situation is not the happiest either.

PR #403, opened 2020-05-20
[pure] [lsp] Try to recover unit testing of pure interface

ejgallego:
PR #289 silently removed important unit tests in the Pure module.

We try to recover them, but indeed, they don't work, as it seems the
library handling API introduced in #289 is lacking in terms of how it
can be used. There seem to be many problems with the new system in
terms of what a language server needs, in particular regarding
assumptions about the underlying file system.

I'm afraid this seems to need significant work.


ejgallego:
In general it would be nice if future PRs don't remove important tests without fixing them.

ejgallego:
@vycastor , I think that some of the refactoring we need for the server should also fix this; in particular the way state is handling needs to be become a bit more abstract as to allow for example file-less parsing.

fblanqui (changes requested):

ejgallego:
Thanks!

PR #404, opened 2020-05-20
add doc on proof ... qed|admit|abort + remove sections subdir

fblanqui:


fblanqui:
The new doc can be visualized on https://github.com/fblanqui/lambdapi/blob/docdir/doc/DOCUMENTATION.md.

doc/DOCUMENTATION.md
```diff
@@ -9,36 +9,34 @@ project.
...
- - [Syntax of terms](sections/terms.md)
+ - [Syntax of terms](terms.md)
 
- - [Commands](sections/commands.md)
+ - [Commands](commands.md)
 
- - [Tactics](sections/tactics.md)
+ - [Tactics](tactics.md)
 
- - [Editing Lambdapi source code with Emacs](sections/emacs.md)
```
gabrielhdt, line 30: There is no more entry for emacs, is it wanted?

doc/terms.md
```diff
@@ -1,13 +1,25 @@
...
 file `dir1/` ... `/dirn/file.lp`. To be used, `dir1.` ... `dirn.file`
 must be required first.
 
+An identifier can be:
+
+  * an ASCII sequence of characters in the regular language `[a-zA-Z_][a-zA-Z0-9_]*`
+
+  * a non-negative integer if the builtins "0" and "+1" have been previously set (see the [command](commands.md) `set builtin`)
+
+  * a unicode x character previously declared using the [command](commands.md) `set declared "x"`
```
gabrielhdt, line 17: I would more simply say
> a unicode character that has been declared using the ...

doc/DOCUMENTATION.md
fblanqui: Yes, it is moved into ui.md.

fblanqui:
@gabrielhdt I took into account your remarks. Thanks. Do you have more?

doc/DOCUMENTATION.md
gabrielhdt: So personally, I'd prefer to keep a slightly longer table of content (it's not that long to read...) to have more accessible pages and less "click paths" to reach a page.

gabrielhdt:
> @gabrielhdt I took into account your remarks. Thanks. Do you have more?

I made one in the above conversation.

doc/DOCUMENTATION.md
fblanqui: Good point. I gave more structure to the toc. Is it better now?

gabrielhdt: Yes, that's nice

PR #406, opened 2020-05-27
Update CONTRIBUTING.md

amelieled:
The purpose of this PR is to discuss the rules for contributing to the project. I only modified the CONTRIBUTING. md file to make explicit some rules that are still implicit.

You tell me what's missing!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,22 +39,11 @@ hook.
 About code, **remember to break the line after `=`, if it is too long**.
 You can find more details in the previous sections.
 
-Importing of librairies
+Importing librairies
 -----------------------
 
 Import a library when you use it many times.
 But to avoid problems with generic names (make, get, ...), you have to put the name of the library before the name of the function.
-
-Naming of variables
--------------------
-
-Type of the variable | Name of the variable
--------------------- | --------------------
-       prop          |        p
-       expo          |        e
-       term          |        t
-       sym           |
-     Sig_state       |        ss
      
 Type annotations and interface
 ------------------------------
@@ -77,14 +66,6 @@ You can do the same thing for nested functions, but it's not mandatory.
 
 Pattern matching
 ----------------
-
-Pattern matching arrows should be aligned to improve readability, especially in the case where almost all of the pattern matching can be seen, as in the
-following.
-```ocaml
-match ... with
-| A(x)   -> ...
-| B(x,y) -> ...
-```
 
 When the action of a pattern does not fit on one line, it should be indented
 by two units (that is, four characters).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,8 +14,20 @@ In the interest of code source uniformity, we ask that:
  - there should be *no trailing spaces* at the end of lines,
  - lines length should be limited to *78 characters* (excluding newline).
 
+You can add the following lines in your .emacs to check the previous constraints:
+```
+(require 'whitespace)
+(setq whitespace-line-column 78)
+(add-hook 'tuareg-mode-hook (function (lambda () (whitespace-mode t))))
+(add-hook 'lambdapi-mode-hook (function (lambda () (whitespace-mode t))))
+```
+and the following line to set the default Emacs window size as the maximum size of a line:
+```
+(add-to-list 'default-frame-alist '(width . 82))
+```
+
 You should at the very least run `make sanity_check` before committing anything.
-Please check you have GNU awk (gawk) installed or another UTF-8 compatible
+*Please check you have GNU awk (gawk) installed* or another UTF-8 compatible
 implementation of the AWK programming language interpreter.
 
 The script `tools/git_hook_helper.sh` helps setting up a git hook to run
@@ -24,6 +36,26 @@ It is encouraged to set up such a hook.
 The script may be called with the `-b` option to include compilation in the
 hook.
 
+About code, **remember to break the line after `=`, if it is too long**.
+You can find more details in the previous sections.
+
+Importing of librairies
+-----------------------
+
+Import a library when you use it many times.
+But to avoid problems with generic names (make, get, ...), you have to put the name of the library before the name of the function.
+
+Naming of variables
+-------------------
+
+Type of the variable | Name of the variable
+-------------------- | --------------------
+       prop          |        p
+       expo          |        e
+       term          |        t
+       sym           |
+     Sig_state       |        ss
+     
 Type annotations and interface
 ------------------------------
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -41,12 +73,12 @@ Note that `ocamldoc` comments should take up as much space as possible,
...
+You can do the same thing for nested functions, but it's not mandatory.
 
+Pattern matching
+----------------
 
-Other conventions
------------------
-
-Pattern matching arrows should be aligned to improve readability, as in the
+Pattern matching arrows should be aligned to improve readability, especially in the case where almost all of the pattern matching can be seen, as in the
```
fblanqui, line 81: Thanks to remove this constraint.

CONTRIBUTING.md
```diff
@@ -24,6 +36,26 @@ It is encouraged to set up such a hook.
...
+
+Naming of variables
+-------------------
+
+Type of the variable | Name of the variable
+-------------------- | --------------------
+       prop          |        p
+       expo          |        e
+       term          |        t
+       sym           |
```
fblanqui, line 56: There's nothing for sym. But is it useful to go into these details? Moreover, a same name can be used with different types. p is used for prop but also for positions, s is used for sym but also for string, etc.

amelieled: Yes, it's a good question! That's why I left a blanck.
This part isn't finished but I'm not sure it's easy to read and very useful.

CONTRIBUTING.md
```diff
@@ -61,3 +93,78 @@ by two units (that is, four characters).
...
+ (* ... *)
+in
+```
+
+Building of examples
+--------------------
+
+Don't forget to build your own examples about your job.
+Moreover, the naming convention is to start the name of an object with a lowercase letter, and the name of a type with an uppercase letter.
+For example:
```
gabrielhdt, line 159: This paragraph is mixing naming conventions and model for pull requests: the two are good ideas, but let's make things clear.

CONTRIBUTING.md
```diff
@@ -14,8 +14,20 @@ In the interest of code source uniformity, we ask that:
  - there should be *no trailing spaces* at the end of lines,
  - lines length should be limited to *78 characters* (excluding newline).
 
+You can add the following lines in your .emacs to check the previous constraints:
+```
+(require 'whitespace)
+(setq whitespace-line-column 78)
```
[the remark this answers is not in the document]
amelieled: Ok maybe, I don't know, but it's not the correct number.
I made some tests and the default value is 80.

CONTRIBUTING.md
```diff
@@ -24,6 +36,26 @@ It is encouraged to set up such a hook.
 The script may be called with the `-b` option to include compilation in the
 hook.
 
+About code, **remember to break the line after `=`, if it is too long**.
+You can find more details in the previous sections.
+
+Importing of librairies
```
ejgallego, line 42: Typo

CONTRIBUTING.md
```diff
@@ -14,8 +14,20 @@ In the interest of code source uniformity, we ask that:
...
+```
+(require 'whitespace)
+(setq whitespace-line-column 78)
+(add-hook 'tuareg-mode-hook (function (lambda () (whitespace-mode t))))
+(add-hook 'lambdapi-mode-hook (function (lambda () (whitespace-mode t))))
+```
+and the following line to set the default Emacs window size as the maximum size of a line:
+```
+(add-to-list 'default-frame-alist '(width . 82))
+```
```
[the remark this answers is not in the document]
amelieled: Yes you right but maybe it can be add in your Emacs mode :)

CONTRIBUTING.md
```diff
@@ -24,6 +36,26 @@ It is encouraged to set up such a hook.
 The script may be called with the `-b` option to include compilation in the
 hook.
 
+About code, **remember to break the line after `=`, if it is too long**.
```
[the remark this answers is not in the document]
amelieled: I'm not sure to completly understand but I think it's a good idea.

CONTRIBUTING.md
amelieled: I agree so you would like to add sub-sections ? sections ?

ejgallego:
Most of the conventions can be enforced by ocamlformat as discussed in other PRs.

A couple of extra comments:
- raising exceptions tends not to scale well in OCaml [Coq counts by thousands the anomalies due to this], I'd avoid that an use `Result.t` [also that tends to produce more efficient code in quite a few cases]. If a function is raising, a useful convention is to suffix it with `_exn`, for example `find_exn`. This is used in large codebases.
- IMHO not having `.mli` files tends to have a pretty bad effect in terms of code organization; once you get several developers writing code, they will invariably end up depending on internals, then you are left with hard-to-maintain code. The usual trick to avoid duplication is to indeed provide the corresponding module types in a `foo_intf.ml` file, and include. This pattern is used pervasively in `core`, etc.... Not having `.mli` files also has bad properties in terms of compilation time, but that's not so much of a problem now for LP.

CONTRIBUTING.md
amelieled: OK but naming schemes could be useful? I'm not sure!

CONTRIBUTING.md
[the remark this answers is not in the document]
fblanqui: Well, the emacs mode is about LP files, not about OCaml files.

CONTRIBUTING.md
```diff
@@ -14,8 +14,20 @@ In the interest of code source uniformity, we ask that:
  - there should be *no trailing spaces* at the end of lines,
  - lines length should be limited to *78 characters* (excluding newline).
 
+You can add the following lines in your .emacs to check the previous constraints:
+```
+(require 'whitespace)
+(setq whitespace-line-column 78)
+(add-hook 'tuareg-mode-hook (function (lambda () (whitespace-mode t))))
+(add-hook 'lambdapi-mode-hook (function (lambda () (whitespace-mode t))))
```
fblanqui, line 22: This should not be here as it is not about OCaml files.

amelieled:
@gabrielhdt, everyone can contribute to this PR :)

fblanqui:
Le 29/05/2020 à 13:32, Emilio Jesús Gallego Arias a écrit :
>
>   * IMHO not having |.mli| files tends to have a pretty bad effect in
>     terms of code organization; once you get several developers
>     writing code, they will invariably end up depending on internals,
>     then you are left with hard-to-maintain code. The usual trick to
>     avoid duplication is to indeed provide the corresponding module
>     types in a |foo_intf.ml| file, and include. This pattern is used
>     pervasively in |core|, etc.... Not having |.mli| files also has
>     bad properties in terms of compilation time, but that's not so
>     much of a problem now for LP.
>
Yes, this is something we should change at some point.

CONTRIBUTING.md
fblanqui: We should not impose too many constraints and keep things simple and easy to adopt.

CONTRIBUTING.md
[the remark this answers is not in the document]
amelieled: Oh yes you right!
The previous lines are just here to help during contributing.

CONTRIBUTING.md
amelieled: Thanks!

CONTRIBUTING.md
amelieled: I have removed it.

CONTRIBUTING.md
amelieled: I have removed it.

amelieled:
Can I delete with PR tomorrow?

fblanqui:
No need to delete it. We can just close it.

PR #408, opened 2020-05-27
Cleaning

amelieled:
A cleaning on documentation.
I checked that each function has the correct name and the corresponding variables compared to their documentation.
Only the comments have changed.


amelieled:
I wish they'd been taken into account.

PR #412, opened 2020-05-29
change semantics of underscore (fix #395)

### LINKED ISSUE #395: Semantics of underscore inside a pattern
The following example in `lambdapi` does not type check:

```
// Generated example file.


constant symbol typ : TYPE

injective symbol term : typ → TYPE

constant symbol I : typ

constant symbol A : typ

constant symbol i0 : term I

constant symbol i1 : term I

constant symbol Path {A : typ} : term A → term A → typ

symbol lamP : Π {A : typ} (f : term I → term A), term (Path {A} (f i0) (f i1))

symbol appP : Π {A : typ} (a : term A) (b : term A), term (Path {A} a b) → term I → term A

rule appP _ $b _ i1 ↪ $b
rule appP $a _ _ i0 ↪ $a

rule lamP (λx, appP _ _ $F x) ↪ $F
```

But if we replace the last rule by 

```
rule lamP (λx, appP $a $b $F x) ↪ $F
```

It works. If I understand correctly this is because an `_` is replaced by `$a x` and the dependency introduces an error which I am not sure to understand why. I had the impression that by typing, the variable x cannot be free in $a or $b so `$a x` or `$a` are the same.

Moreover, I would like to mention that the error message could probably be enhanced.

fblanqui:


src/core/pretty.ml
```diff
@@ -49,7 +49,9 @@ let rec pp_p_term : p_term pp = fun oc t ->
   let out fmt = Format.fprintf oc fmt in
   let rec pp p _ t =
     let pp_env _ ar =
-      if Array.length ar > 0 then out "[%a]" (Array.pp (pp PFunc) ", ") ar
+      match ar with
+      | None -> ()
+      | Some ar -> out "[%a]" (Array.pp (pp PFunc) ", ") ar
```
fblanqui, line 54: Note that, now, when translating Dk files to LP files, pattern variables will always be followed by square brackets even if it has no argument. To change this, one needs to know whether we went through an abstraction or not.

fblanqui: Fixed in 4e9dc0e71f36adf8b0be816a9c292d3e65f15434

gabrielhdt:
Can you remove the quail command that binds `[]` to the box? We'll need to type `[]` rather often.

fblanqui:
Yes, I will do it. But we won't have to write it very often! Why do you think so?

gabrielhdt:
> Yes, I will do it. But we won't have to write it very often! Why do you think so?

You're right, I think I'm biased because the issue was just raised and so I wanted to write it a couple of times recently.

fblanqui:
Hello @gabrielhdt . Do you have other remarks?

fblanqui:
Well, finally, I am going to revert some changes. I will give more details about that soon.

src/core/scope.ml
```diff
@@ -332,7 +345,13 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
                   r.m_rhs_xvars   <- (id.elt, x) :: r.m_rhs_xvars ;
                   x
         in
-        _TEnv (Bindlib.box_var x) (Array.map (scope env) ts)
+        let ts =
+          match ts with
+          | None -> Env.to_tbox env (* [$M] is equivalent to [$M[env]] where
+                                       [env] is the current environment. *)
+          | Some ts -> Array.map (scope env) ts
```
gabrielhdt, line 352: You have twice this piece of code, it can be factorised

src/core/scope.ml
```diff
@@ -408,45 +427,63 @@ let scope_term : expo -> sig_state -> env -> p_term -> term =
...
+    | (ids,a,_)::args ->
+        let acc = match a with None -> acc | Some a -> patt_vars k acc a in
+        (* We compute the number of named variables. *)
+        let rec nb_named_vars ids =
+          match ids with
+          | [] -> 0
+          | None :: ids -> nb_named_vars ids
+          | Some _ :: ids -> 1 + nb_named_vars ids
+        in
+        patt_vars_args (k + nb_named_vars ids) acc args
```
gabrielhdt, line 484: Can you explain briefly what you did here?

fblanqui:
In fact, I am mixing different things in this PR.

Let me summarize the situation in master:
For pattern-variables:
- `$M[]` is forbidden
- `$M` stands for `$M[]`
- `_` stands for `$M[x1;..;xn]` where `M` is a fresh pattern-variable and `x1;..;xn` are the variables of the context.
For meta-variables:
-  `?M[]` is forbidden
- `?M` stands for `?M[]`
- `_` stands for `?M[x1;..;xn]`.

I wanted `?M` to stand for `?M[x1;..;xn]` (like `_`) because we cannot input all the variables by hand when there are many as it is usually the case in proofs.

Then came the issue https://github.com/Deducteam/lambdapi/issues/395 raised by François who suggested to have `_` stand for `$M[]` in a LHS. But this makes `_` to not have the same meaning in a LHS and in a RHS or a term.

I initially thought that it could be interesting to have `$M` stand for `$M[x1;..;xn]` instead of `$M[]`. But, if we do this both in LHS and RHS (https://github.com/Deducteam/lambdapi/pull/412/commits/7788cf3cba39b0b55a5d4657f7c7f1b39ec499bc), many examples get broken because the RHS context is not the same as the LHS context.

So, finally, I propose to do:
For pattern-variables:
- `$M[]` is allowed
- `$M` stands for `$M[]` (`$M` could be forbidden in case of non-empty context)
- `_` stands for `$M[x1;..;xn]`
For meta-variables:
-  `?M[]` is allowed
- `?M` stands for `?M[x1;..;xn]`
- `_` stands for `?M[x1;..;xn]`.

I am going to fix this PR accordingly.

src/core/pretty.ml
```diff
@@ -127,7 +137,7 @@ let pp_p_unif_rule : p_rule pp = fun oc r ->
   in
   let eqs = Unif_rule.p_unpack rhs in
   let pp_sep : unit pp = fun oc () -> Format.fprintf oc ", " in
-  Format.fprintf oc "@[<hov 3>%a@ ↪ %a@]@?"
+  Format.fprintf oc "@[<hov 3>%a ↪ %a@]@?"
```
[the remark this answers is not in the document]
fblanqui: I am not familiar with Format but it was creating too many line breaks.

src/core/scope.ml
fblanqui: The arity of a generated pattern variable (for replacing some `_`) depends on the number of named variables. If you have `\x y,_` you get `_` replaced by `$M[x;y]`, and if you have `\x _,_` you get `_` replaced by `$M[x]`. But I am going to revert this change finally. See https://github.com/Deducteam/lambdapi/pull/412#issuecomment-638232127.

doc/terms.md
```diff
@@ -41,11 +41,11 @@ An identifier can be:
...
  * application is written by space-separated juxtaposition, except for symbol identifiers declared as infix (e.g. `x+y`)
 
- * a meta-variable application `?M[t;u;v]`
+ * a meta-variable application `?M[t;u;v]`. `?M` alone, without arguments between square brackets, is a shorthand for `?M[x1;..;xn]` where `x1;..;xn` are all the variables of the context.
 
- * a pattern-variable application `$M[x;y]` (in rules only)
+ * a pattern-variable application `$M[x;y]` (in rules only). `$M` alone, without arguments between square brackets, is a shorthand for `$M[x1;..;xn]` where `x1;..;xn` are all the variables of the context.
 
- * `_` for an unknown term or a term we don't care about. It will be infered by the system or, inside a proof, replaced by a fresh metavariable generating a new subgoal. It is replaced by a metavariable (or a pattern variable in the case of a rule left-hand side) applied to all the variables of the context.
+ * `_` for an unknown term or a term we don't care about. It is replaced by a fresh metavariable (or a fresh pattern variable in a rule left-hand side) applied to all the variables of the context.
```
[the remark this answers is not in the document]
fblanqui: Yes.

src/core/scope.ml
fblanqui: Not anymore.

fblanqui:
@gabrielhdt @Rehan-MALAK Do you have any other remark?

Do you think that we should enforce the use of empty square brackets in case of a non-empty context? That is, to write `rule lam (λx, app $F[] x) ↪ $F[]` instead of `rule lam (λx, app $F x) ↪ $F`?

gabrielhdt:
> @gabrielhdt @Rehan-MALAK Do you have any other remark?
> 
> Do you think that we should enforce the use of empty square brackets in case of a non-empty context? That is, to write `rule lam (λx, app $F[] x) ↪ $F[]` instead of `rule lam (λx, app $F x) ↪ $F`?

It would make sense. IMO, either we do that, or we should provide very clear and informative error message on unification failure, messages like `$F is equivalent to $F[], perhaps you meant $F[x, y, z,...]?`

fblanqui:
Hi @gabrielhdt . Brackets are now mandatory under binders in LHS. If you have no further remarks, I will merge.

Issue #348, opened 2020-04-10
Emacs mode does not work when opening a file that does not exist yet


fblanqui:

fblanqui:
When we launch emacs with, as argument, a filename that does not exist yet, type some LP code (e.g. symbol N:TYPE) and do Ctrl-C Ctrl-N, emacs loop. The same thing happens when, in emacs, we open a file that does not exist yet. @firewall2142, could you please have a look at this problem?

firewall2142:
I have nothing to say other than what has been discussed in #432. I could make emacs create the file if it doesn't exist, but, as said there, it is a workaround.

PR #414, opened 2020-06-04
Goals display in Emacs

vycastor:
Here is a basic implementation of goals display in Emacs. It writes the current hypotheses and goals at cursor position in a buffer called *Goals* that it opens when calling eglot--signal-proof/goals (I followed the eglot nomenclature for LSP calls).


fblanqui:
Thank you @vycastor ! Some remarks:
- Could the Goals buffer be opened automatically (in the bottom 25% of the current window)?
- Can the buffer be updated when the cursor is moved?
- You can remove the last '-------' line and make the first '---------' line longer (possibly of the size of the window)

fblanqui:
- Concerning the first item, what is important is that the goal buffer is opened when one open a lp file in emacs, so that we don't have to call eglot--signal-proof/goals.This is perhaps related to the issue https://github.com/Deducteam/lambdapi/issues/348.
- Concerning the second item, yes, I mean on real time. An alternative or complementary way would be to offer key bindings, or buttons like in https://proofgeneral.github.io/, to navigate in the proofs.

gabrielhdt:
Perhaps using a key combination to refresh the buffer would be better (à la proof general), we might want to keep a goal in visual while editing another term, and updating automatically the window would prevent this.

fblanqui:
Right @gabrielhdt . But, in fact, the two are not incompatible. In PG, there is an option allowing to update a proof as soon as you finished to write a new proof command. @vycastor I invite you to have a look at PG. I think it would be nice to reuse the same key bindings so that a PG/Coq user could easily use Dk/LP as well. In fact, you could perhaps reuse some of their LISP code.

ejgallego:
IIANM emacs already has some library to control windows layouts, it should help quite a bit to manage the side / bottom buffer.

vycastor:
> IIANM emacs already has some library to control windows layouts, it should help quite a bit to manage the side / bottom buffer.

It indeed does, it's not a problem. 

> Right @gabrielhdt . But, in fact, the two are not incompatible. In PG, there is an option allowing to update a proof as soon as you finished to write a new proof command. @vycastor I invite you to have a look at PG. I think it would be nice to reuse the same key bindings so that a PG/Coq user could easily use Dk/LP as well. In fact, you could perhaps reuse some of their LISP code.

AFAIU, PG is meant to be adapted to each prover, so why wouldn't we do that (at least for the prover-related features)? Do you know any shortcomings? I must say I'm not knowledgeable at all on PG for the moment...

fblanqui:
No shortcomings a priori. Just note that PG is not based on LSP, but this should not be a problem.

fblanqui:
Hi @vycastor . It seems to work very nicely. Thank you! Could you please just change the way goals are printed, as suggested in the meeting two days ago? Currently, we have:
```
y: τ nat
--------------
Goal 77: π ((0 + suc y) = suc (0 + y))
Goal 71: Πx: τ nat, (Πx0: τ nat, π ((x + suc x0) = suc (x + x0))) → Πx0: τ nat, π ((suc x + suc x0) = suc (suc x + x0))
--------------
```
And I prefer to have:
```
y: τ nat
----------------------------------------------------
Goal 77: π ((0 + suc y) = suc (0 + y))

----------------------------------------------------
Goal 71: Πx: τ nat, (Πx0: τ nat, π ((x + suc x0) = suc (x + x0))) → Πx0: τ nat, π ((suc x + suc x0) = suc (suc x + x0))
```

fblanqui:
@vycastor The PR should include some update of doc/emacs.md as well.

vycastor:
It should already include the cosmetic changes you wanted!



> @vycastor The PR should include some update of doc/emacs.md as well.

Ok, I'll do it asap! :)

fblanqui:
Hi @vycastor .
I  do not understand how Alt-up/down work. Would it be possible to visualize the proof-cursor in some way? In PG, I believe the background is of a different color until the proof-cursor. I use "proof-cursor" to distinguish it from the writing-cursor which is the normal blinking cursor used to write/remove text in a buffer.
What is exactly the semantics of Ctrl-Alt-c? At the beginning, the proof-cursor follows the writing-cursor. My understanding is that, when we do Ctrl-Alt-c, we create the proof-cursor. If we do it again, then we remove it. And so on.
When the proof-cursor is on, Alt-up makes it move one command backward, and Alt-down one command forward.
For instance, this is what currently happens when I open Nat.lp:
- On line 55, I have one goal 44.
- I move the writing-cursor down on line 56, I get two goals 48 and 50.
- I do Ctrl-Alt-c.
- When I move the writing-cursor, nothing happens. Good.
- Alt-down/up makes the writing cursor move (this shouldn't) and the proof change (ok).

vycastor:
Ok, fixing this!

fblanqui:
Hi @vycastor . It seems to work well. I just have a few remarks:
- Would it be possible to use other key bindings? It would be nice to use the same key bindings in Emacs and VSCode. Ctrl-up to move backward, Ctrl-down to move forward, Ctrl-enter to move the proof-cursor to the writing-cursor, and Ctrl-Alt-c to toggle the mode.
- Would it be possible to change the color background not just on the line of the proof-cursor but from the top to the line of the proof-cursor like in PG?
- Finally, could you please update your PR so that I can merge it?

gabrielhdt:
> Hi @vycastor . It seems to work well. I just have a few remarks:
> 
>     * Would it be possible to use other key bindings? It would be nice to use the same key bindings in Emacs and VSCode. Ctrl-up to move backward, Ctrl-down to move forward, Ctrl-enter to move the proof-cursor to the writing-cursor, and Ctrl-Alt-c to toggle the mode.
> 
>     * Would it be possible to change the color background not just on the line of the proof-cursor but from the top to the line of the proof-cursor like in PG?
> 
>     * Finally, could you please update your PR so that I can merge it?

Hi @fblanqui and @vycastor,
Regarding the keybindings, `C-Alt-down` and alike are not very emacs-ish, it may be better to use common major mode prefix `C-c` for all commands (and `C-x` is often used for emacs system, such as `C-x C-c` which closes emacs). My suggestions would be
- `C-c C-c` to display goals, because it's quick (`C-c C-d` is ok as far as I'm concerned)
- `C-c C-p` for proof backward (mimicking `C-p` for previous line in emacs)
- `C-c C-n` for proof forward (mimicking `C-n` for next line)
- and why not `C-c C-i` for interactive proof mode.

It seems to me that those keybindings will be easier to remember and more natural for people who use emacs.

editors/emacs/lambdapi-mode.el
```diff
@@ -86,6 +87,108 @@
...
+  (interactive)
+  (save-excursion
+    (let ((line (plist-get proof-line-position :line)))
+      (if interactive-goals
+          (progn
+              (setq proof-line-position (eglot--widening
+                                         (list :line (line-number-at-pos) :character 0)))
+              (goto-line (line-number-at-pos))
+              (hlt-highlight-region (line-beginning-position) (line-end-position)))
+        (progn
```
gabrielhdt, line 169: The else part of an `if` is implicitly a `progn`, so you can remove `progn` and just leave the two expressions.

editors/emacs/lambdapi-mode.el
```diff
@@ -119,13 +222,25 @@
...
 ;;;###autoload
 (add-to-list 'auto-mode-alist '("\\.lp\\'" . lambdapi-mode))
 ;;;###autoload
 (add-to-list 'auto-mode-alist '("\\.dk\\'" . lambdapi-legacy-mode))
 
+;; Keybinding for goals display
+(global-set-key (kbd "C-x C-d") 'lp-display-goals)
+(global-set-key (kbd "C-M-c")   'toggle-interactive-goals)
+(global-set-key (kbd "<M-up>")  'lp-proof-backward)
+(global-set-key (kbd "<M-down>") 'lp-proof-forward)
```
gabrielhdt, line 243: Could you please define the keymaps in `lambdapi-vars`?

editors/emacs/dune
```diff
@@ -10,6 +10,7 @@
  (package lambdapi)
  (section share_root)
  (files
+  (highlight.el as emacs/site-lisp/highlight.el)
```
gabrielhdt, line 13: The dune file has been removed in `master`, it has been replaced by a makefile in ea571d801a139f3fc3281742ff66d37294d703a7.

vycastor: Oh, didn't see that! Thanks.

vycastor: For the sake of testing I had to modify this dune file again, but I assume it'll not be a problem when merging...

editors/emacs/lambdapi-proofs.el
```diff
@@ -0,0 +1,90 @@
+(defconst lp-goal-line-prefix "---------------------------------------------------")
```
gabrielhdt, line 1: Please put this at the top of the file
```elisp
;;; lambdapi-proofs.el --- Proof interactivity for lambadpi-mode -*- lexical-binding: t; -*-
;; SPDX-License-Identifier: CeCILL Free Software License Agreement v2.1
;;; Commentary:
;;
;;
;;; Code:
```

editors/emacs/lambdapi-vars.el
```diff
@@ -83,5 +86,11 @@
 Useful for people who prefer to have the dropdown window display systematically,
 since the window won't display if there are quail candidates.")
 
+;; Keybindings for goals display
+(global-set-key (kbd "C-c C-c") 'lp-display-goals)
+(global-set-key (kbd "C-c C-i") 'toggle-interactive-goals)
+(global-set-key (kbd "C-c C-p") 'lp-proof-backward)
+(global-set-key (kbd "C-c C-n") 'lp-proof-forward)
```
gabrielhdt, line 93: Hmm finally I think it might be better to put this as well into `lambdapi-proofs`, to avoid adding dependencies on this file, sorry.

editors/emacs/dune
gabrielhdt: Yes I see, I think you will only have to delete this file when merging.

vycastor:
Thanks @gabrielhdt for your input!

fblanqui:
Hi @vycastor . There seems to be only 2 comments from @gabrielhdt to fix before we can merge.

vycastor:
Which ones? I tought I fixed them all :p

gabrielhdt:
The two of the last review, the header of `lambdapi-proofs.el` and moving keys definition to `lambdapi-proofs.el` (to avoid requiring anything in `lambdapi-vars.el`).

vycastor:
Oh... I just saw I didn't commit and push them... It should be ok now. Sorry for the inconvenience.

fblanqui:
Your branch is out-of-date now. Could you please update it?

vycastor:
Done!

fblanqui:
Désolé @vycastor mais il semble qu'on a été plus rapide que toi ;-) Il faudrait remettre à jour...

vycastor:
Hop

fblanqui:
Thanks!

Issue #376, opened 2020-04-26
Unification + Interactivity


Rehan-MALAK:
Dear all,
As you may know, some of us are interested in a bit of interactivity in unification.
Whatever it means concretely (unification hints #361, a new tactic language, etc...), this is how lambdapi is currently organized #154 (restricted to modules related to unification, inference, typing, tactics/proofs, ...) and what we probably should have in mind before taking any important design decision.
![modules](https://user-images.githubusercontent.com/39339802/80320468-d86e8080-8816-11ea-8577-46349d76fb27.png)

fblanqui:
Hi @Rehan-MALAK . Woah! How did you get this graph? It's nice. Do you have a solution for #154?

amelieled:
Yes it's wonderful! 
How did you get this graph? It'll be very useful for everyone.

Rehan-MALAK:
It's an awful and non-robust script. And it has to be applied on modified sources because at some point, I'm "grep-ing" on the sources to get the defined symbols... Not really nice... 

1. restrict to a chosen set of interesting modules
2. remove the interesting ``open Module`` in sources so that they are explicitly called and not in scope
3. get the dependencies with ``ocamldep ... | grep``
4. generate the *mli from the *ml and get the constants and functions with ``ocamlc.opt -i ... | grep -e "^val"`` 
5. grep the constants and functions in the dependencies
6. generate the dot files with one colour by module dependency

the script has to be used after building lambdapi with dune (as it used the the src/core/.core.objs/native/*.cmi files)
change LAMBDAPIDIR if needed

```bash
#!/bin/bash

set -o nounset

#uppercase the first letter of a word
function uppercasefirst ()
{
  WORD=$1
  FIRST=${WORD:0:1}
  FIRSTUPPER=${FIRST^^}
  WORDWITHOUTFIRST=${WORD:1}
  echo ${FIRSTUPPER}${WORDWITHOUTFIRST}
}

LAMBDAPIDIR=.
GRAPH=graph
GRAPHDOT=$LAMBDAPIDIR/$GRAPH.dot
GRAPHPNG=$LAMBDAPIDIR/$GRAPH.png
INTERESTING_MODULES=( compile handle infer proof sr tactics typing unif )

# #choose the interesting_modules modules
# cd $LAMBDAPIDIR/src/core
# INTERESTING_MODULES=( )
# for MLFILE in $(ls *.ml) ; do
[...]

Issue #278, opened 2019-11-14
Allow the definition of a symbol using a proof


fblanqui:

fblanqui:
@Rehan-MALAK could you please take care of this?

Rehan-MALAK:
cf #422

PR #422, opened 2020-06-16
[WIP] proof mode with typing/unification goals and a solve tactic

Rehan-MALAK:
WIP unification goals/tactics (cf #376 #278) :

  * unification goals and the usual type goals
  * a single proof mode for symbols with or without term definition, with or without list of tactics
  * there is only one single non-failing call to the unification algorithm before the symbol being added to the signature

######### ideas for later  ########
  * the scoping should return the list of metas
  * duplicate handle_refine => handle_refine handle_refine_nosolve
  * duplicate begin => begin + begin_nosolve 
  * substitute the two previous with a new modifier
  * keep dependency between goals : graph ?
  * modify rules accordingly to what have been done with symbols
  * modifier in the tactics in order to choose when we want or not a call to the default unification solver
  * keeping all the metas of the file and do the meta instanciation (decouples definition and meta resolution) => more drastic change


src/core/parser.ml
```diff
@@ -609,6 +611,9 @@ let parser statement =
 let parser proof =
   ts:tactic* e:proof_end -> (ts, Pos.in_pos _loc_e e)
 
+let parser keyword_proof =
+  _proof_ ts_pe:proof -> ts_pe
```
fblanqui, line 615: These two parsers could be merged into a single one:
```
let parser proof =
  _proof_ ts:tactic* e:proof_end -> (ts, Pos.in_pos _loc_e e)
```

src/core/proof.ml
```diff
@@ -9,45 +9,58 @@ open Print
...
-    type t
+    (** Representation of a proof type goal. *)
+    type goal_typ = (* TODO have to hide this *)
+      { goal_meta : meta  (* Metavariable needing instantiation.          *)
+      ; goal_hyps : Env.t (* Precomputed scope for a suitable term.       *)
+      ; goal_type : term  (* Precomputed type for a suitable term.        *) }
 
-    (** [of_meta m] create a goal from the metavariable [m]. *)
-    val of_meta : meta -> t
+    (** Representation of a general goal : type, unification, special cases *)
```
fblanqui, line 18: There is no special cases anymore, right?

src/core/proof.ml
```diff
@@ -65,40 +78,92 @@ type t = proof_state
     [builtins] may be used by tactics, and have been declared. *)
 let init : Pos.strloc -> term -> t = fun name a ->
   let proof_term = fresh_meta ~name:name.elt a 0 in
-  let proof_goals = [Goal.of_meta proof_term] in
-  {proof_name = name; proof_term; proof_goals}
+  {proof_name = name; proof_term; proof_goals = []}
+
+(** [init builtins name a] returns an initial proof state for a theorem  named
```
fblanqui, line 83: The arguments in the comment need to be fixed.

src/core/handle.ml
```diff
@@ -143,12 +203,13 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
       List.iter Tree.update_dtree syms;
       (ss, None)
-  | P_definition(e,op,x,xs,ao,t) ->
+  | P_definition(e,op,x,xs,ao,t,ts_pe) ->
       (* We check that [x] is not already used. *)
       if Sign.mem ss.signature x.elt then
         fatal x.pos "Symbol [%s] already exists." x.elt;
       (* Desugaring of arguments and scoping of [t]. *)
       let t = if xs = [] then t else Pos.none (P_Abst(xs, t)) in
+(*       let tt = t in *)
```
fblanqui, line 212: to be removed

src/core/tactics.ml
```diff
@@ -85,7 +85,7 @@ let handle_tactic : Sig_state.t -> Proof.t -> p_tactic -> Proof.t =
       | Some gt -> Goal.get_type gt
       | None -> assert false
     in
-    let tt = Scope.scope_term Public ss env t in
+    let tt = Scope.scope_term Privat ss env t in
```
Rehan-MALAK, line 88:
```
constant symbol U : TYPE
private symbol V : U
definition W ≔ V
```

=> Private symbol not allowed here.

```
constant symbol U : TYPE
private symbol V : U
private definition W ≔ V
```

=> typechecks with master but needs here this modification in the scoping (as in this PR, definition uses also the proof mode)

``theorem`` begin opaque, it should be allowed to user any symbols it wants (disregarding the symbol exposition : public/private/...) in the proof script

But a ``definition`` being non-opaque, the scoping of the symbols used during the "proof" should depend on the exposition of the definition itself.

Possibilities for ``handle_tactic`` to know about the definition exposition :

- in sig_state
- new handle_refine parameter
- put it in data proof structure (in handle.ml)

src/core/proof.ml
```diff
@@ -73,39 +73,59 @@ type proof_state =
...
+    typability of [typ] by a sort and checking eventually that term
+    [ter] has type [typ] *)
+let sort_init : term option -> term option -> Goal.t list * term =
+  fun typ ter ->
   let goal_unif x = Goal.GoalUnif x in
-  let (sort, to_solve) = Infer.infer [] a in
+  let (typ,sort, to_solve) = match typ,ter with
+    | Some(typ),Some(ter) ->
+      let to_solve = Infer.check [] ter typ in
+      let sort, to_solve2 = Infer.infer [] typ in
```
Rehan-MALAK, line 98: comments from @barras 
this should be inversed

src/core/proof.ml
```diff
@@ -73,39 +73,59 @@ type proof_state =
...
+      typ, sort, to_solve @ to_solve2
+(*
+    | None,Some(ter) ->
+      let typ,to_solve = Infer.infer [] ter in
+      let sort,to_solve2 = Infer.infer [] typ in
+      typ, sort,to_solve@to_solve2
+*)
+    | None,Some(ter) ->
+      let typ,to_solve = Infer.infer [] ter in
+      let sort = Kind in
```
Rehan-MALAK, line 108: @barras 
sort should not be constrained here.... as we already know that typ is correctly typed by a sort

doc/syntax.bnf
```diff
@@ -17,7 +18,6 @@ reserved "let"
 reserved "open"
 reserved "print"
 reserved "private"
-reserved "proof"
 reserved "proofterm"
 reserved "protected"
 reserved "qed"
```
Rehan-MALAK, line 23: comments from @barras and @fblanqui 

OK for ``begin`` but ``qed`` should then be replaced by ``end``

src/core/proof.ml
```diff
@@ -73,39 +73,59 @@ type proof_state =
...
   let to_solve = (* TODO this shoud be removed *)
     try
       Unif.solve {Unif.empty_problem with to_solve}
     with Unif.Unsolvable -> to_solve
   in
   (* aggregate constr list of type of argument *)
   let goal_sort =
-  match sort with
+  match sort with (* TODO this part is wrong *)
     | Type | Kind -> []
```
Rehan-MALAK, line 123: comments from @barras 

this part has to be rethinked :

from https://github.com/Deducteam/lambdapi/pull/422/commits/a05fba0f7b3cb91bee93b855fce3bfbb0b8be267#diff-5542c3d87c4a7cff86d6eed271815f9cR95 to https://github.com/Deducteam/lambdapi/pull/422/commits/a05fba0f7b3cb91bee93b855fce3bfbb0b8be267#diff-5542c3d87c4a7cff86d6eed271815f9cR128 

the analysis may be in the following order :
- sort = infer typ
- case analysis on the sort
- check if needed typ = infer ter

src/core/parser.ml
```diff
@@ -666,11 +667,11 @@ let parser cmd =
...
   | _rule_ r:rule rs:{_:_with_ rule}*
       -> P_rules(r::rs)
-  | e:exposition? _definition_ s:ident al:arg* ao:{":" term}? "≔" t:term
+  | e:exposition? _definition_ st:statement t:{"≔" term}?
         ts_pe:keyword_proof?
-      -> P_definition(Option.get Terms.Public e,false,s,al,ao,t,ts_pe)
-  | e:exposition? st:statement (ts,pe):proof
-      -> P_theorem(Option.get Terms.Public e,st,ts,pe)
+      -> P_definition(Option.get Terms.Public e,false,st,t,ts_pe)
+  | e:exposition? _theorem_ st:statement ts_pe:keyword_proof
```
Rehan-MALAK: comments from @barras and @fblanqui 

even simpler... theorem is exactly an opaque definition (as it is done in https://github.com/Deducteam/Dedukti) ... so as for a definition, we should be able to give a "proof" lambda-term when the proof is trivial directly like this
```
theorem trivial : typeastheo := triviallambdaproofterm
```

Rehan-MALAK: fixed in 892ebcb

doc/syntax.bnf
Rehan-MALAK: fixed in 75c7b24

src/core/tactics.ml
Rehan-MALAK: fixed in 90fc1e5

src/core/parser.ml
Rehan-MALAK: fixed in 7ff9ab7

src/core/proof.ml
Rehan-MALAK: has been fixed since

src/core/handle.ml
Rehan-MALAK: this part has been cleaned in 3ba6c4d

src/core/proof.ml
Rehan-MALAK: fixed in f6be1ce

src/core/proof.ml
Rehan-MALAK: fixed in badcdf4

src/core/proof.ml
Rehan-MALAK: fixed in badcdf4

src/core/proof.ml
Rehan-MALAK: fixed in badcdf4

src/core/handle.ml
```diff
@@ -133,6 +134,68 @@ let handle_require_as : popt -> sig_state -> Path.t -> ident -> sig_state =
   let path_map = PathMap.add p id.elt ss.path_map in
   {ss with aliases; path_map}
 
+(** [data_proof] returns the datas needed for the proof script check
+   TODO this is going to changes with the homogenization between
+   definition and theorem *)
+let data_proof x a cmd impl expo pdata_expo pos ts pe prop mstrat d goals =
```
fblanqui, line 140: Why do you have two expo arguments?

src/core/handle.ml
```diff
@@ -164,18 +227,19 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
-      if Basics.has_metas true a then
-        begin
-          fatal_msg "The type of [%s] has unsolved metavariables.\n" x.elt;
-          fatal x.pos "We have %s : %a." x.elt pp_term a
-        end;
-      (* Actually add the symbol to the signature and the state. *)
-      out 3 "(symb) %s : %a\n" x.elt pp_term a;
-      (Sig_state.add_symbol ss expo prop mstrat x a impl None,
-       None)
+      let (ts,pe) = match ts_pe with
```
fblanqui, line 230: coding style: match on next line

src/core/handle.ml
```diff
@@ -164,18 +227,19 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+        | None ->
+          let ts = [] in
+          let pe = Pos.make cmd.pos P_proof_end in
+          (ts,pe)
+        | Some(ts,pe) -> (ts,pe)
+      in
+      let sort_unif, _ = Proof.sort_init (Some(a)) None in
+      let goals = sort_unif @ [] in
```
fblanqui, line 238: what's the point of "@ []"?

src/core/handle.ml
fblanqui: idem on line 271

src/core/handle.ml
```diff
@@ -196,115 +260,74 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
         | None    -> (None, List.map (fun (_,_,impl) -> impl) xs)
         | Some(a) ->
-            let a = if xs = [] then a else Pos.none (P_Prod(xs,a)) in
-            (Some(a), Scope.get_implicitness a)
+          let a = if xs = [] then a else Pos.none (P_Prod(xs,a)) in
+          (Some(a), Scope.get_implicitness a)
       in
       let ao = Option.map (scope_basic expo) ao in
-      (* If a type [a] is given, then we check that [a] is typable by a sort
+      (* If a type [ao] is given, then we check that it is typable by a sort
```
fblanqui, line 287: If [ao = Some a]

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
 (** Short synonym for qualified use. *)
 type t = proof_state
 
-(** [init builtins name a] returns an initial proof state for a theorem  named
+(** [init name a] returns an initial proof state for a theorem  named
     [name], which statement is represented by the type [a]. Builtin symbols of
-    [builtins] may be used by tactics, and have been declared. *)
+    [builtins] may be used by tactics, and have been declared.
```
fblanqui, line 78: There is no builtins argument anymore. This should be removed.

src/core/proof.ml
```diff
@@ -9,45 +9,58 @@ open Print
...
+      { goal_meta : meta  (* Metavariable needing instantiation.          *)
+      ; goal_hyps : Env.t (* Precomputed scope for a suitable term.       *)
+      ; goal_type : term  (* Precomputed type for a suitable term.        *) }
 
-    (** [of_meta m] create a goal from the metavariable [m]. *)
-    val of_meta : meta -> t
+    (** Representation of a general goal : type, unification *)
+    type t =
+      | GoalTyp of goal_typ (* The usual proof type goal. *)
+      | GoalUnif of constr (* Two terms we'd like equal in ctxt. *)
```
fblanqui, line 21: typo: in some context

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
...
+    [builtins] may be used by tactics, and have been declared.
+    The list of goals is not initialized *)
 let init : Pos.strloc -> term -> t = fun name a ->
   let proof_term = fresh_meta ~name:name.elt a 0 in
-  let proof_goals = [Goal.of_meta proof_term] in
-  {proof_name = name; proof_term; proof_goals}
+  {proof_name = name; proof_term; proof_goals = []}
+
+(** [typ_init m] returns a goal associated to the meta m *)
+let typ_init : meta -> Goal.t list = fun m ->
```
fblanqui, line 85: rename into goals_of_meta ?

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
...
+  let goal_typ = Goal.GoalTyp (Goal.goal_typ_of_meta m) in
+  [goal_typ]
+
+(** [sort_init typ ter] returns a list of goals corresponding to the
+    typability of [typ] by a sort and checking eventually that term
+    [ter] has type [typ] *)
+let sort_init : term option -> term option -> Goal.t list * term =
```
fblanqui, line 92: rename into goals_of_typ ?

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
...
+  fun typ ter ->
+  let goal_unif x = Goal.GoalUnif x in
```
fblanqui, line 94: can be removed and unfold where it is used: `List.map (fun x -> GoalUnif x) ...`

src/core/proof.ml
```diff
@@ -9,45 +9,58 @@ open Print
...
+    type goal_typ = (* TODO have to hide this *)
+      { goal_meta : meta  (* Metavariable needing instantiation.          *)
+      ; goal_hyps : Env.t (* Precomputed scope for a suitable term.       *)
+      ; goal_type : term  (* Precomputed type for a suitable term.        *) }
 
-    (** [of_meta m] create a goal from the metavariable [m]. *)
-    val of_meta : meta -> t
+    (** Representation of a general goal : type, unification *)
+    type t =
+      | GoalTyp of goal_typ (* The usual proof type goal. *)
```
fblanqui, line 20: If Goal is never open then we could rename GoalTyp into Typ and GoalUnif into Unif, and use Goal.Typ and Goal.Unif instead of Goal.GoalTyp and Goal.GoalUnif.

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
...
+  let goal_typ = Goal.GoalTyp (Goal.goal_typ_of_meta m) in
+  [goal_typ]
+
+(** [sort_init typ ter] returns a list of goals corresponding to the
+    typability of [typ] by a sort and checking eventually that term
+    [ter] has type [typ] *)
+let sort_init : term option -> term option -> Goal.t list * term =
+  fun typ ter ->
+  let goal_unif x = Goal.GoalUnif x in
+  let (typ,sort, to_solve) = match typ,ter with
```
fblanqui, line 95: style:
```
let (typ, sort, to_solve) =
  match (typ, ter) with
```

src/core/handle.ml
Rehan-MALAK: expo is for the symbol that's going to be added to the signature

pdata_expo is the symbol exposition allowed to be used in tactics

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
...
+    | Some(typ),Some(ter) ->
+      let sort, to_solve2 = Infer.infer [] typ in
+      let to_solve =
+        begin
```
fblanqui, line 99: begin/end not useful here

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
...
+          match sort with
+          | Type | Kind -> Infer.check [] ter typ
+          | _ -> assert false
```
fblanqui, line 102: error message required

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
...
+        end
+      in
+      typ, sort, to_solve @ to_solve2
```
fblanqui, line 105: We need to start by solving constraints on the type before solving the constraints on the definition as the second ones depend on the first ones. So, to_solve2 should in fact come first.

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
...
+
+(** [sort_init typ ter] returns a list of goals corresponding to the
+    typability of [typ] by a sort and checking eventually that term
+    [ter] has type [typ] *)
+let sort_init : term option -> term option -> Goal.t list * term =
+  fun typ ter ->
+  let goal_unif x = Goal.GoalUnif x in
+  let (typ,sort, to_solve) = match typ,ter with
+    | Some(typ),Some(ter) ->
+      let sort, to_solve2 = Infer.infer [] typ in
```
fblanqui, line 97: style: `let (sort, to_solve2) = ...`. Idem later.

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
...
+          | _ -> assert false
+        end
+      in
+      typ, sort, to_solve @ to_solve2
+    | None,Some(ter) ->
+      let typ,to_solve = Infer.infer [] ter in
+      let sort,to_solve2 =
+      begin
+        match typ with
+        | Kind -> assert false (* we forbid x := _ -> TYPE *)
```
fblanqui, line 111: should be replaced by an error message

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
...
+        | _ -> let sort, to_solve2 = Infer.infer [] typ in
+          begin
+            match sort with
+            | Type | Kind -> sort,to_solve2
+            | _ -> assert false
```
fblanqui, line 116: should be replaced by an error message

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
...
+          end
+      end
+      in
+      typ,sort, to_solve @ to_solve2
+    | Some(typ),None ->
+      let sort, to_solve2 = Infer.infer [] typ in
+      begin
+        match sort with
+        | Type | Kind -> typ,sort,to_solve2
+        | _ -> assert false
```
fblanqui, line 126: should be replaced by an error message

src/core/proof.ml
```diff
@@ -60,45 +73,132 @@ type proof_state =
...
+  let goal_sort = [goal_unif constr_sort] in
+  (List.map goal_unif to_solve @ goal_sort), typ
+*)
+  let _sort = sort in
+  (List.map goal_unif to_solve), typ
+
+
+(** [unif_init cs] returns a list of unification goals corresponding to a
+    list of unification constraints [cs] *)
+let unif_init : constr list -> Goal.t list = fun cs ->
```
fblanqui, line 147: rename into goals_of_constrs

src/core/tactics.ml
```diff
@@ -12,11 +12,38 @@ open Print
 let log_tact = new_logger 't' "tact" "tactics"
 let log_tact = log_tact.logger
 
+(* Helper functions related to goals *)
```
fblanqui, line 15: should be in Goal

src/core/tactics.ml
```diff
@@ -12,11 +12,38 @@ open Print
...
+let solve ps pos =
+  try
+    let gs_typ,gs_unif = List.partition is_goal_typ ps.proof_goals in
+    let to_solve = List.map cs_of_goal_unif gs_unif in
+    let new_cs = Unif.solve {Unif.empty_problem with to_solve} in
+    let new_gs_unif = List.map goal_unif_of new_cs in
+    {ps with proof_goals = new_gs_unif @ gs_typ}
+  with
+  | Unif.Unsolvable ->
+    wrn pos "Unsolvable, try another unification tactic";
```
fblanqui, line 38: Trying another tactic won't change anything when a constraint is unsolvable. The user should backtrack instead if that's possible. Otherwise the theorem or the definition is definitely wrong or untypable.

src/core/tactics.ml
```diff
@@ -32,40 +59,72 @@ let handle_tactic : Sig_state.t -> Proof.t -> p_tactic -> Proof.t =
...
+
+  match tac.elt with
+  | P_tac_print
+  | P_tac_proofterm
+  | P_tac_query(_)      -> assert false (* Handled above. *)
+  | P_unif_solve        -> solve ps tac.pos
+  | _                   ->
+
+  (* Get the unif goals, the first type goal and the following goals *)
+  let pre_g,g,post_g =
```
fblanqui, line 73: style: hanks to add a space after a comma.

src/core/handle.ml
fblanqui: They are not the same?

fblanqui:
Thanks @francois-lefoulon to tell @Rehan-MALAK what to change in the VSCode files for replacing "proof" by "begin", and "qed" by "end", after the merging of your PR.

src/core/handle.ml
Rehan-MALAK: public definition (expo = Public) => public symbols allowed (pdata_expo = Public)
private definition (expo = Privat) => private symbols allowed (pdata_expo = Privat)
public theorem (expo = Public) => public and private symbols allowed (pdata_expo = Privat)

src/core/handle.ml
Rehan-MALAK: fixed

src/core/handle.ml
Rehan-MALAK: fixed in ed44999

src/core/handle.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in ed44999 (but only here in proof.ml ... there are some helper functions that have not still not been moved in tactics.ml

src/core/proof.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in ed44999

src/core/tactics.ml
Rehan-MALAK: replaced by fatal in ed44999

src/core/tactics.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in ed44999

src/core/proof.ml
Rehan-MALAK: fixed in 1b4a9d9

src/core/handle.ml
fblanqui: I see. Then, it would be good to add this comment somewhere.

fblanqui: Well, the function `data_proof` needs a comment too...

francois-lefoulon:
@Rehan-MALAK 
the file to edit is editors/vscode/syntaxes/lp.tmLanguage.json ; just ctrl+f "proof" and "qed" and replace them by the new keywords.

Rehan-MALAK:
thanks... editors/vscode/syntaxes/lp.tmLanguage.json  was already fixed in 75c7b24 and ee3298e

but I forgot editors/vscode/src/client.ts indeed

src/core/tactics.ml
Rehan-MALAK: fixed in 3054e19

src/core/proof.ml
Rehan-MALAK: fixed in 67fcfa7

Rehan-MALAK:
> Thanks @francois-lefoulon to tell @Rehan-MALAK what to change in the VSCode files for replacing "proof" by "begin", and "qed" by "end", after the merging of your PR.

fixed in 67fcfa7

src/core/handle.ml
Rehan-MALAK: fixed in e6a8788
some arguments have been packed in a type, others have been removed (they were redundant)

src/core/handle.ml
```diff
@@ -133,6 +134,89 @@ let handle_require_as : popt -> sig_state -> Path.t -> ident -> sig_state =
...
 
+(** Symbol properties needed for the signature are packed together *)
+type sig_symbol =
+  { expo   : expo        ;
+    prop   : prop        ;
+    mstrat : match_strat ;
+    x      : ident       ;
+    a      : term        ;
+    impl   : bool list   ;
+    d      : term option ; }
```
fblanqui, line 145: x, a and d are names that are too short. Thanks to replace for instance by ident, typ and def.

src/core/handle.ml
```diff
@@ -133,6 +134,89 @@ let handle_require_as : popt -> sig_state -> Path.t -> ident -> sig_state =
...
+
+(** Wrapper around Sig_state.add_symbol using the sig_symbol type *)
+let add_symbol : sig_state -> sig_symbol -> sig_state =
```
fblanqui, line 148: Why not moving the type sig_symbol to sig_state.ml and redefine Sig_state.add_symbol so that it takes the type `: sig_state -> sig_symbol -> sig_state`?

src/core/handle.ml
```diff
@@ -133,6 +134,89 @@ let handle_require_as : popt -> sig_state -> Path.t -> ident -> sig_state =
...
+  fun ss {expo;prop;mstrat;x;a;impl;d} ->
+  Sig_state.add_symbol ss expo prop mstrat x a impl d
+
+(** [data_proof] returns the datas needed for the symbol or definition
+   [sig_symbol] in the signature and the [goals] we wish to prove with
```
fblanqui, line 153: English: 'to be added" in the signature?

src/core/handle.ml
```diff
@@ -133,6 +134,89 @@ let handle_require_as : popt -> sig_state -> Path.t -> ident -> sig_state =
...
+   the proof script [ts pe]. [pdata_expo] set the authorized exposition
```
fblanqui, line 154: typo: set+s

src/core/handle.ml
```diff
@@ -133,6 +134,89 @@ let handle_require_as : popt -> sig_state -> Path.t -> ident -> sig_state =
...
+   of the symbols used in the proof script : Public (= only public
+   symbols) or Privat (= public and private symbols) *)
+let data_proof sig_symbol cmd pdata_expo ts pe goals =
```
fblanqui, line 157: please follow the coding style by giving the types first

src/core/menhir_parser.mly
```diff
@@ -250,7 +250,10 @@ line:
         | Some(_) -> ms
         | None -> make_pos Lexing.(dummy_pos, dummy_pos) (P_prop(Const)) :: ms
       in
-      make_pos $loc (P_symbol(ms, make_pos $loc(s) s, ps, a))
+      let t =
+        P_symbol(ms, make_pos $loc(s) s, ps, a, None)
```
fblanqui, line 254: coding style: can't this let and the following fit on one line?

src/core/handle.ml
```diff
@@ -133,6 +134,89 @@ let handle_require_as : popt -> sig_state -> Path.t -> ident -> sig_state =
...
+  let x = sig_symbol.x in
+  let a = sig_symbol.a in
+  let d = sig_symbol.d in
+  let pos = x.pos in
+  (* Initialize proof state and save configuration data. *)
+  let st = Proof.init x a in
+  let st = {st with proof_goals = goals} in
+  Console.push_state ();
+  (* Build proof checking data. *)
+  let finalize ss st =
```
fblanqui, line 167: Thanks to recall the type of arguments here: `sig_state -> Proof.t -> sig_state`. `st` is a confusing name here. Thanks to rename it into `ps` for `proof_state`.

src/core/handle.ml
```diff
@@ -196,116 +281,75 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+        | None,Some p_t ->
+            let refine = Pos.make cmd.pos (P_tac_refine(p_t)) in
+            let ts = [refine] in
+            let pe = Pos.make cmd.pos P_proof_end in
+            (ts,pe)
+        | Some(ts,pe),None -> (ts,pe)
+        | Some _, Some _ ->
+            fatal cmd.pos "Proof script and proof_term definition !?"
+        | None, None  ->
+            fatal cmd.pos "NO Proof script and NO proof_term definition !?"
```
fblanqui, line 329: Why these fatal errors? If there is a definition, then the first tactic must be refine. Perhaps, we could instead do something like:
```
let (ts,pe) =
  let (ts,pe) =
    match ts_pe with
    | None -> ([], Pos.make cmd.pos P_proof_end)
    | Some(ts,pe) -> (ts,pe)
  in
  match p_t with
  | Some p_t ->
    let refine = Pos.make cmd.pos (P_tac_refine(p_t)) in
    (refine::ts, pe)
  | None -> (ts,pe)
```

src/core/handle.ml
Rehan-MALAK: fixed in 7e37bcb

src/core/proof.ml
```diff
@@ -9,45 +9,77 @@ open Print
...
+      | Typ of goal_typ (* The usual proof type goal. *)
+      | Unif of constr (* Two terms we'd like equal in ctxt. *)
+
+    let is_typ  = function Typ _ -> true  | Unif _ -> false
+    let is_unif = function Typ _ -> false | Unif _ -> true
+    let typ = function x -> (Typ x)
+    let unif = function x -> (Unif x)
+    let goal_typ_of = function
+      | Typ  gt -> gt
+      | _ -> Console.fatal None "Unif goal instead of a type goal"
```
fblanqui, line 62: I think that fatal is used for user errors. In case of internal error, we should use assert false or invalid_arg for instance here.

fblanqui: Idem below.

src/core/proof.ml
```diff
@@ -60,45 +92,125 @@ type proof_state =
...
+    match typ, ter with
+    | Some(typ),Some(ter) ->
+      let sort, to_solve2 = Infer.infer [] typ in
+      let to_solve =
+        match sort with
+        | Type | Kind -> Infer.check [] ter typ
+        | _ -> Console.fatal pos "%a has type %a (not a sort)."
+                 pp_term typ pp_term sort
+      in
+      typ, sort, to_solve2 @ to_solve
```
fblanqui, line 123: For better readibility, thanks to rename to_solve2 into to_solve1, and to_solve into to_solve2.

fblanqui: Idem below.

src/core/proof.ml
```diff
@@ -60,45 +92,125 @@ type proof_state =
...
+      let sort, to_solve2 =
+        begin
+          match typ with
+          | Kind -> Console.fatal pos "Forbidded definition x := _ -> TYPE"
+          | _ -> let sort, to_solve2 = Infer.infer [] typ in
+            begin
+              match sort with
+              | Type | Kind -> sort,to_solve2
+              | _ -> Console.fatal pos "The term [%a] does not have type \
+                                        [%a]." pp_term ter pp_term typ
```
fblanqui, line 135: Here, the error message should be that the type of `ter` is not typable by a sort.

src/core/handle.ml
Rehan-MALAK: fixed in dfb478e

src/core/handle.ml
Rehan-MALAK: fixed in 7b8c56a

src/core/handle.ml
Rehan-MALAK: fixed in 7b8c56a

src/core/handle.ml
Rehan-MALAK: fixed in 7b8c56a

src/core/handle.ml
Rehan-MALAK: We discussed last time that the (proof + term definition) case should be removed (because we can have a term definition + a unification tactic)
But why not filtering this kind of definition without proof nor term definition
```
definition a : TYPE
```
(with your changes.... this will generate a goal and generate ultimately a ``The proof is not finished`` error)

fblanqui: HI Rehan. What is the problem in this case? The user wants to define a type a but provides no definition: this is an error.

Rehan-MALAK: yes it's an error... so are you ok with this ?

```ocaml
      if ts_pe = None && p_t = None then
        fatal cmd.pos "no proof script nor term definition";
      let (ts,pe) =
        let (ts,pe) =
          match ts_pe with
          | None -> ([], Pos.make cmd.pos P_proof_end)
          | Some(ts,pe) -> (ts,pe)
        in
        match p_t with
        | Some p_t ->
          let refine = Pos.make cmd.pos (P_tac_refine(p_t)) in
          (refine::ts, pe)
        | None -> (ts,pe)
      in
```

src/core/menhir_parser.mly
Rehan-MALAK: fixed in 84e565a4

src/core/handle.ml
Rehan-MALAK: fixed in f2a4219b

src/core/proof.ml
Rehan-MALAK: 0df87258 : 

```ocaml
failwith "Internal error : not a type goal"
...
failwith "Internal error : not a type goal"
```

src/core/proof.ml
Rehan-MALAK: fixed in 0df8725

src/core/proof.ml
Rehan-MALAK: fixed in 0df87258

src/core/handle.ml
fblanqui: Yes, in the case of a definition.

src/core/handle.ml
```diff
@@ -38,15 +38,16 @@ let _ =
...
+    is exited (i.e., when a terminator like “end” is used).  Note that tactics
     do not work on this structure directly,  although they act on the contents
     of its [pdata_p_state] field. *)
 type proof_data =
   { pdata_stmt_pos : Pos.popt (* Position of the proof's statement.  *)
   ; pdata_p_state  : Proof.t  (* Initial proof state for the proof.  *)
   ; pdata_tactics  : p_tactic list (* Tactics for the proof.         *)
   ; pdata_finalize : sig_state -> Proof.t -> sig_state (* Finalizer. *)
-  ; pdata_term_pos : Pos.popt (* Position of the proof's terminator. *) }
+  ; pdata_term_pos : Pos.popt (* Position of the proof's terminator. *)
```
fblanqui, line 49: If this is indeed the position of the proof end, shouldn't we give it a better name like pdata_end_pos?

src/core/handle.ml
```diff
@@ -127,12 +128,85 @@ let handle_modifiers : p_modifier loc list -> (prop * expo * match_strat) =
...
+   or Privat (= public and private symbols) *)
+let data_proof : sig_symbol -> p_command -> expo -> p_tactic list ->
+  p_proof_end loc -> Proof.Goal.t list -> proof_data =
+  fun sig_symbol cmd pdata_expo ts pe goals ->
+  let ident = sig_symbol.ident in
+  let typ   = sig_symbol.typ   in
+  let def   = sig_symbol.def   in
+  let pos = ident.pos in
+  (* Initialize proof state and save configuration data. *)
+  let ps = Proof.init ident typ in
```
fblanqui, line 150: It seems to be the only place where Proof.init is used. Shouldn't we unfold it here and remove Proof.init instead?

src/core/handle.ml
Rehan-MALAK: fixed in 6ab82ed

src/core/handle.ml
Rehan-MALAK: fixed in 8ba333b

fblanqui:
In the doc, the sections on definition and theorem should be removed, and the section on symbol should be updated.

fblanqui:
The keywords theorem and definition should be removed from the grammar (bnf) and in the Emacs and VSCode files as well.

src/core/pretty.ml
```diff
@@ -219,30 +222,48 @@ let pp_command : p_command pp = fun oc cmd ->
...
+      match (t,ts_pe) with
+      | (Some _,_) | (_,Some _) ->
+        let s,args,ao = st.elt in
+        let is_opaq {elt; _} =
+          match elt with
+          | P_opaq(Nonopaque) -> true
+          | _ -> false
+        in
+        let def_or_theo =
+          if List.exists is_opaq ms then "theorem" else "definition"
```
fblanqui, line 236: Shouldn't this part of the code be fixed?

src/core/handle.ml
```diff
@@ -127,12 +128,87 @@ let handle_modifiers : p_modifier loc list -> (prop * expo * match_strat) =
...
+let data_proof : sig_symbol -> p_command -> expo -> p_tactic list ->
+  p_proof_end loc -> Proof.Goal.t list -> proof_data =
+  fun sig_symbol cmd pdata_expo ts pe goals ->
+  let ident = sig_symbol.ident in
+  let typ   = sig_symbol.typ   in
+  let def   = sig_symbol.def   in
+  let pos = ident.pos in
+  (* Initialize proof state and save configuration data. *)
+  let proof_term = fresh_meta ~name:ident.elt typ 0 in
+  let ps = Proof.{proof_name = ident ; proof_term ; proof_goals = goals} in
```
fblanqui, line 151: It seems that, now, proof_term is meaningless in the case of a symbol declaration with no definition and thus goals only due to typing constraints.

fblanqui: In this case, the proof_state itself is questionable in this case:
```
(** Representation of the proof state of a theorem. *)
type proof_state =
  { proof_name     : Pos.strloc  (** Name of the theorem.                 *)
  ; proof_term     : meta        (** Metavariable holding the proof term. *)
  ; proof_goals    : Goal.t list (** Open goals (focused goal is first).  *) }
```
In the previously mentioned case, proof_name and proof_term are meaningless. It's fine but should be documented. Instead, we could use option types.

src/core/handle.ml
```diff
@@ -144,38 +220,117 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+          match p_t,ts_pe with
+          | Some p_t,None ->
+            [Pos.make cmd.pos (P_tac_refine(p_t))], p_end
+          | Some p_t,Some(ts,pe) ->
+            Pos.make cmd.pos (P_tac_refine(p_t))::ts, pe
+          | None,Some(ts,pe) -> ts,pe
+          | None,None -> [],p_end
+        end
+      in
+      let goals,sig_symbol,pdata_expo =
```
fblanqui, line 289: A problem here is that many goals generated during scoping are missing. A simple solution would be to get them all by calling (Basics.get_metas true) on the type and the definition. A better one (that could be implemented later) would be that scoping add in some set the metas it generates on the fly.

fblanqui:
Hi @Rehan-MALAK . This PR is not up to date. Let me know when it is ready to be reviewed.

s1
```diff
@@ -0,0 +1 @@
+br @ Core__Basics 89
```
fblanqui, line 1: What is this file for? I guess it has been committed by mistake, right?

src/core/eval.ml
```diff
@@ -428,3 +428,6 @@ let eq_constr : constr -> constr -> bool = fun (ctx1,t1,u1) (ctx2,t2,u2) ->
   let u2,_ = Ctxt.to_abst ctx2 u2 in
   (eq_modulo [] t1 t2) && (eq_modulo [] u1 u2) ||
   (eq_modulo [] t1 u2) && (eq_modulo [] t2 u1)
+
+let compare_constr : constr -> constr -> int = fun c1 c2 ->
+  if eq_constr c1 c2 then 0 else 1
```
fblanqui, line 433: Is it possible to just do this? After the OCaml doc, comparison functions must satisfy stronger properties:
Following http://caml.inria.fr/pub/docs/manual-ocaml/libref/List.html:
```
val sort_uniq : ('a -> 'a -> int) -> 'a list -> 'a list
Same as List.sort, but also remove duplicates.
see Array.sort for a complete specification
```
and http://caml.inria.fr/pub/docs/manual-ocaml/libref/Array.html:
```
The following must be true for all x, y, z in a:
    cmp x y > 0 if and only if cmp y x < 0
```
Hence, for sort_uniq to work properly, it seems that we should replace 1 by `Stdlib.compare c1 c2`.
Anyway, even if it works with the current definition, it is kind of dangerous to let such a definition without a comment.
For the record, the code of Stdlib.sort_uniq can be seen on https://github.com/ocaml/ocaml/blob/trunk/stdlib/list.ml .

src/core/proof.ml
```diff
@@ -4,103 +4,209 @@ open! Lplib
 open Lplib.Base
 
 open Timed
-open Pos
 open Terms
 open Print
 
 (** Abstract representation of a goal. *)
 module Goal :
   sig
```
fblanqui, line 12: Is the sig really necessary here? Do you need to hide some elements?

src/core/proof.ml
```diff
@@ -4,103 +4,209 @@ open! Lplib
...
-    let compare : t -> t -> int = fun g1 g2 ->
+    let compare : goal_typ -> goal_typ -> int = fun g1 g2 ->
       Meta.compare g1.goal_meta g2.goal_meta
   end
 
 (** Representation of the proof state of a theorem. *)
 type proof_state =
   { proof_name     : Pos.strloc  (** Name of the theorem.                 *)
-  ; proof_term     : meta        (** Metavariable holding the proof term. *)
+  ; proof_term     : meta option (** Metavariable holding the proof term. *)
```
fblanqui, line 90: Perhaps a comment explaining why the proof_term is optional here would be helpful here.

src/core/proof.ml
```diff
@@ -4,103 +4,209 @@ open! Lplib
...
-    [builtins] may be used by tactics, and have been declared. *)
-let init : Pos.strloc -> term -> t = fun name a ->
-  let proof_term = Meta.fresh ~name:name.elt a 0 in
-  let proof_goals = [Goal.of_meta proof_term] in
-  {proof_name = name; proof_term; proof_goals}
+(** [goal_of_meta m] returns a goal associated to the meta m *)
+let goal_of_meta : meta -> Goal.t = fun m ->
+  Goal.Typ (Goal.goal_typ_of_meta m)
+
+(** [goals_of_metas m] returns the goals associated to the metas in a *)
```
fblanqui, line 100: typo: m -> a

fblanqui: Perhaps it would be better to rename goals_of_metas into goals_of_term to avoid confusion with goal_of_meta.

src/core/proof.ml
```diff
@@ -4,103 +4,209 @@ open! Lplib
...
+let goals_of_metas : term -> Goal.t list = fun a ->
+  let goals = ref [] in
+  let add_goal g = goals :=  g::!goals in
+  let add_goal_of_meta m = add_goal (goal_of_meta m) in
+  Basics.iter_meta true add_goal_of_meta a;
+  !goals
+
+(** [goals_of_typ typ ter] returns a list of goals corresponding to the
+    typability of [typ] by a sort and checking eventually that term
+    [ter] has type [typ] *)
```
fblanqui, line 110: Thanks to add that ter and typ should not be both equal to None.

src/core/syntax.ml
```diff
@@ -241,13 +243,18 @@ type p_config =
...
+type p_statement = (ident * p_arg list * p_type option) loc
 
 (** Parser-level representation of modifiers. *)
 type p_modifier =
   | P_mstrat of Terms.match_strat
   | P_expo of Terms.expo
   | P_prop of Terms.prop
+  | P_opaq of Terms.opacity
+
+type proof_meaning =
```
fblanqui, line 255: Thanks to add some comment for explaining the meaning of each constructor.

src/core/tactics.ml
```diff
@@ -8,65 +8,127 @@ open Pos
...
 open Proof
 open Print
+open Timed
 
 (** Logging function for tactics. *)
 let log_tact = new_logger 't' "tact" "tactics"
 let log_tact = log_tact.logger
 
+(** [solve ps pos] calls the default solve algorithm on the unification
+    goals of the proof state [ps] and does not fail *)
```
fblanqui, line 18: You say that this function does not fail but it calls fatal when constraints are unsolvable. Do you want to say that it fails only if some constraint is unsolvable?

src/core/terms.ml
```diff
@@ -38,6 +38,13 @@ type match_strat =
     (** Any rule that filters a term can be applied (even if a rule defined
         earlier filters the term as well). This is the default. *)
 
+(** A symbol is Opaque if its term definition (explicitly given or
+    incrementally builded with a proof-script) should not be added in the
```
fblanqui, line 42: typo: builded -> built

src/core/tactics.ml
```diff
@@ -8,65 +8,127 @@ open Pos
...
     (* Check that [t] is well-typed. *)
     log_tact "proving %a" pp_typing (Env.to_ctxt env, t, a);
-    if not (check t a) then fatal tac.pos "Ill-typed refinement.";
+    let to_solve = Infer.check (Env.to_ctxt env) t a in
+    let gs_unif = List.map Goal.unif to_solve in
+(*
+   TODO : use the new Unif.instantiate ?
+*)
     (* Instantiation. *)
     Meta.set m (Bindlib.unbox (Bindlib.bind_mvar (Env.vars env) (lift t)));
```
fblanqui, line 114: Yes, why not using Unif.instantiate ?

src/core/handle.ml
```diff
@@ -305,66 +336,125 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
-                       in theorems.";
-      (* Scoping the type (statement) of the theorem. *)
-      let a = scope_basic expo a in
-      (* Check that [a] is typable and that its type is a sort. *)
-      Typing.sort_type [] a;
-      (* We check that no metavariable remains in [a]. *)
-      if Basics.has_metas true a then
+  | P_symbol(ms, st, t, ts_pe, e) ->
+    let x,xs,ao = st.elt in
+    (*   : U  := V begin..end *)
```
fblanqui, line 341: What is this comment for? What are U and V?

src/core/handle.ml
```diff
@@ -305,66 +336,125 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+            Some a, metas_a
+          | None -> None, MetaSet.empty
+        end
       in
-      let data =
-        { pdata_stmt_pos = stmt.pos ; pdata_p_state = st ; pdata_tactics = ts
-        ; pdata_finalize = finalize ; pdata_term_pos = pe.pos }
+      (* If a type [ao = Some a] is given, then we check that it is
+         typable by a sort and that [t] has type [a]. Otherwise, we
+             try to infer thetype of [t]. Unification goals are collected *)
```
fblanqui, line 392: typo: thetype -> the type

src/core/handle.ml
```diff
@@ -305,66 +336,125 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+          | Some p_t,None ->
+            [Pos.make cmd.pos (P_tac_refine(p_t))], p_end
+          | Some p_t,Some(ts,pe) ->
+            Pos.make cmd.pos (P_tac_refine(p_t))::ts, pe
+          | None,Some(ts,pe) -> ts,pe
+          | None,None -> [],p_end
+        end
       in
-      (ss, Some(data))
+      let metas_a_t = MetaSet.union metas_a MetaSet.empty in
```
fblanqui, line 407: Why making an union with an empty set?

src/core/handle.ml
```diff
@@ -305,66 +336,125 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+      let typ_goals_a = List.map Proof.Goal.typ typ_goals_a in
+      let proof_term,goals,sig_symbol,pdata_expo =
+        let sort_goals, a = Proof.goals_of_typ x.pos ao t in
+        (* And the main "type" goal *)
+        let proof_term,typ_goal =
+          match e with
+          | Def ->
+            let proof_term = Meta.fresh ~name:x.elt a 0 in
+            let proof_goal = Proof.goal_of_meta proof_term in
+            Some proof_term,[proof_goal] @ typ_goals_a
```
fblanqui, line 421: can be simplified to: ` proof_goal :: typ_goals_a`

src/core/handle.ml
```diff
@@ -305,66 +336,125 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
+                           be used in theorems."
+(*
+                |     _, true , None   , _    , _      ->
+                  fatal cmd.pos "Theorem should have an explicit type !"
+*)
+          (* Definition *)
+(*
+                |     _, false, _      , Const, _      ->
+                  fatal cmd.pos "A definition cannot be a constant."
+*)
```
fblanqui, line 447: These comments can be removed, right?

Rehan-MALAK: fixed

src/core/handle.ml
Rehan-MALAK: fixed

src/core/handle.ml
Rehan-MALAK: fixed

src/core/handle.ml
Rehan-MALAK: fixed

src/core/handle.ml
Rehan-MALAK: removed

src/core/terms.ml
Rehan-MALAK: fixed

src/core/tactics.ml
Rehan-MALAK: yes
fixed

src/core/syntax.ml
Rehan-MALAK: done

src/core/proof.ml
Rehan-MALAK: fixed

src/core/proof.ml
Rehan-MALAK: this function was unused and has been removed

src/core/proof.ml
Rehan-MALAK: fixed

src/core/proof.ml
Rehan-MALAK: I don't know why there was a signature here... As we don't use mli files, everything (but overwritten ocaml symbols) is public. I suppose it's more coherent to remove this signature... I don't have any particular opinion on this. I removed it...

src/core/handle.ml
Rehan-MALAK: yes, this can be done in a separate PR later

s1
Rehan-MALAK: removed

src/core/pretty.ml
```diff
@@ -221,37 +224,47 @@ let pp_command : p_command pp = fun oc cmd ->
       out "require %a as %a" (pp_path cmd.pos) p (pp_path_elt i.pos) i.elt
   | P_open(ps)                      ->
       List.iter (out "open %a" (pp_path cmd.pos)) ps
-  | P_symbol(ms,s,args,a) ->
-      out "@[<hov 2>%asymbol %a"
-        (Format.pp_print_list pp_modifier) ms pp_ident s;
-      List.iter (out " %a" pp_p_arg) args;
-      out " :@ @[<hov>%a@]" pp_p_term a
+  | P_symbol(ms,st,t,ts_pe,_e) (* TODO *) ->
```
fblanqui, line 227: Why this TODO comment? To be removed?

fblanqui:
Hi @gabrielhdt . Do you have any remark concerning the code for Emacs?
@Rehan-MALAK: did you test both the VSCode and the Emacs interfaces?

src/core/eval.ml
fblanqui: For the moment, I propose you to just add a comment with a WARNING saying that this function does NOT satisfy the requirements mentioned above.

gabrielhdt:
> Hi @gabrielhdt . Do you have any remark concerning the code for Emacs?
> @Rehan-MALAK: did you test both the VSCode and the Emacs interfaces?

It looks good to me.

src/core/tactics.ml
```diff
@@ -8,65 +8,128 @@ open Pos
...
 
-  let scope = Scope.scope_term Public ss env in
-  let infer t = Typing.infer (Env.to_ctxt env) t in
-  let check t a = Typing.check (Env.to_ctxt env) t a in
+  (* Obtaining the goal environment and type. *)
+  let scope t =
+    let (env, _a) = match g with
+      | Some gt -> Goal.get_type gt
+      | None -> assert false
+    in
```
fblanqui, line 93: It seems to me that the computation of env, which does not depend on t, should be factorized out of the scope function. This would simplify many calls to scope where env is not used.

Rehan-MALAK:
> Hi @gabrielhdt . Do you have any remark concerning the code for Emacs?
> @Rehan-MALAK: did you test both the VSCode and the Emacs interfaces?

Nope... this has been tested only on emacs

src/core/eval.ml
Rehan-MALAK: fixed (we rely on Stdlib.compare for the non equal case)

src/core/pretty.ml
Rehan-MALAK: fixed (removed)

fblanqui:
@ejgallego @francois-lefoulon This PR introduces a new kind of goals, unification goals. The type of goals has been modified: a goal is either a typing goal like before or a unification goal. The Emacs code has been modified accordingly but not the code for VSCode. Could one of you tell us the modifications to do in the code for VSCode?

ejgallego:
Types sent by the lsp client should match those in TypeScript, for example:
```
export interface Goal {
	gid :  number // goal id
	hyps : Env[] // hypothesis
	type : String // goals
}
```

fblanqui:
The code of the LSP server has been modified to include unification goals, and the client code has been modified on the Emacs side. The client on the VSCode side needs now to be modified as well. Note that a typing goal is something like E |- ? : A while a unification goal is something like E |- t = u.

ejgallego:
Indeed, so if I understand correctly you need to adapt the type declarations that I pointed out, didn't see the concrete commit updating `Lsp_base.json_of_goals` (maybe not in master yet)

fblanqui:
Yes, it is not in master yet but in this PR. But I can't merge this PR without fixing the VSCode client. Otherwise it will be broken.

fblanqui:
I don't know TS and Rehan doesn't have much time this week. Help needed! ;-)

src/pure/pure.mli
```diff
@@ -62,9 +63,10 @@ val initial_state : file_path -> state
     [Cmd_Error] constuctor). *)
 val handle_command : state -> Command.t -> command_result
 
-(** [handle_tactic st tac] evaluated the tactic [tac] in state [st], returning
-    a new proof state (with [Tac_OK]) or an error (with [Tac_Error]). *)
-val handle_tactic : proof_state -> Tactic.t -> tactic_result
+(** [handle_tactic st tac] evaluates the tactic [tac] in state [st], for a
+    symbol or a definition with exposition [expo] returning a new proof state
+    (with [Tac_OK]) or an error (with [Tac_Error]). *)
+val handle_tactic : proof_state -> Terms.expo -> Tactic.t -> tactic_result
```
ejgallego, line 69: Documentation needs update here, what is `Terms.expo`?

ejgallego: I mean the beggining of the doc comment sorry, sorry, the expo parameter is missing.

src/lsp/lp_doc.ml
```diff
@@ -51,19 +51,19 @@ let option_default o1 d =
 let mk_error ~doc pos msg =
   LSP.mk_diagnostics ~uri:doc.uri ~version:doc.version [pos, 1, msg, None]
 
-let process_pstep (pstate, diags) tac =
+let process_pstep (pstate,expo,diags) tac =
```
ejgallego, line 54: This diff suggests that expo should be part of the proof state, IIUC.

src/pure/pure.mli
ejgallego: Yeah it seems that expo should be part of the state.

src/lsp/lsp_base.ml
```diff
@@ -41,15 +41,24 @@ let mk_range (p : Pos.pos) : J.t =
...
+    `Assoc [
+      "typeofgoal", `String "Typ "
+    ; "gid", `Int g_meta.meta_key
+    ; "hyps", `List j_env
+    ; "type", `String (Format.asprintf "%a" Print.pp_term typ)]
+  | Goal.Unif (ctx,t1,t2) ->
+    let constr = Format.asprintf "%a" Print.pp_constr (ctx,t1,t2) in
+    `Assoc [
+      "typeofgoal", `String "Unif"
+    ; "constr", `String constr]
```
ejgallego, line 61: This needs update:
``` 
export interface Goal {
	gid :  number // goal id
	hyps : Env[] // hypothesis
	type : String // goals
}
```
I'm not an expert on TS, but should be easy. I won't do it this week tho as I'm pretty busy with CUDW.

src/core/tactics.ml
fblanqui: Fixed in https://github.com/Deducteam/lambdapi/pull/422/commits/a93d56890b92d4a60a7b2b0f6a9281c9a635fb91

src/lsp/lp_doc.ml
fblanqui: Fixed in https://github.com/Deducteam/lambdapi/pull/422/commits/06a69ca0c614573effeb0eeaa5dbb0938f7ca56a

src/pure/pure.mli
fblanqui: Fixed in https://github.com/Deducteam/lambdapi/pull/422/commits/06a69ca0c614573effeb0eeaa5dbb0938f7ca56a

fblanqui:
Many thanks to @Rehan-MALAK for your work. And thanks to @firewall2142 as well for fixing the VSCode client code.

Rehan-MALAK:
> Many thanks to @Rehan-MALAK for your work. And thanks to @firewall2142 as well for fixing the VSCode client code.

Thanks @fblanqui and @firewall2142 for all your corrections !

PR #423, opened 2020-06-19
Syntax of patterns

gabrielhdt:
Detail the syntax of patterns in the BNF, plus a correction in `terms.md`.

### DIFF
--- a/doc/commands.md
+++ b/doc/commands.md
@@ -128,48 +128,50 @@ Rewriting rules for definable symbols are declared using the `rule` command.
 ```
 rule add zero      $n ↪ $n
 rule add (succ $n) $m ↪ succ (add $n $m)
+rule mul zero      _  ↪ zero
 ```
 
-Note that rewriting rules can also be defined simultaneously, using the `with`
-keyword instead of the `rule` keyword for all but the first rule.
-
-```
-rule add zero      $n ↪ $n
-with add (succ $n) $m ↪ succ (add $n $m)
-```
-
-Pattern variables need to be prefixed by `$`.
+Terms prefixed by the sigil `$` and `_` are pattern variables.
 
 **Higher-order pattern-matching**.
-Lambdapi allows higher-order pattern-matching (not in full generality
-though):
+Lambdapi allows higher-order pattern-matching on patterns à la Miller but modulo
+β-equivalence only (and not βη).
 
 ```
 rule diff (λx, sin $F[x]) ↪ λx, diff (λx, $F[x]) x × cos $F[x]
+```
+
+Patterns can contain abstractions `λx, _` and the user may attach an environment
+made of *distinct* bound variables to a pattern variable to indicate which bound
+variable can occur in the matched term. The environment is a semicolon-separated
+list of variables enclosed in square brackets `[x;y;...]`. For instance, a term
+of the form `λx y,t` matches the pattern `λx y,$F[x]` only if `y` does not
+freely occur in `t`.
+
+```
 rule lam (λx, app $F[] x) ↪ $F // η-reduction
 ```
 
 Hence, the rule `lam (λx, app $F[] x) ↪ $F` implements η-reduction
 since no valid instance of `$F` can contain `x`.
 
-Pattern variables can be applied to distinct bound variables only,
-that is, the terms between `[` and `]` must be distinct bound
-variables only.
+Pattern variables cannot appear at the head of an application: `$F[] x` is not
+allowed. The converse `x $F[]` is.
 
-`$P` without square brackets is a short-hand for `$P[]`. This
-short-hand cannot be used though when `$P` occurs in a rule left-hand
-side below a binder like in the rule η above.
+A pattern variable `$P[]` can be shortened to `$P` when there is no ambiguity,
+i.e. when the variable is not under a binder (unlike in the rule η above).
 
 It is possible to define an unnamed pattern variable with the syntax
-`$_[x,y]`.
+`$_[x;y]`.
 
-If `x` and `y` are the only variables in scope, then `_` is equivalent
-to `$_[x,y]`.
+The unnamed pattern variable `_` is always the most general:
+if `x` and `y` are the only variables in scope, then `_` is equivalent
+to `$_[x;y]`.
 
-In rule left-hand sides, λ-expressions must have no type annotations.
+In rule left-hand sides, λ-expressions cannot have type annotations.
 
 **Important**. In contrast to languages like OCaml, Coq, Agda, etc. rule
- left-hand sides can contain defined symbols:
+left-hand sides can contain defined symbols:
 
 ```
 rule add (add x y) z ↪ add x (add y z)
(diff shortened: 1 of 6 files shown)


fblanqui:
Why not adding some more comments in the doc of the `rule` command? The sentence "Lambdapi allows higher-order pattern-matching (not in full generality though)" could be modified. We could say that we accept any pattern à la Miller but matching is modulo beta only that is, not modulo beta and eta. It would be good to add a few tests for patterns with bound variables applied to something.

gabrielhdt:
Yes, it's a good idea. Meanwhile, I've written an example file which explains all the patterns, which raised a runtime failure during the build process of decision trees.

fblanqui:
Let me know when you want me to merge.

gabrielhdt:
> Let me know when you want me to merge.

I have updated the documentation: you can merge if you're happy with the new one.

doc/commands.md
```diff
@@ -128,48 +128,50 @@ Rewriting rules for definable symbols are declared using the `rule` command.
 ```
 rule add zero      $n ↪ $n
 rule add (succ $n) $m ↪ succ (add $n $m)
+rule mul zero      _  ↪ zero
 ```
 
-Note that rewriting rules can also be defined simultaneously, using the `with`
-keyword instead of the `rule` keyword for all but the first rule.
+Terms prefixed by the sigil `$` and `_` are pattern variables.
```
fblanqui, line 134: What is "sigil"?

doc/commands.md
```diff
@@ -128,48 +128,50 @@ Rewriting rules for definable symbols are declared using the `rule` command.
...
+Matching a ground term against a pattern variable always succeed.
```
fblanqui, line 135: Why adding this sentence?

doc/commands.md
gabrielhdt: Ah, it's symbols used to mark variables in general such as $ in shell, or @ for arrays in Perl &c. <https://en.wikipedia.org/wiki/Sigil_(computer_programming)>

doc/commands.md
gabrielhdt: To give the semantics of the pattern variables

doc/commands.md
```diff
@@ -188,6 +190,16 @@ And they can be non-linear:
...
 
+Note that rewriting rules can also be defined simultaneously, using the `with`
+keyword instead of the `rule` keyword for all but the first rule.
+
+```
+rule add zero      $n ↪ $n
+with add (succ $n) $m ↪ succ (add $n $m)
+```
+
+Considering set of rules allows to maintain confluence while adding rules.
```
fblanqui, line 201: Adding sets of rules allows to maintain confluence.

doc/commands.md
fblanqui: I think it may be confusing. I prefer to remove it. Why talking of closed terms? Even in higher order matching it's relative to a local context. Indeed, consider the following example:
```

symbol A:TYPE
symbol f:(A→A)→A
rule f(λ_,$a[]) ↪ $a
assert λx,f(λ_,x) ≡ λx,x
```

gabrielhdt: Yes indeed, you're right: either it's not correct or it becomes too complicated. But perhaps it should be documented somewhere, how should people know about those things? Do we have any plan to provide a comprehensive documentation?

fblanqui:
For the record, link to the doc: https://github.com/gabrielhdt/lambdapi/blob/patterns-synt/doc/commands.md

doc/commands.md
```diff
@@ -128,48 +128,49 @@ Rewriting rules for definable symbols are declared using the `rule` command.
...
 rule add (succ $n) $m ↪ succ (add $n $m)
+rule mul zero      _  ↪ zero
 ```
 
-Note that rewriting rules can also be defined simultaneously, using the `with`
-keyword instead of the `rule` keyword for all but the first rule.
+Terms prefixed by the sigil `$` and `_` are pattern variables.
+
+**Higher-order pattern-matching**.
+Lambdapi allows higher-order pattern-matching following Miller’s patterns.
```
fblanqui, line 137: Lambdapi allows higher-order pattern-matching on patterns à la Miller but modulo β-equivalence only (and not βη).

doc/commands.md
```diff
@@ -128,48 +128,49 @@ Rewriting rules for definable symbols are declared using the `rule` command.
...
-
-**Higher-order pattern-matching**.
-Lambdapi allows higher-order pattern-matching (not in full generality
-though):
+Patterns can contain abstractions `λx, _` and the user may
+attach an environment made of *distinct* bound variables to a pattern variable.
+The environment is a semicolon-separated list of variables enclosed in square
+brackets `[x;y;...]`.
+A term containing a bound variable `x` only matches pattern variables whose
+environment contain `x`, such as `$F[x]`.
```
fblanqui, line 148: What do you think of: "Patterns can contain abstractions `λx, _` and the user may
attach an environment made of *distinct* bound variables to a pattern variable, to indicate on what bound variables it is allowed to depend.
The environment is a semicolon-separated list of variables enclosed in square
brackets `[x;y;...]`.
For instance, a term of the form `λx y,t` matches the pattern `λx y,$F[x]` only if `y` does not freely occur in `t`." ?

gabrielhdt: I feel that the `it` in
> to indicate on what bound variables it is allowed to depend

is ambiguous. What about, then

> Patterns can contain abstractions `λx, _` and the user may attach an environment made of distinct bound variables to a pattern variable to indicate which bound variable can occur in the matched term. The environment is a semicolon-separated list of variables enclosed in square brackets `[x;y;...]`. For instance, a term of the form `λx y,t` matches the pattern `λx y,$F[x]` only if y does not freely occur in `t`.

and it seems to me that "freely" in the last line is superfluous; or there is a nuance I don't grasp.

fblanqui: Well, strictly speaking, y can occur as bound variable...

fblanqui: Ok for "which bound variable can occur in the matched term".

gabrielhdt: Ah ok, introduced by another binder, I see I see, fine.

doc/commands.md
```diff
@@ -128,48 +128,50 @@ Rewriting rules for definable symbols are declared using the `rule` command.
...
-Pattern variables need to be prefixed by `$`.
-
-**Higher-order pattern-matching**.
-Lambdapi allows higher-order pattern-matching (not in full generality
-though):
+Patterns can contain abstractions `λx, _` and the user may attach an environment
+made of *distinct* bound variables to a pattern variable to indicate which bound
+variable can occur in the matched term. The environment is a semicolon-separated
+list of variables enclosed in square brackets `[x;y;...]`. For instance, a term
+of the form `λx y,t` matches the pattern `λx y,$F[x]` only if y does not freely
```
fblanqui, line 148: y between backquotes

doc/commands.md
fblanqui: Perhaps you can add my example in your test file?

doc/commands.md
gabrielhdt: oups

doc/commands.md
```diff
@@ -188,6 +190,18 @@ And they can be non-linear:
...
+keyword instead of the `rule` keyword for all but the first rule.
+
+```
+rule add zero      $n ↪ $n
+with add (succ $n) $m ↪ succ (add $n $m)
+```
+
+Adding sets of rules allows to maintain confluence.
+
+Examples of patterns are available in the file `patterns.lp` of the test suite.
```
fblanqui, line 203: Add a link to https://github.com/Deducteam/lambdapi/blob/master/tests/OK/patterns.lp ?

gabrielhdt: OK, it worked.

fblanqui:
Thanks to update for merging.

Issue #416, opened 2020-06-09
build instructions lack requirement of emacs


gares:
```
gares@ollypat:~/work-area/lambdapi$ make
File "editors/emacs/dune", line 6, characters 6-16:
6 |     %{bin:emacs} --batch --quick --funcall batch-byte-compile
          ^^^^^^^^^^
Error: Program emacs not found in the tree or in PATH
 (context: default)
GNUmakefile:10: recipe for target 'bin' failed
make: *** [bin] Error 1
```

Issue #411, opened 2020-05-29
recreate make install_emacs


fblanqui:
@gabrielhdt could you please recreate in Makefile a target install_emacs. make install should not install the emacs mode in case it gets broken. On the other hand, the opam install script should call it. The same should be done for vim. What do you think?

gabrielhdt:
OK but does it really solve the issue? Suppose the lambdapi-mode is broken, you'll notice it only after having installed it, so whether you have a separate install command or not does not matter. Or can you explain your workflow?

fblanqui:
Currently the emacs mode is broken. But to test my local branch, I need to do make install for emacs to use my local LP. In this case, it installs the broken emacs mode as well, and erase the installed emacs mode as well which may not have this problem. An alternative would be to tell emacs to use my version of LP and not the installed one.

fblanqui:
I have the version 20151215.1043 installed.

gabrielhdt:
Choosing the version of LP to use in emacs seems to be a better idea to me, but it does not change the fact that once you have a broken emacs mode, there is no way to revert, so you're stuck with the broken mode. And I don't see how having a separate install command helps: you'll remark that it's broken once it's installed and it will already be too late.

Perhaps setting up CI for the emacs mode would prevent this, wdyt?

fblanqui:
Well, if the current mode is broken, I can come back to a commit where it is not and do make install_emacs.

gabrielhdt:
Yes I see, what if we have a `install_bin`, `install_emacs` and `install` calls both?

gabrielhdt:
In fact it is trickier than I thought, dune must install everything, you can't choose if you don't make several packages (determined by `.opam` files). So in fine the best would perhaps be to install the emacs mode through MELPA to control its version (you can select the commit you want). `make install` will still install the broken mode, but your emacs configuration won't load it since it loads the one from melpa. It's not really a solution, but a workaround. But anyway, installing through melpa should be advised since it takes care of the dependencies.

fblanqui:
But how to tell emacs to use your own version of the mode when you want to use it?

gabrielhdt:
What do you mean by your own? The one you're developing?

fblanqui:
When you want to modify and test the emacs mode.

gabrielhdt:
Yes, it's a bit inconvenient. There must be a way with `cask`, but I'm not very familiar with it. For this, I made a `test.sh` that launches an emacs with the new mode loaded.

PR #425, opened 2020-06-19
doc: create a file for each editor

fblanqui:


fblanqui:
Can be visualized on https://github.com/fblanqui/lambdapi/blob/uidoc/doc/DOCUMENTATION.md

fblanqui:
@gabrielhdt About Vim: is it still the case that the user has to do make install_vim when installing LP with opam? See https://github.com/fblanqui/lambdapi/blob/uidoc/doc/vim.md

gabrielhdt:
> @gabrielhdt About Vim: is it still the case that the user has to do make install_vim when installing LP with opam? See https://github.com/fblanqui/lambdapi/blob/uidoc/doc/vim.md

The user has to `make install_vim`, because vim files are not handled by dune. But after #416 and #411, I think that using dune for everything is not a good idea, and I suggest going back to Makefile targets.

fblanqui:
The inconvenient is that you need the sources. Can't we have opam packages "lambdapi-emacs" and "lambdapi-vim" just for the emacs/vim modes?

gabrielhdt:
It would be strange, opam is made to install ocaml packages, not emacs lisp ones... all this would make things confusing, imo.

I don't understand what you mean by "you need the sources", to do what? And when did we not need the sources?

PR #426, opened 2020-06-22
Emacs and vim mode packaging

### LINKED ISSUE #411: recreate make install_emacs

gabrielhdt:
- Use makefile for emacs
- Install vim mode and emacs mode from opam

Fixes #411


fblanqui:
Perhaps you need to update the doc as well.

README.md
```diff
@@ -59,17 +59,18 @@ for more information).
...
 
 **Note:** you can run `lambdapi` without installing with `dune exec -- lambdapi`.
 
-**Note on Emacs:** `make install` installs the `lambdapi-mode`, an Emacs major
-mode for editing `lambdapi` files in `$(opam var share)/emacs/site-lisp`.
-To load the `lambdapi-mode` automatically when editing `*.lp` files, add `(load
-"lambdapi-site-file")` to your `~/.emacs.d/init.el` or `~/.emacs`.
-If `lambdapi-mode` is not activated when editing a file ending in `.lp`, read
-the installation section of the
+**Note on Emacs:** `make install_emacs` installs the `lambdapi-mode`, an Emacs
```
fblanqui, line 68: Instead of having a specific note on Emacs only here, wouldn't it be better to mention that some plugins for several editors are provided, with a link to ui.md?

gabrielhdt: OK, and regarding the documentation, I'd rather centralise information, so I'd like to reduce the content of `ui.md` to a link to `emacs.md` (concerning the emacs part), to avoid spreading information, what do you think?

fblanqui: Isn't already the case (see https://github.com/Deducteam/lambdapi/pull/425)?

gabrielhdt: Ah nice, I was looking at an older version on my computer, sorry.

Issue #331, opened 2020-04-03
Keep rule order as a flag?


gabrielhdt:
Hi,
Would it be sensible to use `keep-rule-order` as a flag (with `set flag ...`) rather than a command line option?

In my case, it would be to create a list membership function, so I want to write
```
symbol Utype: TYPE
symbol type_list: TYPE
symbol cons: ..., symbol nil: ...

set declared "𝔅"
constant symbol 𝔅: TYPE
constant symbol t: 𝔅
constant symbol f: 𝔅

symbol mem: Utype ⇒ type_list ⇒ 𝔅
rule mem &X (cons &X _) → t
rule mem &X (cons _ &L) → mem &X &L
rule mem _ nil                  → f
```

So first, it seems I need to keep rule order, or can I do without? Otherwise, the flag has several advantages:
+ the LSP is informed of the strategy,
+ the strategy can be changed in the file,
+ fewer arguments in command line.

rlepigre:
I think we could do that, let's see what everyone else thinks.

By the way, the rewriting system you are defining is not confluent, so I am wondering if this is the way to go to define the `mem` function. I cannot think of any other way, so maybe this indicates that we need a kind of guard on the second rule to say that `&X` and the wildcard should not be convertible. So something like the following?
```
rule mem  &X (cons &Y &L) when not &X ≡ &Y→ mem &X &L
```
What do you think @fblanqui? Is that necessary?

francoisthire:
> a kind of guard on the second rule to say that &X and the wildcard should not be convertible. 

I don't think this is not decidable and in particular such criterion would not be stable by substitution. The only true way to define a `mem` function is to have an decidable equality on the elements of the container. However, I think that `lambdapi` allows such shortcut is nice for the user.

francoisthire:
There is an easy feature to implement I think is to have a flag and it can be used only once per definable symbol. It fixes many problems that could arise with non-confluent rewrite system.

> the strategy can be changed in the file,

This seems harder to implement and one need to ensure that it is not problematic. Changing the strategy while defining a recursive function could lead maybe to an inconsistency for exemple.


```
rule mem  &X (cons &Y &L) when not &X ≡ &Y → mem &X &L
```

See this PR on Dedukti : https://github.com/Deducteam/Dedukti/pull/74

Raphaël's remark made think that was a bad idea.

fblanqui:
[...]

PR #427, opened 2020-06-24
[WIP] Configuration to enforce rule order

gabrielhdt:
In answer to #331, this pull request allows the user to annotate symbols to change the behaviour of the rewriting engine when it filters terms headed by the symbol.

If the term reduced starts with such an annotated symbol, a rewriting rule defined on this symbol can be applied if and only if the previous rules (in the order of declaration) can not be applied.

A symbol is annotated when it is declared using the `sequential` keyword:
```
sequential symbol mem: ...
```

Remaining to be done
- [x] better variable names (replacing `kro`, find a better name for `keep_rule_order`?)
- [x] documentation
- [x] test files

### DIFF
--- a/doc/commands.md
+++ b/doc/commands.md
@@ -76,30 +76,35 @@ Data types and predicates must be given types of the form
 We recommend to start types and predicates by a capital letter.
 
 **Modifiers:**
- - `constant`: no rule can be added to the symbol
- - `injective`: the symbol can be considered as injective, that is,
- if `f t1 .. tn` ≡ `f u1 .. un`, then `t1`≡`u1`, ..., `tn`≡`un`.
- For the moment, the verification is left to the user.
+- Modifiers for the unification engine,
+  - `constant`: no rule can be added to the symbol
+  - `injective`: the symbol can be considered as injective, that is, if `f t1 ..
+     tn` ≡ `f u1 .. un`, then `t1`≡`u1`, ..., `tn`≡`un`. For the moment, the
+     verification is left to the user.
 
-These modifiers are used to help the unification engine.
+- Exposition markers define how a symbol can be used outside the module where it
+  is defined.
+  - `public` (default): the symbol can be used without restriction
+  - `private`: the symbol cannot be used
+  - `protected`: the symbol can only be used in left-hand side of rewrite rules.
 
-**Exposition markers:**
-Exposition defines how a symbol can be used outside the module where it is
-defined. By default any symbol is _public_, which means it can be used without
-restriction anywhere. There are two exposition markers available:
-- `private`: the symbol cannot be used outside of its module (can be compared to
-  a symbol not in the interface in OCaml);
-- `protected`: the symbol can only be used in left-hand side of rewrite rules
-  outside of its module.
+  Exposition markers obey the following rules: inside a module,
+  - private symbols cannot appear in the type of public symbols;
+  - private symbols cannot appear in the right-hand side of a rewriting rule
+    defining a public symbol
+  - externally defined protected symbols cannot appear at the head of a
+    left-hand side
+  - externally defined protected symbols cannot appear in the right hand side of
+    a rewriting rule
 
-Exposition obeys the following rules: inside a module,
-- private symbols cannot appear in the type of public symbols;
-- private symbols cannot appear in the right-hand side of a rewriting rule
-  defining a public symbol
-- externally defined protected symbols cannot appear at the head of a left-hand
-  side
-- externally defined protected symbols cannot appear in the right hand side of a
-  rewriting rule
+- `sequential`: modifies the pattern matching algorithm. By default, the order
+  of rule declarations is not taken into account. This modifier tells Lambdapi
+  to apply rules defining a sequential symbol in the order they have been
+  declared (note that the order of the rules may depend on the order of the
+  `require` commands). An example can be seen in 
+  [`rule_order.lp`](../tests/OK/rule_order.lp).  
+  *WARNING:* using this modifier can break important properties.
+
 
 **Implicit arguments:** Some function symbol arguments can be declared
 as implicit meaning that they must not be given by the user
--- a/src/core/handle.ml
+++ b/src/core/handle.ml
@@ -102,7 +102,7 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
   | P_open(ps)                  ->
      let ps = List.map (List.map fst) ps in
      (List.fold_left (handle_open cmd.pos) ss ps, None)
-  | P_symbol(e, p, x, xs, a) ->
+  | P_symbol(e, p, st, x, xs, a) ->
       (* We check that [x] is not already used. *)
       if Sign.mem ss.signature x.elt then
         fatal x.pos "Symbol [%s] already exists." x.elt;
@@ -122,7 +122,7 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
         end;
       (* Actually add the symbol to the signature and the state. *)
       out 3 "(symb) %s : %a\n" x.elt pp_term a;
-      (Sig_state.add_symbol ss e p x a impl None, None)
+      (Sig_state.add_symbol ss e p st x a impl None, None)
   | P_rules(rs)                ->
       (* Scoping and checking each rule in turn. *)
       let handle_rule r =
@@ -185,7 +185,7 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
       (* Actually add the symbol to the signature. *)
       out 3 "(symb) %s ≔ %a\n" x.elt pp_term t;
       let d = if op then None else Some(t) in
-      let ss = Sig_state.add_symbol ss e Defin x a impl d in
+      let ss = Sig_state.add_symbol ss e Defin Eager x a impl d in
       (ss, None)
   | P_theorem(e, stmt, ts, pe) ->
       let (x,xs,a) = stmt.elt in
@@ -223,7 +223,7 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
             (* Add a symbol corresponding to the proof, with a warning. *)
             out 3 "(symb) %s (admit)\n" x.elt;
             wrn cmd.pos "Proof admitted.";
-            Sig_state.add_symbol ss e Const x a impl None
+            Sig_state.add_symbol ss e Const Eager x a impl None
         | P_proof_qed   ->
             (* Check that the proof is indeed finished. *)
             if not (Proof.finished st) then
@@ -233,7 +233,7 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
               end;
             (* Add a symbol corresponding to the proof. *)
             out 3 "(symb) %s (qed)\n" x.elt;
-            Sig_state.add_symbol ss e Const x a impl None
+            Sig_state.add_symbol ss e Const Eager x a impl None
       in
       let data =
         { pdata_stmt_pos = stmt.pos ; pdata_p_state = st ; pdata_tactics = ts
--- a/src/core/handle.ml
+++ b/src/core/handle.ml
@@ -73,6 +73,56 @@ let handle_require : bool -> popt -> sig_state -> Path.t -> sig_state =
   ss.signature.sign_deps := PathMap.add p [] !(ss.signature.sign_deps);
   if b then handle_open pos ss p else ss
 
+(** [handle_modifiers ms] verifies that the modifiers in [ms] are compatible.
+    If so, they are returned as a tuple, otherwise, the incompatible modifiers
+    are returned. *)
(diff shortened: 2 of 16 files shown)


src/core/terms.ml
```diff
@@ -99,7 +99,10 @@ type term =
   ; sym_prop  : prop
   (** Property of the symbol. *)
   ; sym_expo  : expo
-  (** The visibility of the symbol. *) }
+  (** The visibility of the symbol. *)
+  ; sym_rord  : bool
```
fblanqui, line 103: Why not bool ref?

src/core/sign.ml
```diff
@@ -335,3 +335,13 @@ let rec dependencies : t -> (Path.t * t) list = fun sign ->
...
+
+(** [keep_rule_order sign id] sets symbol identified by [id] to keep the rule
+    order.
+    TODO enhance documentation *)
+let keep_rule_order: t -> ident -> unit = fun sign id ->
+  match StrMap.find_opt id.elt !(sign.sign_symbols) with
+  | None -> fatal id.pos "Symbol %s not found" id.elt
+  | Some(sym, p) ->
+      let toggled = {sym with sym_rord = true } in
+      sign.sign_symbols := StrMap.add id.elt (toggled, p) !(sign.sign_symbols)
```
fblanqui, line 347: With bool ref, we just need to do `sym.sym_rord := true`.

src/core/tree.ml
```diff
@@ -256,7 +252,9 @@ module CM = struct
     (** Index of next slot to use in [vars] array to store variables. *)
     ; positions : arg list
     (** Positions of the elements of the matrix in the initial term. Note that
-        we rely on the order relation used in sets. *) }
+        we rely on the order relation used in sets. *)
+    ; rule_order: bool
+    (** TODO *) }
```
fblanqui, line 257: Why do you need to change anything here? Isn't it sufficient that a boolean argument is added to compile (and propagated wherever it is necessary)?

gabrielhdt: This is precisely the propagation: to know that we are building the tree of a symbol which keeps rule order.

gabrielhdt: Ah I see what you mean, I'll have a look, but it looks like your idea is better.

gabrielhdt:
Shouldn't we require that if the boolean is set, no rewriting rule may be defined before it is set? Otherwise we could have some declarations type checked without keeping rule order and other typechecked keeping rule order, which seems a bit strange.

fblanqui:
Ah good point. Then, it means that it would be better to replace the set command by a flag attached to the declaration of the symbol itself like injective, for instance `keep_rule_order symbol f:A`. What do you think?

gabrielhdt:
> Ah good point. Then, it means that it would be better to replace the set command by a flag attached to the declaration of the symbol itself like injective, for instance `keep_rule_order symbol f:A`. What do you think?

Lambdapi is getting closer and closer to java... but yes, it seems to be a possibility. But `keep_rule_order` seems way too long doesn't it? What about `topmost` as a keyword,
`injective protected topmost symbol f: ...`

fblanqui:
ok for topmost. I don't know Java very much but, if you have other ideas about the syntax of symbol properties, we can discuss them :-)

gabrielhdt:
Or `prioritised`, to be consistent with the simple past of `protected`?

gabrielhdt:
Francois proposed `sequential`, with which I'm happy

fblanqui:
ok for sequential

doc/commands.md
```diff
@@ -101,6 +101,30 @@ Exposition obeys the following rules: inside a module,
...
+Consider the following rules that compute the membership of an element in a list
+```
+symbol mem: E → List → Bool
+rule mem $e (cons $e  _) ↪ true
+rule mem $e (cons _ $tl) ↪ mem $e $tl
+rule mem _ nil ↪ false
+```
+The function `mem` defined above won’t have the expected behaviour, and will
+always return `false`. This is because, by default, the rewriting engine does
+not take into account the order of declaration of the rules, so given a term
```
fblanqui, line 114: of rule declarations. Hence, given the term

doc/commands.md
```diff
@@ -101,6 +101,30 @@ Exposition obeys the following rules: inside a module,
 - externally defined protected symbols cannot appear in the right hand side of a
   rewriting rule
 
+**Sequential rules:**
```
fblanqui, line 104: I would prefer to put this in modifiers:
sequential: By default, the order of rule declarations is not taken into account. This modifier tells Lambdapi to apply rules defining a sequential symbol in the order they have been declared (note that this may depend on the order of require commands). WARNING: use with caution as this may break important properties.

src/core/parser.ml
```diff
@@ -655,9 +660,10 @@ let parser cmd =
   | _open_ ps:path+
       -> List.iter (get_ops _loc) ps;
          P_open(ps)
-  | e:exposition? p:property? _symbol_ s:ident al:arg* ":" a:term
+  | st:mstrat? e:exposition? p:property? _symbol_ s:ident al:arg* ":" a:term
```
fblanqui, line 663: Could we simplify the grammar (for the user) so that she doesn't need to remember in which order to put all these modifiers? For instance using:
```
(* in syntax.ml *)
type p_modifier = Stra of match_strat | Prop of property | Expo of exposition
```
A symbol declaration would come with a list of modifiers.
Of course, we would then need to check that the list is consistent: fail when there is both protected and private, or constant and injective.

doc/commands.md
fblanqui: Note that this may break confluence and other properties on terms with metavariables when rules are non linear like in your example. Take for instance the term `E=mem ?a (cons ?b nil)`. If at time t=0, ?a and ?b are different, E reduces to `false`. If, later, at time t=1, ?a is instantiated to ?b for some reason, then the application of the rule was wrong...

gabrielhdt: I have slightly reworded, and have changed a bit the doc.

doc/commands.md
```diff
@@ -76,23 +76,27 @@ Data types and predicates must be given types of the form
...
+
+- `sequential`: modifies the pattern matching algorithm. By default, the order
+  of rule declarations is not taken into account. This modifier tells Lambdapi
+  to apply rules defining a sequential symbol in the order they have been
+  declared (note that the order of the rules may depend on the order of the
+  `require` commands). An example can be seen in 
+  [`rule_order.lp`](../tests/OK/rule_order.lp).  
+  *WARNING:* using this modifier can break important properties.
+
+Exposition markers obey the following rules: inside a module,
```
fblanqui, line 99: Why not moving this paragraph on line 90?

src/core/handle.ml
```diff
@@ -73,6 +73,56 @@ let handle_require : bool -> popt -> sig_state -> Path.t -> sig_state =
...
+    let pp_pmloc oc (m: p_modifier loc) =
+      Format.fprintf oc "%a:\"%a\"" Pos.print_short m.pos Pretty.pp_modifier m
+    in
+    fatal_no_pos "%a" (Format.pp_print_list ~pp_sep pp_pmloc) mods
+  in
+  let prop = List.filter is_prop ms in
+  let prop =
+    match prop with
+    | _::_::_ ->
+        fatal_msg "Only one property modifiers can be used, \
```
fblanqui, line 95: typo: one ... modifier (no s)

src/core/handle.ml
```diff
@@ -73,6 +73,56 @@ let handle_require : bool -> popt -> sig_state -> Path.t -> sig_state =
...
+        die prop
+    | [{elt=P_prop(p); _}] -> p
+    | [] -> Defin
+    | _ -> assert false
+  in
+  let expo = List.filter is_expo ms in
+  let expo =
+    match expo with
+    | _::_::_ ->
+        fatal_msg "Only one exposition markers can be used, \
```
fblanqui, line 106: idem

src/core/handle.ml
```diff
@@ -143,13 +196,20 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
...
       (ss, None)
-  | P_definition(e,op,x,xs,ao,t) ->
+  | P_definition(ms, op, x, xs, ao, t) ->
       (* We check that [x] is not already used. *)
       if Sign.mem ss.signature x.elt then
         fatal x.pos "Symbol [%s] already exists." x.elt;
+      (* Verify modifiers. *)
+      let (prop, expo, mstrat) = handle_modifiers ms in
+      if prop <> Defin then
+        fatal_no_pos "Property modifiers cannot be used in definitions.";
```
fblanqui, line 206: `injective` could be accepted. Example: `injective definition double x := x+x`.

src/core/handle.ml
```diff
@@ -196,8 +256,15 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
       let a = if xs = [] then a else Pos.none (P_Prod(xs, a)) in
       (* Obtaining the implicitness of arguments. *)
       let impl = Scope.get_implicitness a in
+      (* Verify modifiers. *)
+      let (prop, expo, mstrat) = handle_modifiers ms in
+      if prop <> Defin then
+        fatal_no_pos "Property modifiers cannot be used in theorem.";
```
fblanqui, line 262: in theorem+s?

src/core/legacy_lexer.mll
```diff
@@ -27,7 +27,7 @@ type token =
   | L_SQB | R_SQB | L_PAR | R_PAR | ARROW | LARROW | FARROW | DEFEQ | COMMA
   | COLON | EQUAL | DOT | EOF
   (* Keywords. *)
-  | KW_DEF | KW_INJ | KW_THM | TYPE | KW_PRV
+  | KW_DEF | KW_INJ | KW_THM | TYPE | KW_PRV | KW_SEQ
```
fblanqui, line 30: Is it correct to modify the legacy parser? The legacy parser should follow the current syntax of Dedukti.

src/core/pos.ml
```diff
@@ -70,10 +70,14 @@ let in_pos : pos -> 'a -> 'a loc =
...
-(** [to_string pos] transforms [pos] into a readable string. *)
-let to_string : pos -> string = fun p ->
+(** [to_string ?print_fname pos] transforms [pos] into a readable string. If
+    [print_fname] is [true] (the default), the filename contained in [pos] is
+    printed. *)
+let to_string : ?print_fname:bool -> pos -> string =
+  fun ?(print_fname=true) p ->
   let {fname; start_line; start_col; end_line; end_col} = Lazy.force p in
   let fname =
+    if not print_fname then "" else
```
fblanqui, line 80: You can avoid `not` by inverting branches.

fblanqui:
For the record, the doc is on https://github.com/gabrielhdt/lambdapi/blob/iss331/doc/commands.md

src/core/legacy_lexer.mll
gabrielhdt: That's a good question. I was thinking that as long as we don't break compatibility, it's ok

src/core/pos.ml
gabrielhdt: Yes, but this form allows the big part of the code to be in the `else` branch, and aligned with the `if`. I agree that introducing a negation can be a bit disturbing.

src/core/legacy_lexer.mll
fblanqui: But who is going to use this syntax? We already have too much work. So, I would prefer not to have more code to maintain if we decide to change things in LP.

amelieled:
What's about inductive type?

fblanqui:
What do you mean? It is not related to inductive types. Induction principle should be defined as Eager (default).

fblanqui:
@gabrielhdt you forgot to remove `--keep-rule-order` in doc/options.md. Thanks to add a word about this in CHANGES.md, and another word on the display of goals in the emacs mode as well (PR by Diego). You can commit directly for this.

amelieled:
> What do you mean? It is not related to inductive types. Induction principle should be defined as Eager (default).
How do I it? If the user write "sequential", I should raise an error?
To be honest, I don't understand why you have merged the type prop and expo (and the optional feature "sequential").
It's not really the same thing and you have created a lot of conflicts on my PR (it's OK now).

fblanqui:
Because it is asking to much to the user to remember in which order to put these modifiers. The order does not matter anymore. An inductive could be private or protected, nothing else for the moment.

amelieled:
OK, I see, you right :)
I will change that in my part !

amelieled:
Hi @gabrielhdt !
Is it normal that you don't change the files .el?

gabrielhdt:
Hmm, you're right, @amelieled. I assumed that the grammar of the emacs mode does not concern keywords. Even though this statement is false with the current code, it may be made true (Emacs shouldn't care whether there's a word before the symbol rule). But anyway, some colours on the new keywords won't hurt. So no, it's not normal...

PR #428, opened 2020-06-24
Opening signature to print symbols

gabrielhdt:
This pull request fixes a bug: symbols of a signature couldn't be printed with the `decision-tree` command.

It seems to me that the usage of `Sign`, `Sig_state` &c. is not very clear, perhaps we could start creating interfaces with those modules?


src/cli/lambdapi.ml
```diff
@@ -79,6 +79,7 @@ let decision_tree_cmd : Config.t -> (Path.t * string) -> unit =
     let sym =
       let sign = Compile.compile false mp in
       let ss = Sig_state.of_sign sign in
+      let ss = Sig_state.open_sign ss sign in
```
fblanqui, line 82: Below, thanks to write `try Sig_state.find_sym ~prt:true ~prv:true false ss (Pos.none ([], sym))` as it is done in other files.

fblanqui: The point was that you were searching a non-qualified symbol from a non-open module. To avoid calling Sig_state.open_sign, you could instead search for a qualified symbol, that is `(mp, sym)` instead of `([],sym)`.

fblanqui:
Concerning the usage of Sign/Sig_state, which doesn't seem related to the current problem though, the rule is that you should never call functions of Sign directly but go through Sig_state instead, unless you know what you do... In fact, Sign should be seen as an internal component of Sig_state.

src/cli/lambdapi.ml
gabrielhdt: Hmm right, but the `mp` doesn't have the adequate type,
```
Error: This expression has type (Path.t * string) Pos.loc
       but an expression was expected of type
         Syntax.qident = (Syntax.p_module_path * string) Pos.loc
       Type Path.t = string list is not compatible with type
         Syntax.p_module_path = (string * bool) list 
       Type string is not compatible with type string * bool
```
so is there something fishy that should be corrected, or opening the module is the way to go? If I understand well, when we create a signature state from a signature, no signature is opened by default? (except the ghost ones)

fblanqui: Right.

src/cli/lambdapi.ml
```diff
@@ -79,8 +79,8 @@ let decision_tree_cmd : Config.t -> (Path.t * string) -> unit =
     let sym =
       let sign = Compile.compile false mp in
       let ss = Sig_state.of_sign sign in
-      let (prt, prv) = (true, true) in
-      try Sig_state.find_sym false ss (Pos.none ([], sym)) ~prt ~prv
+      let ss = Sig_state.open_sign ss sign in
+      try Sig_state.find_sym false ss (Pos.none ([], sym)) ~prt:true ~prv:true
```
fblanqui, line 83: Thanks to put ~prt and ~prv at the beginning like in the other files.

src/cli/lambdapi.ml
fblanqui: What do you think of replacing `mp` by `List.map (fun s -> (s,false)) mp` instead?

fblanqui:
I realize that `lambdapi decision-tree` does not handle escaped identifiers, right? What about ghost symbols?

gabrielhdt:
> I realize that `lambdapi decision-tree` does not handle escaped identifiers, right? What about ghost symbols?

It handles ghost symbols (I used it). For escaped identifiers, it does as well, you can try

```shell
dune exec -- lambdapi decision-tree 'tests.OK.Escaped.{|KIND|}' --lib-root .
```
and see that there is another error... which is due to the fact that you're trying to display the tree of something that doesn't have any rule. I'll fix this one as well in this pull request, it's very easy.

fblanqui:
But what happens if there is a dot inside {|...|} ?...

fblanqui:
Are symbols and module path elements given in command line arguments parsed with parser.ml?

gabrielhdt:
> But what happens if there is a dot inside {|...|} ?...

Ah, well spotted:
```
dune exec -- lambdapi decision-tree 'tests.OK.Escaped.{|KIND.OF.BLUE|}' --lib-root .
File [/home/hondet/workspace/deducteam/lambdapi/tests/OK/Escaped/{|KIND/OF.{lp|dk}] not found.
```

gabrielhdt:
> Are symbols and module path elements given in command line arguments parsed with parser.ml?

No, they're parsed in function `qsym` in `src/cli/lambdapi.ml`. Do you think calling the parser from `parser.ml` would be easy?

src/cli/lambdapi.ml
```diff
@@ -79,11 +79,15 @@ let decision_tree_cmd : Config.t -> (Path.t * string) -> unit =
...
       with Not_found -> fatal_no_pos "Symbol not found."
     in
-    out 0 "%a" Tree_graphviz.to_dot sym
+    if Timed.(!(sym.sym_rules)) = [] then
+      wrn None "Cannot print decision tree: \
```
fblanqui, line 87: Why not simply say: "no dot file generated since symbol ... has no rule" (because we could output an empty dot file instead)?

fblanqui:
We need to define a function like Parser.parse_string calling the parser qident instead of cmds. In fact, Parser.parse_string could perhaps be generalized so that we could call it by specifying which parser we want to use.

gabrielhdt:
> We need to define a function like Parser.parse_string calling the parser qident instead of cmds. In fact, Parser.parse_string could perhaps be generalized so that we could call it by specifying which parser we want to use.

Well, this looks non trivial with Earley, and perhaps we don't want to spend much time on Earley...

fblanqui:
What about just adding:
```
let parse_qident : string -> ast = fun str ->
  Prefix.reset unops; Prefix.reset binops; Prefix.reset declared_ids;
  try Earley.parse_string ~filename:"" qident blank str
  with Earley.Parse_error(buf,pos) ->
    let loc = Pos.locate buf pos buf pos in
    parser_fatal loc "Parse error."
```?

**Comment by gabrielhdt** [comment]
> What about just adding:
> 
> ```
> let parse_qident : string -> ast = fun str ->
>   Prefix.reset unops; Prefix.reset binops; Prefix.reset declared_ids;
>   try Earley.parse_string ~filename:"" qident blank str
>   with Earley.Parse_error(buf,pos) ->
>     let loc = Pos.locate buf pos buf pos in
>     parser_fatal loc "Parse error."
> ```
> 
> ?

Aah, allright seems easier than I thought.

File: src/cli/lambdapi.ml (hunk 1 above)
Comment (gabrielhdt) [reply src/cli/lambdapi.ml:82]: Right, it's what I've done in 826ae97 in fact.

File: src/cli/lambdapi.ml (hunk 3 above)
Comment (gabrielhdt) [reply src/cli/lambdapi.ml:87]: Do you mean without warning? The command does not generate a file, it prints the graph on the standard output.

Comment (fblanqui) [reply src/cli/lambdapi.ml:87]: No, I am just talking about the message.

Comment (gabrielhdt) [reply src/cli/lambdapi.ml:87]: It's meaningless to say that no file has been generated, since it never generates files...

File: src/cli/lambdapi.ml [hunk 4]
```diff
@@ -125,15 +129,23 @@ let lsp_log_file : string Term.t =
...
+  let qsym_conv: (Path.t * string) Arg.conv =
+    let parse s =
+      match Parser.parse_qident s with
+      | Ok({elt = (smp, id); _}) -> Ok(List.map fst smp, id)
+      | Error(p) ->
+          let msg = Format.sprintf "Parse error at %s." (Pos.to_string p) in
+          Error(`Msg(msg))
+    in
+    let print fmt (smp, qid) =
+      let mp = List.map (fun mp -> (mp, false)) smp in
```
fblanqui, line 141: Why do remove the booleans on line 135 and add them (perhaps wrong ones now!) on line 141?

gabrielhdt: Yes, I wanted to ask you what would be a proper management of these booleans

gabrielhdt: I mean, are they useful here?

src/cli/lambdapi.ml
fblanqui: Ah it's printed on stdout now.

src/cli/lambdapi.ml
fblanqui: qsym should return a p_module_path

fblanqui: The boolean says if the enclosing brackets have been removed or not. This is the argument type of find_sym. So it could be important. It seems better that qsym outputs this type.

src/cli/lambdapi.ml
gabrielhdt: Sorry, I misunderstood, it wasn't what I did in 826ae97, but now it's done.

fblanqui:
It looks fine to me. Tell me when it's ready to merge.

gabrielhdt:
I have added a script that tests the `decision-tree` command in the test suite.

gabrielhdt:
Ah the problem with calling the parser is that we can't use ghost symbols (since the parser can't parse their identifiers)

fblanqui:
You could test if the ident starts with `#` for dealing with ghost symbols, and have a special treatment for them.

gabrielhdt:
I just have to update the documentation and it'll be good

gabrielhdt:
There is another issue: `lambdapi.pkg` files are not parsed by the decision-tree command. For this we'd need to have the path to something below the package file. However, for the moment, we input a qualified id, so we do not have a path.

We can use the directory where lambdapi is called, but thus the command depends on the shell context. To avoid this, we may allow the user to input the path of a package file.

fblanqui:
I do not fully understand. Would you have an example?

gabrielhdt:
Perhaps it's more clear if I say that Lambdapi does not know where to find the `lambdapi.pkg` file. If you call `lamdapi decision-tree a.b.c.d`, the `root_path` can be `a.b`, `a`, `a.b.c`, and you depend on the current working directory.

In fact, it can be solved using the `--map-dir` argument.

gabrielhdt:
Regarding my last remarks, there are two ways to sort this out:
1. assuming we're calling `lambdapi decision-tree a.b.c` in directory `dir`, the executable looks for a `lambdapi.pkg` from the current working directory `dir`. If one wants to inspect a symbol whose `lambdapi.pkg` is in a sub-directory, the `--map-dir` option must be used. The downside is that whereas `lambdapi check` is directory independent, this command isn't, so it might be a bit disturbing for the newcomer.
2. to print the symbol `s` of file `foo.lp`, we call `lambdapi decision-tree foo.lp:s`. The advantage is that the command remains independent from the current working directory, so the general behaviour is more coherent. And it might even be a bit easier to implement.

fblanqui:
Hi Gabriel.
In your 2nd proposal, what to do if you want to get the decision tree of a symbol defined in an installed library? The user perhaps doesn't know or doesn't want to write the absolute path to the file installed by opam...
As for the fact that lambdapi check is directory independent to some extent, I am not sure it is a so nice feature...

gabrielhdt:
> Hi Gabriel.
> In your 2nd proposal, what to do if you want to get the decision tree of a symbol defined in an installed library? The user perhaps doesn't know or doesn't want to write the absolute path to the file installed by opam...
> As for the fact that lambdapi check is directory independent to some extent, I am not sure it is a so nice feature...

That's a good remark. So it seems that there are less corner-cases with the first proposal (using `map-dir` if needed).

fblanqui:
Ok, so, let's take your first proposal and just add some warning in the doc. Tell me when you are done with this so that I have a last look at it before merging.

gabrielhdt:
Shouldn't we move all the documentation that is in `src/core/tree_graphviz.ml` to the `docs/`?

fblanqui:
Good idea. But, perhaps, do not re-move it, just copy it and adapt it to MD.

gabrielhdt:
I have revised the doc, it's ready to be merged (unless you have other remarks, @fblanqui)

doc/options.md
```diff
@@ -122,6 +121,71 @@ can first generate a `.xml` file as described below.
 To inspect the `.xml` file generated by Lambdapi, one may use the following
 dummy command: `--termination "cat > output.trs; echo MAYBE"`.
 
+##### Printing decision trees
```
fblanqui, line 124: I propose to put all this in a separate file, and add a link on line 22.

doc/options.md
```diff
@@ -122,6 +121,71 @@ can first generate a `.xml` file as described below.
...
+    is a superset of the free variables `xs`.
+    
+- Triangles represent _stack check_ nodes. The next node is the left
+  child if the stack of arguments is empty, the right child otherwise. These
+  nodes can only appear in trees built for [sequential](commands.md#symbol) 
+  symbols.
+
+*Note for developers:* the ghost symbols can be printed as well, for instance
+```
+lambdapi decision-tree M.#equiv
```
fblanqui, line 185: We don't need to write `M.`, do we?

fblanqui: Sorry, ignore my previous comment...

fblanqui:
@gabrielhdt I just propose to put the doc on decision trees in a separate file. No other remarks. Once done, I will merge. Let me know when it's done.

doc/options.md
gabrielhdt: OK. We should probably change our documentation support to have something more structured than a bunch of markdown files anyway

gabrielhdt:
> @gabrielhdt I just propose to put the doc on decision trees in a separate file. No other remarks. Once done, I will merge. Let me know when it's done.

Done.

fblanqui:
Thanks @gabrielhdt !

PR #434, opened 2020-07-17
Fixed hovers and go to definition

francois-lefoulon:
Go to definition and hovers work, and bound variables are currently ignored.

I wonder if I should change the current keybindings for LP-VSCode, because Ctrl+Arrow is really useful for cursor navigation (it allows to go to the next word) and I don't think the extension is very practical as is ; I would rather use AltGr+Arrow (since all other combinations seem to be reserved by either Ubuntu or VS Code), or any given idea that sounds more practical.


fblanqui:
Hi François. AltGr is usually on the right of the keyboard, so on the same side as arrows. It doesn't seem very convenient. It seems more convenient to use Alt-arrow or Ctrl-Alt-arrow. What do you think?

src/lsp/lp_doc.ml
```diff
@@ -12,6 +12,106 @@
...
+
+module type RangeMap = sig
+
+  type point
+  type interval
+  type 'a t
+
+  val create_point : int -> int -> point
+  val l : point -> int
+  val c : point -> int
```
fblanqui, line 26: l and c are really too short names. Please use line and column instead. Thank you.

src/lsp/lp_doc.ml
```diff
@@ -12,6 +12,106 @@
...
+  val interval_end : interval -> point
+
+  val create_interval : point -> point -> interval
+
+  type cmp = Before | In | After
+
+  val in_range : point -> interval -> cmp
+
+  val add : interval -> 'a -> 'a t -> 'a t
+  val find : point -> 'a t -> 'a option * interval
```
fblanqui, line 38: Why isn't the output type ('a * interval) option?

src/lsp/lp_doc.ml
```diff
@@ -12,6 +12,106 @@
 
 open Core
 module LSP = Lsp_base
+(*
+exception NoPosition of string *)
+
+module type RangeMap = sig
```
fblanqui, line 18: Thanks to add comments and reorder declarations so that to have, whenever possible, type + constructors + destructors. For instance:
```
  type point (** comment *)
  val create_point : int -> int -> point
  val line : point -> int
  val column : point -> int

  type interval (** ... *)
```

src/lsp/lp_doc.ml
```diff
@@ -12,6 +12,106 @@
...
+  type cmp = Before | In | After
+
+  val in_range : point -> interval -> cmp
+
+  val add : interval -> 'a -> 'a t -> 'a t
+  val find : point -> 'a t -> 'a option * interval
+
+  val empty : 'a t
+
+  val rangemap_to_string : ('a * string) t -> string
```
fblanqui, line 42: What is the use of this function? If it is for debug, isn't it better to use Format?

src/lsp/lp_doc.ml
fblanqui: Idem in the implementation below.

src/lsp/lp_doc.ml
```diff
@@ -12,6 +12,106 @@
...
+    then Before
+
+    else if pos.l > pos2.l || (pos.l = pos2.l && pos.c > pos2.c)
+    then After
+
+    else In
+
+  type 'a t = (interval * 'a) list
+
+  let rec find (cursor_pos : point) (l: 'a t) = match l with
```
fblanqui, line 79: coding style: `match` should come on next line. Please check coding style recommendations and see other files for examples.

src/lsp/lp_doc.ml
```diff
@@ -12,6 +12,106 @@
...
+let interval_of_pos (p : Pos.pos) = let open M in
+
+  let data = Lazy.force p in
+  let start : point = create_point data.start_line data.start_col in
+  let finish : point = create_point data.end_line data.end_col in
+
+  create_interval start finish
+
+let interval_of_popt (p : Pos.popt) = match p with
+  | None -> failwith "interval_of_popt : no position given for the token"
```
fblanqui, line 107: invalid_arg instead?

src/lsp/lp_doc.ml
fblanqui: Why not moving RangeMap and M into src/core/pos.ml, and rename them as follows?
```
module type RangeMapSig = ...
module RangeMap : RangeMapSig = ...
```

ejgallego: I'd suggest going even further and placing the rangemap in its own utility module. Then pos could provide the implementation. I think LP doesn't have a `src/lib` , IMHO OCaml's stdlib is a bit limited so at some LP is gonna need a few extra base modules.

src/lsp/lp_doc.ml
ejgallego: Better return and option type and have the clients handle `None`

src/lsp/lp_doc.ml
```diff
@@ -128,6 +232,9 @@ let dummy_loc =
         ; end_col = 2
         }
 
+let concat_map (f:'a -> 'b list) (l: 'a list) : 'b list =
+  List.concat (List.map f l)
+
```
ejgallego, line 237: This should be moved to lplib and eventually replaced by OCaml's own.

src/pure/pure.ml
```diff
@@ -4,11 +4,121 @@ open Extra
...
+      | Syntax.P_query_normalize (term, _) -> qidents_of_p_term term
+      | Syntax.P_query_prover _ -> []
+      | Syntax.P_query_prover_timeout _ -> []
+
+      in f q.elt
+
+  (*This function is used to get the symbol identifiers
+  in the document and then create a map from them*)
+  let get_qidents (doc : t) = (qidents_of_cmd doc)
+
```
ejgallego, line 121: Eventually these extraction functions should be moved to their own file, I dunno how to structure the code tho.

src/pure/pure.mli
```diff
@@ -5,9 +5,10 @@ open Files
 
 (** Abstract representation of a command (top-level item). *)
 module Command : sig
-  type t
+  type t = Syntax.p_command
```
ejgallego, line 8: This should be made abstract again [we used this for debugging]

src/lsp/lp_doc.ml
francois-lefoulon: I'm not sure how I should use Format here. By providing a string version of the map I just need to type LIO.log_error str, but if Format is more practical I'll change that.

src/lsp/lp_doc.ml
francois-lefoulon: This function is not destined to the user but to a tricky fold_left so I don't know how to handle an option here.

francois-lefoulon:
The implementation for token maps using the standard library was harder than expected. The solution was not to consider points as intervals with start = finish, but rather exploring a range map with Map.find_first.

francois-lefoulon:
I eventually found how to make token maps without find_first.

francois-lefoulon:
In this implementation, the in_range function does not seem useful anymore, but it might be in the future so I will let it remain.

src/lplib/cmap.ml
```diff
@@ -0,0 +1,162 @@
...
+of a specific token, which is an interval of points (positions)*)
+(*The presence of t and compare make RangeType an OrderedType (in the sense of Map) *)
+module type RangeType = sig
+
+  type point
+  val make_point : int -> int -> point
+
+  val line : point -> int
+  val column : point -> int
+  type cmp = Before | In | After
```
fblanqui, line 11: thanks to move before in_range

src/lplib/cmap.ml
```diff
@@ -0,0 +1,162 @@
...
+
+  type t
+  type interval = t
+  (*Ensures an interval is well defined (e.g with start < finish)*)
+  val make_interval : point -> point -> interval
+
+  val interval_start : interval -> point
+  val interval_end : interval -> point
+
+  val in_range : point -> interval -> cmp
```
fblanqui, line 23: thanks to move at the end of the signature (with the type cmp)

src/lplib/cmap.ml
```diff
@@ -0,0 +1,162 @@
+(*Type of points and intervals. Used to determine if a cursor is in the range
+of a specific token, which is an interval of points (positions)*)
```
fblanqui, line 2: Thanks to add a space after `(*` and end your sentences with a dot. ;-)

src/lplib/cmap.ml
```diff
@@ -0,0 +1,162 @@
...
+
+  type point = { l : int; c : int }
+  let make_point l c = { l = l; c = c}
+
+  let line pt = pt.l
+  let column pt = pt .c
+  type cmp = Before | In | After
+
+  let point_to_string pt = "Line : "^(string_of_int pt.l)
+  ^"; Column : "^(string_of_int pt.c)^"\n"
```
fblanqui, line 54: Thanks to add spaces around `^`. parentheses are not necessary around `string_of_int pt.c`.

src/lplib/cmap.ml
```diff
@@ -0,0 +1,162 @@
...
+
+  (*Adds the range for a token and the token in a map*)
+  (*Requires all keys to be non overlapping intervals*)
+  (*/!\ Does not ensure consistency if the added keys are overlapping
+  intervals, e.g might change a previously added (key, element) couple*)
+  val add : interval -> 'a -> 'a t -> 'a t
+  (*Given a cursor position, returns the corresponding token*)
+  val find : point -> 'a t -> (interval * 'a) option
+
+  val empty : 'a t
```
fblanqui, line 115: thanks to move before add

src/lplib/cmap.ml
```diff
@@ -0,0 +1,162 @@
...
+
+  val to_string : ('b * string) t -> string
```
fblanqui, line 117: rename b by a to keep the same type variable name as before

fblanqui: this function is strange: why couldn't we print an element of t? I am expecting the type `(`a -> string) -> 'a t -> string`.

src/lplib/cmap.ml
```diff
@@ -0,0 +1,162 @@
...
+end
+
+(*The functor for cursor maps*)
+module MakeCursorMap (Range : RangeType) = struct
```
fblanqui, line 121: `: CursorMapSig` ?

src/lplib/cmap.ml
```diff
@@ -0,0 +1,162 @@
...
+  let interval_to_string interv =
+    let s, f = (interval_start interv), (interval_end interv) in
+    "From :\n"^(point_to_string s)^"To :\n"^(point_to_string f)
+end
+
+(*A structure which maps a position of the cursor to the corresponding token
+in a given text editor with a .lp file open*)
+module type CursorMapSig = sig
+
+  type 'a t
```
fblanqui, line 103: What is `'a` for?

In addition, I would rather name this signature RangeMapType.

src/lplib/cmap.ml
francois-lefoulon: I can't do that (I already tried to), since when I use the functor with type RangeCursorMapSig at the end of the file, it is not compatible with CursorMapSig.

src/lplib/cmap.ml
francois-lefoulon: t is for the map, and 'a is for its values (for instance qident).

fblanqui: Please document

src/core/scope.ml
```diff
@@ -546,8 +546,8 @@ let scope_rule : sig_state -> p_rule -> pre_rule loc = fun ss r ->
   let pr_vars =
     let fn i =
       let name =
-        try Printf.sprintf "v%i_%s" i (Hashtbl.find lhs_names i)
-        with Not_found -> Printf.sprintf "v%i" i
+        try Printf.sprintf "%i_%s" i (Hashtbl.find lhs_names i)
+        with Not_found -> Printf.sprintf "%i" i
```
fblanqui, line 550: Thanks to not change this part of the code (you seem to have forgotten to update).

src/lplib/cmap.ml
```diff
@@ -0,0 +1,172 @@
...
+    (x, y)
+
+  let line pt = pt.l
+  let column pt = pt .c
+
+  let interval_start = fst
+  let interval_end = snd
+
+  type cmp = Before | In | After
+
```
fblanqui, line 61: thanks to remove this kind of empty lines

src/core/scope.ml
francois-lefoulon: Oh right

src/lplib/cmap.ml
francois-lefoulon: Isn't more readable?

francois-lefoulon: I'm not sure which lines I should remove here actually

francois-lefoulon: I'm not sure which lines I should remove here actually

src/lplib/cmap.ml
```diff
@@ -0,0 +1,174 @@
...
+    let line pt = pt.l
+    let column pt = pt .c
+
+    let interval_start = fst
+    let interval_end = snd
+
+    type cmp = Before | In | After
+
+    let in_range pos (pos1, pos2) =
```
fblanqui, line 61: thanks to not use empty lines inside the code of a function unless it is very long

src/lplib/cmap.ml
```diff
@@ -0,0 +1,174 @@
...
+of the other (i.e its finishing point is before
+the starting point of the other). *)
+module Range : RangeType =
+  struct
+    type point = { l : int; c : int }
+    let make_point l c = { l = l; c = c}
+
+    type t = point * point
+    type interval = t
+    let make_interval x y =
```
fblanqui, line 49: perhaps holds on one line only

src/lplib/cmap.ml
```diff
@@ -0,0 +1,174 @@
...
+      else if f1 <= s2
+      then -1
+
+      else if s1 >= f2
+      then 1
+
+      else
+        failwith "Intervals overlap, no inclusion between the two"
+
+    let point_to_string pt = "Line : " ^ string_of_int pt.l
```
fblanqui, line 88: coding style: thanks to add newline after =

src/lplib/cmap.ml
```diff
@@ -0,0 +1,174 @@
+(*Type of points and intervals. Used to determine if a cursor is in the range
```
fblanqui, line 1: Use ocamldoc features for documentation like `(**`, etc. Please check ocamldoc documentation and check the result by doing `make doc` (I believe). If you have questions on ocamldoc, please ask @gabrielhdt

src/lplib/cmap.ml
```diff
@@ -0,0 +1,174 @@
...
+      then Before
+
+      else if pos > pos2
+      then After
+
+      else In
+    
+    let compare (s1, f1) (s2, f2) =
+
+      (*Two intervals are considered equal if one is included in the other. *)
```
fblanqui, line 75: this comment should be in the signature instead

src/lplib/cmap.ml
```diff
@@ -0,0 +1,174 @@
...
+      - the keys for "add" are intervals
+      - the keys for "find" are points.
+    *)
+    type 'a t = (Range.interval * 'a) RangeMap.t
+    type interval = Range.interval
+    type point = Range.point
+
+    let point_to_interval : point -> interval =
+    fun pt ->
+      Range.make_interval pt pt
```
fblanqui, line 137: coding style: please read https://github.com/Deducteam/lambdapi/blob/master/CONTRIBUTING.md

src/lplib/cmap.ml
```diff
@@ -0,0 +1,174 @@
...
+    
+    let find cursor map =
+    let interv = point_to_interval cursor in
+    RangeMap.find_opt interv map
+    
+    let empty = RangeMap.empty
+
+    let add interv elt map = let open Range in
+      (*The interval stored as a value must include the character
+      right after it for display purposes in the text editor. *)
```
fblanqui, line 147: Could you please elaborate? Is it specific to VSCode?

fblanqui: Wouldn't it be better that this is done by the caller?

fblanqui:
Hi @francois-lefoulon . You do not respect very much the coding style of Lambdapi. Please read CONTRIBUTING.md and have a look at other LP files, and update your code so that it looks the same. Thank you.

src/lplib/cmap.ml
```diff
@@ -0,0 +1,174 @@
...
+    (*A map of which keys are intervals. *)
+    module RangeMap = Map.Make(Range)
+  
+    (*Now we need to transform the map so that :
+      - the keys for "add" are intervals
+      - the keys for "find" are points.
+    *)
+    type 'a t = (Range.interval * 'a) RangeMap.t
+    type interval = Range.interval
+    type point = Range.point
```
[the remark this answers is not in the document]
francois-lefoulon: I need types interval and point in the signature for range maps, which can't be made a functor since it is a signature, and with the recommendation of Emilio I made range maps generic so that any implementation of Range can be used. Given these criteria, I don't see any other way to do this.

src/lplib/cmap.ml
francois-lefoulon: I don't know if it is specific to VSCode since emacs doesn't process hover requests for now. I moved that to the caller anyway.

francois-lefoulon:
I have actually read CONTRIBUTING.md a few times already. I think it would benefit some more clarity : 
- nothing is mentioned about empty lines inside functions, what is the issue with lightening the code ?
- nothing is mentionned about naming conventions for modules or module types, nor the ordering of the functions inside them. Plus, the indentation for sig... and and struct... end, which I now copied on the core, should be explained.
- it is said that .mli files don't exist in the source code, which is not true. I decided that I would adopt the advised syntax (let ... : ... = fun ...) if the signature for the function was not already given in an interface or module type ; the code for the language server did not follow the aforementioned syntax anyway, so it would have been even more consistent to never use that syntax, or to change all functions with that syntax.

I'm sorry for the times I made mistakes about the advised syntax or forgot the sanity_check, though.

fblanqui:
No problem. You are right, CONTRIBUTING.md does not contain enough information. https://github.com/Deducteam/lambdapi/pull/406 should provide more once merged. For examples of signatures and implementation, check extra.ml.

fblanqui:
Hi @ejgallego . How are you? I remember that you told us you wanted to review this PR and change a few things before merging it. It would be nice to do this before François' report deadline (September 15). Do you think that you can make it?

ejgallego:
> Hi @ejgallego . How are you? I remember that you told us you wanted to review this PR and change a few things before merging it. It would be nice to do this before François' report deadline (September 15). Do you think that you can make it?

Sorry folks for the delay, finally having a look; will rebase and push some cleanups, likely we can merge just after.

ejgallego:
Ok, so here is what I did:

- I rebased the PR, and I dropped the first part of the PR as it has been integrated in commit 31b164d6 , I did a quite careful review to ensure that nothing was lost, but it could still be the case. I ported the doc changes and dropped the remapping of keybindings from `Ctrl+Right` to `Ctrl+Shift+Right`, as I think the former is nowadays preferred
- Adapted the code to the new inductive constructors
- Tested every commit to check for build / doc problems
- Cleaned up the code a bit [see last commits]
- Move the `Extra` module to the `lplib` library.
- Removed some things that were for OCaml < 4.04 [as we now require 4.05]
- Ported to Dune language 2.0

By the way congratulation on all the nice UI features you got, it is IMHO really nice!

src/lsp/lp_doc.ml
ejgallego: Fixed in my pass.

src/lsp/lp_doc.ml
ejgallego: Fixed in the last pass.

src/lsp/lp_doc.ml
ejgallego: They are now in their own utility; they are not perfect but should be pretty easy to tweak them further.

src/lsp/lp_doc.ml
ejgallego: That's for debug, it could be further tweaked indeed, but not a blocker iMHO.

src/lsp/lp_doc.ml
ejgallego: Solved in all new files by using an automatic formatter.

src/lsp/lp_doc.ml
ejgallego: Actually the function is now used in the implementation of `RangeMap`, passed down to the stdlib's `Map.Make` fucntor. That seems quite tricky.

src/lsp/lp_doc.ml
ejgallego: Done.

src/pure/pure.ml
ejgallego: Done.

src/pure/pure.mli
ejgallego: Done.

src/lplib/cmap.ml
ejgallego: Done.

src/lplib/cmap.ml
ejgallego: Ok now I think.

src/lplib/cmap.ml
ejgallego: Most of this solved.

src/lplib/cmap.ml
ejgallego: Solved.

src/lplib/cmap.ml
ejgallego: Done.

src/lplib/cmap.ml
ejgallego: Solved.

src/lplib/cmap.ml
ejgallego: Fixed too.

src/lplib/cmap.ml
ejgallego: Documented.

src/lplib/cmap.ml
ejgallego: Done by automatic formatter.

src/lplib/cmap.ml
ejgallego: It is using the lex order so it holds, but needs improvement, see the fixme.

src/lplib/cmap.ml
ejgallego: Mostly solved.

src/lplib/cmap.ml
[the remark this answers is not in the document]
ejgallego: Solved.

src/lplib/cmap.ml
ejgallego: Solved.

fblanqui:
@gabrielhdt This is strange: all the OK tests fail.

fblanqui:
error on travis:
```
make $TEST_TARGETS

       utils alias tests/runtest (exit 2)

(cd _build/default/tests && ./utils.exe)

Fatal error: exception Core.Console.Fatal(0, "File [/home/travis/build/Deducteam/lambdapi/_build/default/tests/tests/OK/bool.{lp|dk}] not found.")

       ok_ko alias tests/runtest (exit 1)
```

ejgallego:
> @gabrielhdt This is strange: all the OK tests fail.

This is because the bump to the dune lang 2.0, I can remove that commit.

However, the more I look to the code in `Files.file_to_module` in particular the
```
if String.length p < String.length path then mapping := Some(mp, p)
```
the less I understand it, I'm afraid. Indeed in this case it builds and incorrect module path.

francois-lefoulon:
Thank you for all the corrections @ejgallego !

Out of curiosity, why did you do these updates today ? It is amusing because my presentation is tomorrow, so I downloaded the branch earlier today to install it on my current computer, and then saw that you started making the corrections - were you notified that I downloaded the repository ? Or maybe @fblanqui  just told you that my presentation is tomorrow...

ejgallego:
salut @francois-lefoulon , it was on my todo list since a long time, @fblanqui told me about your presentation too, but it was kinda of a coincidence that I managed to work on this today.

By the way there are no "corrections" per-se, only cleanups and a bit of refactoring; the current code here should be 100% equivalent to the old one. See you tomorrow!

fblanqui:
Why do we still have dune 1.x files? We should move all dune files to dune 2.0. @gabrielhdt or @ejgallego could you please do this?

gabrielhdt:
> Why do we still have dune 1.x files? We should move all dune files to dune 2.0. @gabrielhdt or @ejgallego could you please do this?

OK, I'm looking at it right now.

gabrielhdt:
So I created a pull request at #449 where we can see more properly what works and what doesn't.

ejgallego:
> Why do we still have dune 1.x files? We should move all dune files to dune 2.0. @gabrielhdt or @ejgallego could you please do this?

Not a big deal, dune is compatible with the 1.x files; it was just that I was doing a cleanup.

On a different topic, I have no idea why the test suite is failing...

gabrielhdt:
> > Why do we still have dune 1.x files? We should move all dune files to dune 2.0. @gabrielhdt or @ejgallego could you please do this?
> 
> Not a big deal, dune is compatible with the 1.x files; it was just that I was doing a cleanup.
> 
> On a different topic, I have no idea why the test suite is failing...

I think I found out why, and I fixed it in `dune2`, see the description of #449: same module is imported several times in the test suite, because each `compile_file` call will apply the configuration of `lambdapi.pkg` through `apply_config`, each of which calling `new_lib_mapping` which would fail because the mapping would have already been define by a previous compilation.

ejgallego:
Ok thanks @gabrielhdt ; I think this is mostly ready to go; there are indeed some discussion to be had about the rangemap specification, but IMO this is good for now.

fblanqui:
Hi @ejgallego . The code for tests has been fixed. You can now update your PR. Let me know when it is ready to be merged.

ejgallego:
Rebased, should be good to merge IMO; note however that still `make tests` return error in my setup, but that's the same for master:
```
$ git checkout master
$ dune runtest
       ok_ko alias tests/runtest (exit 1)
(cd _build/default/tests && ./ok_ko.exe)
Testing `Std'.
This run has ID `58AEC567-EFBB-43D0-8361-A950F5305DDA'.
  [OK]          OK          0   OK/uninst_meta_in_rhs.dk.
....
$ echo $?
1
```

gabrielhdt:
> Rebased, should be good to merge IMO; note however that still `make tests` return error in my setup, but that's the same for master:
> 
> ```
> $ git checkout master
> $ dune runtest
>        ok_ko alias tests/runtest (exit 1)
> (cd _build/default/tests && ./ok_ko.exe)
> Testing `Std'.
> This run has ID `58AEC567-EFBB-43D0-8361-A950F5305DDA'.
>   [OK]          OK          0   OK/uninst_meta_in_rhs.dk.
> ....
> $ echo $?
> 1
> ```

That's interesting, is travis supposed to be happy with an exit 1?

fblanqui:
The error that appears in Travis is not this one but a sanity_check error. (By the way, it would be better to do this check at the beginning...). Thanks to do 'make sanity_check' and fix the errors. FYI, we will soon use ocamlformat.

gabrielhdt:
> The error that appears in Travis is not this one but a sanity_check error. (By the way, it would be better to do this check at the beginning...). Thanks to do 'make sanity_check' and fix the errors. FYI, we will soon use ocamlformat.

I saw that it's because of sanity check, but since it returns 1, I would expect that travis fails anyway. The travis script already calls `make sanity_check`

fblanqui:
```
08:45 ~/lambdapi (contribution) make sanity_check
In src/lsp/lp_doc.ml, line 156 more than 78 characters...
In src/lsp/lp_lsp.ml, line 28 more than 78 characters...
In src/lsp/lp_lsp.ml, line 280 more than 78 characters...
In src/lsp/lp_lsp.ml, line 332 more than 78 characters...
```

fblanqui:
There is something strange: only the line 156 is too long. A bug in sanity_check?

gabrielhdt:
> There is something strange: only the line 156 is too long. A bug in sanity_check?

Are you using GNU awk?

fblanqui:
Sorry, this is my mistake. I was looking at the wrong file. I am going to fix the sanity and merge.

fblanqui:
Thanks @francois-lefoulon and @ejgallego !

ejgallego:
Ah, I kinda saw the sanity check issue but didn't get to understand it. Thanks!

francois-lefoulon:
> Thanks @francois-lefoulon and @ejgallego !

I'm glad!

Thanks again for the internship.

PR #436, opened 2020-08-06
Emacs sanity, fixes #435

### LINKED ISSUE #435: Please do not bundle highlight.el
If at all possible, then a repository used to maintain an Emacs package should not bundle any third-party libraries.

Bundling external dependencies might not cause any issues for users who install the package from Melpa, because packages distributed there can filter out unwanted libraries. But some people (including myself) run packages directly from their repositories (e.g. because that makes it much easier to contribute).

For such users the bundled library may end up on the `load-path`, possibly before the independently installed upstream version.  As a result some outdated version can end up being loaded and that can be very confusing and hard to debug.

I recommend that you never bundle any third-party libraries.  If that is not possible, then I recommend that you put those libraries in a separate directory and alongside a file named `.nosearch`. The existence of this file instructs some tools (such as `normal-top-level-add-subdirs-to-load-path`, my [`borg`](https://emacsair.me/2016/05/17/assimilate-emacs-packages-as-git-submodules/) package manager, and the tools I use to maintain the Emacsmirror) to ignore the containing directory.

Thanks for considering this!

<sub>I have written similar messages many many times so I now [use a template](https://help.github.com/en/github/writing-on-github/working-with-saved-replies). I try to adjust it to match each individual case but it might still not apply 100% in all cases.</sub>

gabrielhdt:
This pull request mainly concerns elisp conventions.


fblanqui:
Can I merge @gabrielhdt ?

gabrielhdt:
@fblanqui Yes, you may merge.

Issue #334, opened 2020-04-06
Make (install) and lib/example directory


fblanqui:
Make and make install compiles and installs the files in lib/example. I would prefer to move this example files in tests/OK/.

rlepigre:
The whole point of this example is to show how to install libraries. It made a lot more sense when it was a tiny standard library that you could have built on instead of having your own repository @fblanqui. Moving this to tests makes no sense, if you don't want it just delete it. The current examples are useless as tests (probably already covered).

fblanqui:
Thanks to delete it then.

fblanqui:
There is no lib/ directory anymore.

PR #437, opened 2020-08-10
Package revision

gabrielhdt:
This pull request changes slightly the behaviour of lambdapi regarding packaging to make it more intuitive and portable; both regarding command line usage and the OCaml code. The notable changes are:

- Portability:
  - replaced `install` by `cp` since the latter is posix while the former isn't (see 
    https://pubs.opengroup.org/onlinepubs/9699919799/)
  - makefile generated by `lambdapi init` is POSIX compliant (see 
    https://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html)
- the library root can be defined using a `LAMBDAPI_LIB_ROOT` environment variable, to be independant from Opam;
- the ocaml code has been simplified regarding `lib_root`: it is set using `set_lib_root` which is always called during the configuration phase and nothing more, there is no more `default_lib_root`, `init_lib_root`, `lib_root_path` &c.


docs/module.rst
```diff
@@ -15,19 +15,25 @@ name of the file, but also the position of the file under the *library
...
+By default, all modules are looked up under the library root. However,
 there are cases where the files we want to work with are not (yet!)
 placed under the library root. The typical case is when a library (or
-package) is still under development. In that case, what we can do is map
-the development folder under the library root, similarly to what would
-happen when mounting a volume in our file system. There are two ways of
+package) is still under development. In that case, the development folder can
+be mapped under the library root, similarly to what would
+happen when mounting a volume in a file system. There are two ways of
 doing that, the first one being to use the ``--map-dir MOD:DIR`` command
```
fblanqui, line 37: add a link like for --lib-root?

docs/module.rst
```diff
@@ -36,8 +42,8 @@ Package configuration file
 --------------------------
 
 A package configuration file can be placed at the root of the source
-tree of a library (or package) under development. It should be named
-``lambdapi.pkg``, and contain the following fields (an example is given
+tree of a library (or package) under development. Its name *must* be
+``lambdapi.pkg``, and it must contain the following fields (an example is given
```
fblanqui, line 46: why not directly writing: A package configuration file ``lambdapi.pkg`` can be placed at the root of the source tree of a library (or package) under development. It must contain the following fields (an example is given...

src/core/files.ml
```diff
@@ -142,56 +142,64 @@ module ModMap :
...
+    - the path given as parameter of the [--lib-root] command line argument;
+      if the path given on command line is not a valid (existing) directory,
+      the program terminates with a graceful error message. *)
+let set_lib_root : string option -> unit = fun path ->
+  let open Stdlib in
+  let prefix = Stdlib.ref "/usr/local" in
+  begin match Sys.getenv_opt "OPAM_SWITCH_PREFIX" with
+    | Some(pth) -> prefix := pth
+    | None -> ()
+  end;
```
fblanqui, line 163: here and later, use Option.iter instead ?

docs/module.rst
```diff
@@ -15,19 +15,25 @@ name of the file, but also the position of the file under the *library
 root* (a folder under which the standard library and all packages are
 installed).
 
+By default, the library root is ``/usr/local/lib/lambdapi/lib_root`` or
```
fblanqui, line 18: Is it a valid path under all Linux systems? What about Windows or Mac?

src/cli/init_cmd.ml
```diff
@@ -8,22 +8,26 @@ let with_file : string -> (out_channel -> unit) -> unit = fun file fn ->
 
 let write_makefile : out_channel -> unit = fun oc ->
   Printf.fprintf oc
-    "LP_SRC = $(shell find . -type f -name \"*.lp\")\n\
-     LP_OBJ = $(LP_SRC:%%.lp=%%.lpo)\n\
+    ".POSIX:\n\
+     SRC = nat.lp main.lp\n\
```
fblanqui, line 12: using shell find is not posix?

docs/module.rst
```diff
@@ -15,19 +15,25 @@ name of the file, but also the position of the file under the *library
...
+
 The typical case is when we want to access a module of some installed
 library, for example the standard library. In that case, the module path
 is built using the file path as follows: if our source file is at
 ``<LIB_ROOT>/std/bool.lp`` it is accessed with module path ``std.bool``.
 And if there are nested folders then the module path gets more members.
 File ``<LIB_ROOT>/std/a/b/c/d.lp`` has module path ``std.a.b.c.d``.
 
-By default, every modules are looked up under the library root. However,
+By default, all modules are looked up under the library root. However,
```
fblanqui, line 31: We should underline this sentence.

fblanqui:
It would be nice also to get rid of the lib/ directory and add some lambdapi.pkg file for tests.

src/cli/init_cmd.ml
gabrielhdt: Not it isn't, POSIX makefiles do not implement external shell calls, for instance, in BSD it's performed using `LP_SRC != find ...`

docs/module.rst
gabrielhdt: It must be valid under most Linux system (since it's mentioned in https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) and it's valid for BSD systems (https://en.wikipedia.org/wiki/Unix_filesystem#Conventional_directory_layout) and hence MacOS. I don't know for windows...

src/core/files.ml
gabrielhdt: Very good idea.

gabrielhdt:
> It would be nice also to get rid of the lib/ directory and add some lambdapi.pkg file for tests.

There is already a `lambdapi.pkg` for tests!

fblanqui:
ok. but is it possible to remove lib/ ?

gabrielhdt:
> ok. but is it possible to remove lib/ ?

As simple as `git rm -r lib`

fblanqui:
I vaguely remember strange things about this lib/ directory. The installation of LP was depending on it. Please, check this too.

fblanqui:
See https://github.com/Deducteam/lambdapi/issues/334 .

gabrielhdt:
As I understand it, there shouldn't be any issue: the files in lib/ won't be compiled nor installed anymore.

PR #439, opened 2020-08-25
XTC export fix

gabrielhdt:
Before this pull request, the XTC export processed ghost symbols, which raised an exception. This pull request filters symbols to avoid processing symbols of ghost signatures.

An XTC export unit test has been added as well.

Fixes #438.

PR #440, opened 2020-08-26
Test suite

gabrielhdt:
This pull request
- fixes the test file used introduced in #439 
- excludes ghost symbols from the export in the HRS format
- introduces `Alcotest`.

`dtrees.sh` test script has been test because it mainly tests command line args processing, which is a bit cumbersome to set up in alcotest.


tests/ok_ko.ml
```diff
@@ -0,0 +1,27 @@
...
+let test_ko f () =
+  (* Succeed if compilation fails. *)
+  let r = try (ignore (Compile.compile_file f)); true with _ -> false in
+  Alcotest.(check bool) f r false
+
+let _ =
+  let open Alcotest in
+  let files = Sys.readdir "OK" |> Array.map (fun f -> "OK/" ^ f) in
+  (* TODO: replace the [f] of [test_case] by a keyword, possibly extracted
+     from the file.  *)
```
fblanqui, line 21: What do you mean here?

tests/utils.ml
```diff
@@ -0,0 +1,54 @@
+(** Testing miscellanous utilities provided by Lambdapi, such as decision tree
+    printing, XTC file generation, HRS file generation.
+    REVIEW: testing CLI? *)
```
fblanqui, line 3: line to be removed?

tests/ok_ko.ml
```diff
@@ -0,0 +1,27 @@
+(** Compile files in "OK" and "KO".
+    REVIEW: Most tests could be removed from "lp" files, and tested directly in
+    the code. *)
```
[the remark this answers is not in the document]
gabrielhdt: Should we test lambdapi features through `lp` files, or should we try to use the Ocaml API? I feel that using the ocaml api may target more precisely the feature we want to test, but it may be more cumbersome to set up.

tests/ok_ko.ml
gabrielhdt: This one is not very important, but each test is tagged with a string, which is, for the moment, the filename.

tests/utils.ml
gabrielhdt: Yes indeed.

tests/ok_ko.ml
fblanqui: Why do you think it may be useful to use something else? Anyway, since this is not important, thanks to turn this "TODO" into a "REMARK".

tests/ok_ko.ml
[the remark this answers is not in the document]
fblanqui: If it works well now as it is, no need to spend time on changing this now. Please, turn this into a "REMARK".

tests/ok_ko.ml
gabrielhdt: I've the todo it in my most recent commit, it's not important.

tests/ok_ko.ml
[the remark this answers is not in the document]
gabrielhdt: As with previous remark, I removed it. Do you prefer that I leave it as a remark?

PR #441, opened 2020-08-28
[WIP] New parser

gabrielhdt:
## Synopsis

This pull request replaces the Earley parser by a parsed based on Menhir and [Sedlex](https://github.com/ocaml-community/sedlex). Sedlex is a UTF8 (and UTF16) lexer. The main challenge is to keep the adaptive behaviour of Earley: the syntax is able to define the syntax itself. The syntax is able to declare infix and prefix symbols. To take these declarations into account, applications are parsed as flat lists, and then reordered during scoping by a second parser that transforms a list of terms into a tree of terms. This new parser is defined in `pratt.ml`.

## Syntax modifications
- A termination token has been added. It is currently the semi colon. Hence, a symbol is declared with
   ```symbol foo: bar;```.

- The syntax `λx y z: nat, ...` is not authorised anymore, but `λ(x y z: nat), ...` is, and `λ x : N, t` is as well. There may be ways to fix it and recover previous syntax.

- Unification rules syntax has changed: the left hand side must be enclosed in `begin`, `end`, so
  ```
  set unif_rule $x + $y ≡ 0 ↪ $x ≡ 0; $y ≡ 0
  ```
  becomes
  ```
  set unif_rule $x + $y ≡ 0 ↪ begin $x ≡ 0; $y ≡ 0 end
  ```

- `set declared "?"` is not needed anymore,

- Any (depending on accepted utf8 codepoints) UTF8 identifier is by default valid.  
  **Warning:** string `λx` is now a valid identifier. Hence, expression `λx, t` isn't valid, but `λ x, t` is.

- Symbols can be input as quantifiers using the backquote: the syntax `` `f x, t`` is valid to represent `` f (λ x, t)`` (and a fortiori `` f {T} (λ x, t)``).

- `assert` always take a turnstile (or vdash) to specify (even empty) context, so the syntax is `assert ⊢ t: A`

- the unary minus before a rewrite pattern becomes the keyword `left`

## Migrating with `sed`

- To place semicolons approximately everywhere we want (at least in test files, where most commands fit on one line)
  ```sed
  s/^\([^/].*\)$/\1;/
  ```
- To add a whitespace space after `λ` and `Π` if there is not,
  ```sed
  s/λ\(\w[^,]*\),/λ \1,/g
  s/Π\(\w[^,]*\),/Π \1,/g
  ```

- To add the turnstile after `assert`
  ```sed
  s/assert /assert ⊢ /g
  s/assertnot /assertnot ⊢ /g
  ```

## To be solved
- [x] Precise the range of UTF8 codepoints that we accept
- [ ] Behaviour of non associative symbols?
- [ ] Restore identifier defined quantifiers
- [ ] Update CHANGES.md, syntax.bnf and the user manual
- [x] Update Emacs code
- [x] Update VSCode code


src/core/lp_lexer.ml
```diff
@@ -0,0 +1,216 @@
...
+
+type token =
+  | EOF
+  | L_PAREN | R_PAREN | L_SQ_BRACKET | R_SQ_BRACKET
+  | L_CU_BRACKET | R_CU_BRACKET
+  | APPLY | ARROW | AS | ASSERT | ASSERT_NOT | ASSIGN | ASSOC of Syntax.assoc
+  | AT
+  | BUILTIN | BACKQUOTE
+  | COLON | COMMA | COMPUTE | COMPUTE_TYPE | CONSTANT
+  | DEBUG_FLAGS of (bool * string)
```
fblanqui, line 15: Are parentheses really needed? Idem for ID.

src/core/lp_lexer.ml
```diff
@@ -0,0 +1,216 @@
...
+  let alphabet = [%sedlex.regexp? 'a' .. 'z' | 'A' .. 'Z']
+  let ascii_sub =
+    [%sedlex.regexp? '-' | '\'' | '&' | '^' | '\\' | '*'
+                   | '%' | '#' | '~']
+  (* NOTE: lambda character now counts as a letter, so a sequence λx is an
+     identifier, so [λx t u,] fails on unexpected token [,]. But [λ x t u,]
+     works. *)
+  let letter =
+    [%sedlex.regexp? lowercase | uppercase | ascii_sub
+                   | math | other_math | subscript | superscript
```
fblanqui, line 75: What are math and other_math?

src/core/lp_parser.mly
```diff
@@ -0,0 +1,379 @@
+(* Lambdapi Menhir parser. We need to parse UTF8 - and hence use Sedlex. *)
+(* Using Sedlex (and not Ocamllex) requires to use the "revised" API of *)
+(* Menhir which abstracts the Lexing buffer. *)
```
fblanqui, line 3: Thanks to use a multi-lines comment instead.

src/core/lp_lexer.ml
gabrielhdt: Indeed what these categories are is not really clear. They ought to be defined in ISO standards, but it's a mess to find the precise definitions. The categories are defined on the github page of sedlex https://github.com/ocaml-community/sedlex

src/core/lp_lexer.ml
gabrielhdt: I think they are, we need to represent the identifiers and parentheses in the syntax, don't we?

src/core/lp_lexer.ml
fblanqui: Is it important to restrict the set of identifiers? Why not taking any character?

gabrielhdt: The lexing became very slow, it must be linked to how sedlex works

fblanqui: I see. Thanks to add a comment with the URL you mention so that we remember it.

src/core/lp_lexer.ml
```diff
@@ -0,0 +1,216 @@
...
+    | "with" -> WITH
+    | "rule" -> RULE
+    | "open" -> OPEN
+    | "prover" -> PROVER
+    | "verbose" -> VERBOSE
+    | "prefix" -> PREFIX
+    | "builtin" -> BUILTIN
+    | "symbol" -> SYMBOL
+    | "require" -> REQUIRE
+    | "definition" -> DEFINITION
```
fblanqui, line 196: to be removed

src/core/lp_lexer.ml
```diff
@@ -0,0 +1,216 @@
...
+    | "assert" -> ASSERT
+    | "assertnot" -> ASSERT_NOT
+    | "compute" -> COMPUTE
+    | "type" -> COMPUTE_TYPE
+    | "unif_rule" -> UNIF_RULE
+    | "theorem" -> THEOREM
+    | "in" -> IN
+    | "on" -> SWITCH(true)
+    | "off" -> SWITCH(false)
+    | "qed" -> PROOF_END(Syntax.P_proof_qed)
```
fblanqui, line 178: to be replaced by "end"

src/core/lp_lexer.ml
```diff
@@ -0,0 +1,216 @@
...
+    | "proof" -> PROOF
+    | "refine" -> REFINE
+    | "why3" -> WHY3
+    | "as" -> AS
+    | "assert" -> ASSERT
+    | "assertnot" -> ASSERT_NOT
+    | "compute" -> COMPUTE
+    | "type" -> COMPUTE_TYPE
+    | "unif_rule" -> UNIF_RULE
+    | "theorem" -> THEOREM
```
fblanqui, line 174: To be removed. On the other hand, you need to add "opaque".

src/core/lp_lexer.ml
gabrielhdt: Ah I remember now, they are needed by menhir, who takes constructor with *one* argument, which may be a tuple

src/core/lp_lexer.ml
```diff
@@ -0,0 +1,216 @@
...
+    | "prover_timeout" -> PROVER_TIMEOUT
+    | "reflexivity" -> REFL
+    | "inductive" -> INDUCTIVE
+    | "symmetry" -> SYMMETRY
+    | "rewrite" -> REWRITE
+    | "assume" -> INTRO
+    | "refine" -> REFINE
+    | "simpl" -> SIMPL
+    | "apply" -> APPLY
+    | "proof" -> PROOF
```
fblanqui, line 165: to be replaced by "begin"
side remark: thanks to put all these strings in alphabetical order

src/core/lp_lexer.ml
fblanqui: I see. Thanks to add this in a comment then to remember it.

src/core/lp_parser.mly
```diff
@@ -0,0 +1,379 @@
...
+      { make_pos $loc (P_rules(rs)) }
+  | SET c=config SEMICOLON { make_pos $loc (P_set(c)) }
+  | q=query SEMICOLON { make_pos $loc (P_query(q)) }
+  | EOF { raise End_of_file }
+
+// Environment of a metavariable or rewrite pattern
+env: L_SQ_BRACKET ts=separated_list(SEMICOLON, term) R_SQ_BRACKET { ts }
+
+sterm:
+  // Explicited identifier (with @)
```
fblanqui, line 311: Explicit ?

src/core/lp_parser.mly
```diff
@@ -0,0 +1,379 @@
...
+  | TYPE { make_pos $loc P_Type }
+  // Metavariable
+  | QUESTION_MARK m=ident e=env?
+      { make_pos $loc (P_Meta(fst m, Option.map Array.of_list e)) }
+  // Pattern of rewrite rules
+  | p=patt e=env? { make_pos $loc (P_Patt(p, Option.map Array.of_list e)) }
+  // Parentheses
+  | L_PAREN t=term R_PAREN { make_pos $loc (P_Wrap(t)) }
+  // Explicitly given argument
+  | L_CU_BRACKET t=term R_CU_BRACKET { make_pos $loc (P_Expl(t)) }
```
fblanqui, line 327: What do you think to using square brackets for implicit arguments instead, and reserve curly brackets for (later) denoting subtypes?

gabrielhdt: We can try, and keep it if it doesn't introduce ambiguity

fblanqui:
@firewall2142 Thanks Ashish to review the code for Emacs and VSCode.

src/core/lp_parser.mly
fblanqui: For unification rules, we could use begin/end instead of square brackets?

fblanqui:
@gabrielhdt Thanks to update CHANGES.md and the user manual including syntax.bnf.

gabrielhdt:
> @firewall2142 Thanks Ashish to review the code for Emacs and VSCode.

Emacs code is not yet merged with master

src/core/pos.ml
```diff
@@ -65,17 +40,32 @@ let make : popt -> 'a -> 'a loc =
...
+
 (** [none elt] wraps [elt] in a ['a loc] structure without any specific source
     code position. *)
 let none : 'a -> 'a loc =
   fun elt -> { elt ; pos = None }
 
+(** [cat p1 p2] is the concatenation of positon [p1] and [p2]. If [p1] and
+    [p2] have different filenames, the resulting filename is [None]. *)
+let cat : pos -> pos -> pos = fun p1 p2 ->
+  { fname = if p1.fname <> p2.fname then None else p1.fname
```
fblanqui, line 58: Instead of None, shouldn't we raise Invalid_argument instead?

gabrielhdt: You're probably right indeed

src/core/pratt.ml
```diff
@@ -0,0 +1,120 @@
...
+open Syntax
+
+(** Table containing defined binary operators. *)
+let bin_operators : binop StrHtbl.t = StrHtbl.create 17
+
+(** Table containing defined unary operators. *)
+let una_operators : unop StrHtbl.t = StrHtbl.create 17
+
+module Pratt : sig
+  val expression : ?rbp:priority -> p_term Stream.t -> p_term
```
fblanqui, line 19: What rbp stand for?

src/core/pratt.ml
```diff
@@ -0,0 +1,120 @@
...
+open Lplib
+open Lplib.Extra
+
+open Syntax
+
+(** Table containing defined binary operators. *)
+let bin_operators : binop StrHtbl.t = StrHtbl.create 17
+
+(** Table containing defined unary operators. *)
+let una_operators : unop StrHtbl.t = StrHtbl.create 17
```
fblanqui, line 16: Why not having a single table as an ident cannot be both binary and unary?

src/core/pratt.ml
gabrielhdt: `rbp` stands for `right binding priority`. These are terms introduced in the Pratt parsing algorithm.

src/core/pratt.ml
```diff
@@ -0,0 +1,120 @@
...
+
+module Pratt : sig
+  val expression : ?rbp:priority -> p_term Stream.t -> p_term
+end = struct
+  (** [lbp t] returns the binding power of term [t] (which is 0 if [t] is not
+      an operator). *)
+  let lbp : p_term -> priority = fun {elt=t; _} ->
+    match t with
+    | P_Iden({elt=(_,s); _}, _) ->
+        begin match StrHtbl.find_opt bin_operators s with
```
fblanqui, line 26: Don't you assume here that s has already been scoped? Should be added in the comment.

src/core/pratt.ml
fblanqui: Ok. It's good to define such acronyms in a comment.

src/core/basics.ml
```diff
@@ -205,3 +212,77 @@ let term_of_rhs : rule -> term = fun r ->
     TE_Some(Bindlib.unbox (Bindlib.bind_mvar vars p))
   in
   Bindlib.msubst r.rhs (Array.mapi fn r.vars)
+
+(** Total order on alpha-equivalence classes of terms not containing [Patt],
+   [TEnv] or [TRef].
+@raise Invalid_argument otherwise. *)
+let cmp_term : term Lplib.Extra.cmp =
```
fblanqui, line 219: You didn't merge all of master yet, did you? Let me know when you will have merged all the master please. Thank you.

src/core/pratt.ml
fblanqui: In fact, you seem to do scoping after applying Pratt, don't you? I think that you should scope idents before calling Pratt and use fully qualified names for looking in your table.

src/core/basics.ml
gabrielhdt: Ah yes I have, mistakes slipped through

gabrielhdt: It seems to me that the comparison rendered by github is weird

src/core/pratt.ml
gabrielhdt: It would require mutual dependencies between scoping and pratt parsing, if we want to traverse terms only once. Assuming we have term `(x s)`, we have to call `Pratt.parse` from scoping, then call scoping (of ident) from the Pratt parser. It may not be mutual dependent since we restrict to identifiers. But is there a clear advantage?

fblanqui: Pratt parsing applies on sequences of terms. It only looks at the top of terms. In scoping, in case of an application, you could first do a pre-scoping of idents only, that is, if the term (head or argument) is a P_Iden s, call find_qid and replace s by its fully qualified name (if it is a symbol). This could be done by Pratt directly if you provide Pratt with (as additional argument) a function for knowing whether a P_Iden s is a symbol and, if so, what is its fully qualified name. To avoid calling find_qid twice, we could add a new_constructor P_Symb in p_term. This could be later improved in https://github.com/Deducteam/lambdapi/pull/500.

src/core/basics.ml
fblanqui: What do you mean? You do not get the same thing on your machine?

src/core/pratt.ml
gabrielhdt: I'm not sure I understand why you say that Pratt only looks at the _top_ of terms (or what do you mean by _top_?). But on the overall, I agree with your proposition to provide a more elaborate lookup function to Pratt.

fblanqui: In pratt.ml there is no letrec.

gabrielhdt: Ok ok I see. It doesn't recurse into wrapped terms nor binders is a way to put it formally.

src/core/basics.ml
gabrielhdt: I obtain this diff using `git diff master -- src/core/*`
https://copycat.drycat.fr/?a5d32c929080ba32#D2SNs2W78NjEZGtKaXc1HuMHQLXCbTXWDxfNBBc3jJGH

I'm not sure why though

src/core/lp_parser.mly
gabrielhdt: We could indeed.

src/core/basics.ml
gabrielhdt: Ah I understand. As the merges were not obvious, I used `git checkout --patch master <filepath>`, which directly applies patches to files to go incrementally (rather than merging everything at once). The result is that github is lost: the original version in the presented diff seem so concern master branch at the time I forked, not the new one.

My advice would be to use Emacs' ediff, or go with a plain diff file. It's a bit different, but it works quite well. On the other hand I don't know if I'm being a git heretic using that `checkout --patch` or if github is just not as handy as one would hope.

fblanqui: I don't know. I never used --patch; I always use merge. What if you do merge origin/master now?

gabrielhdt: One way may be to create a new pull request... I'll try out later on

gabrielhdt: > I don't know. I never used --patch; I always use merge. What if you do merge origin/master now?

And I don't want to do that now because I'll have the 100+ test files to edit...

fblanqui: If it's about porting the changes introduced by Rehan, a lot can be done automatically:
proof -> := begin
qed -> end
theorem -> opaque symbol
definition -> symbol

fblanqui: Remark: translation to the new parser syntax should be mostly automatic too. Except for proof/begin, you can replace any command <c> by ;<c>.

src/core/lp_parser.mly
```diff
@@ -0,0 +1,409 @@
...
+proof: BEGIN ts=ttactic* pe=proof_end { ts, pe }
+
+inductive:
+  | i=ident COLON t=term ASSIGN VBAR?
+c=separated_list(VBAR, separated_pair(any_ident, COLON, term))
+      { make_pos $loc (fst i, t, c) }
+
+
+// Top level commands
+command:
```
fblanqui, line 293: Could you please remember me and add in a comment why adding a command terminator is necessary? (You do not require this for tactics.)

gabrielhdt: If we don't do so, Menhir isn't happy. In reality, it's rather rare to not need a terminator, and Earley was more the exception

src/core/lp_parser.mly
gabrielhdt: Using square brackets for implicit arguments introduce an ambiguity with enironments and rewrite patterns, so we may keep the curly brackets (but we can use `begin` `end` for unification rules)

fblanqui: ok

src/core/basics.ml
gabrielhdt: It should, but isn't: for instance, rules shouldn't have a terminator if there are other rules with `with`.

fblanqui: I don't understand. What's the problem is you substitute "rule" by ";rule" (except on the first command of a file) and don't change "with"?

gabrielhdt: True, I'll have to try.

gabrielhdt:
As of b75fa3c, all changes of master have been ported to this branch. The diff obtained with `git diff master` is here, https://gist.github.com/gabrielhdt/ecfcfe0e55d357fc82b41d11b2418cf7

amelieled:
:clap: :clap: :clap:

fblanqui:
It seems that this is not completely up to date yet. For instance, the commit 940662a1894bf69b79bca352923d844d95a12270 doesn't seem merged. Anyway, this is kind of problematic to review your PR. Perhaps you should make a new one?

src/core/pratt.ml
gabrielhdt: I'm not sure about which qid you're talking. Operators contain an identifier, the `+` in `set infix "+" := plus`, and a qid, the `plus`. Here we store the identifier, so that when we encounter a `P_iden("+")`, we are able to know that `"+"` is a defined binary operator. Which qid do you want to have when calling `find_qid`?

fblanqui: Do you agree that a first problem is that your string s may be a variable equal to some operator name (e.g. \+,+*+)? To avoid this, you need to scope identifiers in their environment first.
Then, indeed, as operators is a map on unqualified identifiers, there is no need for having a qualified name. Btw, in case of a qualified name, you could improve the code as you don't even need to look in operators.
But what if the string is a qualified name? After all, if "+" is declared infix in some file A. Why not using A.+ as infix in some other file? This may be necessary if, in a single file, we want to use two distinct + at the same time, say A.+ and B.+. If you scope s first, then you will get its fully qualified name if it is a symbol. But, then, you need operators/pp_hints be maps on qualified names. But this point should probably be discussed in another PR.
Another question, that could be implemented in a separate PR too, is: why not replacing set infix/prefix commands by modifiers and, for instance, write `infix left symbol +:N->N->N`? Or do you think it is better to keep those infix/prefix commands?

amelieled: About "another question", we would like to delete `infix/prefix` commands but not now.
`infix left symbol +:N->N->N` is a good idea (Agda has a shorcut `infixl` for `infix left`), but we can do `infix left symbol _+_ : N->N->N` too.

fblanqui:
Continued in #516 .

PR #449, opened 2020-10-20
Dune 2 + Package management in test suite

gabrielhdt:
Upping to dune2.

Test suite now relies on the package files (`lambdapi.pkg`) for requirements, which is cleaner.

The behaviour of `new_lib_mappings` is slightly changed: when the module path is already mapped, we silently do nothing, rather than raising an exception. This will be necessary in the context of interoperability, where we want to compile several files, which will potentially re import several times the same files.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,4 +1,4 @@
-(lang dune 1.11)
+(lang dune 2.1)
 (name lambdapi)
 
 (using menhir 1.0)
--- a/src/core/files.ml
+++ b/src/core/files.ml
@@ -66,13 +66,13 @@ module ModMap :
         (including the root). *)
     exception Already_mapped
 
-    (** [set_root path m] sets the library root of [m] to be [path]. If it was
-        already set, the the exception [Already_mapped] is raised. *)
+    (** [set_root path m] sets the library root of [m] to be [path].
+        @raise Already_mapped if library root of [m] is already set. *)
     val set_root : file_path -> t -> t
 
     (** [add mp path m] extends the mapping [m] by associating the module path
-        [mp] to the file path [path]. In the case where [mp] is already mapped
-        in [m], the exception [Already_mapped] is raised. *)
+        [mp] to the file path [path].
+        @raise Already_mapped when [mp] is already mapped in [m]. *)
     val add : Path.t -> file_path -> t -> t
 
     (** Exception raised if an attempt is made to use the [get] function prior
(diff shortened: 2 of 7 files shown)


dune-project
```diff
@@ -1,4 +1,4 @@
-(lang dune 1.11)
+(lang dune 2.7)
```
ejgallego, line 1: No need to go so far in the version, IMHO 2.1 should work fine for us now.

ejgallego:
Looks good, but I think you need to update the opam file too.

ejgallego:
If you feel like it you can actually have the opam file generate by dune, see https://dune.readthedocs.io/en/stable/opam.html#generating-opam-files

src/core/files.ml
```diff
@@ -210,8 +210,7 @@ let new_lib_mapping : string -> unit = fun s ->
   let file_path = Filename.realpath file_path in
```
fblanqui, line 210: `try ... with Invalid_argument(_)` is missing here

src/core/files.ml
```diff
@@ -210,8 +210,7 @@ let new_lib_mapping : string -> unit = fun s ->
...
   let new_mapping =
     try ModMap.add module_path file_path !lib_mappings
-    with ModMap.Already_mapped ->
-    fatal_no_pos "Module path [%a] is already mapped." Path.pp module_path
+    with ModMap.Already_mapped -> !lib_mappings
```
fblanqui, line 213: It means that the exception Already_mapped becomes useless to raise in the code of add. It would be better to change the code of add instead. The implications are not clear though.

fblanqui: Could you please take advantage of this to fix the typo in the comment on line 70 ("the the")?

fblanqui: @ejgallego made some remark about this part of the code in https://github.com/Deducteam/lambdapi/pull/434#issuecomment-712387097 . @gabrielhdt do you have a better understanding ?

fblanqui: In https://github.com/Deducteam/lambdapi/pull/434#issuecomment-713365488, @gabrielhdt said:

> I think I found out why, and I fixed it in `dune2`, see the description of #449: same module is imported several times in the test suite, because each `compile_file` call will apply the configuration of `lambdapi.pkg` through `apply_config`, each of which calling `new_lib_mapping` which would fail because the mapping would have already been define by a previous compilation.

It seems better to fix your tests/ok_ko.ml and tests/utils.ml files. Instead of calling compile_file, you should better call apply_config only once by doing for instance:
```
(* ok_ko.ml *)
let _ = Files.set_lib_root (Some(".."))
let _ = Package.apply_config "OK/bool.lp"
let compile fname = Compile.compile false (Files.file_to_module fname)
...
```

gabrielhdt: Yes, I agree indeed. But I wonder therefore whether `compile_file` is useful: we can't call it twice, so is it worth it to make function?

gabrielhdt: > @ejgallego made some remark about this part of the code in [#434 (comment)](https://github.com/Deducteam/lambdapi/pull/434#issuecomment-712387097) . @gabrielhdt do you have a better understanding ?

No, unfortunately, I don't.

gabrielhdt:
And by the way, line 290 of `files.ml` seems completely useless, doesn't it?

fblanqui:
Right. Can you remove it please?

PR #453, opened 2020-10-28
[lsp] Don't error fatally for symbols without a location.

### LINKED ISSUE #451: regression bug with emacs lsp mode
it seems that tests/OK/unif_hint.lp (unification hint via ``set unif_rule <patt> ≡ <patt> ↪ <term> ≡ <term>``) doesn't work anymore with emacs mode since #434

are you able to reproduce this or is it just me ?

### LINKED ISSUE #452: [lsp server] column offsets don't follow LSP specification mandated UTF-16, creating problems in emacs client
@amelieled 
Hi ! Do you use emacs or vscode ?
Have you been able to use emacs lsp mode on tests/OK/inductive_type.lp
(it's freezing on my laptop)
thanks

can someone reproduce this behaviour or is it just me ?

PS : not related to #451 as neither 9e080cd7 4a54066f 8c5755f6 74de6107 seem to work

ejgallego:
This was some leftover cleanup from #434 , the code here was not
robust.

We should anyways overhaul the whole error handling code of the server
when we have a bit of time, as it is pretty primitive owning to its
prototypical nature. Errors like this should not be so silent.

Fixes #451 , fixes #452

### DIFF
--- a/src/lsp/lp_doc.ml
+++ b/src/lsp/lp_doc.ml
@@ -24,12 +24,6 @@ let interval_of_pos : Pos.pos -> Range.t = fun p ->
   let finish : point = make_point data.end_line data.end_col in
   make_interval start finish
 
-let interval_of_popt : Pos.popt -> Range.t = fun p ->
-  match p with
-  | None -> invalid_arg "Position option with 'None' can't be converted
-      to interval"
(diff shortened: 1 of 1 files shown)


fblanqui:
Thanks @ejgallego

PR #454, opened 2020-10-31
[WIP] Check typing before metavariable instanciation

Rehan-MALAK:
Currently in master, when there is a meta ``?m : a`` and we would like instantiate ``m := t``, we do the typechecking ``t : a`` only in the refine tactic.

@barras @fblanqui suggested that we should do it whenever there is an instantiation (in lambdapi, it's only in the unification algorithm and it's disjoint from the inference algorithm) as there are unifications and instantiations all the time, with symbols and definitions, not just with theorems

this is especially important before considering #422 (unification in proof mode used for both symbol, definition, and theorem) 

it turns out that typechecking in the unification algorithm revealed several subtle bugs in context propagation (during unification but also during inference)

also , typechecking during unification, makes the role of the precomputed meta type ``a`` (and the associated constraints) much more important. We can't initialise always with dummy types as we used to do in scoping before : ``?m:Kind`` or ``?m:?n && ?n:Type`` or ...

### DIFF
--- a/src/core/eval.ml
+++ b/src/core/eval.ml
@@ -218,7 +218,8 @@ and tree_walk : dtree -> ctxt -> stack -> (term * stack) option =
         List.iter fn env_builder;
         (* Complete the array with fresh meta-variables if needed. *)
         for i = env_len - xvars to env_len - 1 do
-          let t = make_meta [] Kind in
+          let x = make_meta [] Type in
+          let t = make_meta [] x in
           let b = Bindlib.raw_mbinder [||] [||] 0 mkfree (fun _ -> t) in
           env.(i) <- TE_Some(b)
         done;
--- a/src/core/eval.ml
+++ b/src/core/eval.ml
@@ -418,3 +418,12 @@ let eval : Syntax.eval_config -> ctxt -> term -> term = fun c ctx t ->
   | (HNF , None   ) -> hnf ctx t
   (* TODO implement the rest. *)
   | (_   , Some(_)) -> wrn None "Number of steps not supported."; t
+
+(** Function comparing two constraints *)
+let eq_constr : constr -> constr -> bool = fun (ctx1,ta1,tb1) (ctx2,ta2,tb2) ->
+  let ta1,_ = Ctxt.to_prod ctx1 ta1 in
+  let tb1,_ = Ctxt.to_prod ctx1 tb1 in
+  let ta2,_ = Ctxt.to_prod ctx2 ta2 in
+  let tb2,_ = Ctxt.to_prod ctx2 tb2 in
+  (eq_modulo [] ta1 ta2) && (eq_modulo [] tb1 tb2) ||
+  (eq_modulo [] ta1 tb2) && (eq_modulo [] tb1 ta2)
--- a/src/core/eval.ml
+++ b/src/core/eval.ml
@@ -111,7 +111,7 @@ and whnf_stk : ctxt -> term -> stack -> term * stack = fun ctx t stk ->
       | Some(t) -> Stdlib.incr steps; whnf_stk ctx t stk
       | None    ->
       (* Otherwise try rewriting using decision tree. *)
-      match tree_walk !(s.sym_tree) ctx stk None with
+      match tree_walk !(s.sym_tree) ctx stk with
       (* If no rule is found, return the original term *)
       | None        -> st
       | Some(t,stk) -> Stdlib.incr steps; whnf_stk ctx t stk
@@ -179,9 +179,8 @@ and eq_modulo : ctxt -> term -> term -> bool = fun ctx a b ->
     fails,  the stack [stk] may be imperatively updated since a reduction step
     taken in elements of the stack is preserved (this is done using
     {!constructor:Terms.term.TRef}). *)
-and tree_walk :
-  dtree -> ctxt -> stack -> term option -> (term * stack) option =
-  fun tree ctx stk meta_type ->
+and tree_walk : dtree -> ctxt -> stack -> (term * stack) option =
+  fun tree ctx stk ->
   let (lazy capacity, lazy tree) = tree in
   let vars = Array.make capacity Kind in (* dummy terms *)
   let bound = Array.make capacity TE_None in
@@ -218,16 +217,8 @@ and tree_walk :
         in
         List.iter fn env_builder;
         (* Complete the array with fresh meta-variables if needed. *)
-        let make_t =
-          begin
-            match meta_type with
-            | Some meta_type -> (function () -> make_meta [] meta_type)
-            | None -> (function () -> let x = make_meta [] Type in
-                make_meta [] x)
-          end
-        in
         for i = env_len - xvars to env_len - 1 do
-          let t = make_t () in
+          let t = make_meta [] Kind in
           let b = Bindlib.raw_mbinder [||] [||] 0 mkfree (fun _ -> t) in
           env.(i) <- TE_Some(b)
         done;
@@ -415,6 +406,15 @@ let rec hnf : ctxt -> term -> term = fun ctx t ->
      let x,t = Bindlib.unbind t in
      Abst(a, Bindlib.unbox (Bindlib.bind_var x (lift (hnf ctx t))))
   | t         -> t
+
+(** [type_app a ts] returns the type of [add_args x ts] where [x] is any term of type [a], if it exists. *)
+let rec type_app : ctxt -> term -> term list -> term option =
+  fun ctx a ts ->
+  match (whnf ctx a), ts with
+  | Prod(_,b), t :: ts -> type_app ctx (Bindlib.subst b t) ts
+  | _, [] -> Some a
+  | _, _ -> None
+
 
 (** [eval cfg ctx t] evaluates the term [t] in the context [ctx] according to
     configuration [cfg]. *)
--- a/src/core/eval.ml
+++ b/src/core/eval.ml
@@ -420,10 +420,10 @@ let eval : Syntax.eval_config -> ctxt -> term -> term = fun c ctx t ->
   | (_   , Some(_)) -> wrn None "Number of steps not supported."; t
 
 (** Function comparing two constraints *)
-let eq_constr : constr -> constr -> bool = fun (ctx1,ta1,tb1) (ctx2,ta2,tb2) ->
-  let ta1,_ = Ctxt.to_prod ctx1 ta1 in
-  let tb1,_ = Ctxt.to_prod ctx1 tb1 in
-  let ta2,_ = Ctxt.to_prod ctx2 ta2 in
-  let tb2,_ = Ctxt.to_prod ctx2 tb2 in
-  (eq_modulo [] ta1 ta2) && (eq_modulo [] tb1 tb2) ||
-  (eq_modulo [] ta1 tb2) && (eq_modulo [] tb1 ta2)
+let eq_constr : constr -> constr -> bool = fun (ctx1,t1,u1) (ctx2,t2,u2) ->
+  let t1,_ = Ctxt.to_prod ctx1 t1 in
+  let u1,_ = Ctxt.to_prod ctx1 u1 in
+  let t2,_ = Ctxt.to_prod ctx2 t2 in
+  let u2,_ = Ctxt.to_prod ctx2 u2 in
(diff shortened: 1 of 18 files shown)


src/core/unif.ml
```diff
@@ -68,12 +68,23 @@ let try_rules : ctxt -> term -> term -> constr list option = fun ctx s t ->
...
   try
+    let a_s,constrs_s = Infer.infer ctx s in
+    let a_t,constrs_t = Infer.infer ctx t in
+    let meta_type =
+      match constrs_s,constrs_t with
+      | ([],_) -> Some a_s
+      | (_,[]) -> Some a_t
+      | (_,_) ->
+        if !log_enabled then
+          log_unif (yel "tree walk with defaut meta array");
```
fblanqui, line 79: defaut -> default

src/core/unif.ml
```diff
@@ -162,19 +173,53 @@ let instantiation : ctxt -> meta -> term array -> term ->
...
+let g_type_check = Stdlib.ref TypeCheckInstanciation
+
 (** [instantiate ctx m ts u] check whether, in a problem [m[ts]=u], [m] can be
     instantiated and, if so, instantiate it. *)
-let instantiate : ctxt -> meta -> term array -> term -> bool =
-  fun ctx m ts u ->
+let instantiate : ctxt -> meta -> term array ->
+  term -> constr list -> bool =
+  fun ctx m ts u initial ->
+  let type_meta_app m ts =
```
fblanqui, line 185: move this to basics.ml:
```
(** [type_app a ts] returns the type of [add_args x ts] where [x] is any term of type [a], if it exists. *) 
let rec type_app : ctxt -> term -> term list -> term option =
      match (Eval.whnf ctx a), ts with
      | Prod(_,b), t :: ts -> type_app (Bindlib.subst b t) ts
      | _, [] -> Some a
      | _, _ -> None
```
Then replace `    let m_app = type_meta_app m (Array.to_list ts) in` by:
```
    let m_app =
      match type_app ctx !(m.meta_type) (Array.to_list ts) with
      | Some a -> a
      | None -> assert false
  in
```
```

File: src/core/unif.ml [hunk 3]
```diff
@@ -184,19 +229,32 @@ let rec solve : problem -> constr list = fun p ->
...
+    else
+      match try_rules ctx t1 t2 with
+      | None     -> solve {p with unsolved = (ctx,t1,t2) :: p.unsolved}
+      | Some([]) -> assert false
+      (* Unification rules generate non empty list of unification
+         constraints *)
+      | Some(cs) ->
+        let lost_constrs = []
+(*
+          try (* TODO to get constraints on the type of implicit argument *)
```
fblanqui, line 251: what are those lines (let and comment) for?

src/core/unif.ml
```diff
@@ -207,9 +265,9 @@ and solve_aux : ctxt -> term -> term -> problem -> constr list =
 
   let decompose () =
     (* Propagate context *)
-    let add_arg_pb l a b = (ctx,a,b)::l in
+    let add_arg_pb a b l = (ctx,a,b)::l in
     let to_solve =
-      try List.fold_left2 add_arg_pb p.to_solve ts1 ts2
+      try List.fold_right2 add_arg_pb ts1 ts2 p.to_solve
```
fblanqui, line 270: Why do you change fold_left to fold_right (which is not tail-recursive)?

tests/utils.ml
```diff
@@ -58,5 +58,4 @@ let _ =
   let open Alcotest in
   run "Utils" [ ("hrs", [test_case "bool" `Quick test_hrs])
               ; ("xtc", [test_case "bool" `Quick test_xtc])
-              ; ("dtree", [ test_case "bool" `Quick test_dtree
-                          ; test_case "ghost" `Quick test_dtree_ghost ]) ]
```
fblanqui, line 62: perhaps keep in comment saying that it needs to be fixed

tests/ok_ko.ml
```diff
@@ -18,7 +18,10 @@ let _ =
   | None -> assert false
   | Some(f) -> Package.apply_config f;
   let open Alcotest in
-  let files = Sys.readdir "OK" |> Array.map (fun f -> "OK/" ^ f) in
+  let files = Sys.readdir "OK" |> Array.map (fun f -> "OK/" ^ f)
+  |> Array.to_list
+  |> List.filter (function f -> f <> "OK/unif_hint.lp")
+  |> Array.of_list in
```
fblanqui, line 24: thanks to add a comment saying that this is temporary
thanks to use fun instead of function everywhere, for uniformity

src/core/eval.ml
```diff
@@ -179,8 +179,9 @@ and eq_modulo : ctxt -> term -> term -> bool = fun ctx a b ->
     fails,  the stack [stk] may be imperatively updated since a reduction step
     taken in elements of the stack is preserved (this is done using
     {!constructor:Terms.term.TRef}). *)
-and tree_walk : dtree -> ctxt -> stack -> (term * stack) option =
-  fun tree ctx stk ->
+and tree_walk :
+  dtree -> ctxt -> stack -> term option -> (term * stack) option =
```
fblanqui, line 183: Thanks to update the comment  of tree_walk

src/core/eval.ml
```diff
@@ -217,8 +218,16 @@ and tree_walk : dtree -> ctxt -> stack -> (term * stack) option =
         in
         List.iter fn env_builder;
         (* Complete the array with fresh meta-variables if needed. *)
+        let make_t =
+          begin
+            match meta_type with
+            | Some meta_type -> (function () -> make_meta [] meta_type)
+            | None -> (function () -> let x = make_meta [] Type in
```
fblanqui, line 225: why not puting () outside: `let make_t () = ....` ?
```
let make_t () =
  match meta_type with
    (* FIXME? why using [] and not ctxt ? *)
            | Some meta_type -> make_meta [] meta_type
            | None -> make_meta [] (make_meta [] Type)
in
```
@gabrielhdt what do you think of using ctxt instead of []?

src/core/scope.ml
```diff
@@ -194,29 +194,35 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
...
+  (* Create a new metavariable of type [TYPE] for a missing domain. *)
+  and _Meta_Type : env -> tbox = fun env ->
+    let vs = Env.to_tbox env in
+    let a = Env.to_prod_box env _Type in
+    let m = _Meta_full (fresh_meta_box a (Array.length vs)) vs in
+    (* Sanity check: only variables of [env] free in [m] if not in RHS. *)
+    match md with
+    | M_RHS(_) -> m
+    | _        ->
+      assert (Bindlib.is_closed (Bindlib.bind_mvar (Env.vars env) m)); m
```
fblanqui, line 206: Is this check really useful? If it fails, it means that there is a bug in Bindlib, right?

src/core/eval.ml
fblanqui: Actually, I don't understand what is this metatype argument. Could you please explain?

src/core/unif.ml
```diff
@@ -68,12 +68,23 @@ let try_rules : ctxt -> term -> term -> constr list option = fun ctx s t ->
   let exception No_match in
   let open Unif_rule in
   try
+    let a_s,constrs_s = Infer.infer ctx s in
+    let a_t,constrs_t = Infer.infer ctx t in
```
fblanqui, line 72: I don't understand why you need to do type inference to apply unif_rules.

fblanqui: As already mentioned, in fact, I don't understand why you need to change tree_walk by adding an extra argument. If those two things are an attempt to fix the problem with `set unif_rule fst $p ≡ $f ↪ $p ≡ pair $f $s`, they should be removed as I think this is a bug in the handling of unif_rules.

Rehan-MALAK: I will removed everything in the PR related to set unif_rule and we will think about it in another PR later

It was indeed an attempt (but not sufficient) to fill an array in tree_walk with the correct type instead of Kind, but this has to be investigated further

src/core/unif.ml
Rehan-MALAK: here the rationale is to decompose f a b = f a' b' to a = a'  and b = b' instead of b = b' and a = a'

keeping the order, the algorithm study the equations related to implicit arguments first

src/core/unif.ml
Rehan-MALAK: this part has been removed

src/core/unif.ml
Rehan-MALAK: can't use Eval in Basics....

I have put type_app in Eval.... WDYT ?

src/core/unif.ml
Rehan-MALAK: this part has been removed

tests/utils.ml
Rehan-MALAK: fixed

tests/ok_ko.ml
Rehan-MALAK: fixed

src/core/eval.ml
Rehan-MALAK: this part has been removed 

the rationale was to fill an array in tree_walk with a precomputed type instead of a meta of type Kind (cf also comment below)

src/core/eval.ml
Rehan-MALAK: this part has been removed... but the question of the context : ctxt vs []  has remains relevant

src/core/unif.ml
fblanqui: Ok. And you should open another issue on this topic: the application of unification rules does not create well typed metavariables. @gabrielhdt

src/core/unif.ml
fblanqui: Thanks to add a comment here. Indeed, with dependent types, it is a good idea to keep the order.

src/core/unif.ml
fblanqui: Ah, right. What about putting it in infer.ml as it is related to type inference?

src/core/eval.ml
```diff
@@ -407,6 +407,16 @@ let rec hnf : ctxt -> term -> term = fun ctx t ->
      Abst(a, Bindlib.unbox (Bindlib.bind_var x (lift (hnf ctx t))))
   | t         -> t
 
+(** [type_app a ts] returns the type of [add_args x ts] where [x] is any
```
fblanqui, line 410: I suggest to move this function to infer.ml if that is possible.

tests/ok_ko.ml
Rehan-MALAK: why github doesn't write yellow circled Outdated ?

src/core/eval.ml
Rehan-MALAK: fixed

src/core/unif.ml
Rehan-MALAK: I see that you have already done that. Thanks

src/core/unif.ml
Rehan-MALAK: fixed

src/core/unif.ml
Rehan-MALAK: fixed

fblanqui:
@Rehan-MALAK Travis failed. Could you please fix this?

fblanqui:
@gabrielhdt we got no feedback on which test has failed :-(

gabrielhdt:
It's really strange, locally, when I execute `make tests` it returns code exit 2, and when I run commands sequentially, I only have 0 exit codes.

gabrielhdt:
I found a strange workaround: put line `dune exec -- tests/dtrees.sh` before `dune test`. I'm afraid I don't understand what happens.

fblanqui:
@gabrielhdt could you please open an issue about this problem?
@Rehan-MALAK could you please implement this workaround in your PR so that we can merge it?

gabrielhdt:
Oops, my bad, it doesn't solve anything, it comes from an error on my side.

gabrielhdt:
The problem is solved in #459 (the script would exit silently if lamdapi command failed, before saying it failed).

fblanqui:
@Rehan-MALAK I just merged #459. Thanks to update your PR.

fblanqui:
Thanks @Rehan-MALAK !

PR #457, opened 2020-11-04
Purifying

### LINKED ISSUE #450: hangover OK KO tests with alcotest test suite
Dear all !
We used to test each OK / KO test as separate shell commands.
Now, if I understand #449 correctly, the ok_ko.ml is a big executable linked against lambdapi 
(Core module) so one can theoretically do unit tests against the ocaml code.

We are still using these OK / KO tests doing I/O, parsing, scoping, typing.... to check inference, unification, tactics... as if they were unit tests without side effect. Maybe this confusion is a pragmatic choice as it is more enjoyable to write lambdapi files than unit test ocaml code.
And we can still add later proper unit tests.
  
But for now, each OK / KO test has consequences on the following ones.

For example, if I add the following tests/OK/univ2.lp
```
set debug +ui
```
all, the following tests will be done with ``set debug +ui`` (and tests/OK/triangular.lp will take ages)

And we could imagine much much more problematics examples...

gabrielhdt:
This pull request proposes an attempt to provide a pure interface. The first step is to modify only entry points, capturing side effects.

Fixes #450

### DIFF
--- a/src/core/compile.ml
+++ b/src/core/compile.ml
@@ -120,13 +120,21 @@ let _ =
   in
   Stdlib.(Parser.require := require)
 
-(** Pure wrappers around compilation functions. *)
+(** Pure wrappers around compilation functions. Functions provided perform the
+    same computations as the ones defined earlier, but restores the state when
+    they have finished. An optional state can be passed as argument to chnage
+    the settings. *)
 module Pure : sig
-  val compile : bool -> Path.t -> Sign.t
-  val compile_file : file_path -> Sign.t
+  val compile : ?st:Console.state -> bool -> Path.t -> Sign.t
+  val compile_file : ?st:Console.state -> file_path -> Sign.t
 end = struct
-  let app2 : ('a -> 'b -> 'c) -> ('a * 'b) -> 'c = fun f (x, y) -> f x y
+  let compile ?st force path =
+    let f (force, path) =
+      Option.iter apply_state st;
+      compile force path
+    in
+    pure_apply f (force, path)
 
-  let compile force path = pure_apply (app2 compile) (force, path)
-  let compile_file = pure_apply compile_file
+  let compile_file ?st =
+    pure_apply (fun f -> Option.iter apply_state st; compile_file f)
 end
--- a/src/core/compile.ml
+++ b/src/core/compile.ml
@@ -119,3 +119,14 @@ let _ =
     with e -> Console.pop_state (); raise e
   in
   Stdlib.(Parser.require := require)
+
+(** Pure wrappers around compilation functions. *)
+module Pure : sig
+  val compile : bool -> Path.t -> Sign.t
+  val compile_file : file_path -> Sign.t
+end = struct
+  let app2 : ('a -> 'b -> 'c) -> ('a * 'b) -> 'c = fun f (x, y) -> f x y
+
+  let compile force path = pure_apply (app2 compile) (force, path)
+  let compile_file = pure_apply compile_file
+end
--- a/src/core/compile.ml
+++ b/src/core/compile.ml
@@ -122,19 +122,33 @@ let _ =
 
 (** Pure wrappers around compilation functions. Functions provided perform the
     same computations as the ones defined earlier, but restores the state when
-    they have finished. An optional state can be passed as argument to chnage
-    the settings. *)
+    they have finished. An optional library mapping or state can be passed as
+    argument to change the settings. *)
 module Pure : sig
-  val compile : ?st:State.t -> bool -> Path.t -> Sign.t
-  val compile_file : ?st:State.t -> file_path -> Sign.t
+  val compile : ?lm:string -> ?st:State.t -> bool -> Path.t -> Sign.t
+  val compile_file : ?lm:string -> ?st:State.t -> file_path -> Sign.t
 end = struct
-  let compile ?st force path =
-    let f (force, path) =
-      Option.iter State.apply st;
-      compile force path
-    in
-    pure_apply f (force, path)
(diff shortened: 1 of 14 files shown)


src/core/console.ml
```diff
@@ -221,31 +221,34 @@ let reset_default : unit -> unit = fun () ->
...
   StrMap.iter reset Stdlib.(!boolean_flags)
 
+(** Settings used to compile files. *)
+type state = { st_verbose: int
+             (** Verbosity level. *)
+             ; st_loggers: (char * bool) list
+             (** Loggers enabled. *)
+             ; st_bflags: bool StrMap.t
+             (** Boolean flags. *) }
+
```
ejgallego, line 231: Nit: wrapping records into a module is common style in some OCaml projects, so that you do
```ocaml
module State = struct
   type t = { verbose : int
                ; ...
```
This has quite a few advantages and the compiler also provides some special support for record scoping.

gabrielhdt: We can even bring `apply_state`, `push_state` &c into the module in that case I guess

src/core/compile.ml
```diff
@@ -109,13 +109,46 @@ let compile_file : file_path -> Sign.t = fun fname ->
...
+module Pure : sig
+  val compile : ?lm:string -> ?st:State.t -> bool -> Path.t -> Sign.t
+  val compile_file : ?lm:string -> ?st:State.t -> file_path -> Sign.t
+end = struct
+
+  (* [pure_apply_cfg ?lm ?st f] is function [f] but pure (without side
+     effects).  The side effects taken into account occur in
+     {!val:Console.State.t}, {!val:Files.lib_mappings} and in the meta
+     variable counter {!module:Terms.Meta}. Arguments [?lm] allows to set the
+     libray mappings and [?st] sets the state. *)
```
fblanqui, line 136: typo: libray -> library

gabrielhdt:
If I understand well, this pull request kind of generates redundancy with the `Pure` module, isn't it correct @ejgallego?

src/core/compile.ml
```diff
@@ -109,13 +109,46 @@ let compile_file : file_path -> Sign.t = fun fname ->
...
+  let pure_apply_cfg : ?lm:string -> ?st:State.t -> ('a -> 'b) -> 'a -> 'b =
+    fun ?lm ?st f x ->
+    let libmap = !lib_mappings in
+    State.push ();
+    Option.iter new_lib_mapping lm;
+    Option.iter State.apply st;
+    let res = f x in
+    State.pop ();
+    Terms.Meta.reset_key_counter ();
```
fblanqui, line 145: Ideally, this counter should be reset at the beginning of each top-level command, so in handle_cmd when proof_data = None. Could this be done?

gabrielhdt: Are you happy with 694e8f4?

src/core/handle.ml
```diff
@@ -436,6 +436,7 @@ let handle_cmd : sig_state -> p_command -> sig_state * proof_data option =
     let (tm, ss) = time (handle_cmd ss) cmd in
     if Stdlib.(tm >= !too_long) then
       wrn cmd.pos "It took %.2f seconds to handle the command." tm;
+    begin match snd ss with None -> Meta.reset_key_counter () | _ -> () end;
```
fblanqui, line 439: Why not doing this on line 436 in handle.ml or in line 62 in compile.ml (no need to test if there is a proof_data in both cases)?

gabrielhdt: Ah so I think I do not understand your comment about `proof_data`, since there is no `proof_data` on line 436.

fblanqui: Right. There is need for testing proof_date, contrary to what I said in my first comment. But after looking at the code in more details, I realized that there was no need to have and test a proof data. Does it make sense? Now I don't know where it is best to reset: in compile or in handle. What do you think?

gabrielhdt: I think it's better to reset in handle, so that compile only wraps `handle_cmd` to take care of exceptions, and so we know clearly who does what. Does it make sense?

gabrielhdt: Well reading again the code, not really since `handle` in compile already performs computation related to tactics.

gabrielhdt: Maybe we should ask @ejgallego, since he may use `handle_cmd` to parse incrementally a file. For us, it doesn't matter much since we always compile files completeley.

fblanqui: So what do you propose finally?

gabrielhdt: We can place it in `compile`, to keep side effects closer to entry points (and avoid putting them deep into computations)

fblanqui:
Thanks @gabrielhdt

ejgallego:
> If I understand well, this pull request kind of generates redundancy with the `Pure` module (the `Pure` library of lambdapi), isn't it correct @ejgallego?

Indeed, I think a bit. Ideally we would make things converge a bit more, as you have observed the APIs have different requirements, for example w.r.t. incremental building.

In the UI side, we really want the exposed API to be functional, as it is common to perform speculative execution, work on proofs in parallel, etc... There are some tradeoffs tho, for example, debug and top-level options such as paths etc... can be set once when creating the initial state.

Keep in mind that an incremental document processing engine works as a DAG, so where to place data that is susceptible of change is a bit again of a compromise. For example, if we add paths in the initial state, a workspace modification would invalidate the deps of the initial state so the whole document would have to be rebuilt; as a gain, you get a nicer DAG. In this case it seems sensible to do so, but our mileage may vary.

PR #459, opened 2020-11-04
Fixing test script

### LINKED ISSUE #458: Test script failing silently
It appeared in pull request #454 that `make tests` can fail while having all checks succeeding. In the aforementioned case.

gabrielhdt:
Fixes #458


fblanqui:
Thanks @gabrielhdt

PR #465, opened 2020-11-05
Unification rule application

### LINKED ISSUE #456: Bug in unification rule application
Unification rules can contain in their RHS pattern variables that do not occur in the LHS. At rule application time, these extra pattern variables are replaced by fresh metavariables. Currently, those freshly created metavariables are declared of type KIND. This is incorrect. They should be of type yet another metavariable of type TYPE, like it is done in the scoping of underscores in terms.

gabrielhdt:
Fixes #456.

### DIFF
--- a/src/core/eval.ml
+++ b/src/core/eval.ml
@@ -218,7 +218,8 @@ and tree_walk : dtree -> ctxt -> stack -> (term * stack) option =
         List.iter fn env_builder;
         (* Complete the array with fresh meta-variables if needed. *)
         for i = env_len - xvars to env_len - 1 do
-          let t = make_meta [] Kind in
+          let mt = make_meta ctx Type in
+          let t = make_meta ctx mt in
           let b = Bindlib.raw_mbinder [||] [||] 0 mkfree (fun _ -> t) in
           env.(i) <- TE_Some(b)
         done;


gabrielhdt:
Actually the test in `unif_hint` still fails on constraints such as `z ≡ (?37:(?36:TYPE))`, is it normal? Shouldn't the algorithm be able to instantiate `?37` by `z` and `?36` by `Nat`?

fblanqui:
The problem is that, after #454, a metavariable M can be instantiated by u only if u has the type expected for M. In your example, we are left with the constraints 19=nat and 37=z, where 19:18 and 37:36, but we have no constraint saying that 18=U and 36=Nat.
After rewriting the constraint z = fst nat (pair 33 35 z) into pair 33 35 z = pair 19 z 37, we decompose and get 33=19, 35=z and z=37.
We could solve the problem if decompose added also equalities on types: 32=18, 34=Nat, Nat=36, because there is also a constraint 32=U.

fblanqui:
For the record, the file unif_hint.lp can be found here: https://github.com/Deducteam/lambdapi/blob/89ab81c8ceef62262ac850ace4d2a8cd71560be4/tests/OK/unif_hint.lp

fblanqui:
Conclusion: solving this problem requires to think again on the unification algorithm and modify it...

fblanqui:
@gabrielhdt Do you agree to merge this PR right away?

gabrielhdt:
Yes

Issue #466, opened 2020-11-05
Missing constraints in a unification problem


fblanqui:
See https://github.com/Deducteam/lambdapi/pull/465.

fblanqui:
https://github.com/Deducteam/lambdapi/blob/89ab81c8ceef62262ac850ace4d2a8cd71560be4/tests/OK/unif_hint.lp fails because, after #454, a metavariable M can be instantiated by u only if u has the type expected for M. In unif_hint.lp, we are left with the constraints 19=nat and 37=z, where 19:18 and 37:36, but we have no constraint saying that 18=U and 36=Nat.
After rewriting the constraint z = fst nat (pair 33 35 z) into pair 33 35 z = pair 19 z 37, we decompose and get 33=19, 35=z and z=37.
We could solve the problem if decompose added also equalities on types: 32=18, 34=Nat, Nat=36, because there is also a constraint 32=U.

PR #467, opened 2020-11-05
Fixing scoping for unification rules

### LINKED ISSUE #455: Bug in unification rule scoping
When scoping the RHS of a unification rule, underscores (and thus implicit arguments) should be substituted by fresh pattern variables.
The rationale is that a rule (including a unification rule) added in a signature must not contain any metavariable.

gabrielhdt:
Fixes #455. Introduce a new scoping mode for unification rules.

### DIFF
--- a/src/core/scope.ml
+++ b/src/core/scope.ml
@@ -76,19 +76,22 @@ type mode =
       { m_rhs_prv             : bool
       (** True if {!constructor:Terms.expo.Privat} symbols are allowed. *)
       ; m_rhs_data            : (string, tevar) Hashtbl.t
-      (** Environment for variables that we know to be bound in the RHS. *)
-      ; mutable m_rhs_vars_nb : int
+      (** Environment for variables that we know to be bound in the RHS. *) }
+  (** Scoping mode for rewriting rule right-hand sides. *)
+  | M_URHS of
+      { mutable m_urhs_vars_nb : int
       (** Number  of  distinct  variables  in  the rewriting  rule,  including
           variables  only in  the  RHS. It  is initialised  to  the number  of
           (distinct) variables in the LHS and incremented each time a variable
           of the RHS that was not in the LHS is scoped. *)
-      ; mutable m_rhs_xvars   : (string * tevar) list
+      ; mutable m_urhs_xvars : (string * tevar) list
       (** Variables scoped that  were not in the LHS. This  field is only used
           in  unification  rules and  is  updated  imperatively for  each  new
           variable scoped. A couple [(n, v)]  is the name of the variable with
           the variable itself. The name is needed to ensure that two variables
-          with the same name are scoped as the same variable. *) }
-  (** Scoping  mode for  rewriting rule right-hand  sides. During  scoping, we
+          with the same name are scoped as the same variable. *)
+      ; m_urhs_data : (string, tevar) Hashtbl.t }
+  (** Scoping mode for unification rule right-hand sides. During  scoping, we
       always have [m_rhs_vars_nb = m_lhs_size + length m_rhs_xvars]. *)
 
 (** [get_implicitness t] gives the specified implicitness of the parameters of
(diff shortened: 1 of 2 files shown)


src/core/scope.ml
```diff
@@ -253,6 +256,15 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
...
         find_qid false false ss env qid
+    | (P_Wild, M_URHS(data)) ->
+      let x =
+        let name = Printf.sprintf "v%i" data.m_urhs_vars_nb in
+        let x = Bindlib.new_var te_mkfree name in
+        data.m_urhs_vars_nb <- data.m_urhs_vars_nb + 1;
+        data.m_urhs_xvars <- (name, x) :: data.m_urhs_xvars;
+        x
+      in
+      _TEnv (Bindlib.box_var x) [||]
```
fblanqui, line 267: Why is the array argument empty? You should take the environment into account. After the doc: "_ for an unknown term or a term we don’t care about. It is replaced by a fresh metavariable (or a fresh pattern variable in a rule left-hand side) applied to all the variables of the context." For instance, in `\x:N,\Pi y:_,a`, _ is equivalent to `&Y[x]` where Y is fresh.

src/core/scope.ml
```diff
@@ -499,9 +529,9 @@ type pre_rule =
   ; pr_xvars_nb : int
   (** Number of variables that appear in the RHS but not in the LHS. *) }
 
-(** [scope_rule ss r] turns a parser-level rewriting rule [r] into a rewriting
+(** [scope_rule u ss r] turns a parser-level rewriting rule [r] into a rewriting
```
fblanqui, line 532: u -> ur, like in the code below
explanation of ur missing

src/core/scope.ml
gabrielhdt: Indeed it's a mistake.

src/core/scope.ml
```diff
@@ -499,9 +529,10 @@ type pre_rule =
   ; pr_xvars_nb : int
   (** Number of variables that appear in the RHS but not in the LHS. *) }
 
-(** [scope_rule ss r] turns a parser-level rewriting rule [r] into a rewriting
-    rule (and the associated head symbol). *)
-let scope_rule : sig_state -> p_rule -> pre_rule loc = fun ss r ->
+(** [scope_rule ur ss r] turns a parser-level rewriting rule [r], or a
+    unification rule if [ur] is true, into a rewriting rule (and the
```
fblanqui, line 533: into a pre-rule

fblanqui:
Thanks to add Rehan's example in tests/OK to check the PR.

gabrielhdt:
His example did not check anything, or only visually. The file `unif_hint.lp` still fails because of #466. So there is no systematic way to check that it works. To convince yourself, you can run `dune exec -- lambdapi check tests/OK/unif_hint.lp --verbose 3` with `set flag "print_implicits" on` in the file to see that there is no meta anymore.

Rehan-MALAK:
Just to clarify :
-there was no new example, it was just a shortened version of unif_hint.
-unif_hint needs at least three things (and maybe more) to work when we typecheck meta instanciation : change Kind for the correct type (#465), change scoping (#467) and unification is missing constraints (#466)

we removed #465 and #466 during review of #454 because :
- the way I was adding constraints had to be investigated more (it was a bit artificial for this specific example)
- I had no modifications for rule RHS as it's done in #467

fblanqui:
It's not only visual. You should actually be able to check that it computes correctly. Just add a conversion test between the lhs and the rhs. This test should fail in master.

gabrielhdt:
I'm afraid I can't manage to create such an example. I tried something like
```
symbol T: TYPE
injective symbol El: T → TYPE
symbol iota: T
symbol f {t: T} (x: El t): El t
symbol g {t: T} (x: El t): El t
symbol h {t: T} (x: El t): El t
set unif_rule f $x ≡ g $y ↪ h $x ≡ h $y
symbol x: El iota
rule h {iota} _ ↪ x
constant symbol eq {t: T}: El t → El t → TYPE
symbol refl {t: T} (x: El t): eq x x
set flag "print_implicits" on
assert refl (f x): eq (f x) (g x)
```

but in both cases, I end up with `Cannot solve h ?1 x ≡ h ?3 x.`. So the correction seems tricky to test since occurrences of the bug were likely shadowed by the fact that the type checker couldn't infer an implicit argument?

fblanqui:
Right. Currently we cannot ask to evaluate a term without type checking it first, and unification rules use symbols inaccessible to users and with no type.

fblanqui:
@gabrielhdt could you please update this branch so that we can merge it ?

gabrielhdt:
It is updated, isn't it?

gabrielhdt:
Sorry, it wasn't indeed, it is now.

fblanqui:
Thanks @gabrielhdt

PR #469, opened 2020-11-11
[package] Workaround realpath failing with non-existing files

### LINKED ISSUE #432: Emacs LSP mode with goals display is freezing with new files
Hi !

If I do 

```bash
echo "constant symbol A : TYPE" > tests/OK/new.lp
```

then I open emacs with the usual ``.emacs`` configuration

```emacs
(require 'opam-user-setup "~/.emacs.d/opam-user-setup.el")
(load "lambdapi-site-file")
```

I can open emacs in lambdapi lsp mode with the goals window

```bash
emacs tests/OK/new.lp
```
and do some lambdapi development.

But if I start a new file 

```bash
emacs tests/OK/new2.lp
```
emacs freezes when I try to type

```lambdapi
constant symbol A : TYPE
```
and in fact anything followed by ``<Enter>``

Can you reproduce this ?

This may be related to #348

ejgallego:
Fixes #432 , introduced in #289

This whole code needs more investigation as the logic seems quite
strange, in particular I think the core of LP should not mess with
filesystem low-level details, but hoping for a quick fix.

Please test in emacs as I don't happen to have the setup right now.

Note that Vscode won't actually initialize the server until the file is saved, we should check what the lsp protocol says, maybe it is a bug that emacs is sending `didOpen` for a non-existing file?

### DIFF
--- a/src/core/package.ml
+++ b/src/core/package.ml
@@ -75,7 +75,12 @@ let read : file_path -> config_data = fun fname ->
     [Sys_error] is raised. Note that [fname] is first normalized with a call
     to [Filename.realpath]. *)
 let find_config : file_path -> file_path option = fun fname ->
-  let fname = Filename.realpath fname in
+  (* workaround #432 , this whole code should be reworked tho *)
+  let fname =
+    if Sys.file_exists fname
+    then Filename.realpath fname
+    else fname
+  in
   let fname =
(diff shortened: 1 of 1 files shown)


gabrielhdt (approved):

fblanqui:
Thanks @ejgallego !

PR #470, opened 2020-11-12
Restoring environment upon failure

gabrielhdt:
Environment would not be restored when `compile` failed.


fblanqui:
This completes https://github.com/Deducteam/lambdapi/pull/457, right?

gabrielhdt:
Indeed it does

PR #473, opened 2020-12-09
CI: add ocaml 4.10.2 and remove intermediate ocaml releases

fblanqui:


gabrielhdt:
I guess the purpose of that is to speed up Travis' tests, isn't it? I think we should at least keep version 4.07.0 as the oldest version supported. Another way to speed up tests (and which would make more sense I think), is to divide tests between the "fast" ones and "slow" ones, with `alcotest` (changing `` `Quick`` to `` `Slow``).

fblanqui:
This is one reason but not the only one. I believe that the OCaml versions _._.k with k>0 are bug fixes of the base version _._.0. So it does not seem very useful to test both _._.max and the versions _._.k with k<max. What do you think?

ejgallego:
Sounds reasonable. By the way I keep hearing a lot of trouble with Travis new limits, so at some point CI should move to github actions.

gabrielhdt (approved):

amelieled:
Hi @ejgallego !
Do you think github actions is better than gitlab CI?

ejgallego:
> Do you think github actions is better than gitlab CI?

Given that the project is on github already, and the current CI load, I think so; github actions are based on azure and has pretty good integration / performance / free tier. Gitlab's CI has very interesting features [like allowing you to have your own workers], but requires using a bot to integrate with github.

PR #474, opened 2020-12-10
output assertions on verbose 1

fblanqui:


fblanqui:
@gabrielhdt do you know where the default verbose level is set? In console.ml, it is set to 1 when the ref verbose is created but this seems to be overridden to 0 later but I don't know where. In config? By cmdliner?

gabrielhdt:
Have you seen that there is both `default_verbose` and `verbose` in `console.ml`? And that there is `set_default_verbose`?

fblanqui:
Yes but I do not see where and when it is set.

amelieled:
src/cli/lambdapi.ml:    Timed.(verbose := 0); (* To avoid printing the "Checked ..." line *) ?
tests/dtrees.sh:LAMBDAPI="lambdapi decision-tree --verbose 0 \ ?

fblanqui:
No, these functions are called for printing decision trees only, aren't they?

amelieled:
It seems to be right, but thanks to `grep -r "verbose" *`, there is no others lines where the reference `verbose` is changed (at 0).

amelieled:
But I think @gabrielhdt right: thanks to `grep -r "set_default_verbose" *, you can find the line `Option.iter set_default_verbose cfg.verbose;` and since cfg.verbose = None, this line do nothing.
I don't know how what is a `configuration` but maybe you can change `verbose = None` by `verbose = 3`

gabrielhdt:
I'm not sure I understand fully the problem here, @fblanqui, could you tell us what you want to do precisely? As pointed out by Amelie, the verbosity level can be changed by altering the reference, so I don't understand what fails. Perhaps the design of this verbosity setting can be clarified?

fblanqui:
The problem is that, when created, verbose is set to 1 but, later, it is reset to 0. I don't understand where.

amelieled:
It's a bit confusing because there are a lot of `config` in the code (why3, inductive, file, data, etc.)

gabrielhdt:
> Maybe because of:
> 
> ```
> (** Default configuration for commands that do not support all flags. *)
> let default_config =
>   { gen_obj     = false
>   ; lib_root    = None
>   ; map_dir     = []
>   ; verbose     = None
>   ; no_warnings = false
>   ; debug       = ""
>   ; no_colors   = false
>   ; too_long    = infinity
>   ; confluence  = None
>   ; termination = None }
> ```
> 
> ? (See my previous message).
> I don't understand why `no_warnings = false` (instead of `no_wrn`) and `verbose = None` (instead of `verbose`)

As I understand it, we set `no_warnings` to false to enable warnings by default, and `verbose = None` to use the default verbosity level.

@fblanqui, in this pull request, you want to change the behaviour so that assertion results are output only on verbose mode 3, don't you? I'll pull your branch and tweak around to see what I manage to do.

fblanqui:
Sorry for the confusion. I must have made mistakes in my tests. Every thing is fine now.

amelieled:
Why don't define:
```
type verbose_level = low | default | high
```
and the function:
```
let int_of_verbose_level l = match l with
 | low -> 1
 | default -> 2
 |  high -> 3
```
to give more meaning?
I mean, I don't know the difference between the levels 2 and 3, and it's could be important for the user.

In the future, if you would like to increase the number of levels, the code will be more meaningful (and it will be easier to do changes than now).
For example, a "proof" level can be defined, or "warning" one, to disable the proof goals or warnings.

gabrielhdt:
> Why don't define:
> 
> ```
> type verbose_level = low | default | high
> ```
> 
> and the function:
> 
> ```
> let int_of_verbose_level l = match l with
>  | low -> 1
>  | default -> 2
>  |  high -> 3
> ```
> 
> to give more meaning?
> I mean, I don't know the difference between the levels 2 and 3, and it's could be important for the user.
> 
> In the future, if you would like to increase the number of levels, the code will be more meaningful (and it will be easier to do changes than now).
> For example, a "proof" level can be defined, or "warning" one, to disable the proof goals or warnings.

I see what you mean, @amelieled. Indeed it makes sense, but it should be a pull request on its own, so for now, an issue may be created, so that we can handle it later.

Issue #462, opened 2020-11-05
Display problem in vscode


fblanqui:
With code 1.50.1, strange characters are written above and below "No goals".

fblanqui:
![Capture d’écran_2020-12-11_16-27-27](https://user-images.githubusercontent.com/10852505/101922076-fd5d0b80-3bcd-11eb-83be-a14e0ebf61bb.png)

PR #477, opened 2020-12-11
Update VSCode client in PR #422

firewall2142:
* Updated syntax highlighting
* Updated client.ts

### DIFF
--- a/editors/vscode/src/client.ts
+++ b/editors/vscode/src/client.ts
@@ -405,7 +405,7 @@ function getGoalsEnvContent(goals : Goal[]){
     for(let i=0; i < goals.length; i++) {
         
         // check if this is a Type Goal
-    	// extra space is not a typo
+    	// extra space is not a typo, this is defined in lsp_base.ml
         if (goals[i].typeofgoal == "Typ "){
             
             let curGoal = goals[i] as TypGoal;
--- a/editors/vscode/src/client.ts
+++ b/editors/vscode/src/client.ts
@@ -400,12 +400,12 @@ function getGoalsEnvContent(goals : Goal[]){
     let codeHyps : String = ""; //hypothesis HTML code
     let codeGoals : String = ""; //goals HTML code
     let codeEnvGoals : String = ""; //result code HTML
-    let goalNo = 0;
+    let typGoalNo = 0;
     
     for(let i=0; i < goals.length; i++) {
         
         // check if this is a Type Goal
-	// extra space is not a typo
+    	// extra space is not a typo
         if (goals[i].typeofgoal == "Typ "){
             
             let curGoal = goals[i] as TypGoal;
@@ -428,7 +428,7 @@ function getGoalsEnvContent(goals : Goal[]){
             }
 
             let numGoalcode = `<label class="numGoal">`
-                + goalNo + `</label>`;
+                + typGoalNo + `</label>`;
 
             let typeGoal = `<span class="goal">`
                 + curGoal.type + `</span>`;
@@ -439,14 +439,15 @@ function getGoalsEnvContent(goals : Goal[]){
 
             codeHyps = codeHyps + `</div>`;
 
-            goalNo++;
+            typGoalNo++;
 
             let codeSep = `<hr/>`;
             codeEnvGoals = codeEnvGoals + "" + codeHyps + codeSep + codeGoals;
 
         } else {    // case if this is a Unification Goal
             let curGoal = goals[i] as UnifGoal;
-            codeEnvGoals = codeEnvGoals + "" + `<hr/>` + curGoal.constr
+            codeEnvGoals = codeEnvGoals + "" + `<hr/>` +
+                           `<span class="goal">`+curGoal.constr+`</span>`
         }
     }
 
--- a/editors/vscode/src/client.ts
+++ b/editors/vscode/src/client.ts
@@ -522,7 +522,7 @@ export interface Goal {
 }
 
 export interface UnifGoal extends Goal {
-    constr : String // constraint
+    constr : String
 }
 
 export interface TypGoal extends Goal {
--- a/editors/vscode/src/client.ts
+++ b/editors/vscode/src/client.ts
@@ -405,7 +405,7 @@ function getGoalsEnvContent(goals : Goal[]){
     for(let i=0; i < goals.length; i++) {
         
         // check if this is a Type Goal
-        if ("gid" in goals[i]){
+        if (goals[i].typeofgoal == "Typ "){
             
             let curGoal = goals[i] as TypGoal;
             codeHyps = `<div class="hypothesis">`;
--- a/editors/vscode/src/client.ts
+++ b/editors/vscode/src/client.ts
@@ -520,7 +520,7 @@ export interface Env {
 }
 
 export interface Goal {
-    typeofgoal : String // type of goal
+    typeofgoal : String // type of goal, values defined in lsp_base.ml
 }
 
 export interface UnifGoal extends Goal {
--- a/editors/vscode/src/client.ts
+++ b/editors/vscode/src/client.ts
@@ -405,6 +405,7 @@ function getGoalsEnvContent(goals : Goal[]){
     for(let i=0; i < goals.length; i++) {
         
         // check if this is a Type Goal
+	// extra space is not a typo
         if (goals[i].typeofgoal == "Typ "){
             
             let curGoal = goals[i] as TypGoal;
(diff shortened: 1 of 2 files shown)


editors/vscode/src/client.ts
```diff
@@ -506,6 +518,14 @@ export interface Env {
 }
 
 export interface Goal {
+    typeofgoal : String // type of goal
```
fblanqui, line 521: Isn't String a too big type? There are only two possible values: UnifGoal and TypGoal. Is it possible to have enumerated types or just a boolean, or is it useful to have a String?

editors/vscode/src/client.ts
```diff
@@ -400,41 +400,53 @@ function getGoalsEnvContent(goals : Goal[]){
     let codeHyps : String = ""; //hypothesis HTML code
     let codeGoals : String = ""; //goals HTML code
     let codeEnvGoals : String = ""; //result code HTML
-
+    let goalNo = 0;
+    
     for(let i=0; i < goals.length; i++) {
+        
+        // check if this is a Type Goal
+        if ("gid" in goals[i]){
```
fblanqui, line 408: This test looks to much like a hack. You seem to use the fact that in UnifGoal there is no "gid" field, but this could change in the future. It would be better to use a more robust test.

editors/vscode/src/client.ts
fblanqui: typeofgoal doesn't seem to be used anywhere. Why?

editors/vscode/src/client.ts
fblanqui: Wouldn't typeofgoal be useful?

editors/vscode/src/client.ts
fblanqui: I see. UnifGoal and TypGoal add fields to Goal. So both UnifGoal and TypGoal have the field typeofgoal. Right?

firewall2142: I kept `typeofgoal` as string, because it was used to parse the incoming JSON

editors/vscode/src/client.ts
firewall2142: I agree, but "Typ " had an extra space. I thought it was a bug. Anyway I will make the changes.

editors/vscode/src/client.ts
firewall2142: yes both of them will have the `typeofgoal` field

fblanqui: I see. Could you please add in a comment that the values are defined in lsp_base.ml?

editors/vscode/src/client.ts
```diff
@@ -400,41 +400,53 @@ function getGoalsEnvContent(goals : Goal[]){
     let codeHyps : String = ""; //hypothesis HTML code
     let codeGoals : String = ""; //goals HTML code
     let codeEnvGoals : String = ""; //result code HTML
-
+    let goalNo = 0;
+    
     for(let i=0; i < goals.length; i++) {
+        
+        // check if this is a Type Goal
+        if (goals[i].typeofgoal == "Typ "){
```
fblanqui, line 408: Thanks to add in a comment that the space is not a typo and that this is defined in lsp_base.ml.

editors/vscode/syntaxes/lp.tmLanguage.json
```diff
@@ -44,7 +44,7 @@
     },
 
     "signature-commands": {
-      "match": "(definition|as|in|let|declared|builtin|proof|qed|rule|symbol|theorem|with)\\s+([^\\s+]*)",
+      "match": "(begin|end|definition|as|in|let|declared|builtin|rule|symbol|with)\\s+([^\\s+]*)",
```
fblanqui, line 47: The keyword "begin" has the first 3 letters in one color and the last 2 letters in another color. Perhaps you need to put the keyword "begin" after the keyword "in". If this is the case, thanks to add a comment saying that the order is important and mentioning this example.

firewall2142: This is weird, that didn't happen in my vscode

fblanqui:
When trying tests/OK/symbol_syntax.lp there are a number of problems:
- I open tests/OK/symbol_syntax.lp: every thing looks fine
- I click on the first line: I have strange symbols written in the Goals panel (this is https://github.com/Deducteam/lambdapi/issues/462)
- I do Ctrl-Right: the next 2 lines become green
- I do Ctrl-Left: I go back only 1 line and cannot go back further using Ctrl-Left - there is something wrong with Ctrl-Left
- I do Ctrl-Right until line 4 gets green: a goal is printed - I don't understand why - there should be no goal at this point - I guess that the printed goal is in fact the goal of line 6

firewall2142:
Here is the log from LSP server, it showed that there was a goal:

```
433 [read]: {
434           "jsonrpc": "2.0",
435           "id": 3,
436           "method": "proof/goals",
437           "params": {
438             "textDocument": {
439               "uri":
440                 "file:///home/ashish/Software/forked_lambdapi/lambdapi/tests/OK/symbol_syntax.lp"
441             },
442             "position": { "line": 3, "character": 0 }
443           }
444         }
445 [get_node_at_pos]: call: false l:3 p:0 / file:///home/ashish/Software/forked_lambdapi/lambdapi/tests/OK/symbol_syntax.lp, 6:0-10:3
446 [get_node_at_pos]: call: false l:3 p:0 / file:///home/ashish/Software/forked_lambdapi/lambdapi/tests/OK/symbol_syntax.lp, 5:0-27
447 [get_node_at_pos]: call: true l:3 p:0 / file:///home/ashish/Software/forked_lambdapi/lambdapi/tests/OK/symbol_syntax.lp, 4:0-18
448 [send]: {
449           "jsonrpc": "2.0",
450           "id": 3,
451           "result": {
452             "goals": [
453               {
454                 "typeofgoal": "Typ ",
455                 "gid": 0,
456                 "hyps": [],
457                 "type": "U → U"
458               }
459             ]
460           }
461         }

```

fblanqui:
Hi @Rehan-MALAK . Do you have any idea of what is going wrong here?

editors/vscode/syntaxes/lp.tmLanguage.json
fblanqui: Could you get the version 1.52.0?

fblanqui:
Perhaps there is something wrong in the associations between positions and goals on the LP or LSP side?

editors/vscode/syntaxes/lp.tmLanguage.json
firewall2142: Didn't see that, I should refresh before asking.

I still couldn't get the same error. 
Its weird, as both `begin` and `in` are in the same `signature-commands` so  they should have the same colour

ejgallego:
> Perhaps there is something wrong in the associations between positions and goals on the LP or LSP side?

It could well be, is there unicode involved? We can have a look next week I think.

fblanqui:
Thanks @ejgallego to help us! There is no unicode involved. This is tests/OK/symbol_syntax.lp. A very simple and short file.

fblanqui:
I believe I found the problem. In https://github.com/Deducteam/lambdapi/blob/92df4fa8028021da3560dab7bfc142099d4766ca/src/core/handle.ml#L428-L439, we implicitly add a proof to every symbol declaration/definition. I will try to fix this on Monday.

editors/vscode/src/client.ts
```diff
@@ -400,41 +400,54 @@ function getGoalsEnvContent(goals : Goal[]){
     let codeHyps : String = ""; //hypothesis HTML code
     let codeGoals : String = ""; //goals HTML code
     let codeEnvGoals : String = ""; //result code HTML
-
+    let goalNo = 0;
+    
     for(let i=0; i < goals.length; i++) {
```
fblanqui, line 405: goalNo and i are equal. why not using i only?

fblanqui:
Isn't there something wrong with https://github.com/Deducteam/lambdapi/blob/76c6cb76019d4b7d792dd611ce6c18509e7e314f/editors/vscode/src/client.ts#L463 as there is no style.css file.

editors/vscode/syntaxes/lp.tmLanguage.json
fblanqui: Sorry, this is my mistake: I was not testing the right branch.

fblanqui:
It's not really a problem that these temporary goals are printed although it would be nice to get rid of them at some point. But I currently don't know the code of the LSP server enough to do that by myself. So I propose to let this aside for the moment and open an issue about this later.
So, to conclude, only a few remarks remain to be fixed before merging:
https://github.com/Deducteam/lambdapi/pull/477/files#r542501283
https://github.com/Deducteam/lambdapi/pull/477/files#r541018485

editors/vscode/src/client.ts
firewall2142: `goalNo` only gets increments for type goals. I changed the variable name to `typGoalNo`

fblanqui:
Thank you @firewall2142 !

Issue #448, opened 2020-10-16
security vulnerability in VSCode interface


fblanqui:
Github found a vulnerable dependency:
Known  low severity security vulnerability detected in minimist < 0.2.1 defined in package-lock.json.
You have Dependabot security updates enabled, so we are attempting to create a pull request to update the vulnerable dependency.

https://github.com/Deducteam/lambdapi/blob/9e080cd73ab5d655086197fcdee5a8290b87ce71/editors/vscode/package-lock.json#L214-L219

https://github.com/Deducteam/lambdapi/network/alert/editors/vscode/package-lock.json/minimist/open

fblanqui:
Should we add some code scanning tool? See https://github.com/Deducteam/lambdapi/security/code-scanning

fblanqui:
Hello @francois-lefoulon . Would you mind to make a PR and test it to fix this problem?

francois-lefoulon:
I don't have access to my computer until late this evening ; can it wait?

Should I just delete the minimist part in package-lock? (this isn't my work so I don't know what it is for)

fblanqui:
Of course. There is no hurry. In https://github.com/Deducteam/lambdapi/network/alert/editors/vscode/package-lock.json/minimist/open, they propose:
```
Remediation

Upgrade minimist to version 0.2.1 or later. For example:

"dependencies": {
  "minimist": ">=0.2.1"
}

or…

"devDependencies": {
  "minimist": ">=0.2.1"
}
````
I don't know what is minimist. If it is not useful anymore, you can remove it.

firewall2142:
This can be closed as `minimist` is a development dependency. It won't get bundled with the extension.

fblanqui:
Ok, but I would like the warning to disappear from npm and github...

PR #480, opened 2020-12-14
fix editor issues #452, #462, #463, #448 and some other minor fixes

### LINKED ISSUE #462: Display problem in vscode

firewall2142:
#462 -  Every new line character was encoded as the sequence `\n\xef\xbf\xbc` in mutli-line strings. 
The variables `header` and `footer` inside `buildGoalsContent(goals: Goal[])` function were initialized with multi-line strings. 
I am curious why this happened, even copy pasting the code in github comments had those strange characters. 

#463 - added alpha values

#448 - removed tslint from devDependencies (tslint depends on mkdirp, mkdirp depended upon minimist)

Other minor fixes:
* Moved css styles to seperate file
* Fixed the proof line

I think this was all for the editors for now.


fblanqui:
Thank you @firewall2142 for this new PR!

fblanqui:
With the Light color theme (white background), the goals are difficult to read. Could you please print hypothesis and goals in black in this case?

fblanqui:
Thanks @firewall2142 to update your PR so that we can merge it.

firewall2142:
@fblanqui I think this PR can be merged automatically without any issues

firewall2142:
Oh I see. I had to update, my bad. I am not so experienced with github

fblanqui:
Thank you @firewall2142 !

editors/vscode/src/client.ts
```diff
@@ -29,23 +29,23 @@ export function activate(context: ExtensionContext) {
     //___Declaration of workspace variables___
 
     //Position of the proof cursor : colored highlights show until which point the proof was surveyed
-    let proofState : Position = new Position(1, 0);
+    let proofState : Position = new Position(0, 0);
```
ejgallego, line 32: Nice catch!

firewall2142: Thanks :-)

editors/vscode/src/client.ts
```diff
@@ -487,6 +489,10 @@ function buildGoalsContent(goals : Goal[]) {
     footer =` </p>
 ￼       </body>
 ￼       </html>`;
+
+    // remove character sequences caused by multiline strings
+    header = header.replace(/[\u{0080}-\u{FFFF}]/gu,"");
+    footer = footer.replace(/[\u{0080}-\u{FFFF}]/gu,"");
```
ejgallego, line 493: Interesting; indeed we should improve the document type that LP is sending to the client. I have some ideas here which could provide for very cool display [see for example what jsCoq goal buffer can do already] Let's hopefully discuss on a call.

ejgallego:
Nice work, thanks @firewall2142 !

PR #491, opened 2020-12-17
Fix #393

### LINKED ISSUE #393: Type inference does not fail when abstracting over a kind
Thanks to @barras for this issue.
LP accepts `type λ_,TYPE`! The result is `?0 → KIND` which is not typable.
The problem is in https://github.com/Deducteam/lambdapi/blob/4ec298ad8425409fd535b4a4df00be3a60279554/src/core/infer.ml#L139-L146
We should check that the inferred product is typable.
It is perhaps sufficient to check that b is not KIND.

fblanqui:


fblanqui:
PR based on a wrong branch, subsumed by https://github.com/Deducteam/lambdapi/pull/493.

Issue #487, opened 2020-12-17
Change organization of source files ?


amelieled:
It's probably the good time to change the file system hierarchy in order to increase the readability of this github (and the modularity too).

fblanqui:
Why?

amelieled:
"to increase the readability of this github (and the modularity too)."

fblanqui:
I don't think that this is currently an issue and that this is very urgent to do that.

amelieled:
Why do you close it even if it's not urgent?

amelieled:
In fact, it is, and to be honest with you, I discussed this morning with @gabrielhdt and @EmilieGrienenberger about that, so it's not completely my idea.

amelieled:
Moreover, it's a discussion, or a proposition.

fblanqui:
Then, it would be good to provide arguments for such a change now.

gabrielhdt:
In my opinion, such a reorganistion may indeed not be urgent when we consider lambdapi as a command line prover, or a proof checker tied to a LSP server.

However, in the team we are more and more using lambdapi as library, may it be to implement a coercion insertion algorithm in my case, to port DkMeta to Lambdapi; or, in the near future, to port Universo to Lambdapi. When we lambdapi as a library, we become heavily depenent on what the code OCaml code exposes, and the datatypes. The current lack of specification has a quite annoying consequence on the code of these tools since any part of lambdapi may change anytime. There have already been examples of such tools becoming broken because of such modifications.

Even regarding lambdapi code, it may help in the future because dependencies are getting blurred, and non obvious. For instance, the flow of a file is to be parsed, using code of `syntax.ml`, then to be scoped with `scope.ml` to terms of `terms.ml`. Therefore, I think (maybe it's only me) that intuitively, we would expect `terms.ml` to perhaps depend on `syntax.ml`. But actually it's the opposite. And it may be the case that not having clear interfaces increases the entanglement of modules.

As is said in [The Art of Unix Programming](http://www.catb.org/~esr/writings/taoup/html/ch01s06.html#id2877537),
> The only way to write complex software that won't fall on its face is to hold its global complexity down — to build it out of simple parts connected by well-defined interfaces, so that most problems are local and you can have some hope of upgrading a part without breaking the whole.

More information can be found in [Chapter 4](http://www.catb.org/~esr/writings/taoup/html/modularitychapter.html), for the curious.

[...]

PR #492, opened 2020-12-17
Hierarchy of tests

amelieled:
This PR is about the hierarchy of tests (Issue #487 ).
More details should be discussed.


gabrielhdt:
Could you provide the hierarchy you are planning to implement in the pull request description? So that we can discuss it beforehand.

amelieled:
Yes, of course @gabrielhdt ! :)
So until now, I've done:
   * parsing_tests/  (empty file and escaped identifier)
   * queries_tests/  (assert / assertnot / type / compute)
   * require_open_tests/ (require / require as / open / and paths)
   * rule_tests/             (on rules : arity / free variable in LHS / patterns / non linearity / order instanciation performance )
   * set_option_tests/  (test suite on set option)
   * inductive_tests/    (on inductive types)
   * symbol_tests/       (on symbols)
   * tactics_tests/        (on tactics)
   * term_tests/ (implicit argument / lambda construction / let construction / wildcard )

Maybe, I can add the name of testing files (for each category) in the file description_*.

Now, we need to discuss on files are listed on the file named "theory".

amelieled:
Problems:
 * Files which their name begins by a number (related to an issue so).
 * Maybe we can create the folder "theory_tests" or "examples_tests" (or remove redundancies)
 * I don't understand the goal of these testing files:
                   bug0.dk
                bug2.dk
                bug.dk
                raphael.dk
                sujective.dk
                symbol.lp

gabrielhdt:
Thanks @amelieled. I'm quite happy with your categories. I'll be picky, but I'd remove the suffix `_test`, because we know these are all tests, we're already in the `tests` folder... I'd rename `require_open` into `require`, `rule` into `rewriting`, perhaps `symbol` into `declaration`? Perhaps `term` into `scoping`?

> Files which their name begins by a number (related to an issue so).

I'd put that into parsing I think. Or another `system` category. Or even create a `miscellaneous` category, don't bother, it will still be better that what we have.

Yes, `examples` would be fine I think.

For the remaining files, I'll have a look later, don't worry.

amelieled:
> I'll be picky, but I'd remove the suffix `_test`, because we know these are all tests, we're already in the `tests` folder... 
Yes, sure! Same thing for "description_blabla".
> I'd rename `require_open` into `require`, 
`import` ? I think `require` could be ambiguous.
> `rule` into `rewriting`, 
Done.
> perhaps `symbol` into `declaration`? 
`inductive` and `rule` are declaration too?
> Perhaps `term` into `scoping`?
Done.
> Files which their name begins by a number (related to an issue so).
> I'd put that into parsing I think. Or another `system` category. Or even create a `miscellaneous` category, don't bother, it will still be better than what we have.
The problem is that it breaks the links on the "issue" discussions. Moreover, some of them are about parsing, or tactic, etc.
One category for them doesn't seem satisfying.
> Yes, `examples` would be fine I think.
Done.

gabrielhdt:
> `import` ? I think require could be ambiguous.

Indeed.

> The problem is that it breaks the links on the "issue" discussions.

What do you mean?

> One category for them doesn't seem satisfying.

Rome wasn't built in a day, as we say. It's not a big deal if we have some files not categorised yet.

fblanqui:
I'm closing this PR because it is too old now.

PR #493, opened 2020-12-17
Fix type inference and remove typing.ml

### LINKED ISSUE #393: Type inference does not fail when abstracting over a kind

fblanqui:


gabrielhdt (commented):
Thanks a lot @fblanqui! You haven't removed yet `typing.ml`, have you? Otherwise, I totally approve the work done here, I feel that the code is much clearer. To link back with the discussion on the module reorganisation, it may be way easier to specify an interface for the type checker, since the only entry point is `Infer.infer_noexn` (rather than having a file that call another one &c.). And it seems also clearer to me that `Unif` defines the type inference from `Infer` with unification included. I'm very happy with this. I just suggest some style modifications regarding naming of modules.

PR #495, opened 2020-12-19
Auto generation of opam file

### LINKED ISSUE #305: OCaml dependencies
Using OCaml in version 4.05.1, I have an error while trying to compile the project

```
      ocamlc lp-lsp/.lp_lsp_lib.objs/lp_lsp_lib__Lsp_base.{cmi,cmti} (exit 2)
(cd _build/default && /home/fthire/.opam/4.05.0/bin/ocamlc.opt -w @a-4-29-40-41-42-44-45-48-58-59-60-40 -strict-sequence -strict-formats -short-paths -keep-locs -g -bin-annot -I lp-lsp/.lp_lsp_lib.objs -I lp-lsp/.lp_lsp_lib.objs/.private -I /home/fthire/.opam/4.05.0/lib/base -I /home/fthire/.opam/4.05.0/lib/base/caml -I /home/fthire/.opam/4.05.0/lib/base/shadow_stdlib -I /home/fthire/.opam/4.05.0/lib/bindlib -I /home/fthire/.opam/4.05.0/lib/biniou -I /home/fthire/.opam/4.05.0/lib/bytes -I /home/fthire/.opam/4.05.0/lib/cmdliner -I /home/fthire/.opam/4.05.0/lib/earley/core -I /home/fthire/.opam/4.05.0/lib/easy-format -I /home/fthire/.opam/4.05.0/lib/menhirLib -I /home/fthire/.opam/4.05.0/lib/num -I /home/fthire/.opam/4.05.0/lib/ppx_inline_test/config -I /home/fthire/.opam/4.05.0/lib/ppx_inline_test/runtime-lib -I /home/fthire/.opam/4.05.0/lib/result -I /home/fthire/.opam/4.05.0/lib/sexplib0 -I /home/fthire/.opam/4.05.0/lib/timed -I /home/fthire/.opam/4.05.0/lib/why3 -I /home/fthire/.opam/4.05.0/lib/yojson -I src/.core.objs -I src/pure/.pure.objs -no-alias-deps -opaque -open Lp_lsp_lib -o lp-lsp/.lp_lsp_lib.objs/lp_lsp_lib__Lsp_base.cmi -c -intf lp-lsp/lsp_base.mli)
File "lp-lsp/lsp_base.mli", line 18, characters 26-29:
Error: Unbound type constructor J.t
GNUmakefile:11: recipe for target 'bin' failed
make: *** [bin] Error 1
fthire@elitebook:~/Git/lambdapi$ make clean
fthire@elitebook:~/Git/lambdapi$ make
      ocamlc lp-lsp/.lp_lsp_lib.objs/lp_lsp_lib__Lsp_base.{cmi,cmti} (exit 2)
(cd _build/default && /home/fthire/.opam/4.05.0/bin/ocamlc.opt -w @a-4-29-40-41-42-44-45-48-58-59-60-40 -strict-sequence -strict-formats -short-paths -keep-locs -g -bin-annot -I lp-lsp/.lp_lsp_lib.objs -I lp-lsp/.lp_lsp_lib.objs/.private -I /home/fthire/.opam/4.05.0/lib/base -I /home/fthire/.opam/4.05.0/lib/base/caml -I /home/fthire/.opam/4.05.0/lib/base/shadow_stdlib -I /home/fthire/.opam/4.05.0/lib/bindlib -I /home/fthire/.opam/4.05.0/lib/biniou -I /home/fthire/.opam/4.05.0/lib/bytes -I /home/fthire/.opam/4.05.0/lib/cmdliner -I /home/fthire/.opam/4.05.0/lib/earley/core -I /home/fthire/.opam/4.05.0/lib/easy-format -I /home/fthire/.opam/4.05.0/lib/menhirLib -I /home/fthire/.opam/4.05.0/lib/num -I /home/fthire/.opam/4.05.0/lib/ppx_inline_test/config -I /home/fthire/.opam/4.05.0/lib/ppx_inline_test/runtime-lib -I /home/fthire/.opam/4.05.0/lib/result -I /home/fthire/.opam/4.05.0/lib/sexplib0 -I /home/fthire/.opam/4.05.0/lib/timed -I /home/fthire/.opam/4.05.0/lib/why3 -I /home/fthire/.opam/4.05.0/lib/yojson -I src/.core.objs -I src/pure/.pure.objs -no-alias-deps -opaque -open Lp_lsp_lib -o lp-lsp/.lp_lsp_lib.objs/lp_lsp_lib__Lsp_base.cmi -c -intf lp-lsp/lsp_base.mli)
File "lp-lsp/lsp_base.mli", line 18, characters 26-29:
Error: Unbound type constructor J.t
```

Trying to install lambdapi-lsp from opam, I have an error that my version of OCaml should be `>= 4.06'. However, one of the dependencies of the project (earley, timed or bindlib I forget) require a version of OCaml `< 4.07`. Hence, for compilation, only the version 4.06 is supported?

gabrielhdt:
Generation of `.opam` is handed out to dune. Emacs and Vim files can now be installed using `dune install`. The documentation on Vim mode has also been updated.

Fixes #305


dune-project
```diff
@@ -2,3 +2,30 @@
...
+ (name lambdapi)
+ (synopsis "Proof assistant based on the λΠ-calculus modulo rewriting")
+ (description "Proof assistant based on the λΠ-calculus modulo rewriting,
+mostly compatible with Dedukti.
+
+This package contains both the `lambdapi` proof assistant and the `lp-lsp`
+Language Server Protocol helper for IDEs to check lambdapi files.")
+ (depends
+  (ocaml (>= 4.07.0))
+  (dune (>= 2.1.0))
```
ejgallego, line 23: This is automatically added from the dune-lang version I think.

ejgallego:
Looks good, note that most project still keep the opam files in git as `opam pin` may look for them, rules here are a bit confusing tho.

ejgallego:
So indeed we should test that `opam pin` works ok before going ahead with this.

dune-project
gabrielhdt: Indeed it is, well seen.

gabrielhdt:
> So indeed we should test that `opam pin` works ok before going ahead with this.

`opam pin add .` is happy without any `lambdapi.opam` file.

dune-project
```diff
@@ -2,3 +2,29 @@
 (name lambdapi)
 
 (using menhir 1.0)
+
+(generate_opam_files true)
+(source (github Deducteam/lambdapi))
+(license CeCILL2.1)
+(authors "Rodolphe Lepigre <lepigre@mpi-sws.org>"
+         "Frédéric Blanqui <frederic.blanqui@inria.fr>"
+         "Emilio Jesús Gallego Arias <e@x80.org>")
```
fblanqui, line 11: There are contributions by other people. So it would be better to put no names at all. Contributors can be found out on github.

editors/vim/syntax/lambdapi.vim
```diff
@@ -127,7 +127,7 @@ syntax match Keyword "@"
 syntax match Keyword "\["
 syntax match Keyword "\]"
 syntax match Keyword ":"
-syntax match Keyword "⇒"
+syntax match Keyword "↪"
 syntax match Keyword "→"
 syntax match Keyword "∀"
```
fblanqui, line 132: ∀ is not a LP keyword, nor [ and ]. On the other hand are missing: ↪ , Π, ≔

editors/vim/syntax/lambdapi.vim
```diff
@@ -143,9 +143,9 @@ syntax match Constant "&\(\<\h\w*\>\|\({|\([^|]\|\(|[^}]\)\)*|*|}\)\)"
 syntax match PreProc  "?\(\<\h\w*\>\|\({|\([^|]\|\(|[^}]\)\)*|*|}\)\)"
 
 " Abbreviations.
+abbreviate --> ↪
 abbreviate -> →
-abbreviate => ⇒
-abbreviate !  ∀
+abbreviate !  Π
 abbreviate (! (∀
```
fblanqui, line 149: should be removed

editors/emacs/dune
```diff
@@ -0,0 +1,19 @@
+; Byte compile file that is loaded at startup of Emacs
```
fblanqui, line 1: English: at Emacs startup

docs/ui/vim.rst
```diff
@@ -1,6 +1,26 @@
...
+  ``make install_vim`` in the ``lambdapi`` repository.
+
+Installing with Opam
+
+  If Lambdapi is installed with Opam or using ``dune build`` from the
+  sources, then the line
+
+  .. code:: vim
+
+     set rtp+=~/.opam/OCAML_VERSION/share/vim
```
fblanqui, line 22: shouldn't we write `$OPAM_SWITCH_PREFIX/share/vim` instead?

README.md
```diff
@@ -57,16 +57,15 @@ for more information).
 make               # Build lambdapi.
 make doc           # Build the documentation.
```
fblanqui, line 58: Remark: there are now two kinds of documentation. make doc generates html pages for the source code and, in docs/, make html generates html pages for the user manual. It would be nice to indicate the two in the top Makefile: make src_doc and make usr_doc?
Btw, I don't like very much that our top Makefile is called GNUmakefile. Is there a good reason for not calling it Makefile?

docs/ui/vim.rst
gabrielhdt: I didn't know that `$OPAM_SWITCH_PREFIX` was defined, thanks for the tip.

editors/vim/syntax/lambdapi.vim
gabrielhdt: Well I think we can remove parentheses as well.

README.md
gabrielhdt: We can rename the user documentation into "manual" as well, if it helps distinguishing the two.

For the makefile, we can write `GNUMakefile` to specify that it uses some GNU features, so that BSD make won't look for it. In our case, the `$(shell ...)` constructs are specific to GNU, so it makes sense. But we can also rename it simply to Makefile, I don't mind; and I guess nobody does.

amelieled: "manual" is too generic? 
"usr_doc" and "dev_doc"?
Renamed "GNUmakefile" by "Makefile" is a good idea :)

fblanqui:
@gabrielhdt Do you plan to update this PR so that we can merge it?
Remarks: as lambdapi.opam is deleted, README.md should refer to some dune file instead for listing the dependencies.

ejgallego:
I have confirmed upstream that the `opam` file should still remain in the tree [as this PR seems to do]

gabrielhdt:
> I have confirmed upstream that the `opam` file should still remain in the tree [as this PR seems to do]

All right, thanks a lot, I was wondering indeed.

ejgallego:
> All right, thanks a lot, I was wondering indeed.

This is a limitation of opam which may be lifted but not before opam 2.2

fblanqui:
@gabrielhdt What is the status of this PR?

gabrielhdt:
Currently, continuous integration fails because the build target of the emacs mode requires emacs, and is started by `dune build`. We should find a way to either not build the emacs mode automatically, or conditionally on the availability of emacs.

ejgallego:
> Currently, continuous integration fails because the build target of the emacs mode requires emacs, and is started by `dune build`. We should find a way to either not build the emacs mode automatically, or conditionally on the availability of emacs.

You can use `(enabled_if )` or place the emacs mode into its own package, note that enabled_if may require to bump the version of dune language .

gabrielhdt:
> > Currently, continuous integration fails because the build target of the emacs mode requires emacs, and is started by `dune build`. We should find a way to either not build the emacs mode automatically, or conditionally on the availability of emacs.
> 
> You can use `(enabled_if )` or place the emacs mode into its own package, note that enabled_if may require to bump the version of dune language .

It seems to me that the predicate language of dune is insufficient, or can we write a predicate that is true if the emacs binary is in the PATH?

ejgallego:
> It seems to me that the predicate language of dune is insufficient, or can we write a predicate that is true if the emacs binary is in the PATH?

I think you should rather use the condition "the opam package `conf-emacs` is available".

gabrielhdt:
> I think you should rather use the condition "the opam package conf-emacs is available".

It seems that the `enabled_if` only allows boolean expressions made of variables from <https://dune.readthedocs.io/en/stable/concepts.html#variables>, which is not enough for our use case. Or am I missing something?

ejgallego:
> It seems that the `enabled_if` only allows boolean expressions made of variables from https://dune.readthedocs.io/en/stable/concepts.html#variables, which is not enough for our use case. Or am I missing something?

Actually I thought there was a way to have enabled_if depending on an installed opam package but it seems indeed there is not; this would work
```
(enabled_if (= env:BUILD_EMACS=false true))
```
but that's a bit messy [tho the wrapper in make can set the env var correctly, in opam indeed you can set the variable conditionally on `conf-emacs`. Cleaner is to have a separate package instead of the conditional bit, but adds other overhead.

ejgallego:
Not really sure about all the possibilities here so it could be worth asking dune devs.

gabrielhdt:
> > It seems that the `enabled_if` only allows boolean expressions made of variables from https://dune.readthedocs.io/en/stable/concepts.html#variables, which is not enough for our use case. Or am I missing something?
> 
> Actually I thought there was a way to have enabled_if depending on an installed opam package but it seems indeed there is not; this would work
> 
> ```
> (enabled_if (= env:BUILD_EMACS=false true))
> ```
> 
> but that's a bit messy [tho the wrapper in make can set the env var correctly, in opam indeed you can set the variable conditionally on `conf-emacs`. Cleaner is to have a separate package instead of the conditional bit, but adds other overhead.

This doesn't work because if `BUILD_EMACS` is false, dune complains that it doesn't know how to build the `elc` target (since the rule is disabled). For now, I've disabled emacs mode precompilation.

fblanqui:
@gabrielhdt let me know when this PR is ready for merging

gabrielhdt:
If we're happy with these modifications, it can be merged.

ejgallego:
> `dune @install` which installs all packages available in the workplace.

This is fishy, the generated dune file should call `dune -p pkg` for the corresponding package.

ejgallego:
> This is fishy, the generated dune file should call `dune -p pkg` for the corresponding package.

Indeed just checked in some projects I have here and install is called in release mode with the proper package scoping.

dune-project
```diff
@@ -2,3 +2,27 @@
...
+This package contains both the `lambdapi` proof assistant and the `lp-lsp`
+Language Server Protocol helper for IDEs to check lambdapi files.")
+ (depends
+  (ocaml (>= 4.07.0))
+  (menhir (>= 20200624))
+  (alcotest :with-test)
+  (bindlib (>= 5.0.1))
+  (timed (>= 1.0))
+  (pratter (>= 1.1))
+  (why3 (>= 1.3.1))
```
fblanqui, line 25: We currently need to restrict why3 to <= 1.3.3. See https://github.com/Deducteam/lambdapi/pull/608.

gabrielhdt:
> > This is fishy, the generated dune file should call `dune -p pkg` for the corresponding package.
> 
> Indeed just checked in some projects I have here and install is called in release mode with the proper package scoping.

Ah I see what the problem was, the test target depends on install which does not use `-p`, I'll try again using packages, thanks @ejgallego !

fblanqui:
Note that the definition of identifiers in vim is not correct anymore. What can we do about that?

fblanqui:
A question: since lambdapi.opam is now generated, why do you want to keep under git?

ejgallego:
> A question: since lambdapi.opam is now generated, why do you want to keep under git?

Was answered above; in short, opam requires it due to `opam pin` among others.

gabrielhdt:
> A question: since lambdapi.opam is now generated, why do you want to keep under git?

I think opam still needs to access the file statically to operate, it doesn't use the `dune-project` yet, and it doesn't generate the file (yet).

tests/dune
```diff
@@ -1,4 +1,4 @@
 (tests
  (names utils ok_ko)
- (deps (glob_files OK/*.lp) (glob_files KO/*.lp) lambdapi.pkg)
+ (deps (glob_files OK/*.lp) (glob_files "OK/a b/*.lp") (glob_files KO/*.lp) lambdapi.pkg)
  (libraries lambdapi.core lambdapi.handle lambdapi.tool alcotest))
```
fblanqui, line 4: Alt-ergo <= 2.3.3 is needed for running tests. Shouldn't this dependency appear somewhere?

gabrielhdt: It's more a binary than a caml library isn't it?

fblanqui: What do you mean?

ejgallego: It only makes sense to appear as a dependency if Dune may know how to build it [for example by composition]

fblanqui:
This needs a slight update after the merge of #608. Then, I can merge.

gabrielhdt:
I'm trying to merge things but I have issues compiling wh3 1.4.0 on my machine.

lambdapi.opam
```diff
@@ -19,7 +19,7 @@ depends: [
   "bindlib" {>= "5.0.1"}
   "timed" {>= "1.0"}
   "pratter" {>= "1.1"}
-  "why3" {>= "1.3.1" & <= "1.3.3"}
```
gabrielhdt: The `dune-project` file should be changed, not this one.

fblanqui: Do you want me to make the change?

gabrielhdt: If I manage to compile why3, yes

gabrielhdt: Ah sorry, I misread, yes, you can make the change :)

fblanqui:
@gabrielhdt Did you succeed in installing why3 1.4.0 on your machine?

gabrielhdt:
It's compiling, but it looks good.

fblanqui:
make fails after make clean because lambdapi-site-file.elc is not removed by make clean. Shouldn't we have make clean remove this file?

gabrielhdt:
> make fails after make clean because lambdapi-site-file.elc is not removed by make clean. Shouldn't we have make clean remove this file?

Perhaps `dune clean` should handle it? Otherwise we have to go to the opam prefix and do it ourselves indeed.

fblanqui:
Not in the opam prefix; it's about local files only. In fact, there should a clean target in editors/emacs, and in editors/vscode?, called from the main Makefile.

gabrielhdt:
> Not in the opam prefix; it's about local files only. In fact, there should a clean target in editors/emacs, and in editors/vscode?, called from the main Makefile.

Indeed, I created a recursive call.

fblanqui:
@gabrielhdt I see that you renamed ./GNUmakefile into Makefile. Is there a reason not to do it for editors/emacs/GNUmakefile as well?

gabrielhdt:
Indeed, not really, I'll rename everything.

fblanqui:
Don't forget tools/rec_to_lp/GNUmakefile too.

fblanqui:
So, now, it's ready for merging, right?

gabrielhdt:
Yes it is.

gabrielhdt:
Ah one might add in `CONTRIBUTING` that `lambdapi.opam` shouldn't be edited

fblanqui:
Ok, waiting for your modif on CONTRIBUTING and CHANGES too.

fblanqui:
@gabrielhdt Don't you want to edit CHANGES too?

gabrielhdt:
Of course, I forgot, thanks

Issue #494, opened 2020-12-19
Allow focus on unification goals


fblanqui:
functions that need to be fixed:
- Proof.focus_goal
- Tactics: all tactics should take a goal as argument and fail if this is not the right kind of goal
- client.ts: in getGoalsEnvContent, remove typGoalNo and use i instead

PR #499, opened 2020-12-21
fix client.ts for #494

firewall2142:
- fixes client.ts for #494
- fixes jump to next proof for vscode


fblanqui:
@firewall2142 Thanks to update your branch so that I can merge it.

editors/vscode/src/client.ts
```diff
@@ -187,7 +187,7 @@ function nextProofPosition(document: TextDocument, proofState : Position, direct
     //Looking for the next line with "proof" in it (or the beggining/end of file)
     let line : string = document.lineAt(current).text;
 
-    while(!line.includes("proof") && current + direction >= 1 && current + direction <= document.lineCount - 1) {
+    while(!(/^\s*opaque\s+symbol/.test(line)) && current + direction >= 1 && current + direction <= document.lineCount - 1) {
```
fblanqui, line 190: What do you want to test here exactly? The beginning of a proof? If so, then it should be `while(!line.includes("begin")`.

firewall2142: done

PR #504, opened 2020-12-24
fix #348

### LINKED ISSUE #348: Emacs mode does not work when opening a file that does not exist yet

firewall2142:


firewall2142:
Merry Christmas🎄!
The path was only used to find the directory with `lambdapi.pkg` and hence get the module name. So, I removed a few checks which were ensuring the file and the directories the file resided in existed.

Things should also work with: `emacs ~/exist1/nonexist1/nonexist2/nonexistfile.lp`

amelieled:
Thanks @firewall2142 and merry Christmas too!
I hope you have fixed the problem.

Issue #501, opened 2020-12-23
Problem with `proofterm` and `print`


amelieled:
It seems to be not possible to print the initial goal (with `print`) and the final proof-term (with `proofterm`).
For example:
```
opaque symbol eq_sym {a} (x y:τ a) : π (x=y) → π (y=x) ≔
begin
  print                     <-- fail
  assume a x y xy
  proofterm
  print
  symmetry
  proofterm
  print
  apply xy
  proofterm                 <-- fail
end
```

fblanqui:
Thanks to provide a standalone file.

amelieled:
It's just an example. You can try by yourself with any files.

fblanqui:
Bad answer. When you raise an issue, you should provide a standalone file so that we can reproduce the issue.

amelieled:
When it's a specific one, yes, but it's not the case here.

fblanqui:
It is. Why don't you try to be more helpful to identify what is actually the problem? It is common practice to ask for standalone and possibly minimal examples. See for instance https://github.com/coq/coq/issues/new .

amelieled:
I know this "common practice" and the most important is that my explanation is self-sufficient. 
You can take the file tests/OK/natproofs.lp for example, but it's better if you take another example in order to, maybe, find a counter-example that proofs my generalization isn't correct.

PR #505, opened 2020-12-28
fix a number of problems wrt focus

fblanqui:


fblanqui:
Should fix https://github.com/Deducteam/lambdapi/issues/494 and https://github.com/Deducteam/lambdapi/issues/501.

PR #510, opened 2021-01-03
bug in PR#504

firewall2142:
The file path should've been expanded in package.ml in #504. This caused `lambdapi check natproofs.lp` to get stuck.

### DIFF
--- a/src/core/files.ml
+++ b/src/core/files.ml
@@ -266,16 +266,7 @@ let file_to_module : string -> Path.t = fun fname ->
       fatal_no_pos "Expected any of: %a." pp_exp valid_extensions
     end;
   (* Normalizing the file path. *)
-  let fname =
-    let rec normalize fname =
-      if Sys.file_exists fname then
-        Filename.realpath fname
-      else
-        let dirnm = Filename.dirname fname in
-        let basenm = Filename.basename fname in
-        Filename.concat (normalize dirnm) basenm
-    in
-    normalize fname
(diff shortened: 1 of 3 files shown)


src/core/package.ml
```diff
@@ -75,6 +75,15 @@ let read : file_path -> config_data = fun fname ->
     [Sys_error] is raised. Note that [fname] is first normalized with a call
     to [Filename.realpath]. *)
 let find_config : file_path -> file_path option = fun fname ->
+  let rec normalize fname =
```
fblanqui, line 78: You already defined this function in files.ml. This should be factorized by adding a function in filename.ml.

firewall2142: done

src/core/files.ml
```diff
@@ -266,16 +266,7 @@ let file_to_module : string -> Path.t = fun fname ->
...
-      if Sys.file_exists fname then
-        Filename.realpath fname
-      else
-        let dirnm = Filename.dirname fname in
-        let basenm = Filename.basename fname in
-        Filename.concat (normalize dirnm) basenm
-    in
-    normalize fname
+  let fname = Filename.normalize fname
   in
```
fblanqui, line 270: Thanks to put the "in" on the same line as "let" as everything holds on a single line in this case.

firewall2142: done

fblanqui:
Thanks @firewall2142 !

PR #512, opened 2021-01-05
Implicitly call tac_solve at the beginning of proof scripts

fblanqui:


fblanqui:
Perhaps this PR also solves the problem that, in symbol declarations with no proof script, the (implicit) goals were printed.

Issue #447, opened 2020-10-16
Remove compilation option -w -27 due to Earley


fblanqui:
https://github.com/Deducteam/lambdapi/blob/9e080cd73ab5d655086197fcdee5a8290b87ce71/src/core/dune#L15

Rehan-MALAK:
I would be very happy to find a solution for this... even a very crappy hackish one.
Let the compiler find for us unused variable avoid so much stupid errors in ocaml.

Issue #261, opened 2019-10-24
Parsing entries one by one for big files


francoisthire:
The current behavior of lambdapi is: 1. parse the file to generate a list of entries, 2. type check the file. 

This is becoming an issue with the importation of Isabelle/HOL into lambdapi. For example, @01mf02 is generating files of size 250mb. In this case, the size of the AST is getting big and the memory explodes at parsing stage.

For type checking, all the entries do not need to be in the memory, type checking could be done entry by entry. Sure, you need to store something to type check other entries, but in HOL for example, the proof can be thrown away.

01mf02:
How about making parse_file return a Seq instead of a list? I've used this technique in the past, and it yielded very good results performance-wise.

rlepigre:
Well, the memory remains linear in the size of the file, so I would not say that it "explodes". And even if we parse line by line that will remain the case since we typically store a new symbol or rewriting rule at each step.

The bigger problem is that Earley cannot handle files of that (unreasonable) size. And the Lambdapi syntax is definitely not meant for that kind of application.

Anyway, what do you big files look like? Are they formed of many short commands? Few bit commands? A mix of both?

francoisthire:
Why parsing a file line by line is a problem with Earley?

rlepigre:
That is doable I guess, it's just more complicated code.

francoisthire:
In the long term, this will be an issue anyway. When terms are generated from a program, in general we don't need all the machinery provided by Earley. Instead, we only need a simple syntax with maybe holes for implicit parameters.

fblanqui:
@01mf02, could you please provide an URL towards your file so that we can have a look at it?

01mf02:
Certainly, @fblanqui. Here is the output for [HOL.Groups](http://cl-informatik.uibk.ac.at/users/mfaerber/dk_hol_groups.lp.bz2), which Dedukti is able to parse on my machine (typechecking fails still ATM), and here is the output for [HOL.Hilbert_Choice](http://cl-informatik.uibk.ac.at/users/mfaerber/dk_hol_hilbert_choice.lp.bz2), on which Lambdapi currently chokes.

fblanqui:
Hi @01mf02 .

I see that, in your big file, you use `definition proofXXX : A := p` for what looks like proofs. Couldn't you instead do the following: `theorem thmXXX : A proof refine p qed`? (I checked on some example and it worked.)

[...]

Issue #515, opened 2021-01-06
Emacs/VSCode: navigation in files command-by-command


fblanqui:
Currently navigation is done line by line, which is incorrect when a command spreads out on several lines (e.g. inductive/rule ... with ...). Navigation should be command by command. Note that there may be several commands on the same line as well. Perhaps the introduction of a command terminator ";" in the new parser will make things easier.

ejgallego:
Shouldn't matter too much, the lsp server already can map a position to the concrete sentence, so you could just call lp to obtain sentence information.

PR #516, opened 2021-01-07
Implementation of new parser

gabrielhdt:
This pull request is identical to the the other one, but hopefully cleaner regarding github's interface. See #441 for more information.

It fixes https://github.com/Deducteam/lambdapi/issues/447.


fblanqui:
editors/vscode/media/styles.css is removed. I guess it is a mistake during merge, right?

src/core/compile.ml
```diff
@@ -17,14 +17,20 @@ let gen_obj = Stdlib.ref false
...
+  | false -> Parser.Legacy.parse_file fname
 
 (** [compile force path] compiles the file corresponding to [path], when it is
     necessary (the corresponding object file does not exist,  must be updated,
     or [force] is [true]).  In that case,  the produced signature is stored in
     the corresponding object file. *)
 let rec compile : bool -> Path.t -> Sign.t = fun force path ->
-  let base = Files.module_to_file path in
+  (* Remove escaping quotes from module paths. *)
+  let remove_quote p =
```
fblanqui, line 28: Move to lp_lexer together with is_escaped?

src/core/pos.ml
```diff
@@ -17,34 +13,13 @@ and pos_data =
 (* NOTE laziness is essential on big files (especially those with long lines),
    because computing utf8 positions is expensive. *)
```
fblanqui, line 14: If you remove lazyness, this comment should be removed as well. But are you sure you want to remove lazyness on positions?

src/core/pratt.ml
```diff
@@ -0,0 +1,133 @@
...
+(** Type of operators with additional data. *)
+(* REVIEW: simplify the type into [Unary of float * assoc] and remove [unop]
+   and [binop] *)
+type operator =
+  | Unary of unop
+  | Binary of binop
+
+(** Table containing all registered binary and unary operators that may appear
+    in terms parsed by {!val:Pratt.expression}. *)
+let operators : operator StrHtbl.t = StrHtbl.create 17
```
fblanqui, line 21: This map should evolve with the state (Timed). In fact, pratt should depend on the sig_state which already contains the operator map in the field pp_hints.

src/core/scope.ml
```diff
@@ -105,17 +106,23 @@ let rec get_implicitness : p_term -> bool list = fun t ->
...
+(** [get_parsed_args t] decomposes the parser level term [t] into a spine
+    [(h,args)], when [h] is the term at the head of the application and [args]
+    is the list of all its arguments with infix and prefix operators added
+    using {!module:Pratt}.  Note that sugared applications (e.g., infix
+    symbols) are not expanded, so [h] may still be unsugared to an
+    application. *)
+(* REVIEW: this function is one of the few that use the Pratt parser, and the
+   term is converted from appl to list in [Pratt.parse], then rebuilt into
+   appl node (still by Pratt.parse), then again decomposed into a list by the
+   function. We may make [Pratt.parse] to return already a list of terms. *)
```
fblanqui, line 118: Yes, why not indeed? Concerning the name, why not "get_pratt_args" instead, to avoid confusion?

fblanqui:
I am wondering: why not renaming the files legacy_* into dk_*?

src/core/pos.ml
gabrielhdt: Lazyness would be handled by the lexer, Sedlex doesn't use it, so we don't.

src/core/pratt.ml
gabrielhdt: I do completely agree with that. In fact, I'm trying to externalise completely the Pratt part, so that we just call a library. And the operator are in the `signature` of the `sig_state`, right?

src/core/scope.ml
gabrielhdt: It's just not done yet, there is no reason.

gabrielhdt:
> I am wondering: why not renaming the files legacy_* into dk_*?

We can indeed.

src/core/pratt.ml
fblanqui: Ok but the point is about the correct update of the operators table. Why not removing operators and passing pp_hints as argument to the pratt functions?

fblanqui: The operators are in the pp_hints filed of the sig_state, not in the signature.

src/core/pratt.ml
```diff
@@ -0,0 +1,133 @@
...
+      not an operator). *)
+  let lbp : p_term -> priority = fun pt ->
+    match get_ident pt with
+    | Some(s,_) ->
+        begin match StrHtbl.find_opt operators s with
+          | Some (Binary(_, _, bp, _))
+          | Some (Unary(_, bp, _)) -> bp
+          | None -> assert false
+        end
+    | _ -> assert false (* [t] must be an operator *)
```
fblanqui, line 51: In the comment, you say that it returns 0 if it is not an operator but, here, you assert false.

editors/emacs/lambdapi-mode.el
```diff
@@ -2,7 +2,7 @@
 
 ;; Copyright (C) 2020 Deducteam
 
-;; Author: Rodolphe Lepigre, Gabriel Hondet, Ashish Barnawal
+;; Authors: Ashish Barnawal, Diego Riviero, Gabriel Hondet, Rodolphe Lepigre
```
fblanqui, line 5: This diff should disappear if you merge master in your branch.

fblanqui: Reminder: style.css is still missing as well.

src/core/handle.ml
```diff
@@ -174,32 +178,34 @@ let handle_inductive_symbol :
     do not work on this structure directly,  although they act on the contents
     of its [pdata_p_state] field. *)
 type proof_data =
-  { pdata_stmt_pos : Pos.popt (** Position of the proof's statement. *)
+  { pdata_stmt_pos : Pos.popt (** Position of the declared symbol. *)
```
fblanqui, line 181: This diff should disappear when merging master.

src/core/pos.ml
```diff
@@ -90,24 +79,27 @@ let to_string : ?print_fname:bool -> pos -> string =
     Printf.sprintf "%s%d:%d-%d" fname start_line start_col end_col
 
 (** [print oc pos] prints the optional position [pos] to [oc]. *)
-let print : Format.formatter -> popt -> unit = fun ch p ->
+let print : Format.formatter -> pos option -> unit = fun ch p ->
```
fblanqui, line 82: why not using popt? idem line 89.

gabrielhdt:
As can be seen, I modified the use of the pratt algorithm. Now I use the `pratter` library (with the `pure` branch), which is under development (accessible at https://github.com/gabrielhdt/pratter), and the pratt parsing algorithm has indeed access to the signature state. So there is no additional operator table anymore.

Your remark about qualified identifiers and resolution is not yet solved though (regarding bound variables which have the same name as an infix operator, `\+, + * +`), everything happen in the module given as argument to the pratt parser. We may indeed, as you suggested in your previous comment (in the other pull request) use an other construction in `p_terms`.

fblanqui:
The other constructor for p_term is an optim for not calling find_qid twice. It does not need to be implemented now.

src/core/scope.ml
gabrielhdt: In fact, to do this properly, I think we should use n-ary applications in `p_terms` and binary in `terms`. What do you think?

fblanqui: As this implies many changes, I suggest to leave this for another PR.

gabrielhdt:
It seems that this pull request is in near final state. The errors are due to the processing of non associative symbols, do we still want to keep them? In a sense, we may force symbols to be left or right associative, since non associativity may be a consequence of the type of symbols.

I also have to publish the new version of `pratter`, which shouldn't take long (publishing version 0.1.1 took an evening).

fblanqui:
Could you please tell which test files fail?

gabrielhdt:
`rewrite.lp`, `printA.lp`, `printC.lp`, `328.lp`, `builtin_zero_succ.lp`, `printB.lp` and `inductive_type.lp`. I think they all fail because of non associative symbols (I checked two or three files among them), they fail on `assert false` of line 65 of `pratt.ml`.

src/core/pratt.ml
```diff
@@ -0,0 +1,113 @@
...
+    | None -> (
+        match StrMap.find_opt s ss.Sig_state.binops with
+        | Some sym -> (
+            match Terms.SymMap.find_opt sym ss.Sig_state.pp_hints with
+            | Some (Infix (_, _, _, qid)) -> qid
+            | _ -> assert false )
+        | None -> qid )
+
+
+  module Pratt_terms : Pratter.SUPPORT
```
fblanqui, line 44: @gabrielhdt Could you please add an URL to the definition of SUPPORT?

src/core/pretty.ml
```diff
@@ -12,10 +12,60 @@ open Console
 open Pos
 open Syntax
 
+(** [is_keyword s] returns [true] if [s] is a keyword (defined by the lexer)
+    in {!module:Lp_lexer}. *)
+let is_keyword : string -> bool = fun s ->
```
fblanqui, line 17: Should be moved to lp_lexer.

fblanqui: For efficiency, you could use a map or hashtable.

src/core/scope.ml
fblanqui: Just, would you mind to rename "get_parsed_args" into, say, "get_pratt_args"?

src/core/pratt.ml
fblanqui: Currently, this does not match https://github.com/gabrielhdt/pratter/ ...
Perhaps, while this is still under development, it would easier and faster to include pratter.ml, and later remove it and use the opam version instead?

gabrielhdt: I've published the version used now, so the repository is up to date. I'll add the file anyway.

src/core/lp_lexer.ml
```diff
@@ -0,0 +1,277 @@
...
+exception SyntaxError of strloc
+
+(** Lexer for Lambdapi syntax. *)
+module Lp_lexer : sig
+
+  val is_identifier : string -> bool
+  (** [is_identifier s] is [true] if [s] is a valid identifier. *)
+
+  val is_keyword : string -> bool
+  (** [is_keyword s] retunrs [true] if [s] is a keyword. *)
```
fblanqui, line 49: typo: retunrs

src/core/pratter.ml
```diff
@@ -0,0 +1,131 @@
...
+  (** The table is used to store available operators. *)
+
+  val get_unary : table -> term -> priority option
+  (** [get_unary tbl t] returns the priority, or binding power of operator
+      identified [t] if it is a unary operator, or [None]. *)
+
+  val get_binary : table -> term -> (priority * associativity) option
+  (** [get_binary tbl t] returns the priority (or binding power) and
+      associativity of operator [t], or [None] if [t] isn't a binary
+      operator. *)
```
fblanqui, line 40: Why not having a single function
```
type op = Una | Bin of associativity
val get_sym : table -> term -> (op * priority) option
```
?

src/core/pratt.ml
fblanqui: Thanks @gabrielhdt . It would be still very useful to add an URL here anyway, or at the top of the file, using `@see <...>`.

src/core/pratt.ml
```diff
@@ -0,0 +1,113 @@
...
+          | Some(Prefix((_, prio, _))) -> Some(prio)
+          | _ -> None )
+      | _ -> None
+
+    let make_appl (tbl, env) t u =
+      let pos =
+        match t.pos, u.pos with
+        | Some(pt), Some(pu) -> Some(cat pt pu)
+        | None, _ | _, None -> None
+      in
```
fblanqui, line 89: You could move this cat function on popt to pos.ml.

src/core/pratt.ml
```diff
@@ -0,0 +1,113 @@
...
+     [Map.find_opt]s. Things may be optimised by modifying the data structure
+     of the signature state, notably making the qualified identfier to which
+     operators are linked available in the [unops] field of the signature
+     state. *)
+  let find_qid : Sig_state.t -> Env.t -> qident -> qident =
+    fun ss env ({elt=(mp, s); _} as qid) ->
+    try
+      if mp <> [] then raise Not_found;
+      ignore (Env.find s env);
+      qid
```
fblanqui, line 28: Why not having find_qid return a `qident option` and here, in the case of a variable, replace `ignore (Env.find s env); qid` by `Some (Env.find s env)`?

src/core/pratt.ml
```diff
@@ -0,0 +1,113 @@
...
+    with Not_found ->
```
fblanqui, line 29: Here, you know that it cannot be a variable. So it must be a symbol. You can then call Sig_state.find_sym to properly know which sym it actually is. Then, to know whether it is unary or binary, you can look into pp_hint.

fblanqui: By the way, we may perhaps get rid of unops and binops as these maps have been introduced for Earley? pp_hints seems to be enough.

src/core/pratt.ml
```diff
@@ -0,0 +1,113 @@
...
+          match StrMap.find_opt suffix tbl.binops with
+          | None -> None
+          | Some(sym) ->
+          match Terms.SymMap.find_opt sym tbl.pp_hints with
+          | Some(Infix((_, assoc, prio, _))) -> (
+              let assoc =
+                match assoc with
+                | Assoc_left -> Pratter.Left
+                | Assoc_right -> Pratter.Right
+                | Assoc_none -> assert false
```
fblanqui, line 65: Non-associative infix symbols like `::` (cons) on lists are important and useful. We need to handle them. I don't understand why this is a problem. Could you please elaborate on that?

src/core/pratter.ml
```diff
@@ -0,0 +1,131 @@
+(* FIXME: use opam version instead *)
+(** This modules defines a functor whose image is a parser for terms with
+    applications, binary and unary operators. These terms are specified in the
+    argument of the functor.
+
+    The algorithm implemented is an extension of the Pratt parser. The Sunting
+    Yard algorithm could also be used.
+    @see <https://dev.to/jrop/pratt-parsing>
+    @see <https://effbot.org/zone/simple-top-down-parsing.htm> *)
```
fblanqui, line 9: This link is broken at the moment.

fblanqui: Another link that may be useful is: https://en.wikipedia.org/wiki/Operator-precedence_parser#Pratt_parsing

src/core/pratt.ml
fblanqui: Well, `::` is perhaps not a very good example as it could be considered as right associative: 1::2::[] = 1::(2::[]).
If you have c:T->U->V and want x c y c z = x c (y c z), you need x:T, y:T, z:U and V=U, so c:T->U->U.
Conversely, if you want x c y c z = (x c y) c z, you need x:T, y:U, z:U and V=T, so c:T->U->T.

fblanqui: Files where you mentioned a problem:
rewrite.lp: = is declared as infix; no associativity is possible here; yet this notation is VERY important! ;-)
328.lp: `::` declared infix with no associativity (could be fixed by declaring it right associative)
builtin_zero_succ.lp: + is declared infix with no associativity (yet); note that someone may want to declare an infix notation, later prove that it is associative and add associativity as a rule?
inductive_type.lp: = is declared infix like in rewrite.lp
(I cannot find the files printA.lp, printB.lp and printC.lp that you also mentioned.)

src/core/pratt.ml
gabrielhdt: Yes we could, it looks like the applicative functor use of the option, or something of the king

src/core/pratt.ml
gabrielhdt: > Well, `::` is perhaps not a very good example as it could be considered as right associative: 1::2::[] = 1::(2::[]).
> If you have c:T->U->V and want x c y c z = x c (y c z), you need x:T, y:T, z:U and V=U, so c:T->U->U.
> Conversely, if you want x c y c z = (x c y) c z, you need x:T, y:U, z:U and V=T, so c:T->U->T.

But what should be the behaviour of non associative symbols? In the previous parser, parsing failed on an input `x c y c z`. Do we keep this? When you say `x c y c z = x c (y c z)`, `c` is right associative, right?

src/core/pratt.ml
gabrielhdt: Well `Env.find s env` returns a `tvar`, not a `qident`, so your `Some (Env.find s env)` is ill-typed, isn't it?

src/core/pretty.ml
gabrielhdt: I'll see if I can use back the lexer's `match%sedlex` to factorise code.

src/core/pratt.ml
fblanqui: Yes. If c is not associative, x c y c z should generate a syntax error.

src/core/pratter.ml
fblanqui: Remark: it may look like this forbids to have identifiers that can be both unary and binary symbols depending on the number of arguments they are applied to, like - for instance. But we could later implement this as well by modifying the sign type and the Sign.find function.

src/core/pratt.ml
fblanqui: Ah, right. It should be `ignore (Env.find s env); None`.

src/core/pratt.ml
gabrielhdt: Indeed, the sig state may be lightened

gabrielhdt: In its current form, `Sig_state.find_sym` doesn't fetch the qid of operators, if I'm not mistaken, that's why I had to create this function.

fblanqui: Right, it returns a sym, but you can build a qualified ident from a sym.

fblanqui: Btw, it seems to me that many things could simplified in Sign.t and Sig_state.t, but perhaps in another PR. First, sign_idents could be removed. Then, sign_symbols, sign_unops, sign_binops and sign_quants could be merged into a single sign_symbols:(sym*popt*pp_hint)StrMap.t ref if notations only comes with symbol declarations. Otherwise, if notations can be defined afterwards, sign_unops, sign_binops and sign_quants could be merged into a single sign_pp_hints:pp_hint SymMap.t ref. In Sig_state.t, I think that unops and binops can be removed, because an identifier is identified as a symbol thanks to Sign.sign_symbols in case of a qualified identifier, or thanks to Sig_state.in_scope in case of an unqualified identifier. Then, you can use Sig_state.pp_hints to know if what is its notation.

gabrielhdt: It doesn't even return a sym, I think that calling `find_sym` on, say, `+` which has been declared infix just raises `Not_found`. The Earley parser would add the tuple `(op, prop, qid)` (which would be `("+", 1.0, plus)` in our case) into the syntax, and scoping would only use the `qid` (so the `plus`) rather than the `op`. Here we'd like to keep a mapping from `"+"` to `plus` additionally.

src/core/pratter.ml
gabrielhdt: It may require quite a change in the pratt parser...

src/core/pratt.ml
fblanqui: I don't understand. Currently, in master, in case of a P_Iden, scope_head calls Scope.find_qid which calls Sig_state.find_sym which returns a sym. If the ident is not qualified, find_sym looks in in_scope which from a string returns a sym. And, in case of P_UnaO and P_BinO, which disappear in your PR, scope_head calls Sig_state.find_sym directly. But Earley maps unqualified identifiers corresponding to unary or binary operators in scope to some pp_hint-like data. See Parser.get_ops. Note that, while sign_unops/binops maps strings to sym+pp_hint-like data, the sym part is ignored. It could have returned this sym field as well so that there is no need to call find_sym later.

src/core/pratter.ml
fblanqui: Are you talking of my first question or my second remark?

src/core/pratt.ml
gabrielhdt: In case of `P_UnaO` and `P_BinO`, `scope_head` calls `Sig_state.find_sym` using the `qid` associated to the binary operator. In our case, we do not have the `qid` associated with the binary operator available anymore. So what I say is that we have to fetch the qid associated to the `"+"` identifier using the signature state. Does it make sense?

src/core/pratter.ml
gabrielhdt: The second indeed.

src/core/pretty.ml
gabrielhdt: After some thought, I'm not sure which would be the most efficient and the clearer, so I suggest that I leave a note, and we'll optimise in a later pull request.

src/core/pratt.ml
fblanqui: But why do you want to make a distinction between identifiers? If you have an unqualified ident that is not a variable, you can look at in_scope to get the corresponding sym and then look at pp_hints to know if it is unary, binary, etc. You can also build a qualified ident out of this sym like in pp_qualified if you need it.

gabrielhdt: But my point is that you won't get the symbol associated to `"+"`, it is not in `in_scope`, if I understand correctly.

fblanqui: Then, this is wrong and should be fixed.

src/core/pratt.ml
```diff
@@ -0,0 +1,94 @@
...
+            let qid = find_op_qid tbl env id in
+            Option.map (Sig_state.find_sym ~prt:true ~prv:true false tbl) qid
+          in
+          let f sym =
+            match Terms.SymMap.find_opt sym tbl.pp_hints with
+            | Some(Infix(_, assoc, prio, _)) -> Some(Pratter.Bin assoc, prio)
+            | Some(Prefix(_, prio, _)) -> Some (Pratter.Una, prio)
+            | _ -> None
+          in
+          Option.bind f sym )
```
fblanqui, line 61: find_sym and find_opt in pp_hints are done twice. It should be possible to improve this a little bit by partly unfolding find_op_qid.

src/core/pratt.ml
```diff
@@ -0,0 +1,94 @@
...
+              | Some(qid) -> qid
+              | None -> qid
+            in
+            P_Iden(qid, b)
+          | _ -> e
+        in
+        Pos.map f
+      in
+      let t = build t in
+      let u = build u in
```
fblanqui, line 84: After all, is it worth doing this, as find_sym will be called anyway? To really avoid multiple calls to find_sym, we should replace P_Iden by new constructors P_Var:strloc->p_term and P_Sym:sym->p_term.

src/core/parser.ml
```diff
@@ -320,409 +23,120 @@ let add_prefix : string -> string -> string = fun p s ->
...
+    let (ids, last) = List.split_last ids in
+    Ok(ids, fst last)
+with Invalid_id(i, pos) -> Error(i, pos)
+
+(** Parsing legacy (Dedukti2) syntax. *)
+module Dk : PARSER = struct
+  let parse_lexbuf : string -> Lexing.lexbuf -> Syntax.ast =
+    fun fname lexbuf ->
+    Stdlib.(DkLexer.filename := fname);
+    let lines = ref [] in
```
fblanqui, line 116: Why not renaming lines as cmds like for Lp?

src/core/parser.ml
```diff
@@ -320,409 +23,120 @@ let add_prefix : string -> string -> string = fun p s ->
...
+    let lexer = LpLexer.lexer lexbuf in
+    let cmds = ref [] in
+    try
+      while true do
+        let cmd = parse lexer in
+        cmds := cmd :: !cmds;
+      done;
+      assert false (* Should raise end of file before *)
+    with
+    | End_of_file -> List.rev !cmds
```
fblanqui, line 53: Why do you parse command by command and then have to reverse? Why not defining an entry in LpParser for handling a list of commands?
Idem for Dk below.
Remark: later we could improve all this by interleaving parsing and handling: indeed we could handle each command once it is parser; in this case, there is no need for building a list of commands.

gabrielhdt: From memory, parsing a list of commands raised an error in the parser generation (something like end of stream conflict). I'll check that again.

src/core/pratt.ml
gabrielhdt: Well seen, all this can simply be removed indeed.

src/core/parser.ml
gabrielhdt: Parsing and handling are interleaved in 88e97d5 using a stream rather than a list. Does this suit you?

src/core/handle.ml
```diff
@@ -462,15 +468,15 @@ let handle_cmd : sig_state -> p_command ->
             (* Make sure the operator has a fully qualified [qid]. *)
             let unop = (s, prio, with_path sym.sym_path qid) in
             out 3 "(conf) %a %a\n" pp_symbol sym pp_hint (Prefix unop);
-            add_unop ss s (sym, unop)
+            add_unop ss (Pos.none s) (sym, unop)
```
fblanqui, line 471: Use cmd.pos instead ? Idem for binop below.

src/core/lpLexer.ml
```diff
@@ -0,0 +1,279 @@
...
+  (** [is_identifier s] is true if [s] is an identifier. *)
+  let is_identifier : string -> bool = fun s ->
+    let lexbuf = Sedlexing.Utf8.from_string s in
+    match%sedlex lexbuf with
+    | id -> true
+    | _ -> false
+
+  let is_keyword : string -> bool = fun s ->
+    (* NOTE this function may be optimised using a map, a hashtable, or using
+       [match%sedlex]. *)
```
fblanqui, line 126: Do you plan to do it in this PR?

gabrielhdt: I don't think so, to do it properly, we'd have to time things to be sure we optimise correctly. And it remains easy to modify and doesn't impact the rest, so I don't see it as urgent.

src/core/handle.ml
gabrielhdt: I was rather thinking of modifying the unop and binop types to include the position of the symbol declaration, and have `unop = string loc * priority * qident`

src/core/sig_state.ml
```diff
@@ -109,24 +107,30 @@ let add_symbol : sig_state -> sig_symbol -> sig_state * sym =
   ({ss with in_scope; pp_hints}, s)
 
 (** [add_unop ss n x] generates a new signature state from [ss] by adding a
-   unary operator [x] with name [n]. *)
-let add_unop : sig_state -> string -> (sym * unop) -> sig_state =
+    unary operator [x] with name [n]. This name is added to the scope. *)
+let add_unop : sig_state -> string loc -> (sym * unop) -> sig_state =
```
fblanqui, line 111: You can use 'strloc'.

src/core/pratt.ml
```diff
@@ -0,0 +1,64 @@
...
+      it doesn't recurse into abstractions or implications and alike. *)
+end = struct
+
+  open Lplib
+  open Pos
+
+  (** [find_sym ss env qid] is [{!val:Sig_state.find_sym} ss qid], but
+      doesn't fail if [qid] is in the environment [env]. It fetches private
+      and protected symbols as well. *)
+  let find_sym : Sig_state.t -> Env.t -> qident -> Terms.sym option =
```
fblanqui, line 23: Why not unfold this function if it is not used outside this module and if it is called only once?

fblanqui:
Nice! This should fix https://github.com/Deducteam/lambdapi/issues/261, 
right?

Le 18/01/2021 à 16:14, Gabriel Hondet a écrit :
>
> Parsing and handling are interleaved in 88e97d5 
> <https://github.com/Deducteam/lambdapi/commit/88e97d59113d50057e7423541302fb0332c000ce> 
> using a stream rather than a list. Does this suit you?
>

fblanqui:
So, it is the case? (I would like to know if we can avoid List.rev in 
pure.ml.)

Le 18/01/2021 à 10:51, Gabriel Hondet a écrit :
> From memory, parsing a list of commands raised an error in the parser 
> generation (something like end of stream conflict). I'll check that again.

fblanqui:
Is https://github.com/Deducteam/lambdapi/issues/359 fixed by your PR?

src/core/lpLexer.ml
gabrielhdt: Finally I kept your proposition, I'm not sure there's a clear gain in propagating the position of the string.

gabrielhdt:
> This should fix #261, right?

It _should_ indeed.

> So, it is the case?

Yes it is. We may be able to do something by tweaking further, but I'm not sure. Perhaps we can avoid `List.rev` in `pure.ml` using the stream?

And yes, issue #359 is fixed! I've added a test file.

src/core/lpLexer.ml
fblanqui: Well, it helps to give a more precise position. But it is not very important indeed.

fblanqui: At least, you could use mem_sorted instead of just mem.

gabrielhdt: Ah I used the wrong review thing, it's obviously about the use of `cmd.pos` in handle...

fblanqui:
I am wondering: why not removing Syntax.P_config_ident, Sign.sign_idents, Sig_state.unops, Sig_state.binops? They should be useless now.
And, as already mentioned in a previous comment, why not replacing sign_unops, sign_binops and sign_quants by a single sign_pp_hints:pp_hint StrMap.t ref?
Btw, in the definition of pp_hint, we should replace Syntax.unop by Pratter.priority, and Syntax.binop by Pratter.priority*Pratter.associativity. The types unop and binop are useful for handling the set prefix/infix commands only.

src/core/pratter.ml
fblanqui: The link seems permanently broken. Thanks to remove it

gabrielhdt:
`Sig_state.unops` and `Sig_state.binops` have already been removed. All about idents are useless as well indeed. And simplifications are yet to be done.

On the other hand, I'm not sure, _in fine_ that we'll be able to retrieve the previous "quantifier" syntax: it seems to me that they make the grammar ambiguous, and they can't be parsed efficiently.

fblanqui:
Concerning the quantifier notation, for the moment, you proposed to use backquote, but backquote is also used in emacs for entering various kinds of letters (greek letters, etc.). So shouldn't we use something else?

gabrielhdt:
> Concerning the quantifier notation, for the moment, you proposed to use backquote, but backquote is also used in emacs for entering various kinds of letters (greek letters, etc.). So shouldn't we use something else?

Yes, but it doesn't use the `quant` in the sig_state: terms are parsed directly as applications. All machinery related to quantifiers isn't used anymore currently.

For the backquote, completion can be inhibited with `C-q`, so one may type ``C-q `f`` to input `` `f ``

gabrielhdt:
> I am wondering: why not removing Syntax.P_config_ident, Sign.sign_idents, Sig_state.unops, Sig_state.binops? They should be useless now.
> And, as already mentioned in a previous comment, why not replacing sign_unops, sign_binops and sign_quants by a single sign_pp_hints:pp_hint StrMap.t ref?
> Btw, in the definition of pp_hint, we should replace Syntax.unop by Pratter.priority, and Syntax.binop by Pratter.priority*Pratter.associativity. The types unop and binop are useful for handling the set prefix/infix commands only.

It requires to change the definition of `eq_pp_hint`

fblanqui:
It is not very convenient to have to use C-q for something we may have to do often... Let's try to find something else.

fblanqui:
@gabrielhdt did you update CHANGES.md, syntax.bnf and the user manual?

src/core/lpParser.mly
```diff
@@ -0,0 +1,417 @@
...
+  /* | LAMBDA xs=arg_ident+ COLON a=term COMMA t=term */
+  /*     { make_pos $loc (P_Abst([xs, Some(a), false], t)) } */
+  | LET x=ident a=arg* b=preceded(COLON, term)? ASSIGN t=term IN u=term
+      { make_pos $loc (P_LLet(fst x, a, b, t, u)) }
+
+binder:
+  | id=ident COMMA t=term
+      { make_pos $loc (P_Abst([[Some(fst id)], None, false], t)) }
+  | id=ident COLON a=term COMMA t=term
+      { make_pos $loc (P_Abst([[Some(fst id)], Some(a), false], t)) }
```
fblanqui, line 395: I propose you to rename arg into arg_list and factorize the BACKQUOTE, PI and LAMBDA cases as follows:
```
term:
  | BACKQUOTE q=term_ident b=binder
    { make_pos $loc (P_Appl(q, Pos.make b.pos (P_Abst(fst b.elt, snd b.elt)))) }
  | PI b=binder { make_pos $loc (P_Prod(fst b.elt, snd b.elt)) }
  | LAMBDA b=binder { make_pos $loc (P_Abst(fst b.elt, snd b.elt)) }
  ...
binder:
  | xs=arg_list+ COMMA t=term { make_pos $loc (xs, t) }
  | x=arg_ident COLON a=term COMMA t=term
    { make_pos $loc ([[x], Some(a), false], t) }
```

fblanqui:
Le 19/01/2021 à 10:08, Gabriel Hondet a écrit :
> Yes, but it doesn't use the |quant| in the sig_state: terms are parsed 
> directly as applications.

I see.


> All machinery related to quantifiers isn't used anymore currently.

Well, it is still needed to print terms the way they are entered by the 
user. So Print.pp_quantifier should be fixed.

fblanqui:
Perhaps, one solution could be to have the lexer depends on the 
sig_state and, in case of an id, return a new token QUANT instead of ID. 
Note that \lambda and \Pi could be seen as predefined quantifiers. Then, 
in the grammar we could replace the rules on LAMBDA, PI and BACKQUOTE by 
simply QUANT binder.

Le 19/01/2021 à 09:32, Gabriel Hondet a écrit :
> On the other hand, I'm not sure, /in fine/ that we'll be able to 
> retrieve the previous "quantifier" syntax: it seems to me that they 
> make the grammar ambiguous, and they can't be parsed efficiently.

gabrielhdt:
As of 767cb3c
- Parsing fails on not fully applied operators (such as `3 +`),
- structures `sign.quants`, `sign.unops` and `sign.binops` have been merged.

If you have no more suggestion, @fblanqui, I'll edit the BNF (looking at [Obelisk](https://github.com/Lelio-Brun/Obelisk)), publish the adequate version of Pratter, edit the changelog, and voilà.

src/core/compile.ml
```diff
@@ -73,14 +75,16 @@ let rec compile : bool -> Path.t -> Sign.t = fun force path ->
...
-      let not_prv_fst k s_ = not_prv k (fst s_) in
-      sign.sign_symbols := StrMap.filter not_prv_fst !(sign.sign_symbols);
-      sign.sign_builtins := StrMap.filter not_prv !(sign.sign_builtins);
-      sign.sign_unops := StrMap.filter not_prv_fst !(sign.sign_unops);
-      sign.sign_binops := StrMap.filter not_prv_fst !(sign.sign_binops);
+      let not_prv sym = not (Terms.is_private sym) in
+      sign.sign_symbols :=
+        StrMap.filter (fun _ s -> not_prv (fst s)) !(sign.sign_symbols);
+      sign.sign_builtins :=
+        StrMap.filter (fun _ s -> not_prv s) !(sign.sign_builtins);
```
fblanqui, line 85: There should be no filtering here as there is no notion of private builtins.

src/core/sign.ml
```diff
@@ -15,17 +15,22 @@ type inductive =
   { ind_cons  : sym list  (** List of constructors                 *)
   ; ind_prop  : sym       (** Induction principle on propositions. *) }
 
+(** Syntactic sugar properties of symbols. They are linked to symbols to
+    provide syntax extensions to these symbols. *)
+type synt_sugar =
```
fblanqui, line 20: What about calling it "notation"?

src/core/sign.ml
```diff
@@ -15,17 +15,22 @@ type inductive =
...
+  | Prefix of unop (** Prefix (or unary) operator. *)
+  | Infix of binop (** Infix (or binary) operator. *)
```
fblanqui, line 22: What about replacing unop by priority, and binop by priority*associativity, as you don't need the other elements of unop and binop that are useful for handling the set prefix/infix commands only?

src/core/sign.ml
```diff
@@ -149,11 +149,12 @@ let link : t -> unit = fun sign ->
...
   PathMap.iter gn !(sign.sign_deps);
   sign.sign_builtins := StrMap.map link_symb !(sign.sign_builtins);
-  let hn (s,h) = (link_symb s, h) in
-  sign.sign_unops := StrMap.map hn !(sign.sign_unops);
-  sign.sign_binops := StrMap.map hn !(sign.sign_binops);
+  let lsy (sym, h) = link_symb sym, h in
+  sign.sign_syntax :=
+    (* Keys of the mapping are linked *)
+    SymMap.to_seq !(sign.sign_syntax) |>
+    Seq.map lsy |> SymMap.of_seq;
```
fblanqui, line 156: Why not using a SymMap.fold instead?

src/core/pratter.ml
```diff
@@ -0,0 +1,159 @@
+(** This modules defines a functor whose image is a parser for terms with
+    applications, binary and unary operators. These terms are specified in the
+    argument of the functor.
+
+    The algorithm implemented is an extension of the Pratt parser. The
+    Shunting Yard algorithm could also be used.
+    @see <https://matklad.github.io/2020/04/13/simple-but-powerful-\
+pratt-parsing.html>
```
fblanqui, line 8: Thanks to remove the \ and put the URL on one line so that we can easily copy and paste it. make sanity will not complain in this case. ;-)

gabrielhdt: Anyway that file will eventually go away since it's part of the pratter library (and the line is not cut in the library)

src/core/sign.ml
gabrielhdt: Well it's really similar. I don't have any real preference.

gabrielhdt: And I think neither solution is better than the other, performance wise.

src/core/sign.ml
gabrielhdt: We may need the string for printing purposes, don't we? Otherwise we won't remember, for instance, which is the associated infix representation of `plus` when we have `set infix "+" := plus`.

gabrielhdt: And do you think we should propagate those modifications on the `pp_hint` of `Sig_state`? I must admit that all the `pp_hint` functions are not really clear to me.

fblanqui: First question: right but is it a good idea to keep the set infix/prefix commands as they currently are? Until now, these commands had two functions: 1) declare a non-regular identifier like set declared, 2) map this non-regular identifier to a (qualified) regular one. The function 1) is useless now. So, I think that we should also change the set infix/prefix commands. Instead of having: set infix/prefix ... string_lit := qident, we should have: set infix/prefix ... qident, the meaning of which would simply be add a new notation for qident. What do you think?

Second question: yes.

fblanqui:
Don't forget the user manual as well.

src/core/sign.ml
gabrielhdt: Ah yes. Could we leave this as a later pull request, provided that we create an issue for the record?

src/core/sig_state.ml
```diff
@@ -160,14 +157,15 @@ let update_pp_hints_from_symbols :
...
+          | Sign.Prefix unop ->
+              if s == sym then raise (Hint (Prefix unop))
+          | _ -> ()
         in
-        StrMap.iter fn_binop !(sign.sign_binops);
-        let fn_unop _ (s,unop) =
-          if s == sym then raise (Hint (Prefix unop))
-        in
-        StrMap.iter fn_unop !(sign.sign_unops);
+        SymMap.iter fn !(sign.sign_syntax);
```
fblanqui, line 168: It seems to me that we can simplify this function a lot now that we have sign_syntax and pp_hints are attached to symbols instead of strings. In fact, it can be reduced (and thus unfolded) to:
```
SymMap.fold SymMap.add sign.sign_syntax pp_hints
```
If you take sign_syntax:pp_hint SymMap.t.

src/core/sig_state.ml
```diff
@@ -109,32 +106,32 @@ let add_symbol : sig_state -> sig_symbol -> sig_state * sym =
...
+  {ss with in_scope; pp_hints}
 
 (** [add_builtin ss n s] generates a new signature state from [ss] by mapping
    the builtin [n] to [s]. *)
 let add_builtin : sig_state -> string -> sym -> sig_state = fun ss name sym ->
   Sign.add_builtin ss.signature name sym;
   let builtins = StrMap.add name sym ss.builtins in
   let add_pp_hint hint =
-    SymMap.add sym hint (remove_pp_hint ss.builtins name ss.pp_hints)
+    SymMap.add sym hint (remove_pp_hint sym ss.pp_hints)
```
fblanqui, line 134: Can be simplified to `SymMap.update sym (function _ -> Some hint) ss.pp_hints`, and then remove_pp_hint can be removed.

src/core/sign.ml
```diff
@@ -15,17 +15,22 @@ type inductive =
...
 type t =
   { sign_symbols  : (sym * Pos.popt) StrMap.t ref
   ; sign_path     : Path.t
   ; sign_deps     : (string * rule) list PathMap.t ref
   ; sign_builtins : sym StrMap.t ref
-  ; sign_unops    : (sym * unop ) StrMap.t ref
-  ; sign_binops   : (sym * binop) StrMap.t ref
-  ; sign_idents   : StrSet.t ref
-  ; sign_quants   : SymSet.t ref
+  ; sign_syntax   : notation SymMap.t ref
```
fblanqui, line 32: I propose to use a pp_hint SymMap.t instead of a notation SymMap.t. It will ease interaction with sig_state. Then we can move the definition of pp_hint in sign.ml.

src/core/sig_state.ml
```diff
@@ -109,32 +106,32 @@ let add_symbol : sig_state -> sig_symbol -> sig_state * sym =
...
-  Sign.add_unop ss.signature name x;
-  let unops = StrMap.add name sym ss.unops in
-  let pp_hints = remove_pp_hint ss.unops name ss.pp_hints in
+    unary operator [x] with name [n]. This name is added to the scope. *)
+let add_unop : sig_state -> strloc -> (sym * unop) -> sig_state =
+  fun ss name (sym, unop) ->
+  Sign.add_unop ss.signature sym unop;
+  let in_scope = StrMap.add name.elt (sym, name.pos) ss.in_scope in
+  let pp_hints = remove_pp_hint sym ss.pp_hints in
   let pp_hints = SymMap.add sym (Prefix unop) pp_hints in
```
fblanqui, line 115: remove/add can be simplified to `SymMap.update sym (function _ -> Some (Prefix unop)) ss.pp_hints`.

src/core/sig_state.ml
```diff
@@ -109,32 +106,32 @@ let add_symbol : sig_state -> sig_symbol -> sig_state * sym =
...
-  Sign.add_binop ss.signature name x;
-  let binops = StrMap.add name sym ss.binops in
-  let pp_hints = remove_pp_hint ss.binops name ss.pp_hints in
+    binary operator [x] with name [n]. This name is added to scope. *)
+let add_binop : sig_state -> strloc -> (sym * binop) -> sig_state =
+  fun ss name (sym, binop) ->
+  Sign.add_binop ss.signature sym binop;
+  let in_scope = StrMap.add name.elt (sym, name.pos) ss.in_scope in
+  let pp_hints = remove_pp_hint sym ss.pp_hints in
   let pp_hints = SymMap.add sym (Infix binop) pp_hints in
```
fblanqui, line 125: remove/add can be simplified to `SymMap.update sym (function _ -> Some (Infix binop)) ss.pp_hints`.

src/core/sign.ml
fblanqui: Ok, I can take care of these two items in a next PR. Just, could you please remove the type notation, move the type pp_hint to sign.ml, and take sign_syntax:pp_hint SymMap.t ref?

src/core/sig_state.ml
gabrielhdt: Isn't `Map.update e (fun _ -> Some v) m` the same as `Map.add e v m`?

src/core/sign.ml
gabrielhdt: Would you agree with renaming `pp_hint` into `notation`? I feel it as less technical, and more communicative in some way.

src/core/sig_state.ml
fblanqui: Good question. I thought it was different in term of memory consumption but http://caml.inria.fr/pub/docs/manual-ocaml/libref/Map.S.html is not very clear about that. What about https://github.com/ocaml/ocaml/blob/trunk/stdlib/map.ml ?

fblanqui: You seem to be right. So we could simplify all these calls even further! :-)

src/core/sig_state.ml
```diff
@@ -104,37 +59,34 @@ let add_symbol : sig_state -> sig_symbol -> sig_state * sym =
     | None -> ()
   end;
   let in_scope = StrMap.add x.elt (s, x.pos) ss.in_scope in
-  let pp_hints = remove_pp_hint_eq ss.in_scope x.elt Unqual ss.pp_hints in
-  let pp_hints = SymMap.add s Unqual pp_hints in
+  let pp_hints = SymMap.add s Unqual ss.pp_hints in
```
fblanqui, line 62: Do we really need this as Not_found => Unqual?

src/core/sign.ml
```diff
@@ -236,8 +238,17 @@ let add_symbol : t -> expo -> prop -> match_strat -> strloc -> term ->
     ; sym_expo ; sym_tree = ref Tree_types.empty_dtree
     ; sym_mstrat = ref sym_mstrat }
   in
+  sign.sign_syntax := SymMap.add sym Unqual !(sign.sign_syntax);
```
fblanqui, line 241: Idem. Really needed if Not_found => Unqual?

src/core/sign.ml
```diff
@@ -335,25 +344,27 @@ let add_rule : t -> sym -> rule -> unit = fun sign sym r ->
...
-  sign.sign_builtins := StrMap.add s sym !(sign.sign_builtins)
-
-(** [add_unop sign unop sym] binds the unary operator [op] to [sym] in [sign].
-    If [unop] was previously bound, the previous binding is discarded. *)
-let add_unop : t -> string -> (sym * unop) -> unit = fun sign s sym ->
-  sign.sign_unops := StrMap.add s sym !(sign.sign_unops)
+  sign.sign_builtins := StrMap.add s sym !(sign.sign_builtins);
+  if s = "0" then sign.sign_syntax := SymMap.add sym Zero !(sign.sign_syntax);
+  if s = "+1" then
+    sign.sign_syntax := SymMap.add sym Succ !(sign.sign_syntax)
```
fblanqui, line 350: Thanks to use a match s instead.

src/core/sign.ml
```diff
@@ -236,8 +238,17 @@ let add_symbol : t -> expo -> prop -> match_strat -> strloc -> term ->
     ; sym_expo ; sym_tree = ref Tree_types.empty_dtree
     ; sym_mstrat = ref sym_mstrat }
   in
+  sign.sign_syntax := SymMap.add sym Unqual !(sign.sign_syntax);
   sign.sign_symbols := StrMap.add s.elt (sym, s.pos) !(sign.sign_symbols); sym
 
+(** [strip_private sign] removes private symbols from signature [sign]. *)
+let strip_private : t -> unit = fun sign ->
```
fblanqui, line 245: I cannot see where it is used.

src/core/sig_state.ml
gabrielhdt: Ah indeed, I did not see it. I'll add a note somewhere.

gabrielhdt: Or should we have `notation_of` return a `notation option` with `None` being the current `Unqual`?

src/core/sign.ml
gabrielhdt: In `Compile.compile`. I thought better to move these operations in `Sign` because they access the internals of the `sign` datatype.

src/pure/pure.ml
```diff
@@ -31,8 +31,13 @@ let parse_text : state -> string -> string -> Command.t list * state =
...
     let ast =
-      if old_syntax then Legacy_parser.parse_string fname s
-      else Parser.parse_string fname s
+      let strm =
+        if old_syntax then Parser.Dk.parse_string fname s
+        else Parser.parse_string fname s
+      in
+      let cmds = Stdlib.ref [] in
+      Stream.iter (fun c -> Stdlib.(cmds := c :: !cmds)) strm;
+      List.rev Stdlib.(!cmds)
```
fblanqui, line 40: I do not remember if you already answered me: can we avoid this List.rev by having a parser entry for a list of commands? If not, what is the problem?

fblanqui:
This PR is almost ready to be merged. The only missing things are: CHANGES.md, syntax.bnf and the user manual.

src/pure/pure.ml
gabrielhdt: A parser for a list of commands cannot be trivially made, it raises an end of stream conflict.

fblanqui:
Don't we need to modify things in vscode as well (because commands must be terminated by ";" now)?

src/core/lpParser.mly
```diff
@@ -0,0 +1,417 @@
...
+  | REFINE t=term { make_pos $loc (P_tac_refine(t)) }
+  | REFL { make_pos $loc P_tac_refl }
+  | SYMMETRY { make_pos $loc P_tac_sym }
+  | WHY3 s=STRINGLIT? { make_pos $loc (P_tac_why3(s)) }
+  | q=query { make_pos $loc (P_tac_query(q)) }
+  | FAIL { make_pos $loc P_tac_fail }
+
+// Terminated tactic. The semicolon isn't necessary, it is added for
+// uniformity.
+ttactic: t=tactic SEMICOLON? { t }
```
fblanqui, line 206: It will be better to make semi-colons mandatory here, for https://github.com/Deducteam/lambdapi/issues/515, but this may be done in a separate PR.

src/pure/pure.ml
fblanqui: Ok. Could you please add a comment about that in the code so that we remember why there is a List.rev?

src/core/lpParser.mly
gabrielhdt: I think we can easily parse based on the tactics keywords. But it may be more homogeneous to use semicolons to end tactics as well.

amelieled:
Are you sure it's a good idea that the parsers depend on Terms.ml ?
It's due to the addition of `modifiers`.

src/core/lpParser.mly
amelieled: In the future, I hope, tacticals should be added, so between the tactics, it will be possible to have different separators.
Don't forget that.

gabrielhdt:
> Are you sure it's a good idea that the parsers depend on Terms.ml ?
> It's due to the addition of `modifiers`.

Ah you're right. It seems to me that the best solution would be to transfer those modifiers to the world of `p_term`s, and then perhaps make `terms` depend on `p_terms`. This would allow to separate cleanly parsing from the rest.

CHANGES.md
```diff
@@ -1,3 +1,60 @@
...
+
+**Syntax modifications:**
+
+- The semi colon as termination token has been added. Hence, a symbol is
+  declared with symbol `foo: bar;`.
+
+- The syntax `λx y z: nat, ...` is not authorised anymore, but 
+  `λ(x y z: nat), ...` is, and `λ x : N, t` is as well.
+
+- Unification rules syntax has changed: the left hand side must be enclosed in
```
fblanqui, line 15: right-hand side

CHANGES.md
```diff
@@ -1,3 +1,60 @@
...
+  `begin ... end`, so
+  ```
+  set unif_rule $x + $y ≡ 0 ↪ $x ≡ 0; $y ≡ 0
+  ```
+  becomes
+  ```
+  set unif_rule $x + $y ≡ 0 ↪ begin $x ≡ 0; $y ≡ 0 end
+  ```
+- `set declared "?"` is not needed anymore,
```
fblanqui, line 24: has been removed since a large class of unicode characters is now allowed for identifiers

CHANGES.md
```diff
@@ -1,3 +1,60 @@
...
+- `assert` always take a turnstile (or vdash) to specify (even empty) context,
+  so the syntax is `assert ⊢ t: A`
+
+- the unary minus before a rewrite pattern becomes the keyword left
+
+**Code modifications:**
+
+- Parsing and handling are interleaved: the parser returns a stream of parsed
+  commands. Requesting an item of the stream parses one command in the file.
+
```
fblanqui, line 44: (except when using the LSP server)

docs/commands.rst
```diff
@@ -39,15 +45,15 @@ module. It can also be combined with the ``require`` command.
...
+   open booleans;
+   require open church.sums;
 
 ``symbol``
 ----------
 
 Allows to declare or define a symbol as follows:
 
-``symbol`` *modifiers* *identifier* *parameters* [``:`` *type*] [``≔`` *term*] [``begin`` *proof* ``end``]
+``symbol`` *modifiers* *identifier* *parameters* [``:`` *type*] [``≔`` *term*] {[``begin`` *proof* ``end``] | ``;``}
```
fblanqui, line 56: shouldn't it be ``` [``begin`` *proof* ``end``] ``;`` ```?

docs/commands.rst
```diff
@@ -297,55 +304,40 @@ Examples of patterns are available in ``tests/OK/patterns.lp``.
...
 
 ::
 
-   set quantifier ∀ // : Π {a}, (T a → Prop) → Prop
-   compute ∀ {a'} (λx:T a,p) // prints ∀x:T a,p
-   compute ∀ (λx:T a, p) // prints ∀x,p
-   type ∀x,p // quantifiers can be written as such
+   set quantifier ∀; // : Π {a}, (T a → Prop) → Prop
+   compute ∀ {a'} (λx:T a,p); // prints `∀x:T a,p
+   compute ∀ (λx:T a, p); // prints `∀x,p
```
fblanqui, line 338: space missing after \forall

fblanqui: btw, do you print a space after \lambda, \Pi and quantifiers in print.ml ?

docs/commands.rst
```diff
@@ -376,9 +368,9 @@ Examples:
 
 ::
 
-   set unif_rule Bool ≡ T $t ↪ $t ≡ bool
-   set unif_rule $x + $y ≡ 0 ↪ $x ≡ 0; $y ≡ 0
-   set unif_rule $a → $b ≡ T $c ↪ $a ≡ T $a'; $b ≡ T $b'; $c ≡ arrow $a' $b'
+   set unif_rule Bool ≡ T $t ↪ begin $t ≡ bool end
+   set unif_rule $x + $y ≡ 0 ↪ begin $x ≡ 0; $y ≡ 0 end
+   set unif_rule $a → $b ≡ T $c ↪ begin $a ≡ T $a'; $b ≡ T $b'; $c ≡ arrow $a' $b' end
```
fblanqui, line 373: semi-colons missing at the end

docs/devel.rst
```diff
@@ -12,9 +12,9 @@ List of dependencies:
 * Dune (>= 1.2.0)
```
fblanqui, line 12: All this is a duplication of README.md. why to keep it? I think that all the things related to dependencies and compilation should be written in README.md only.

docs/syntax.bnf
```diff
@@ -1,203 +1,128 @@
...
-  | "set" "infix" {"left" | "right"}? <float_lit> <string_lit> "≔" <qident>
-  | "set" "quantifier" <qident>
-  | "set" "builtin" \"<builtin>\" "≔" <qident>
-  | "set" "unif_rule" <unif_rule>
-
-// Entry point: list of commands
-<main> ::= <command>*
+<ident> ::= ID
+
+<path> ::= ID (DOT ID)*
```
fblanqui, line 3: This suggests that we can have spaces between ID and DOT. Is it really the case? If so, this should be fixed. Qualified identifiers should be dealt with at the lexer level.

docs/syntax.bnf
```diff
@@ -1,203 +1,128 @@
...
+        | L_PAREN <arg_ident>+ COLON <term> R_PAREN
+        | L_CU_BRACKET <arg_ident>+ [COLON <term>] R_CU_BRACKET
+
+<rw_patt_spec> ::= <term>
+                 | IN <term>
+                 | IN <ident> IN <term>
+                 | <term> IN <term> [IN <term>]
+                 | <term> AS <ident> IN <term>
+
+<rw_patt> ::= L_SQ_BRACKET <rw_patt_spec> R_SQ_BRACKET
```
fblanqui, line 26: It would be nicer to unfold rw_patt and rename rw_patt_spec into rw_patt.

docs/syntax.bnf
```diff
@@ -1,203 +1,128 @@
...
+
+<modifier> ::= CONSTANT
+             | INJECTIVE
+             | PROTECTED
+             | PRIVATE
+             | SEQUENTIAL
+             | OPAQUE
+
+<float_of_int> ::= FLOAT
+                 | INT
```
fblanqui, line 49: Don't you mean float_or_int ("or" instead of "of")?

docs/syntax.bnf
```diff
@@ -1,203 +1,128 @@
...
+           | SIMPL
+           | REWRITE [ASSOC] [<rw_patt>] <term>
+           | REFINE <term>
+           | REFL
+           | SYMMETRY
+           | WHY3 [STRINGLIT]
+           | <query>
+           | FAIL
+
+<ttactic> ::= <tactic> [SEMICOLON]
```
fblanqui, line 39: It would be nicer to unfold this entry.

docs/syntax.bnf
```diff
@@ -1,203 +1,128 @@
...
+         | PI <binder>
+         | LAMBDA <binder>
+         | LET <ident> <arg>* [COLON <term>] ASSIGN <term> IN <term>
+
+<binder> ::= <arg>+ COMMA <term>
+           | <arg_ident> COLON <term> COMMA <term>
+
+<rule> ::= <term> REWRITE <term>
+
+<unification> ::= <term> EQUIV <term>
```
fblanqui, line 123: Could you please rename unification into equation?

docs/syntax.bnf
```diff
@@ -1,203 +1,128 @@
...
+          | <qident>
+          | WILD
+          | TYPE
+          | QUESTION_MARK <ident> [<env>]
+          | <patt> [<env>]
+          | L_PAREN <term> R_PAREN
+          | L_CU_BRACKET <term> R_CU_BRACKET
+          | INT
+
+<aterm> ::= <sterm>+
```
fblanqui, line 109: It seems to me that this should be the opposite: what you call sterm's are atomic terms (aterm's), and what you call aterm's are s-expressions. See https://en.wikipedia.org/wiki/S-expression.

src/core/print.ml
```diff
@@ -186,7 +186,7 @@ and pp_term : term pp = fun oc t ->
           match unfold b with
           | Abst(a,b) ->
               let (x,p) = Bindlib.unbind b in
-              out oc "%a%a" pp_symbol s pp_var x;
+              out oc "`%a%a" pp_symbol s pp_var x;
```
fblanqui, line 189: space missing after symbol

fblanqui: idem in pp_head in the Abst and Prod cases

docs/syntax.bnf
gabrielhdt: In fact no, because the menhir rule takes a float or an int and returns a float, so it's either a float of an int or the identity.

docs/syntax.bnf
gabrielhdt: Ah true, true.

docs/syntax.bnf
fblanqui: Ok but from a user point of view you can write a float or an int, not just an int as suggested by the name float_of_int!

docs/syntax.bnf
fblanqui: Idem for pattern variables $ID and metavariables ?ID.

firewall2142:
> Don't we need to modify things in vscode as well (because commands must be terminated by ";" now)? @firewall2142 perhaps you have some suggestions about that?

Nothing needs to be changed in vscode because currently it is configured only to colour keywords.

fblanqui:
@gabrielhdt Following https://github.com/Deducteam/lambdapi/pull/532/files you probably need to modify editors/vscode/lp.configuration.json if you add multi-line comments /* ... */.

docs/syntax.bnf
gabrielhdt: Ah I was reading simple terms and applied terms, your interpretation makes more sense

docs/syntax.bnf
gabrielhdt: That's true, but I wonder whether this automated translation is what we want. The old BNF with tokens unfolded to their representation was better, even though we have to maintain it, I think.

docs/commands.rst
gabrielhdt: We don't need semicolons after an `end`

docs/commands.rst
gabrielhdt: Aaah well seen, I'd have forgotten that.

docs/commands.rst
gabrielhdt: Here again we have an `end`, so there is no need for a semicolon

docs/devel.rst
gabrielhdt: I agree

firewall2142:
Will the tactics be separated by `;` too?

gabrielhdt:
> Will the tactics be separated by `;` too?

Should they? For the moment it's optional, but we can make semicolons mandatory.

docs/commands.rst
fblanqui: But it would be better for uniformity.

fblanqui:
Would it better for interfaces to detect the end of a command or tactic? See https://github.com/Deducteam/lambdapi/issues/515

firewall2142:
> Should they? For the moment it's optional, but we can make semicolons mandatory.

For the interactive mode, if we plan to implement tactic-by-tactic navigation then without semicolon it seems almost impossible to tell the end of tactics for the editors.

It will also help with indentation problem we have in emacs. Maybe we can fix that by using a simpler indentation strategy in SMIE, because currently it seems to create more problems with indentation than it helps. But I don't know SMIE very well right now, so I am not very sure about that.

If it is possible for the lambdapi to know as soon as a tactic ends, we could use formatting from the LSP server (See [textDocument/onTypeFormatting](https://microsoft.github.io/language-server-protocol/specification#textDocument_onTypeFormatting)).

CHANGES.md
```diff
@@ -1,3 +1,61 @@
+### Unreleased
+
+#### Parser (2021-01-??)
+
+Replace Earley by Menhir, Pratter and Sedlex
+
+**Syntax modifications:**
+
+- The semi colon as termination token has been added. Hence, a symbol is
+  declared with symbol `foo: bar;`.
```
fblanqui, line 10: Perhaps it would be better to say that, now, a semi-colon must terminate every command and every tactic as well.

CHANGES.md
```diff
@@ -1,3 +1,58 @@
...
+  N, t` with a single variable.
+
+- In unification rules, the right hand-side must now be enclosed in
+  `begin ... end`, so
+  ```
+  set unif_rule $x + $y ≡ 0 ↪ $x ≡ 0; $y ≡ 0
+  ```
+  becomes
+  ```
+  set unif_rule $x + $y ≡ 0 ↪ begin $x ≡ 0; $y ≡ 0 end;
```
fblanqui, line 22: Finally, I realize that using begin/end for unification rules interferes with the navigation commands in VSCode and Emacs where begin is used to jump to the next/previous proof. So, we should use something else. What did you propose in the first place? Curly brackets? Square brackets?

docs/syntax.bnf
fblanqui: We could use a sed command on the generated file?

fblanqui:
@gabrielhdt I fixed some mistakes in the lexer, the grammar, the parser, the emacs mode, and the vscode mode. I sorted tokens everywhere to avoid mistakes. Thanks to have a look.

fblanqui:
One thing that remains to do is: regenerate syntax.bnf. Did you add a make entry to do that? Do you want to use sed on the generated file to replace tokens by their syntactic representation?

gabrielhdt:
> @gabrielhdt I fixed some mistakes in the lexer, the grammar, the parser, the emacs mode, and the vscode mode. I sorted tokens everywhere to avoid mistakes. Thanks to have a look.

Looks good to me.

> One thing that remains to do is: regenerate syntax.bnf. Did you add a make entry to do that? Do you want to use sed on the generated file to replace tokens by their syntactic representation?

Ah yes we may do that

src/core/lpLexer.ml
```diff
@@ -30,10 +30,7 @@ type token =
   | INDUCTIVE
   | INFIX
   | INJECTIVE
-  (*| LEFT replaced by ASSOC(Pratter.Left) *)
```
fblanqui: @gabrielhdt Thanks to put again these comments both in the parser and the lexer. It helps check that no keyword is missing or added.

fblanqui:
@gabrielhdt I don't think that we should remove syntax.bnf. And thanks to add a target for syntax_dedukti.bnf too.

fblanqui:
@firewall2142 Do you know why comments need to be defined in two different files: lp.configuration.json and syntaxes/p.tmLanguage.json ?

firewall2142:
> @firewall2142 Do you know why comments need to be defined in two different files: lp.configuration.json and syntaxes/p.tmLanguage.json ?

The latter file is for syntax highlighting and the former changes behaviour of shortcuts and other actions

src/core/lpLexer.ml
fblanqui: No need finally.

fblanqui:
Every thing seems fine now. I will merge soon unless you have other modifications to do @gabrielhdt .

gabrielhdt:
> Every thing seems fine now. I will merge soon unless you have other modifications to do @gabrielhdt .

All right, just could you put the content of the Changelog into the commit message? Rather than having the list of commits.

fblanqui:
A BIG thank to @gabrielhdt for his work on the new parser!

firewall2142:
@gabrielhdt `p_command.pos` seems to have the line numbers always 0. Can you please fix this? Thanks.

fblanqui:
Hello @firewall2142 . Could you please open an issue?

PR #517, opened 2021-01-07
fix problem with pdata_stmt_pos

fblanqui:
Problem introduced by 95a6427a8dcc94e24bd6ab0b598cc6abdd057900 / PR https://github.com/Deducteam/lambdapi/pull/505.

### DIFF
--- a/src/core/handle.ml
+++ b/src/core/handle.ml
@@ -174,7 +174,7 @@ let handle_inductive_symbol :
     do not work on this structure directly,  although they act on the contents
     of its [pdata_p_state] field. *)
 type proof_data =
-  { pdata_stmt_pos : Pos.popt (** Position of the proof's statement. *)
+  { pdata_stmt_pos : Pos.popt (** Position of the declared symbol. *)
   ; pdata_p_state  : proof_state (** Proof state. *)
   ; pdata_tactics  : p_tactic list (** Tactics. *)
   ; pdata_finalize : sig_state -> proof_state -> sig_state (** Finalizer. *)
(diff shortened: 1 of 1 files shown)


fblanqui:
What is weird is that, while pdata_stmt_pos was said to be the "position of the proof's statement", it in fact must be set to the position of the declared symbol. Otherwise the Emacs and VSCode interfaces get broken and do not print goals anymore! Because Lp_lsp.get_node_at_pos does not provide the right goal id in this case. It's kind of mysterious for me as I don't know this part of the code. But it looks like the LSP server code makes not explicit assumptions on positions. It would be nice to clarify this. @ejgallego do you have any idea about that?

fblanqui:
@firewall2142 Could you please check that Emacs and VScode work well with this branch, and that goals are well displayed?

firewall2142:
Yes, it works now. Earlier the initial goal's `popt` shadowed the whole `n.goals` list in line 194. See end of lsp logs [https://pastebin.com/raw/DbSVrnrK](https://pastebin.com/raw/DbSVrnrK) (between BEGIN GOALS and END GOALS) 

```
lp_lsp.ml
-----------
188 
189 let rec get_goals ~doc ~line ~pos =
190   let node = get_node_at_pos doc line pos in
191   let goals = match node with
192     | None -> None
193     | Some n ->
194         List.find_opt (fun (_, loc) -> in_range ?loc (line,pos)) n.goals in
195   match goals with
196     | None -> begin match node with
197               | None   -> None
198               | Some _ -> get_goals ~doc ~line:(line-1) ~pos:0 end
199     | Some (v,_) -> Some v
```

fblanqui:
Thank you @firewall2142 .

PR #518, opened 2021-01-08
Logs UI

firewall2142:
New:
- Query results can be viewed on hover
- Logs are now displayed in both vscode and emacs
- Emacs: window layout can be easily changed
- ~~Emacs: Debug menu on the menu-bar~~
- ~~Vscode: debug menu in Run (in the activity bar)~~
- ~~LSP Server : added `proof/debugFlags` notification~~


src/core/queries.ml
```diff
@@ -50,45 +53,55 @@ let handle_query : Sig_state.t -> proof_state option -> p_query -> unit =
...
   | P_query_infer(pt, cfg) ->
       let env = Proof.focus_env ps in
       let ctxt = Env.to_ctxt env in
       let scope = Scope.scope_term Public ss env in
       let t = scope pt in
       let a = Infer.infer Unif.solve_noexn pt.pos ctxt t in
-      out 1 "(infr) %a : %a\n" pp_term t pp_term (Eval.eval cfg ctxt a)
+      out 1 "(infr) %a : %a\n" pp_term t pp_term (Eval.eval cfg ctxt a);
+      Some (Format.asprintf "%a : %a" pp_term t pp_term
+              (Eval.eval cfg ctxt a))
```
fblanqui, line 84: It would be better to factorize the result as follows: `let res = Format.asprintf ... in out 1 "..."; Some res`. Idem in the other queries.

fblanqui: No sorry. I mean the computations only: `let a = Eval.eval cfg ctxt a in ...`

src/core/queries.ml
```diff
@@ -98,14 +111,22 @@ let handle_query : Sig_state.t -> proof_state option -> p_query -> unit =
...
       if h <> Unqual then out 1 " [%a]" pp_hint h;
       (match !(sym.sym_def) with
-      | Some t -> out 1 " ≔ %a\n" pp_term t
+      | Some t ->
+          out 1 " ≔ %a\n" pp_term t;
+          Some (Format.asprintf " ≔ %a" pp_term t)
       | None ->
           out 1 "\n";
-          List.iter (fun r -> out 1 "%a\n" pp_rule (sym, r)) !(sym.sym_rules))
+          List.iter (fun r -> out 1 "%a\n" pp_rule (sym, r)) !(sym.sym_rules);
```
fblanqui, line 119: I realize that, to avoid a useless List.iter if nothing needs to be printed, it would be better to change this line as well as follows:
```
let rule oc r = Format.fprintf oc "%a\n" pp_rule (sym, r) in
out 1 "%a" (List.pp rule "") !(sym.sym_rules)
```

src/core/queries.ml
```diff
@@ -50,45 +53,55 @@ let handle_query : Sig_state.t -> proof_state option -> p_query -> unit =
...
   | P_query_prover_timeout(n) ->
-      Timed.(Why3_tactic.timeout := n)
+      Timed.(Why3_tactic.timeout := n);
+      None
   | P_query_print(None) ->
       (match ps with
        | None -> fatal q.pos "Not in a proof."
-       | Some ps -> out 1 "%a" Proof.pp_goals ps)
+       | Some ps -> out 1 "%a" Proof.pp_goals ps;
+                    Some (Format.asprintf "%a" Proof.pp_goals ps))
```
fblanqui, line 104: print with no argument is used for printing goals. So, here, I guess you can send None instead.

src/core/queries.ml
```diff
@@ -7,8 +7,11 @@ open Terms
 open Print
 open Proof
 
+(** Result of query displayed on hover in the editor*)
+type q_res = string option
```
fblanqui, line 11: Perhaps you can just call it result and use Queries.result outside.

src/lsp/lp_lsp.ml
```diff
@@ -198,11 +220,48 @@ let rec get_goals ~doc ~line ~pos =
               | Some _ -> get_goals ~doc ~line:(line-1) ~pos:0 end
     | Some (v,_) -> Some v
 
+let get_logs ~doc ~line ~pos =
+  let rec fold_left_while f cond acc l =
+    match l with
+    | x :: _ when not (cond x) -> acc
+    | x :: xs -> fold_left_while f cond (f acc x) xs
+    | [] -> acc
```
fblanqui, line 228: This should probably be moved to lplib/list.

src/lsp/lp_lsp.ml
```diff
@@ -391,6 +450,10 @@ let dispatch_message ofmt dict =
     protect_dispatch "didClose"
       (do_close ofmt) params
 
+  | "proof/debugFlags" ->
+    protect_dispatch "debugFlags"
+      (do_debugFlags ofmt) params
+
```
fblanqui, line 456: Could you please explain what is the goal of this message? Is it in the LSP standard?

fblanqui:
Hi @ejgallego . Your review on this PR would be very much appreciated as it affects the LSP server.

src/lsp/lp_lsp.ml
```diff
@@ -422,8 +485,8 @@ let main std log_file =
   LIO.debug_fmt := F.formatter_of_out_channel debug_oc;
 
   (* XXX: Capture better / per sentence. *)
-  let lp_oc = open_out "log-lp.txt" in
-  let lp_fmt = F.formatter_of_out_channel lp_oc in
+  (* let lp_oc = open_out "log-lp.txt" in *)
```
fblanqui, line 488: @ejgallego What was log-lp.txt for as there is also /tmp/lambdapi_lsp_log.txt ?

src/core/queries.ml
firewall2142: Sorry, I don't understand what you meant. What is `List.pp`?

src/lsp/lp_lsp.ml
firewall2142: This is used to change the debug flags. It is not in the LSP standard.

src/core/queries.ml
amelieled: See  `lambdapi/src/lplib/list.ml` to find the function `pp`. ;)

firewall2142: Oh. Thanks!

src/core/compile.ml
```diff
@@ -60,13 +60,17 @@ let rec compile : bool -> Path.t -> Sign.t = fun force path ->
...
         match p with
         | None       -> ss
         | Some(data) ->
             let (st,ts) = (data.pdata_p_state, data.pdata_tactics) in
             let e = data.pdata_expo in
-            let st = List.fold_left (Tactics.handle_tactic ss e) st ts in
+            let st =
+              (List.fold_left
+                (fun st tac -> fst (Tactics.handle_tactic ss e st tac))
+                st ts)
```
fblanqui, line 72: The surrounding parentheses are useless ;-)

src/lsp/lp_lsp.ml
```diff
@@ -55,11 +55,33 @@ let doc_table : (string, _) Hashtbl.t = Hashtbl.create 39
...
+  let flags = string_field "flags" params in
+  LIO.log_error "debugFlags" ("flags="^flags);
+  let open Console in
+  let logger_keys = List.map (fun {logger_key; _} -> logger_key) !loggers in
+  (* let verbose = int_field "verbose" params in *)
+  Timed.Time.restore @@ Pure.get_initial_time ();
+  Console.set_debug false @@ String.of_list logger_keys;
+  Console.set_default_debug flags;
+  (* Console.set_default_verbose verbose; *)
+  Pure.set_initial_time ();
```
fblanqui, line 75: I am worried about this part of the code. I believe that set_initial_time should be called only once. Resetting time erase the environment, that is, everything that was read before! Why do you need to reset time here?

firewall2142: I change the default values of debug flags, set initial time (to set the default values of debug flags) then check the opened files again to generate the log messages according to the newly set flags. 
Also, I just realized that I forgot to implement the verbose part :/  will do it today

fblanqui: But we may not want to get all the debug messages from the beginning. Some times we want those messages only on a specific command. I think that we need to think more about how we want to handle debug flags first.

docs/ui/emacs.rst
```diff
@@ -196,16 +196,20 @@ display goals while doing a proof. To display the result of ``compute``,
...
+(Do ``M-x customize-group lambdapi``).
 The layout can be refreshed with ``C-c C-r``.
 
 
+Debug
+'''''
+
+The debug flags can be set from the LambdaPi-Debug menu
+
+
```
fblanqui, line 212: Should probably be removed now.

docs/ui/vscode.rst
```diff
@@ -147,4 +147,13 @@ Hover
 Hovering symbols provides their type. This feature is still
 experimental.
 
+Logs
+----
+
+Logs are displayed in a terminal window. The debug flags can be set
+from the debug flags menu. The menu can be found by selecting Run in
+the Activity Bar (in the left by default). If you want, you can drag
+the menu to another place.
```
fblanqui, line 156: Things related to the debug flags menu should probably be removed now?

editors/emacs/lambdapi-mode.el
```diff
@@ -195,6 +175,72 @@ for the goals buffer using the lambdapi-goals-window-* variables."
...
+  :group 'lambdapi)
+
+(defcustom lambdapi-window-Y-ratio 0.8
+  "Ratio of width taken in vertical split during window layout"
+  :type '(float)
+  :group 'lambdapi)
+
+(defcustom lambdapi-window-layout '(v lambdapi-window-Y-ratio
+                                      lambdapi--temp-buffer-name
+                                      (h 0.5 "*Goals*" "*lp-logs*"))
```
fblanqui, line 194: Could you please set the default layout as follows?
________
|               |
|               | text 70%
|_______ |
|               | goals 25%
|_______ |
|_______ | queries & logs 5%

firewall2142: done

src/core/queries.ml
```diff
@@ -98,14 +112,23 @@ let handle_query : Sig_state.t -> proof_state option -> p_query -> unit =
...
+          Some (Format.asprintf " ≔ %a" pp_term t)
       | None ->
           out 1 "\n";
-          List.iter (fun r -> out 1 "%a\n" pp_rule (sym, r)) !(sym.sym_rules))
+          let rule oc r = Format.fprintf oc "%a\n" pp_rule (sym, r) in
+          out 1 "%a" (List.pp rule "") !(sym.sym_rules);
+          Some (
+            List.fold_left (^) ""
+              (List.map (fun r -> Format.asprintf "%a\n" pp_rule (sym, r))
+                !(sym.sym_rules))))
```
fblanqui, line 125: May be simplified as `Some (Format.asprintf "%a" (List.pp rule "") !(sym.sym_rules))`.

src/core/queries.ml
```diff
@@ -50,45 +54,55 @@ let handle_query : Sig_state.t -> proof_state option -> p_query -> unit =
...
+      None
   | P_query_prover_timeout(n) ->
-      Timed.(Why3_tactic.timeout := n)
+      Timed.(Why3_tactic.timeout := n);
+      None
   | P_query_print(None) ->
       (match ps with
-       | None -> fatal q.pos "Not in a proof."
-       | Some ps -> out 1 "%a" Proof.pp_goals ps)
+       | None -> None
```
fblanqui, line 103: Sorry but, finally, I think that we should keep `| None -> fatal q.pos "Not in a proof."` as this is incorrect to use `print` with no argument outside a proof.

src/lsp/lp_lsp.ml
```diff
@@ -432,7 +463,7 @@ let main std log_file =
     let com = LIO.read_request stdin in
     LIO.log_object "read" com;
     process_input oc com;
-    F.pp_print_flush lp_fmt (); flush lp_oc;
```
fblanqui, line 435: Thanks to keep `flush lp_oc` in a comment: `(*flush lp_oc*)`. Idem a few lines below.

src/lsp/lp_lsp.ml
```diff
@@ -442,7 +474,6 @@ let main std log_file =
     LIO.log_error "[BT]" bt;
     F.pp_print_flush !LIO.debug_fmt ();
     flush_all ();
-    close_out debug_oc;
-    close_out lp_oc
```
fblanqui, line 446: Thanks to keep `close_out lp_oc` in comment. Then, it should be possible to merge, even if a few things are not completely clear to me on the LSP server side.

firewall2142: done

firewall2142: Mainly, I replaced the channel associated with `lp-log.txt` with a buffer (`Lp_doc.lp_logger`). Then when the commands are evaluated they write logs to the buffer. I take the the contents of the buffer and store them along with the position of the command in `Lp_doc.t.logs`.

PR #521, opened 2021-01-11
Use Github action for CI

### LINKED ISSUE #478: Migrate continuous integration
Some problem arise from the use of Travis (see #473). It seems we should change the continuous integration system. The alternatives seem to be either github actions, or an inria based ci.

vblot:
This is a CI script using github actions. The tests are performed against several ocaml versions and the `.opam` directory is cached to speed up the process. Should fix #478 .


fblanqui:
Thanks @vblot ! Don't we need to remove the file .travis.yml?

vblot:
I don't know, what is/was its use?

fblanqui:
This file was used for CI on Travis.

fblanqui:
Is it possible to automatically have links from the PR pages to the results of CI?
Currently, these links point to Travis. For instance, on https://github.com/Deducteam/lambdapi/pull/521 (the current PR), if you click on "Show all checks", you will see a link to Travis.
In addition, if you look at the commits on a branch, for instance on https://github.com/Deducteam/lambdapi/pulls, https://github.com/Deducteam/lambdapi/commits/master or https://github.com/Deducteam/lambdapi/pull/518/commits, there is a green tick if CI worked or a red cross if CI failed.

vblot:
Has the corresponding part on travis been deleted as well?

fblanqui:
Currently, there is a wekhook to Travis: https://github.com/Deducteam/lambdapi/settings/hooks
And there is currently nothing on https://github.com/Deducteam/lambdapi/actions/new

vblot:
Ok so I suppose we can remove the webhook to travis, right? For your other question: I think these links will be generated as soon as the PR is merged. Indeed, they appear on my clone: https://github.com/vblot/lambdapi.

fblanqui:
Hopefully. But wait that we merge #516 et #518 before removing travis webhook.

vblot:
Ok I suppose the new CI checks will just be added to the existing travis check as soon as this is merged.

vblot:
But if we include deletion of .travis.yml in this PR, then we would need to wait until we are ready to get rid of travis before merging.
So we have 2 options:
- we do not delete .travis.yml in this PR, so we can merge it now,
- we delete .travis.yml in this PR, but we have to wait for #516 and #518 to be merged before merging the current PR.
What do you want to do?

fblanqui:
If there are no other remarks, I propose to merge this PR as it is.

amelieled:
As @vblot said, we can delete travis file, right?

fblanqui:
Right. @vblot do you mind to update your PR?

vblot:
Since #516 and #518 are merged, I suppose so, yes.

vblot:
Done.

fblanqui:
Thanks @vblot !

vblot:
Welcome, let's see how it goes in the long term.

fblanqui:
Results are available on https://github.com/Deducteam/lambdapi/actions . We should probably remove the webhook to Travis now, right?

vblot:
Yes, indeed.

fblanqui:
Problem: there is no action generated when @gabrielhdt pushes on https://github.com/Deducteam/lambdapi/pull/535...

fblanqui:
@vblot The current workflow is: `on: [push, workflow_dispatch]`. Shouldn't we add something else here to check every new commit in PRs?

vblot:
I think you're right, it should be `on: [push, workflow_dispatch, pull_request]`, see https://docs.github.com/en/actions/reference/events-that-trigger-workflows.

fblanqui:
Perhaps we should restrict to a few types only?
```
on:
  pull_request:
    types: [opened, synchronize, edited, reopened]
```

vblot:
Yes that's probably more reasonable indeed.

fblanqui:
So, we should have something like:
```
on:
  pull_request:
    types: [opened, synchronize, edited, reopened]
  push:
  worklow_dispatch:
```
I am not sure about the syntax. Do you mind to make a PR to fix the workflow script?

vblot:
Actually the default is only `types: [opened, synchronize, reopened]`:
```
Note: By default, a workflow only runs when a pull_request's activity type is opened, synchronize, or reopened. To trigger workflows for more activity types, use the types keyword.
```
but as you suggest it's probably better to have it also on edits, I think.

vblot:
Here it is: #537.

fblanqui:
FYI I just deactivated (not deleted) the Travis webhook.

amelieled:
But Travis is always expected as you can see in PR #530

vblot:
I think this is because this particular CI was triggered before the removal of `.travis.yml`.

amelieled:
Frédéric just opened a new PR (#541 ) and there is the same thing.

vblot:
My guess is that it's because `.travis.yml` was present at the forking point of #541 (which is fdbd9bbfb97f8ef7d38d78254d515979c8d8308f). We could try opening a PR from a fork point that is beyond 8ee26bb7268a8092162ca4046febfaf4c64c2299 if we want to be sure.

amelieled:
Yes I do it (See #542 ) but same problem.

amelieled:
Maybe the answer is here: https://github.com/travis-ci/travis-ci/issues/8988
Good luck!

vblot:
Nice catch, that was it!

fblanqui:
Thanks @amelieled !

amelieled:
And thanks @vblot ! :)
Bye bye Travis ! :)

PR #524, opened 2021-01-14
why3 load driver API minor change after cfbdd9c2

Rehan-MALAK:
to keep in mind for the day this will be in stable why3


fblanqui:
Thanks @Rehan-MALAK !

fblanqui:
Included in #608. Thank @Rehan-MALAK !

PR #526, opened 2021-01-21
Fix goal printing in LSP server

fblanqui:
Because printing depends on the current state.

### DIFF
--- a/editors/vscode/src/client.ts
+++ b/editors/vscode/src/client.ts
@@ -409,7 +409,7 @@ function getGoalsEnvContent(goals : Goal[]){
         
         // check if this is a Type Goal
     	// extra space is not a typo, this is defined in lsp_base.ml
-        if (goals[i].typeofgoal == "Typ "){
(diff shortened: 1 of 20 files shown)


editors/vscode/src/client.ts
```diff
@@ -407,12 +407,28 @@ function getGoalsEnvContent(goals : Goal[]){
         codeHyps = `<div class="hypothesis">`;
         codeGoals = `<div class="pp_goals">`;
         
-        // check if this is a Type Goal
-    	// extra space is not a typo, this is defined in lsp_base.ml
-        if (goals[i].typeofgoal == "Typ "){
+        if (goals[i].typeofgoal == "Typ"){
```
fblanqui, line 410: Hi @firewall2142 . Is there a way to factorize these two identical loops below? (Both types of goals now have a hyps field.)

firewall2142: I made a PR for that : #527

PR #530, opened 2021-01-24
Deleting of a dependence in scope.ml

amelieled:
Until now, scope.ml depends on rewrite.ml, but it's artificial.
You can merge it if you are satisfied.


gabrielhdt (approved):
Seems good to me. Simplifying dependencies may ease further development, and we see that `terms.ml` already define many types, so this modification does not seem disruptive at all.

fblanqui:
Thanks @amelieled for your proposal. There is perhaps something even better to do: factorizing the types Syntax.p_rw_patt and rw_patt as follows:
```
type ('term,'binder) rw_patt =
  | Rw_Term           of 'term
  | Rw_InTerm         of 'term
  | Rw_InIdInTerm     of 'binder
  | Rw_IdInTerm       of 'binder
  | Rw_TermInIdInTerm of 'term * 'binder
  | Rw_TermAsIdInTerm of 'term * 'binder
```
Then, p_rw_patt = (p_term, p_ident*p_term) rw_patt
and rw_patt = (term, tbinder).

amelieled:
Yes, we can do that too, but where do you want to put the definition of `('term, 'binder) rw_patt`?

fblanqui:
in terms.ml

amelieled:
Yes it is but are you sure it's a good idea that Syntax.ml (or the parsers) depends on Terms.ml ?

fblanqui:
Why not? To improve that though, one thing we could perhaps do is put the basic types that are common to terms and syntax into a file, say common.ml.

amelieled:
The problem is: Terms.ml depends on Tree_types.ml. So the parsers depend on rewriting engine!
Why not about the file `common.ml`, except if there is only this type.

fblanqui:
@amelieled What is the status of this PR?

fblanqui:
This PR has already been included in another one, probably #535. The factorization of types is done in #640 .

amelieled:
Yes, but the factorization of types isn't a good idea, as I said before.

fblanqui:
Why? Note that rw_patt is now defined in Parsing.Syntax. Core does not depend on it.

amelieled:
It's normal because Parsing depends on Core...

fblanqui:
Right, the kernel does not depend on the concrete syntax. I guess that it is similar in Coq for instance.

amelieled:
Right, but here, the concrete syntax depends on the kernel.

fblanqui:
In your PR, you proposed to move rw_patt from rewrite to term so that scope doesn't depend on rewrite anymore. This has been done.
In a later comment, you were wondering why parsing depends on tree_type. This is because parsing depends on term (parsing generates terms), and term depends on tree_type. term depends on tree_type because, in LP, symbols are shared and contain their rewriting rules. Why should we change that?
Finally, you wrote that factorizing Term.rw_patt and Syntax.p_rw_patt_aux into a single polymorphic type Syntax.rw_patt "isn't a good idea" without explaining why, although the code changed since your previous comment. This factorization reduces the current dependencies since Term.rw_patt isn't used in core/ but in handle/ only. Why isn't this a good idea?

amelieled:
There is a confusion between type `p_term` and type `term`.
It's quite strange that the parsing depends on the kernel. The parsing syntax is based on the type `p_term`, and the kernel on the type `term`. The separation had been done in PR #535 , but has since been broken. If it is restored, I think the latter commit are problematic as well.

PR #532, opened 2021-01-26
correct vscode comment configuration

firewall2142:
`C-/` should comment the line/region correctly in vscode


fblanqui:
Thanks @firewall2142 !

PR #535, opened 2021-01-31
Segment source code

gabrielhdt:
This pull request suggest a segmentation of the source code into libraries to lower the number of interdependencies, and possibly ease the definition of interfaces.

For the moment, four libraries are added

- `Common` that defines miscellaneous modules used a bit everywhere such as `Pos` or `Console`
- `Parsing` that defines all that is used for parsing.
- `Handle` that uses the core of lambdapi to typecheck elements of the syntax, modules or commands.
- `Tool` that provides some other tools, such as XTC export or graphviz visualisation of rewriting trees.


fblanqui:
@gabrielhdt Thanks to update and add an entry into CHANGES.md summarizing the changes.

amelieled (dismissed):
Nice!
Note : In CHANGES.md, `Core` library divided into libraries
           divided into -> from ... extract 2 libraries

docs/structure.rst
```diff
@@ -27,14 +27,35 @@ Overview of directories and files
...
+  -  ``pretty.ml``: pretty print the abstract syntax (used to
+     convert old Dedukti files into Lambdapi files)
+
+  -  legacy parser:
+
+     -  ``legacy_lexer.ml``: lexer for the old Dedukti syntax
+     -  ``legacy_parser.ml``: interface of the parser for the old
+        Dedukti syntax
+     -  ``menhir_parser.ml``: parser using
+        `menhir <http://gallium.inria.fr/~fpottier/menhir/>`__
```
fblanqui, line 50: these filnames are not correct anymore

docs/structure.rst
```diff
@@ -27,14 +27,35 @@ Overview of directories and files
 
 -  ``libraries/``: libraries of Dedukti files (see GNUmakefile)
 
+- ``src/backbone/``: miscellaneous modules and libraries
```
fblanqui, line 30: the directory lplib is missing now (not fixed in previous PR)

fblanqui: This file has not been updated for quite some time and many files are missing, have been deleted or renamed.

Missing files: builtin, inductive, package, pratt, unif_rule, sig_state

Removed files: typing

Thanks to make sure that every existing file is listed and that every listed file exists.

fblanqui: Thanks to update CONTRIBUTING.md as well to indicate that this file should be kept up to date.

docs/structure.rst
```diff
@@ -1,127 +1,128 @@
...
+      *  ``Makefile``
 
--  ``libraries/``: libraries of Dedukti files (see GNUmakefile)
+*  ``libraries/``: libraries of Dedukti files (see GNUmakefile)
 
--  ``src/core/``: source code of Lambdapi
+* ``src/lplib/``: extensions of the Ocaml standard library for Lambdapi
 
-   -  utilities:
+  *  ``stubs.c``: C implementation for ``Extra.Filename.realpath``.
```
fblanqui, line 32: lplib/range* files are missing. Should they be in lplib or backbone?

gabrielhdt: Ah good question, I'm not sure.

fblanqui:
Perhaps we can take advantage of this PR to do a few renamings?
- remove plural forms in filenames to be more uniform: terms.ml -> term.ml, queries -> query (except tree_types.ml)
- rename a few files: files -> path, stubs.c -> realpath.c 
- integrate basics into term

gabrielhdt:
> Perhaps we can take advantage of this PR to do a few renamings?
> 
>     * remove plural forms in filenames to be more uniform: terms.ml -> term.ml, queries -> query (except tree_types.ml)
> 
>     * rename a few files: files -> path, stubs.c -> realpath.c
> 
>     * integrate basics into term

Ah right, why not.

amelieled:
rewrite -> rewrite_tactic ?
query -> query_handle ?
tactic -> tactic_handle ?
handle -> cmd_handle ?

But please, don't merge `basics` with `term`.
I think it's a good idea to split some at this point, like `syntax.ml` into `p_term.ml` and `p_basic.ml`.
Moreover, it could be nice to rename `pretty.ml` into  `p_print.ml`, and to underline the difference between `env.ml` and `ctxt.ml`, `ctxt.ml` and `p_ctxt.ml`, or `p_env.ml` (current `env.ml`) and `env.ml`.

fblanqui:
No, thanks.

gabrielhdt:
> rewrite -> rewrite_tactic ?
> query -> query_handle ?
> tactic -> tactic_handle ?
> handle -> cmd_handle ?
> 
> But please, don't merge `basics` with `term`.
> I think it's a good idea to split some files at this point, like `syntax.ml` into `p_term.ml` and `p_basic.ml`.
> Moreover, it could be nice to rename `pretty.ml` into `p_print.ml`, and to underline the difference between `env.ml` and `ctxt.ml`, `ctxt.ml` and `p_ctxt.ml`, or `p_env.ml` (current `env.ml`) and `env.ml`.

Your proposed names are rather long and irrelevant, but it's true that it's a bit confusing to have a rewriting engine and a `rewrite.ml` file. But we may be able to create another library which gathers all `handle` related things and `compile` and that depends on `Core`, so we may want to wait for this before operating successive renamings.

amelieled:
Yes you right @gabrielhdt it's better to do :
handle/
   tactic.ml
   query.ml
   cmd.ml (current handle.ml)
   compile.ml
   
 Maybe we can do the same thing about tactic :
 tactics/
      rewrite.ml
      why3.ml

gabrielhdt:
> Yes you right @gabrielhdt it's better to do :
> `handle/`
> 
>     * `tactic.ml`
> 
>     * `query.ml`
> 
>     * `cmd.ml` (current `handle.ml`)
> 
>     * `compile.ml`
> 
> 
> Maybe we can do the same thing about tactic :
> `tactics/`
> 
>     * `rewrite.ml`
> 
>     * `why3.ml`

In a first place, I suggest that we group everything related to handling into a sub library (so `Compile`, `Handle`, `Query`, `Rewrite` &c.), but I'd prefer to do this in another PR. So we may operate benign renamings here and then merge.

gabrielhdt:
> Perhaps we can take advantage of this PR to do a few renamings?
> 
>     * remove plural forms in filenames to be more uniform: terms.ml -> term.ml, queries -> query (except tree_types.ml)
> 
>     * rename a few files: files -> path, stubs.c -> realpath.c
> 
>     * integrate basics into term

I've replaced `Terms` and `Queries` and `subs.c` with `realpath.c`. There is already a submodule `Path` in `Files`, so we may not want to rename `Files` with `Path`. On the integration of `Basics` into `Term`, I'm not sure either that would help.

@fblanqui do you agree to merge this pull request?

fblanqui:
package.ml is missing in structure.rst

I don't like very much the name "backbone": it suggest much more than what it actually is. It would be better to find another name.

gabrielhdt:
> package.ml is missing in structure.rst
> 
> I don't like very much the name "backbone": it suggest much more than what it actually is. It would be better to find another name.

Yes, I do totally agree. `Support`, `Common`, `Util`, `Tool`? I'd go for `Util` or `Support`.

fblanqui:
common seems best

fblanqui:
The renaming tactics -> tactic is missing. And what about files -> module?

fblanqui:
I agree.

Le 01/02/2021 à 16:05, amelieled a écrit :
>
> Yes you right @gabrielhdt <https://github.com/gabrielhdt> it's better 
> to do :
> handle/
> tactic.ml
> query.ml
> cmd.ml (current handle.ml)
>
command.ml
>
> compile.ml
>
> Maybe we can do the same thing about tactic :
> tactics/
>
tactic/ (no plural)
>
> rewrite.ml
> why3.ml
>
>
why not also having:

- in tool/ : tree_graphviz.ml, sr.ml, xtc.ml, hrs.ml, pretty.ml ?

additional remarks:

- basics.ml should be renamed into libterm.ml

- pratt.ml should be in parsing/

- package.ml should be in cli/

gabrielhdt:
> I agree. Le 01/02/2021 à 16:05, amelieled a écrit :
> Yes you right @gabrielhdt <https://github.com/gabrielhdt> it's better to do : handle/ tactic.ml query.ml cmd.ml (current handle.ml)
> command.ml
> compile.ml Maybe we can do the same thing about tactic : tactics/
> tactic/ (no plural)
> rewrite.ml why3.ml
> why not also having: - in tool/ : tree_graphviz.ml, sr.ml, xtc.ml, hrs.ml, pretty.ml ? additional remarks: - basics.ml should be renamed into libterm.ml - pratt.ml should be in parsing/ - package.ml should be in cli/

These separations may be more delicate to make, because of dependencies, but it's a good idea. ` Pratt` cannot be in parsing because it depends on `Sig_state` and `Env` (for good reasons), and morally, its closer to scoping than parsing of files.

amelieled:
> - in tool/ : tree_graphviz.ml, sr.ml, xtc.ml, hrs.ml, pretty.ml ?

I agree about `xtc.ml` and `xtc.ml` (in `tool/competitor_format` folder ?)
About `tree_graphviz.ml` and `sr.ml` ~ rewriting engine ?
And for `pretty.ml`, it's for `p_term` so in `parsing/` ?

> - basics.ml should be renamed into libterm.ml

Yes, and we can extract `lib_p_term.ml` from syntax.ml ?

> - pratt.ml should be in parsing/

See Gabriel's message.

> - package.ml should be in cli/

If it's possible (dependencies)

gabrielhdt:
> > * in tool/ : tree_graphviz.ml, sr.ml, xtc.ml, hrs.ml, pretty.ml ?
> 
> I agree about `xtc.ml` and `xtc.ml` (in `tool/competitor_format` folder ?)
Well let's not have such a complex filesystem hierarchy...
> About `tree_graphviz.ml` and `sr.ml` ~ rewriting engine ?
Shouldn't `sr.ml` stay in `Core`? Isn't it part of the kernel?
> And for `pretty.ml`, it's for `p_term` so in `parsing/` ?
For the moment, `pretty.ml` is in parsing, I think it fits, but it fits also in `tool`, I have no strong opinion.
> 
> > * basics.ml should be renamed into libterm.ml
> 
> Yes, and we can extract `lib_p_term.ml` from syntax.ml ?
Well, let's keep that for later perhaps, I'd keep this PR as a "renaming everything" thing.
> 
> > * pratt.ml should be in parsing/
> 
> See Gabriel's message.
> 
> > * package.ml should be in cli/
> 
> If it's possible (dependencies)
My concern is that `cli` is really about processing the command line and taking the right action, `package.ml` does a bit more. In my opinion it should be with `Files` (or `Module` as we'll rename it). Does it make sense?

fblanqui:
@amelieled I don't want to multiply sub-folders and files. There are already many.
@gabrielhdt ok for package.ml (don't forget it in structure.rst) and pratt.ml
concerning pretty: it's mainly used for --beautify, so it's really like a tool
concerning sr: like sn and cr checking, this can be externalized; it's not really part of the kernel although the kernel may assume that rules indeed satisfy sr and cr. And later, if we just want to use the rewriting engine without doing any type verification (for doing like dkmeta for instance), we can simply disconnect sr and type checking as well.

gabrielhdt:
> @amelieled I don't want to multiply sub-folders and files. There are already many.
> @gabrielhdt ok for package.ml (don't forget it in structure.rst) and pratt.ml
> concerning pretty: it's mainly used for --beautify, so it's really like a tool
> concerning sr: like sn and cr checking, this can be externalized; it's not really part of the kernel although the kernel may assume that rules indeed satisfy sr and cr. And later, if we just want to use the rewriting engine without doing any type verification (for doing like dkmeta for instance), we can simply disconnect sr and type checking as well.

`Pretty` cannot be put into `Tool` because layers that depend on `P_term` may want to display them, and hence they require `Pretty`. For instance, `Pratt` may use `Pretty` to display which term raised a priority conflict.

fblanqui:
Remark: move inductive to handle/ ?

fblanqui:
Questions:
- Why not moving proof, rewrite and why3_tactic into handle/ ?
- Why not moving why3_tactic into tools/why3 ?

fblanqui:
Sorry but, finally, wouldn't it be better to put inductive in tool/ ?

amelieled:
If we do that, `handle.ml` goes too in `tool/`.
Moreover, what's it the difference between `tools/` and `src/tool`? I don't understand

fblanqui:
Le 04/02/2021 à 09:59, amelieled a écrit :
>
> If we do that, |handle.ml| goes too in |tool/|.
>
Why? sr is already in tool/ .


> Moreover, what's it the difference between |tools/| and |src/tool|? I 
> don't understand
>
tools is external to lambdapi, except perhaps gen_version.ml that we 
perhaps could move to cli

but perhaps we could rename tools or src/tool into something else

gabrielhdt:
I was wondering whether we'd put `Inductive` in `Tool`, but have no opinion. The fact that we hesitate likely means that we're not yet sure of the meaning of the `Tool` folder. If I'm correct, for you, @amelieled, this library contains external programs that are not needed by anything, except the command line interface. For you @fblanqui, it containts pieces of code that are not part of the kernel, and that may be used for anything, while `Handle` is dedicated to build a proper signature. I have no clear idea, and perhaps for the moment it's not a problem to hesitate, and I'm already quite satisfied with this PR anyway.

For `generate_tests.ml`, yes it would be a good idea, but perhaps in another PR? We'd be able to create quick tests and long tests thanks to Alcotest.

PR #551, opened 2021-02-04
[doc] [ui] Nits

### LINKED ISSUE #417: lambdapi and vscodium
I'm trying to use codium and lambdapi, I'm following the instruction in https://github.com/Deducteam/lambdapi/blob/master/doc/ui.md

- The first problem is that the makefile in editors/vscode/ installs the extension to `~/.vscode/extensions/lp-vscode` while vscodium uses `~/.vscode-oss/extensions/lp-vscode`. I guess this is expected, it may be mentioned in the doc how to override this path by passing a variable to make
- the "manage" button of the vscode extension uses a wrong filer for the settings, namely `@ext:mines paristech / inria.vscode-lp`, that returns no result. I could find the settings of the extension by removing the filter
- I've installed lambdapi system wide and `lambdapi lsp` starts. Still the extension seems to not run it (it gives no error, but does not react to changing a `.lp` file and saving it). I've tried both setting `lp.path` to `lambdapi lsp` and writing a simple shell script that execs that command and set `lp.path` to the absolute path of that script. Am I doing something wrong? Is there a way to debug this?

ejgallego:
- actually we didn't use any of vscoq files [except for the syntax?],
  but we started from the vscode plugin tutorial, note that vscoq
  doesn't use lsp

- comment on where the link is created, fixes #417

### DIFF
--- a/docs/ui/vscode.rst
+++ b/docs/ui/vscode.rst
@@ -1,8 +1,11 @@
 `VSCode`_
 =========
 
-There is an extension for VSCode >= 1.37 derived from VSCoq. To install
-it do from the ``lambdapi`` repository:
+Lambdapi provides an extension for VSCode editing (>= 1.37) . The
+extension was developed from a bare-bones LSP template, as lambdapi
+implements the LSP standard more or less to the letter, but includes
+some extra features to handle proofs and goals. To install it from the
+``lambdapi`` repository do:
 
 .. code:: bash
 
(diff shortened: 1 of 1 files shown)


fblanqui:
Thanks @ejgallego . Should we remove https://github.com/Deducteam/lambdapi/blob/master/editors/vscode/THANKS.md ?

ejgallego:
> Should we remove https://github.com/Deducteam/lambdapi/blob/master/editors/vscode/THANKS.md ?

I dunno, maybe ? I think VsCoq counts for this as "related work", I did have a look to it extensively, but we first did Atom [with lsp] and later moved to vscode, so it doesn't really have a direct derivation in terms of code except maybe for some grammar file that I think has been removed.

When doing the initial mode for vscode there was some cross setup with work by Vincent Laporte on the Coq side, but it was really just instantiating the extension tutorial [but these tricky bootstrap details are hard to get right]

PR #552, opened 2021-02-04
[parser] Workaround buggy sedlex `from_string` lexbuf creation

### LINKED ISSUE #549: p_command.pos has line numbers always 0
`p_command.pos` has line numbers always 0

ejgallego:
This is a really ugly hack, but until this is not fixed upstream we
can do little more.

Fixes #549

I let you folks report this issue `sedlex` upstream.

### DIFF
--- a/src/core/parser.ml
+++ b/src/core/parser.ml
@@ -35,12 +35,23 @@ module type PARSER = sig
 end
 
 module Lp : PARSER = struct
+
+  (* Needed to workaround serious bug in sedlex, see #549 *)
+  let lp_lexbuf_fixup fname lexbuf =
+    let pos = Lexing.
+                { pos_fname = fname
+                ; pos_lnum = 1
+                ; pos_bol = 0
+                ; pos_cnum = 0 } in
+    Sedlexing.set_position lexbuf pos
+
   let stream_of_lexbuf :
     ?inchan:in_channel -> string -> Sedlexing.lexbuf ->
     (* Input channel passed as parameter to be closed at the end of stream. *)
     Syntax.p_command Stream.t =
     fun ?inchan fname lexbuf ->
       Sedlexing.set_filename lexbuf fname;
+      lp_lexbuf_fixup fname lexbuf;
       let parse =
         MenhirLib.Convert.Simplified.traditional2revised LpParser.command
       in


fblanqui:
Thanks @ejgallego but it doesn't seem to fix all the problems: VSCode still doesn't print any goal (see for instance tests/OK/natproof.lp). Idem for Emacs. On the other hand, Emacs does not get frozen anymore.

ejgallego:
> Thanks @ejgallego but it doesn't seem to fix all the problems: VSCode still doesn't print any goal (see for instance tests/OK/natproof.lp). Idem for Emacs. On the other hand, Emacs does not get frozen anymore.

It did something weird to me at first open, but then it worked, both in natproofs and append.

ejgallego:
There is something bizarre going on, but I'm not sure it is due to the parser. For example right now in natproofs I can only see the goals for the second part, the log seems pretty confusing, but basically `proof/goals` doesn't return back.

firewall2142:
```
2460           }
2461         }
2462 [get_node_at_pos]: call: false l:6 p:0 / file:///home/ashish/lambdapi/tests/OK/natproofs.lp, 202:0-209:4
2463 [get_node_at_pos]: call: false l:6 p:0 / file:///home/ashish/lambdapi/tests/OK/natproofs.lp, 192:0-200:4
2464 [get_node_at_pos]: call: false l:6 p:0 / file:///home/ashish/lambdapi/tests/OK/natproofs.lp, 183:0-190:4
2465 [get_node_at_pos]: call: false l:6 p:0 / file:///home/ashish/lambdapi/tests/OK/natproofs.lp, 174:0-181:4
2466 [get_node_at_pos]: call: false l:6 p:0 / file:///home/ashish/lambdapi/tests/OK/natproofs.lp, 165:0-172:4
2467 [get_node_at_pos]: call: false l:6 p:0 / file:///home/ashish/lambdapi/tests/OK/natproofs.lp, 155:0-162:4
2468 [get_node_at_pos]: call: true l:6 p:0 / 1:0-153:43  <<<<<<<
2469 [get_node_at_pos]: call: false l:5 p:0 / file:///home/ashish/lambdapi/tests/OK/natproofs.lp, 202:0-209:4
2470 [get_node_at_pos]: call: false l:5 p:0 / file:///home/ashish/lambdapi/tests/OK/natproofs.lp, 192:0-200:4
```
line `2468` seems strange. 

In `natproofs.lp`
```
152 // Example 17: Very trivial SSReflect example.
153 symbol silly_axiom : Π m n, P (eq nat m n);
154
```
I think the position isn't correct

fblanqui:
It seems to be related to https://github.com/Deducteam/lambdapi/issues/553.

fblanqui:
@firewall2142 As a consequence of this patch, don't we need to decrement line numbers by 1 in the clients?

firewall2142:
> @firewall2142 As a consequence of this patch, don't we need to decrement line numbers by 1 in the clients?

Don't think so as everything looks ok

ejgallego:
> @firewall2142 As a consequence of this patch, don't we need to decrement line numbers by 1 in the clients?

The clients receive LSP-mandated positions, so it is the server that needs to ensure it is sending the right ones.

PR #557, opened 2021-02-08
update vscode API

fblanqui:


fblanqui:
@ejgallego @firewall2142 It's a long time it has not been update.

ejgallego:
I guess we didn't need any of the "proposed" features yet.

Does this bump the require version number for vscode?

fblanqui:
I don't know. I was just wondering whether we shouldn't keep this file in sync?

ejgallego:
> I don't know. I was just wondering whether we shouldn't keep this file in sync?

I don't recall all the details, but we keep this infrastructure at hand to try experimental vscode stuff, so yup it is in principle good to update it, but likely we don't really test it unless we enable the experimental flag, which is off by default for obvious reasons.

PR #560, opened 2021-02-09
Split some files and improve handling of escaped identifiers

fblanqui:


fblanqui:
@gabrielhdt note that, in this PR, I remove ident parsing. Is it fine for you?

gabrielhdt:
I think that the specifications are not clear enough, and we never know what we want to achieve clearly with these naming schemes. But the code is quite simpler so it seem to make sense.

fblanqui:
@gabrielhdt I succeeded in fixing everything I think. The point was Path.of_string.

gabrielhdt:
Nice! Could you describe somewhere how escaped identifiers are managed now? It could be as documentation in source code.

fblanqui:
Well, escaped identifiers are completely transparent now. You don't have to care about them. The only thing to care about is when you want to turn an identifier into a filename. In this case, you need to remove "{|" and "|}". This is done in module.ml.

fblanqui:
In this PR, I also rename everything related to module paths into just modules to make more abstract what is a module at least in the vocabulary. Hence, Path is renamed into Mod, sym_path into sym_mod, sign_path into sign_mod, etc.
@gabrielhdt Any comment before merging ?

src/parsing/lpLexer.ml
```diff
@@ -358,29 +338,29 @@ and nom_comment : lexbuf -> unit = fun buf ->
...
+       is preferred over using anything for identifiers. *)
+
+    | '?', uid -> UID_META(tail 1 buf)
+    | '$', uid -> UID_PAT(tail 1 buf)
+
+    | '@', uid -> ID_EXPL([tail 1 buf])
+    | '@', qid -> ID_EXPL(Mod.of_string (tail 1 buf))
+
+    (*| '`', id -> ID_QUANT(Mod.of_string (tail 1 buf), false)
+    | '`', '@', id -> ID_QUANT(Mod.of_string (tail 2 buf), true)*)
```
gabrielhdt, line 352: Do you want to keep these comments?

src/parsing/lpParser.mly
```diff
@@ -378,31 +377,32 @@ sterm:
...
+    {
+      let b = make_pos $loc(b) (P_Abst(fst b, snd b)) in
+      make_pos $sloc (P_Appl(q, b))
+    }
+//| i=ID_QUANT b=binder
+//  {
+//    let q = make_pos $loc(i) (P_Iden(qid_of_mod $loc(i) (fst i), snd i)) in
+//    let b = make_pos $loc(b) (P_Abst(fst b, snd b)) in
+//    make_pos $sloc (P_Appl(q, b))
+//  }
```
gabrielhdt, line 390: Do you want to keep these comments as well?

src/parsing/pretty.ml
```diff
@@ -7,45 +7,41 @@
...
-  fprintf ff "%a%a" (List.pp path_elt "") mp ident (Pos.make pos id)
+let qident : p_qident pp = fun ppf {elt=(mp,s); pos} ->
+  if LpLexer.is_keyword s then
+    fatal pos "Identifier [%s] is a Lambdapi keyword." s
+  else match mp with
+       | [] -> pp_print_string ppf s
+       | _::_ -> fprintf ppf "%a.%s" Mod.pp mp s
 
-let modifier : p_modifier pp = fun ff {elt; _} ->
+let modifier : p_modifier pp = fun ppf {elt; _} ->
```
gabrielhdt, line 36: It would be preferable that you don't change the names of the variables if it isn't necessary, it has already been renamed from `pp` to `ff` in the past. It requires to sync back branches that are not yet merged and demands extra work.

src/common/escape.ml
```diff
@@ -0,0 +1,47 @@
+(** Escaped identifiers "{|...|}". *)
+
+(** [is_beg_escaped s] tells if [s] starts with '{'. *)
+let is_beg_escaped : string -> bool = fun s ->
+  String.length s > 0 && s.[0] = '{'
```
gabrielhdt, line 5: Don't we have to check if it starts with `{|`?

src/parsing/pretty.ml
fblanqui: ppf is the name used in the code of the OCaml library itself. I don't remember why I used ff at some point.

src/common/escape.ml
fblanqui: Right. This is a slight optimization.

fblanqui:
@gabrielhdt FYI there a few more changes in the last commits:
- move some functions from console.ml to extra.ml
- move logging functions from console.ml to common/debug.ml 
- add printing functions in debug.ml
- fix Escape.split
- add Extra.files
- improve ok_ko.ml using Extra.files
- add test files for escaped identifiers

src/parsing/lpParser.mly
fblanqui: No, I will remove them before merging (if I do not remember ;-))

src/common/escape.ml
fblanqui: Ah but you are right; it could be wrong. Thanks! I'm going to fix it.

fblanqui:
FYI, in the last commits, I revert the renaming path -> mod and split module.ml into path.ml and library.ml.

fblanqui:
I think that I am at the end of this PR. Finally, in addition to improving the handling of escaped identifiers, it splits a number of files, and do some renamings. Everything is summarized in CHANGES.md. I believe that it clarifies a number of things.

fblanqui:
@gabrielhdt There is something that I do not plan to include in this PR but that I would to discuss is console.ml.
There is something strange about this file. Every thing is defined using Stdlib.ref. Why not using Timed.ref and move the state type defined in pure.ml into console?

tests/OK/a b/c.lp
```diff
@@ -0,0 +1,4 @@
+symbol {||}:TYPE;
```
gabrielhdt, line 1: It would help that test files have a description of what they test.

fblanqui:
@gabrielhdt I fixed another subtle point in the use of escaped identifiers in paths: `{|p|}` should be equivalent to `p` when `p` is a regular identifier!

gabrielhdt:
So the path `{|p|}.A` and `p.A` are equal? If it's the case, it should be documented somewhere!

fblanqui:
Right. And I found another difficulty. It is that, if {|p|}.A = p.A, then we should also have p.{|A|}=p.A, right? So my fix is not sufficient I think. But I have a simpler solution in mind: unescape all identifiers at lexing time and, when we need to print an identifier, escape those that need to be escaped. I'm going to test this today.

gabrielhdt:
Indeed, I think the most important is to have a clear and predictable behaviour, so that the user isn't confused.

fblanqui:
@gabrielhdt I did what I said (unescape at lexing time) and fix printing accordingly.
One thing that could be improved is printing in library.ml: as library is in common, we cannot check whether an identifier needs to be escaped or not. A solution perhaps would be to split lexing in two parts: a lexer for identifiers in common, and the rest.
I don't know whether this is possible, but this is not very urgent.

PR #563, opened 2021-02-12
Bidirectional type checker

gabrielhdt:
This pull request provides a new bidirectional type checker. It allows to refine terms, provide a coercion mechanism. It is based on <https://arxiv.org/abs/1202.4905>.


src/core/refiner.ml
```diff
@@ -0,0 +1,321 @@
...
+open Timed
+open Term
+
+let log = Debug.new_logger 'z' "refi" "Refiner"
+let log = log.logger
+
+(** Module that provide a lookup function to the refiner. *)
+module type LOOKUP = sig
+  val lookup : ctxt -> term -> term -> ((term -> term) * constr list) option
+  (** [lookup a b] returns a coercion from [a] to [b] in context [ctx], if it
```
fblanqui, line 12: [lookup ctx a b]

Could you also explain what is returned?

gabrielhdt: For the moment, `lookup a b` returns a function `f` and constraints `eqs` such that if `t` is a term of type `a`, then `f t` is of type `b` if `eqs` are solvable.

But this behaviour might change.

fblanqui: Thanks to add this in the comment.

fblanqui:
You renamed infer into refiner. I would prefer to keep infer to avoid any confusion with the refine tactic.

src/core/refiner.ml
gabrielhdt: It has been changed, now the coercion returns a Bindlib binder.

gabrielhdt:
I wonder whether we could keep `Infer` as a type checker and `Refiner` as the refiner, that is, `Infer` would be the kernel part for typechecking, not needing unification nor coercions. We would thus have the property that if `Refiner.infer t = (t', a)`, then `Infer.check t' a` is true.

src/core/refiner.ml
fblanqui: The comment is still not very clear. Why not saying something like (if I understand correctly): "[lookup a b] returns a pair [Some(coerce,cs)] where [coerce] is a term binder that is a coercion from terms of type [a] to terms of type [b], and [cs] is a list of constraints for [subst coerce (Vari x)] to be of type [b] when [x] is a fresh variable of type [a], if such a coercion exists. It returns [None] otherwise."?

src/core/refiner.ml
```diff
@@ -0,0 +1,316 @@
...
+          (* Unbind [u] and get new context with [x: t_ty ≔ t] *)
+          let x, u, ctx = Ctxt.unbind ctx t_ty (Some t) u in
+          (* Infer type of [u'] and refine it. *)
+          let u, u_ty = infer ctx u in
+          let u_ty = Bindlib.(u_ty |> lift |> bind_var x |> unbox) in
+          let u_ty = Bindlib.subst u_ty t in
+          let u = Bindlib.(u |> lift |> bind_var x |> unbox) in
+          (LLet (t_ty, t, u), u_ty)
+      | Abst (dom, b) ->
+          let dom, _ = type_enforce ctx dom in
```
fblanqui, line 196: type_enforce enforces dom to be typable by a sort while, here, dom should be of type Type. Why not enforcing dom to be of type Type?

src/core/refiner.ml
```diff
@@ -0,0 +1,316 @@
...
+          let x, b, ctx = Ctxt.unbind ctx dom None b in
+          let b, range = infer ctx b in
+          let b = Bindlib.(lift b |> bind_var x |> unbox) in
+          let range = Bindlib.(lift range |> bind_var x |> unbox) in
+          (Abst (dom, b), Prod (dom, range))
+      | Prod (dom, b) ->
+          let dom, dom_s = type_enforce ctx dom in
```
fblanqui, line 203: type_enforce enforces dom to be typable by a sort while, here, dom should be of type Type. Why not enforcing dom to be of type Type?

src/core/refiner.ml
```diff
@@ -0,0 +1,316 @@
...
+          let x, b, ctx = Ctxt.unbind ctx dom None b in
+          let b, b_s = type_enforce ctx b in
+          let b = Bindlib.(lift b |> bind_var x |> unbox) in
```
fblanqui, line 206: could be moved just before (Prod (dom, b), s) line 219.

src/core/refiner.ml
```diff
@@ -0,0 +1,316 @@
...
+            if !(Debug.log_enabled) then
+              log (Extra.gre "Coerced \"%a\" to \"%a\"") Print.pp_term t
+                Print.pp_term out;
+            out
+        | None ->
+            (if !Debug.log_enabled then log (Extra.red "Failed coercion"));
+            unif ctx a b; t
+
+    (** [type_enforce ctx a] refines type [a] in context [ctx]. [a] must be a
+        type. *)
```
fblanqui, line 102: This comment does not clearly explain what is returned. [type_enforce ctx a] returns a pair [(a',s)] where [a'] is a refinement of [a] and [s] a sort so that [a'] is of type [s].

src/core/refiner.ml
```diff
@@ -0,0 +1,316 @@
...
+module Make : functor (_ : LOOKUP) -> S =
+functor
+  (L : LOOKUP)
+  ->
+  struct
+    let constraints = Stdlib.ref []
+
+    exception NotTypable
+
+    let unif : ctxt -> term -> term -> unit =
```
fblanqui, line 66: Comment explaining what this function does missing.

src/core/refiner.ml
fblanqui: Idem on lines 133, 141 and 203.

src/core/refiner.ml
```diff
@@ -0,0 +1,316 @@
...
+          | _ -> default () )
+      | LLet (t_ty, t, u) ->
+          let t_ty, _ = type_enforce ctx t_ty in
+          let t = force ctx t t_ty in
+          let x, u, ctx = Ctxt.unbind ctx t_ty (Some t) u in
+          let ty = Bindlib.(lift ty |> bind_var x |> unbox) in
+          let ty = Bindlib.subst ty t in
+          let u = force ctx u ty in
+          let u = Bindlib.(lift u |> bind_var x |> unbox) in
+          LLet (t_ty, t, u)
```
fblanqui, line 148: What is the advantage of having special cases for Abst and LLet? Because it looks like code duplication. Is it more efficient in practice?

fblanqui: I tested your branch and, actually, this is slower: make tests takes 11.863s with no special cases instead of 13.861s with special cases. So, it seems better not to have special cases.

gabrielhdt: However these special cases allow to infer more concise type, and take advantage of the user defined types. Since we intend to have an interactive tool, it makes sense to keep them, in my opinion. Perhaps we want a "batch" mode that would skip those cases?

gabrielhdt: And in some cases (an example would be welcome), it may generate simpler unification problems.

src/core/refiner.ml
gabrielhdt: I'm not sure that `let x : T := t in u` requires `T` to be of type `Type`, we may have `let Term_0 : term -> TYPE := Term z in ...`.

src/core/refiner.ml
fblanqui: Perhaps, but a time increase of 17% seems really too much.

src/core/refiner.ml
fblanqui: Right. Some tests are necessary though, to make sure that it does not break some invariant somewhere.

src/core/refiner.ml
gabrielhdt: I found that these cases are required to insert coercions into the domain of abstractions, and into the type annotation of local lets.

gabrielhdt: Nope, my mistake, coercions are inserted thanks to
```ocaml
      | Abst (dom, b) ->
          let dom, _ = type_enforce ctx dom in
```
in `infer`

gabrielhdt:
Superseeded by #696

PR #564, opened 2021-02-13
Emacs/VSCode: Interactive mode

### LINKED ISSUE #513: VSCode: Ctrl-Up doesn't work at the top of the file
On tests/OK/natproofs.lp after a few Ctrl-Down/Right and finally Ctrl-Up as much as possible, the top two lines stay green.

### LINKED ISSUE #514: VSCode: Stay green in spite of changes.
If you change a green line, this line stays green.

### LINKED ISSUE #515: Emacs/VSCode: navigation in files command-by-command

### LINKED ISSUE #519: Emacs/VSCode: do not allow navigation after an error
Perhaps it would be better to mark lines where there is an error in a more visible way as well, for instance by using a red background.

firewall2142:
Changes:
- Interactive mode for Emacs
- Buttons for Proof Navigation in Emacs
- Navigate by commands/tactics in Emacs and VScode
- Evaluated region shrinks on edit in Emacs and VScode
- Evaluated region changes colour after error in Emacs and VScode
- LSP server sends logs only from last command/tactic
- Few minor corrections in LSP server
- Improve VSCode indentation

Fixes #519 
Fixes #515 
Fixes #514
Fixes #513 

TODO:
- [x] Syntax highlight needs to be updated in Emacs


fblanqui:
Thank you very much @firewall2142 ! That looks very nice.

You seem to fix several issues. It would be nice to mention them in the first comment.

Concerning Emacs:

It would be nice to take advantage of this PR to update the user manual as well.
- For instance, "Lambdapi handles only single-line comments with //" is not correct anymore. Does master or the current PR handle multi-lines comments `/* ... */` in Emacs as well?
- We need to remove things related to quickrun.
- What is the effect of C-C C-i? Is it still useful?

I tested your branch and got the following problems:
- When I start modifying something in the green zone, the cursor is moved at the end of the previous command, so that I need to move back to the position where I started to edit the command to finish my modifications.
I don't know how you handle this at the moment but we should distinguish two positions: the cursor position used to edit the document, and the position of the end of the green zone.

Concerning VSCode:

The font doesn't seem to be the same in the goal panel than in the other panels. And some unicode characters are displayed very differently. For instance `⸬`.

src/lsp/lp_lsp.ml
```diff
@@ -166,13 +166,41 @@ let get_textPosition params =
   let line, character = int_field "line" pos, int_field "character" pos in
   line, character
 
+let closest_before (line, pos) (objs: ('a * Pos.popt) list) =
```
fblanqui, line 169: Thanks to add a comment before explaining what this function is intended to do.

firewall2142: done

firewall2142:
>     * For instance, "Lambdapi handles only single-line comments with //" is not correct anymore. Does master or the current PR handle multi-lines comments `/* ... */` in Emacs as well?

Syntax highlight needs to be updated in Emacs, I haven't yet added it to the PR

>     * What is the effect of C-C C-i? Is it still useful?

Yes, it toggles interactive mode

> Concerning VSCode:
> 
> The font doesn't seem to be the same in the goal panel than in the other panels. And some unicode characters are displayed very differently. For instance `⸬`.

I changed the font to use same as the editor. It can be modified in `media/styles.css`

fblanqui:
@firewall2142 Could you please explain what is the difference between C-c C-i and C-c C-c? I do not understand what is the interactive mode anymore, as there is none in VSCode, right?

firewall2142:
> @firewall2142 Could you please explain what is the difference between C-c C-i and C-c C-c? I do not understand what is the interactive mode anymore, as there is none in VSCode, right?

`C-c C-c` evaluates upto cursor.
`C-c C-i` is the electric terminator mode in Proof General, it evaluates upto cursor whenever you enter `;` or `begin`. I thought interactive-goals meant that in elisp code. Should we also have it in vscode?

fblanqui:
Thanks for the explanation. It would be good to put it in the user manual. But what do you think of calling this "automatic-checking mode" and map it to C-c C-a instead?

firewall2142:
> But what do you think of calling this "automatic-checking mode" and map it to C-c C-a instead?

Don't know, as there is also automatic check done by the LSP server.

I was also looking at why it was called electric terminator in PG and found the below links.

[https://www.emacswiki.org/emacs/Electricity](https://www.emacswiki.org/emacs/Electricity) and from [https://github.com/emacs-mirror/emacs/blob/master/lisp/electric.el](https://github.com/emacs-mirror/emacs/blob/master/lisp/electric.el)
```
;; "Electric" has been used in Emacs to refer to different things.
;; Among them:
;;
;; - electric modes and buffers: modes that typically pop-up in a modal kind of
;;   way a transient buffer that automatically disappears as soon as the user
;;   is done with it.
;;
;; - electric keys: self inserting keys which additionally perform some side
;;   operation which happens to be often convenient at that time.  Examples of
;;   such side operations are: reindenting code, inserting a newline,
;;   ... auto-fill-mode and abbrev-mode can be considered as built-in forms of
;;   electric key behavior.
```

fblanqui:
Another problem I found out is: goals are not printed in Emacs (didn't check VSCode yet) when there are several tactics on the same line. See for instance the end of the file tests/OK/inductive.lp.

firewall2142:
That's strange, it works on my computer. Maybe you need to scroll?

fblanqui:
Le 17/02/2021 à 10:26, Ashish Barnawal a écrit :
>
>     But what do you think of calling this "automatic-checking mode"
>     and map it to C-c C-a instead?
>
> Don't know, as there is also automatic check done by the LSP server.
>
What do you mean?


> I was also looking at why it was called electric terminator in PG and 
> found the below links.
>
> https://www.emacswiki.org/emacs/Electricity and from 
> https://github.com/emacs-mirror/emacs/blob/master/lisp/electric.el
>
> |;; "Electric" has been used in Emacs to refer to different things. ;; 
> Among them: ;; ;; - electric modes and buffers: modes that typically 
> pop-up in a modal kind of ;; way a transient buffer that automatically 
> disappears as soon as the user ;; is done with it. ;; ;; - electric 
> keys: self inserting keys which additionally perform some side ;; 
> operation which happens to be often convenient at that time. Examples 
> of ;; such side operations are: reindenting code, inserting a newline, 
> ;; ... auto-fill-mode and abbrev-mode can be considered as built-in 
> forms of ;; electric key behavior. |
>
>
Do you suggest to call it "electric mode" instead, and map it to C-c C-e?

fblanqui:
Sorry, this must have been my mistake. It works indeed.

Le 17/02/2021 à 12:31, Ashish Barnawal a écrit :
>
> That's strange, it works on my computer. Maybe you need to scroll?
>

fblanqui:
On the other hand, there are still problems with the cursor when we modify something in the green zone.
For instance, in tests/OK/inductive.lp:
- If I put my cursor on position (699,12) and do C-c C-c, the green zone stops on line 697 instead of going up to line 699.
- If now I go to position (696,10) to insert "12" in order to rename "n" into "n12", then bad things happen: "1" is inserted at the right position but "2" is inserted at position (695,4), and the cursor finally goes to position (692,3).

fblanqui:
Btw, would you mind to update `editors/vscode/package-lock.json`? When I compile the vscode interface, I get a few modifications. I guess that you get the same, right?
[package-lock.json.txt](https://github.com/Deducteam/lambdapi/files/5996722/package-lock.json.txt)

firewall2142:
> Do you suggest to call it "electric mode" instead, and map it to C-c C-e?

Yes, but I've only used PG so I don't know if there is another common name for it.

> If I put my cursor on position (699,12) and do C-c C-c, the green zone stops on line 697 ...

Those were fixed in earlier commits, so your repo might not be updated?

fblanqui:
Right, sorry. I must have forgotten to do 'make install_emacs'. So what 
remains to do in this PR?

Le 17/02/2021 à 19:29, Ashish Barnawal a écrit :
>
>     If I put my cursor on position (699,12) and do C-c C-c, the green
>     zone stops on line 697 ...
>
> Those were fixed in earlier commits, so your repo might not be updated?
>

firewall2142:
Nothing unless the name of interactive mode needs to be changed

fblanqui:
Yes, I would prefer another name like "electric" as you suggested.

firewall2142:
Also, `C-c C-e` is sometimes used by people to directly go to their `init.el` file (I use it and I saw some other people use it too). So I suggest we use something else like `C-c C-t`

fblanqui:
A BIG thank @firewall2142 !

PR #583, opened 2021-03-01
Various bug fixes

### LINKED ISSUE #573: Allow pattern matching on abstraction/product domains ?
```
constant symbol Set : TYPE;

constant symbol prop : Set;
symbol Prop ≔ τ prop;
injective symbol π : Prop → TYPE;

constant symbol ∀ {a} : (τ a → Prop) → Prop;
set quantifier ∀;

set unif_rule π $p ≡ Π x:$t, π $q[x] ↪ [ $p ≡ `∀ x:$t, $q[x] ];
// fails with: Annotation not allowed in a LHS.
```

### LINKED ISSUE #579: Performance with growing context
I have experienced performances to be very much impacted when the context grows in a theorem and its proof. I show a simplfied example where the same hypothesis is added 20 times to the context. In reality, when a different hypothesis is added, the problem arrives much faster.

The files are given with the .txt extension (so that I can upload them), to test it yourself you should rename them with the .lp extension. The example files use the (adapted) Lambdapi encoding of the first order logic found in `tests/OK/firstOrder.dk`.

On my machine, on file [or_elim_fast.txt](https://github.com/Deducteam/lambdapi/files/6012436/or_elim_fast.txt), checking with Lambapi terminates in about 0.05s.
On file [or_elim_long_2.txt](https://github.com/Deducteam/lambdapi/files/6012437/or_elim_long_2.txt), where we add the same hypothesis 20 times,  it takes approximately 13 seconds to terminate. 

It seems to me that this problem did not occur in my project in a previous version of Lambdapi, about a year ago. I did not manage to check because I don't know how to compile older versions anymore (before mid 2020).

### LINKED ISSUE #580: Induction tactic and inductive symbols
The last proof named `length_app` of `inductive.lp` fails on the branch `refiner` with the error message `τ is not an inductive type` on line 720.

This happens because the inductive symbol that is stored in `sign_ind` is `List`, but there is a reduction rule `rule List ↪ τ list;` and hence the function `ind_data` looks whether `τ list` is an inductive type, which is not the case because only `List` is in `sign_ind`, and symbols are not compared up to convertibility. I'm not completely sure, but I think that there is no error on the master branch because the type checker is not bidirectional.

One solution would be to use an association list rather than a `SymMap` for `sign_ind`, and using `assoc_eq` with the convertibility relation as equality.

fblanqui:
- allow matching on abstraction/product type annotations (fix #573)
- Infer: do not check constraint duplication and return constraints in the order they have been added (fix #579)
- inductive type symbols are now declared as constant (fix #580)
- fix parsing and printing of unification rules
- Extra.files returns only the files that satisfy some user-defined condition
- tests/ok_ko.ml: test only .dk and .lp files
- Pretty: checking that identifiers are LP keywords is now optional (useful for debug)

### DIFF
--- a/tests/OK/inductive.lp
+++ b/tests/OK/inductive.lp
@@ -77,7 +77,7 @@ assert p ptrue pfalse ⊢ ind_B p ptrue pfalse true ≡ ptrue;
 assert p ptrue pfalse ⊢ ind_B p ptrue pfalse false ≡ pfalse;
 
 constant symbol bool : Set;
-rule B ↪ τ bool;
+rule τ bool ↪ B;
 
 ////////////////// False (empty type)
 
@@ -103,7 +103,7 @@ assert p pz psucc ⊢ ind_N p pz psucc z ≡ pz;
 assert p pz psucc n ⊢ ind_N p pz psucc (succ n) ≡ psucc n (ind_N p pz psucc n);
 
 constant symbol nat : Set;
-rule N ↪ τ nat;
+rule τ nat ↪ N;
 
 set builtin "0"  ≔ z;
 set builtin "+1" ≔ succ;
@@ -386,7 +386,7 @@ assert p pnil pcons n l
 ⊢ ind_List p pnil pcons (cons n l) ≡ pcons n l (ind_List p pnil pcons l);
 
 constant symbol list : Set;
-rule List ↪ τ list;
+rule τ list ↪ List;
 
 ///////////////// Pairs of natural numbers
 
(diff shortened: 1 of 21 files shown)


tests/OK/inductive.lp
```diff
@@ -641,31 +641,6 @@ assert p pnil pcons a n x v
 ⊢ ind_Vec p pnil pcons _ _ (Vcons a n x v)
 ≡ pcons a n x v (ind_Vec p pnil pcons a n v);
 
-////////////////// Type Bush (nested inductive type)
```
gabrielhdt, line 644: If I remember correctly, this was an important example for Amelie's work, shouldn't we at least keep some record of it?

fblanqui: The problem is that the induction principle generated by this type definition doesn't seem to be usable. More investigations need to be done on this kind of type definitions. With the current code, this is not a reasonable example to keep at the moment. It will be easy to add it again later if needed.

fblanqui:
Remark: this PR allows matching on abstraction/product type annotations.

gabrielhdt:
Generally, I think I'd be in favour of smaller pull requests, it helps integrating upstream changes into the different branches.

fblanqui:
Ok, in the future, I'll try to split my PRs in smaller pieces.

PR #586, opened 2021-03-04
remove set keyword

fblanqui:


fblanqui:
Any comment on this PR ?

gabrielhdt:
You may want edit the vim syntax file as well.

amelieled:
I have a question: why change the syntax again? It's annoying for other works, and not really change important thing for you.

fblanqui:
@gabrielhdt the vim mode does not handle multi-lines comments. Do you want to fix it? (no need)

PR #590, opened 2021-03-05
Fix admit command

gabrielhdt:
When "admit" is used to finish a proof, each metavariable that remain in
the type of the symbol is substituted with a fresh axiom.

Note that this pull request branches off of #589.


src/handle/command.ml
```diff
@@ -407,31 +407,79 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
...
-                 pp_uid id;
-               fatal pe.pos "We have %a : %a." pp_uid id pp_term a);
-            (match t with
-             | Some(t) when LibTerm.Meta.has true t ->
-                 fatal_msg
-                   "The definition of %a has unsolved metavariables.\n"
-                   pp_uid id;
-                 fatal pe.pos "We have %a : %a ≔ %a."
-                   pp_uid id pp_term a pp_term t
-             | _ -> ());
```
gabrielhdt, line 422: This chunk of code has just been moved to the case `P_proof_end`.

src/handle/command.ml
```diff
@@ -407,31 +407,79 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
...
+                  let type_m = LibTerm.Meta.get true a in
+                  let metas =
+                    match t with
+                    | Some t -> MetaSet.union type_m (LibTerm.Meta.get true t)
+                    | None -> type_m
+                  in
+                  let dep m_a m_b =
+                    (* If [m_a] depends on [m_b], then it is greater than
+                       [m_b], else they are equal. *)
+                    if LibTerm.Meta.depends m_a m_b then 1 else 0
```
fblanqui, line 427: dep does not satisfy the conditions required for a comparison function. See http://caml.inria.fr/pub/docs/manual-ocaml/libref/Array.html.
```
 The following must be true for all x, y, z in a :
    cmp x y > 0 if and only if cmp y x < 0
    if cmp x y >= 0 and cmp y z >= 0 then cmp x z >= 0
```

src/handle/command.ml
```diff
@@ -407,31 +407,79 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
...
+                     pp_uid id;
+                   fatal pe.pos "We have %a : %a." pp_uid id pp_term a);
+                (match t with
+                 | Some(t) when LibTerm.Meta.has true t ->
+                     fatal_msg
+                       "The definition of %a has unsolved metavariables.\n"
+                       pp_uid id;
+                     fatal pe.pos "We have %a : %a ≔ %a."
+                       pp_uid id pp_term a pp_term t
+                 | _ -> ());
```
fblanqui, line 482: I think that metavariable checking is useless now as, normally, all metavariables are in ps.

src/handle/command.ml
```diff
@@ -407,31 +407,79 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
...
                 if finished ps then
                   wrn pe.pos "The proof is finished. Use 'end' instead.";
+                (* Get a list of metas, ordered by dependency. *)
+                let metas =
+                  let type_m = LibTerm.Meta.get true a in
+                  let metas =
+                    match t with
+                    | Some t -> MetaSet.union type_m (LibTerm.Meta.get true t)
+                    | None -> type_m
+                  in
```
fblanqui, line 423: The metas to consider are those of ps.

src/handle/command.ml
```diff
@@ -407,31 +407,79 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
...
+                       [m_b], else they are equal. *)
+                    if LibTerm.Meta.depends m_a m_b then 1 else 0
+                  in
+                  List.stable_sort dep (MetaSet.to_seq metas |> List.of_seq)
+                in
+                (* [add_axiom ss m] adds an axiom for meta [m] in signature
+                   state [ss]. *)
+                let add_axiom ss m =
+                  let name =
+                    "ax_" ^ Option.get (string_of_int m.meta_key) m.meta_name
```
fblanqui, line 435: Two axioms could get the same name as the meta key counter is reset at each command. You should use a specific counter for each file instead.
But, even with this, such a name could be in conflict with a user name introduced before or later. I see no solution at the moment. For the moment, perhaps use a prefix less likely to be used like "__ax_".

src/handle/command.ml
```diff
@@ -407,31 +407,79 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
...
+                     the explicit substitution of the metavariable. *)
+                  let meta_value =
+                    let vars =
+                      let names =
+                        Array.init m.meta_arity
+                          (fun i -> name ^ "_v" ^ (string_of_int i))
+                      in
+                      Bindlib.new_mvar mkfree names
+                    in
+                    let v = Array.to_list vars |> List.map mkfree in
```
fblanqui, line 456: As done in other files (grep Array.init), this can be simplified to:
```
let make_var i = Bindlib.new_var mkfree (Printf.sprintf "x%i" i) in
let v = Array.init arity make_var
```

src/handle/command.ml
```diff
@@ -407,31 +407,79 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
...
+                    let ax = LibTerm.add_args (Symb sym) v in
+                    Bindlib.(lift ax |> bind_mvar vars |> unbox)
```
fblanqui, line 458: You could avoid lifting by building boxed terms from the beginning (use _Appl_symb).

src/handle/command.ml
fblanqui: Yes, but the remark is still valid.

src/handle/command.ml
gabrielhdt: These properties should be guaranteed now, if I'm not mistaken.

src/core/libTerm.ml
```diff
@@ -186,6 +186,36 @@ module Meta = struct
...
+
+  (** [m_a < m_b] if [m_b] depends on [m_a] transitively, such that the least
+      element has less dependencies. It is the transitive closure of the
+      inverse of [depends]. *)
+  let ( < ) : t -> t -> bool = fun m_a m_b ->
+    (* [depends m_a m_b] is true if metavariable [m_b] occurs in the type of
+       [m_a] ([m_a] depends on [m_b]). *)
+    let depends m_a m_b: bool = occurs m_b !(m_a.meta_type) in
+    depends m_b m_a ||
+    MetaSet.exists (fun m -> depends m m_a) (get true !(m_b.meta_type))
```
fblanqui, line 198: < doesn't seem transitive as it is not recursive. Moreover, it would be easier to understand to not permute the arguments of < and depends.
```
let rec lt m n = let t = !(n.meta_type) in occurs m t || MetaSet.exists (lt m) (get true t)
```
But it is not very efficient as occurs is defined with iter (false), and get as well. Moreover, we do not need to generate the set of all metas to check dependency. A single iter (true) should be enough.
```
let lt =
  let exception Found in
  fun m ->
  let f n = if n == m then raise Found in
  fun n ->
  try iter true f !(n.meta_type); false with Found -> true
```
In fact this is the same function as occurs but with iter true instead of iter false.
-> you could add a boolean argument to occurs and use occurs true instead.

src/handle/command.ml
```diff
@@ -407,38 +405,90 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
...
+                let metas =
+                  let ms =
+                    let f = function Typ g -> Some g.goal_meta | _ -> None in
+                    List.filter_map f ps.proof_goals
+                  in
+                  let dep m_a m_b =
+                    if m_a == m_b then 0 else
+                    if LibTerm.Meta.(m_a < m_b) then -1 else
+                    if LibTerm.Meta.(m_b < m_a) then 1 else 0
+                    (* Incomparable meta variables are considered equal. *)
```
fblanqui, line 424: This looks wrong. Imagine you have a2 > a1 and b incomparable with both a1 and a2. Then, dep a2 b = dep b a1 = 0. So, by transitivity, we should have a2 = a1...
The problem is that your dependency order can only be partial. So you can sort metavariables with such an ordering.
We therefore need another strategy.
Why not simply iterating other metavariables, recording those already seen, and raise an exception when we encounter a metavariable already seen?
```
exception Cannot_admit
(* ms is the set of metas already seen *)
let rec admit_meta ms m =
  if MetaSet.mem m ms then raise Cannot_admit;
  iter true (admit ms) !(m.meta_type);
  ... (* code adding a new symbol f of type !(m.meta_type) and setting the value of m to f *)
in
let admit_goal g =
  match g with
  | Unif _ -> ()
  | Typ gt ->
  let m = gt.goal_meta in
  match !(m.meta_value) with
  | Some _ -> ()
  | None -> admit_meta MetaSet.empty gt.goal_meta
in
List.iter admit_goal ps.proof_goals
```

src/core/libTerm.ml
gabrielhdt: This function is recursive because `get` retrieves all metavariables recursively I think, you may see the unit test below that checks this. And for the order of arguments, the idea is that the minimal argument doesn't depend on anything. Good idea for `occurs`.

src/handle/command.ml
gabrielhdt: It looks like metavariables that are in the definition may not end up in `ps`

src/handle/command.ml
gabrielhdt: I implemented something based on this idea. We should note that we shouldn't obtain the `Cannot_admit` because the typechecker should produce admissible terms.

src/handle/command.ml
fblanqui: If it is the case, this is a bug.

gabrielhdt: For information, typing goals from definition are added using the refine tactic on the definition

fblanqui:
@gabrielhdt can I merge this PR now or do you want to modify it?
I am not completely sure whether it is a good idea to assert false when we cannot admit a meta, but we can leave this for the moment and fix it later if needed.

gabrielhdt:
You may merge it :)

On 12 March 2021 07:52:46 CET, "Frédéric Blanqui" ***@***.***> wrote:
***@***.*** can I merge this PR now or do you want to modify it?

PR #594, opened 2021-03-08
emacs mode: update keywords list

fblanqui:


fblanqui:
Remark 1: sequential is missing the emacs grammar.
Remark 2: there is some kind of duplication between lambdapi-vars and lambdapi-smie for the definition of keywords. Couldn't this be avoided?

fblanqui:
@gabrielhdt do you want to fix the emacs grammar in this PR or later?
Hello @firewall2142 . Perhaps you have a suggestion about Remark 2 above?

firewall2142:
>  Hello @firewall2142 . Perhaps you have a suggestion about Remark 2 above?

We could use only the variables defined in lambdapi-vars when we fix the indentation

PR #596, opened 2021-03-11
fix #595

### LINKED ISSUE #595: Performances with both new variables and hypotheses in context
In some cases, performances seem to be impacted when adding new variable names, even if those variables are unused.

I give an example of this using the first order logic encoding found in `tests/OK/firstOrder.dk`. The file is given with the `.txt` extension (so that I can upload it), to test it yourself you should rename it with the `.lp` extension.

Checking [and_elim.txt](https://github.com/Deducteam/lambdapi/files/6109630/and_elim.txt) takes about 0.5s on my machine.
Uncommenting the commented lines in this examples effectively introduces 20 fresh, unused variables. Checking time then goes up to approximately 12 seconds.

fblanqui:


fblanqui:
@gabrielhdt wrote: "I don't understand why the logger is needed, shouldn't be the logging be done in the function calling infer_noexn? It seemed to me that each logger should be used exclusively in the file they are defined."
I am not sure the place you are talking about. I guess https://github.com/Deducteam/lambdapi/blob/08755211c073f4216c34f8dd7864ce3b874415b4/src/core/infer.ml#L185-L186.
The reason of these modifications is to print the results of type inference/checking and constraint solving with debug +h without all the details of recursive calls which you get with debug +i/u. Note also that metavariable instantiation calls type-checking.

gabrielhdt:
Exactly, in that case, shouldn't you call the logger where the `infer_noexn` is called rather than giving the logger as argument? In my opinion, these arguments tend to clutter the code. Furthermore, in my opinion, timing function calls should be performed with a profiler, or timing in caml should be done locally only, and not pushed on the code mainstream, because there is no reason to time type checking and not, for instance, evaluation; so it would lead to call all functions with `time_of` eventually, which may obscure the code.

All this to say that I'm not sure on the proper way to include debugging facilities in the code, and I'm not convinced that what is done here is scalable and eases maintainability.

gabrielhdt:
And for the problem of having too much information on debug, we may introduce debugging levels, so that you may specify how much information you want. I can take care of that if you agree.

fblanqui:
Infer_noexn is called from many places so that its debugging and timing code would be duplicated too much I think. So I prefer to keep it like this for the moment. For the logger argument, I remove it in https://github.com/Deducteam/lambdapi/pull/601. As for adding a debug level for each debug flag, this looks like a good idea indeed.

fblanqui:
Remark: in that case, you could change the grammar of debug flags (in command line arguments as well) since + and - are not needed anymore. Indeed, +h is equivalent to h n with n>0, and -h is equivalent to h 0.

PR #601, opened 2021-03-16
fix and improve inverse image computation and unification rules handling

### LINKED ISSUE #342: Improve inversion of injective functions
Consider this code:

```
require open tests.lambdapi_examples.notation

theorem my_first_proof : π (true = true)
proof
qed
```
I obtain:
```
Cannot solve [B] ≡ [τ ?57].
Cannot solve [B] ≡ [τ ?57].
Unable to infer a sort for [π (true = true)].
```
Of course, I have the same thing with:

```
theorem my_first_proof : π (eq true true)
proof
qed
```

The only way to solve this problem is writing explicitly the implicit type:

```
theorem my_first_proof : π (eq {bool} true true)
proof
qed
```

notation.lp:

```
// Proposition type
constant symbol Prop : TYPE      // Type of propositions
set declared "π"
injective symbol π : Prop ⇒ TYPE // Interpretation of propositions in TYPE

// Set type
constant symbol Set : TYPE       // Type of set codes
set declared "τ"
injective symbol τ : Set ⇒ TYPE  // Interpretation of set codes in TYPE
                      
// Boolean type
constant symbol bool : Set       // Type of booleans
constant symbol B : TYPE
rule τ bool       → B
constant symbol true : B
constant symbol false : B
                        
// Nat type
constant symbol nat  : Set       // Type of natural numbers
constant symbol N : TYPE
rule τ nat        → N
constant symbol O : N
constant symbol S : N ⇒ N
set builtin "0"  ≔ O             // Enabling builtin natural number literals
set builtin "+1" ≔ S

// Leibniz equality
constant symbol eq {a} : τ a ⇒ τ a ⇒ Prop
set infix 1 "=" ≔ eq
constant symbol eq_refl {a} (x:τ a) : π (x = x)
constant symbol eq_ind {a} (x y:τ a) :
                       π (x = y) ⇒ ∀P, π (P y) ⇒ π (P x)

// Set builtins for the rewrite tactic
set builtin "P"     ≔ π
set builtin "T"     ≔ τ
set builtin "eq"    ≔ eq
set builtin "refl"  ≔ eq_refl
set builtin "eqind" ≔ eq_ind

// Properties of Leibniz equality
theorem eq_sym {a} (x y:τ a) : π (x=y) ⇒ π (y=x)
proof
  assume a x y xy
  symmetry
  apply xy
qed
```

### LINKED ISSUE #466: Missing constraints in a unification problem

fblanqui:
- fix and improve in `inverse.ml` the computation of the inverse image of a term wrt an injective function (no unification rule is needed anymore in common examples, fix #342)
- fix management of "initial" constraints in unification (initial is now a global variable updated whenever a new constraint is added)
- when applying a unification rule, add constraints on types too (fix #466)
- turn `Infer.make_prod` into `Infer.set_to_prod`
- add pp_constrs for printing lists of constraints
- make time printing optional
- improve visualization of debugging data using colors:
  . blue: top-level type inference/checking
  . magenta: new constraint
  . green: constraint to solve
  . yellow: data from signature or context
  . red: instantiations (and handled commands)

### DIFF
--- a/src/core/unif.ml
+++ b/src/core/unif.ml
@@ -472,7 +472,7 @@ let solve : ?type_check:bool -> problem -> constr list =
   fun ?(type_check=true) p ->
   if !log_enabled then log_hndl "solve %a" pp_problem p;
   Stdlib.(do_type_check := type_check; initial := p.to_solve);
-  time_of logger_hndl (fun () -> solve p)
+  time_of (fun () -> solve p)
 
 (** [solve_noexn problem] attempts to solve [problem]. If there is
    no solution, the value [None] is returned. Otherwise [Some(cs)] is
--- a/src/core/unif.ml
+++ b/src/core/unif.ml
@@ -80,8 +80,8 @@ let do_type_check = Stdlib.ref true
 let initial : constr list Stdlib.ref = Stdlib.ref []
 let is_initial c = List.exists (LibTerm.eq_constr c) Stdlib.(!initial)
 
-(** [instantiate ctx m ts u] check whether, in a problem [m[ts]=u], [m] can be
-    instantiated and, if so, instantiate it. *)
+(** [instantiate ctx m ts u] check whether, in a problem [m[ts] ≡ u], [m] can
+   be instantiated and, if so, instantiate it. *)
 let instantiate : ctxt -> meta -> term array -> term -> bool =
   fun ctx m ts u ->
   if !log_enabled then log_unif "try instantiate";
@@ -147,7 +147,7 @@ let add_constr : constr list -> constr -> constr list =
       | _ -> c::cs
 
 (** [add_to_unsolved t1 t2 p] checks whether [t1] is equivalent to [t2]. If
-   not, then it tries to apply unification rules on the problem [t1 = t2]. If
+   not, then it tries to apply unification rules on the problem [t1 ≡ t2]. If
    no unification rule applies then it adds [t1 = t2] in the unsolved problems
    of [p]. *)
 let add_to_unsolved : ctxt -> term -> term -> problem -> problem =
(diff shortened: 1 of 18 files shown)


fblanqui:
@gabrielhdt Let me know if you want to have a look. Otherwise I will merge it.

src/core/unif.ml
```diff
@@ -68,24 +67,30 @@ let instantiation : ctxt -> meta -> term array -> term ->
     match nl_distinct_vars ctx ts with
     | None -> None
     | Some(vs, map) ->
+        (*if !log_enabled then
+          log_unif "variables: %a" (Array.pp pp_var " ") vs;*)
```
gabrielhdt, line 71: Have you forgotten to remove this comment?

src/core/unif.ml
```diff
@@ -68,24 +67,30 @@ let instantiation : ctxt -> meta -> term array -> term ->
...
 
 (** Checking type or not during meta instanciation. *)
 let do_type_check = Stdlib.ref true
 
-(** [instantiate ctx m ts u] check whether, in a problem [m[ts]=u], [m] can be
-    instantiated and, if so, instantiate it. *)
-let instantiate : ctxt -> meta -> term array -> term -> constr list -> bool =
-  fun ctx m ts u initial ->
+(** Initial set of constraints. *)
+let initial : constr list Stdlib.ref = Stdlib.ref []
```
gabrielhdt, line 80: Do you think that you could provide an interface for this module? Because I guess that this reference should not be used outside of `unif.ml` right?

gabrielhdt (commented):
Regarding logging, perhaps we could have the loggers to print time systematically if `print_time` is set? I think your colour code is a good idea, but it will be hard to remember when coding, so either we set up an abstraction layer, or it should be documented in the man page.

src/core/unif.ml
fblanqui: Right. The same applies to do_type_check. But I am not sure it is worth doing an interface, at least for now.

fblanqui:
I added color codes in CONTRIBUTING.

PR #602, opened 2021-03-17
delay computation of query results

fblanqui:
The computation of query results as strings added in #518 slowed down performances in a very significant way.
Delay these computations whose results are required in the LSP server only allows to recover the lost efficiency.


fblanqui:
make | 15faf45f | master | PR
tests  | 7s          | 12s       | 7s
focalide | 15s    |  15s      | 15s
holide | 7m30s  |              | 7m15s

PR #608, opened 2021-03-22
compatibility with why3 1.4.0

fblanqui:


fblanqui:
This includes the patch proposed in https://github.com/Deducteam/lambdapi/pull/524.

fblanqui:
We cannot get the available provers anymore. See https://gitlab.inria.fr/why3/why3/-/issues/564.

fblanqui:
This PR works with why3 1.4.0 only but CI fails with this constraint. I don't understand why (it works on my machine). Perhaps @vblot do you have an idea?

vblot:
My guess is that in the cached opam files, why3 version 1.4.0 does not exist. If you move the line `opam update` (and maybe `opam upgrade`) above `opam install` (and maybe above `opam pin add` as well) I think it should work.

fblanqui:
Thanks @vblot ! It works indeed.

PR #618, opened 2021-03-26
Introduce new_tvar = Bindlib.new_var mkfree

fblanqui:


fblanqui:
Hello @gabrielhdt . What do you think of these changes?

src/core/term.ml
```diff
@@ -379,11 +379,21 @@ type tevar = term_env Bindlib.var
...
+let of_tvar : tvar -> term = fun x -> Vari(x)
 
-(** [te_mkfree x] injects the [Bindlib] variable [x] in a {!type:term_env}. *)
-let te_mkfree : tevar -> term_env = fun x -> TE_Vari(x)
+(** [new_tvar s] creates a new [tvar] of name [s]. *)
+let new_tvar : string -> tvar = Bindlib.new_var of_tvar
+
+(** [new_tvar_ind s i] creates a new [tvar] of name [s ^ string_of_int i]. *)
+let new_tvar_ind : string -> int -> tvar = fun s i ->
+  new_tvar (Printf.sprintf "%s%i" s i)
```
gabrielhdt, line 390: What do you think of adding the integer as an optional argument to `new_tvar`?

gabrielhdt (approved):
I only have one suggestion, otherwise it seems good to me.

src/core/term.ml
fblanqui: What do you have in mind? Where would you use this?

gabrielhdt: Having `new_tvar: string -> ?ind:int -> tvar`

fblanqui: I understand that you want to factorize but these two functions are used differently: new_tvar_ind is used in conjunction with Array.init.

PR #623, opened 2021-03-28
No warning when variable does not occur in type of definition

### LINKED ISSUE #341: Wrong warning in scoping

gabrielhdt:
Fixes #341. There is no warning raised when a variable does not appear in the type of a definition nor in the term that defines the symbol.


src/handle/command.ml
```diff
@@ -351,18 +353,20 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
      | _, _, _, _ -> ());
     (* Build proof data. *)
     let data =
-      (* Scope keeping position, with [expo] parsed above. *)
-      let scope_p a = Pos.make a.pos (scope expo a) in
+      (* Scope keeping position, with [expo] parsed above without warning . *)
```
fblanqui, line 356: This comment is confusing: it seems better to remove " without warning ".

src/handle/command.ml
```diff
@@ -351,18 +353,20 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
...
               if p_sym_arg = [] then pt
-              else let pos = Pos.(cat (end_pos p_sym_nam.pos) pt.pos) in
-                   Pos.make pos (P_Abst(p_sym_arg, pt))
+              else
+                let pos = Pos.(cat (end_pos p_sym_nam.pos) pt.pos) in
+                Pos.make pos (P_Abst(p_sym_arg, pt))
             in
-            Some pt, Some (scope_p pt)
+            (* Show warnings if there is no type definition. *)
+            Some pt, Some (scope_p (p_sym_typ = None) pt)
```
fblanqui, line 369: || p_sym_arg <> []

src/handle/command.ml
```diff
@@ -376,7 +380,24 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
             in
             (Some(a), Scope.get_implicitness a)
       in
-      let ao = Option.map scope_p ao in
+      (* Show warnings if it's not a definition. *)
+      let ao = Option.map (scope_p (p_sym_trm = None)) ao in
```
fblanqui, line 384: || p_sym_arg <> []

src/handle/command.ml
```diff
@@ -351,18 +353,20 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
      | _, _, _, _ -> ());
     (* Build proof data. *)
     let data =
-      (* Scope keeping position, with [expo] parsed above. *)
-      let scope_p a = Pos.make a.pos (scope expo a) in
+      (* Scope keeping position, with [expo] parsed above without warning . *)
+      let scope_p warn a = Pos.make a.pos (scope ~warn expo a) in
```
fblanqui, line 357: Why not writing:
```
let warn = p_sym_arg <> [] && p_sym_typ <> None && p_sym_trm <> None in
let scope_p a = Pos.make a.pos (scope ~warn expo a)
```

src/handle/command.ml
```diff
@@ -376,7 +380,24 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
...
+      (* Show warnings if it's not a definition. *)
+      let ao = Option.map (scope_p (p_sym_trm = None)) ao in
+      (* In case of a definition, warn if variables are named but not used
+         inside the type or the definition. *)
+      let rec binders_warn ty te =
+        match ty, te with
+        | Prod(_, by), Abst(_, be) ->
+            let x, by, be = Bindlib.unbind2 by be in
+            if not (Bindlib.occur x (lift by))
+            && not (Bindlib.occur x (lift be)) then
```
fblanqui, line 392: No lifting is necessary!
```
let x, ty, te = Bindlib.unbind2 by be in
if not (Bindlib.occur x by) && not (Bindlib.occur x be) then
  wrn pos "Variable [%a] could be replaced by [_]." pp_var x;
binders_warn ty te
```

fblanqui: Remember to always print identifier by using pp_var or pp_sym in order to handle escaped identifiers properly. Never print variable names using Bindlib.name_of.

src/handle/command.ml
```diff
@@ -376,7 +380,24 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
...
+              wrn pos "Variable [%s] could be replaced by [_]."
+                (Bindlib.name_of x);
+            binders_warn by be
+        | _ -> ()
+      in
+      ( match ao, t with
+        | Some(a), Some(t) -> binders_warn a.elt t.elt
```
fblanqui, line 399: when warn

src/handle/command.ml
gabrielhdt: It's not well typed, `Bindlib.occur : 'a var -> 'a box`!

gabrielhdt: Ah but `binder_constant` can be used instead, don't you think?

fblanqui: Right, sorry. I was thinking in binder_occur or binder_constant indeed.

src/handle/command.ml
gabrielhdt: We want to remove warning when the list of arguments is not empty, and we have both a type and a definition, which would render `warn = not (p_sym_arg <> [] && p_sym_typ = Some _ && p_sym_trm = Some _)`, so `(p_sym_arg <> [] || p_sym_trm = None || p_sym_typ = None)`? Or am I missing something?

src/handle/command.ml
gabrielhdt: We want to run this code when warning in scoping have been disabled, so it would rather be `when not warn`, if I'm not mistaken.

src/handle/command.ml
fblanqui: p_sym_arg = []

gabrielhdt: It's even worse, I think we want `p_sym_arg = [] || p_sym_typ = None && p_sym_trm <> None || p_sym_typ <> None && p_sym_trm = None`, namely, we want to use normal warning if there is only a type or only a term, and if there is no argument, then we use normal warning anyway.

fblanqui: I don't understand. We want warnings when there is no parameters or, if there are parameters, when there is no type or no definition. No need to check that, when there is no type, there is a definition, and vice versa, because there must be one.

gabrielhdt: Ah indeed.

src/core/scope.ml
```diff
@@ -106,6 +106,17 @@ let rec get_implicitness : p_term -> bool list = fun t ->
...
 
+(** [_Meta_Type ~is_rhs env] create a new metavariable of type [TYPE] for a
+    missing domain in environment [env] in a RHS if [~is_rhs] is true. *)
+let _Meta_Type : is_rhs:bool -> env -> tbox = fun ~is_rhs env ->
+  let vs = Env.to_tbox env in
+  let a = Env.to_prod_box env _Type in
+  let m = _Meta_full (Meta.fresh_box a (Array.length vs)) vs in
+  (* Sanity check: only variables of [env] free in [m] if not in RHS. *)
+  if not is_rhs then
+    assert (Bindlib.is_closed (Bindlib.bind_mvar (Env.vars env) m));
```
fblanqui, line 117: Do you think that it is really useful? I do not see the point of this check, and why not doing it if we are in a RHS? We should simply try to remove it and see whether the tests work.

fblanqui: I tested: everything works.

fblanqui: So _Meta_Type can be reduced to:
```
  let _Meta_Type : env -> tbox = fun env ->
    let vs = Env.to_tbox env in
    let a = Env.to_prod_box env _Type in
    _Meta_full (Meta.fresh_box a (Array.length vs)) vs
```

fblanqui: This could be moved to libTerm:
```
(** Create a new metavariable of type [TYPE] in the environment [env]. *)
let _Meta_Type : env -> tbox = fun env ->
  let vs = to_tbox env in
  let arity = Array.length vs in
  let tm = to_prod_box env _Type in
  _Meta_full (Meta.fresh_box tm arity) vs

(** [fresh_meta_tbox env] creates a _Meta tbox from a fresh metavariable whose
   type is a fresh metavariable of type [to_prod env Type]. *)
let fresh_meta_tbox : env -> tbox = fun env ->
  let vs = to_tbox env in
  let arity = Array.length vs in
  let tm =
    let x = Meta.fresh_box (to_prod_box env _Type) arity in
    to_prod_box env (_Meta_full x vs)
  in
  _Meta_full (Meta.fresh_box tm arity) vs
```

src/core/scope.ml
```diff
@@ -156,279 +151,312 @@ let scope : mode -> sig_state -> env -> p_term -> tbox = fun md ss env t ->
...
+  if Timed.(!log_enabled) then log_scop "%a" Pretty.term t;
+  (* Extract the spine. *)
+  let p_head, _ = get_pratt_args ss env t in
+  let scope_b cons xs t =
+    let warn = false in
+    Bindlib.unbox (scope_binder ~warn md ss cons env xs (Some t)) in
+  match p_head.elt with
+  | P_Abst(xs,t) -> scope_b _Abst xs t
+  | P_Prod(xs,t) -> scope_b _Prod xs t
+  | _ -> scope_term expo ss env sgm t
```
gabrielhdt, line 458: I've changed your code here: in the case of constant definitions, the assert false would trigger. Perhaps you intended to check before calling `scope_params` that the list of parameters is not empty?

gabrielhdt: Otherwise we can just split into two functions, one that scopes types and the other that scopes terms. In fact I prefer this alternative.

fblanqui: I don't know exactly why but your proposal was failing to pass all tests.

fblanqui:
@gabrielhdt I fixed your PR. Your review is welcome.

gabrielhdt:
> @gabrielhdt I fixed your PR. Your review is welcome.

I saw that, quite a fix! Thank you, I'll review it.

gabrielhdt:
I have committed two small changes, you may want to review them. Otherwise it's fine for me.

editors/emacs/Makefile
```diff
@@ -31,6 +31,7 @@ endif
 .PHONY: install
 install: $(SRC) lambdapi-site-file.elc
 	$(INSTALL) -m 644 -t $(INSTALL_DIR) $^
+	$(MAKE) clean
```
gabrielhdt, line 34: Could you please add a comment to say what it's for? I guess it's because of the `.install` files, but it's unusual to clean automatically after an install.

fblanqui: Because make fails after a make install_emacs.

fblanqui: Is it fine?

gabrielhdt: I guess, we can't fix this now, so let's keep it and we'll have a look at it later

fblanqui:
In https://github.com/Deducteam/lambdapi/pull/623/commits/b1734593955aa6762ffec9e5165aca286f29b72d, I replaced the expo argument used in M_Term scoping by a boolean telling whether private symbols are allowed. This clarifies the role of this argument and is more in line with the other scoping modes.

PR #624, opened 2021-03-29
fix #621

### LINKED ISSUE #621: Problem with id : {|00|} becomes {|0|}
As far as I'm concerned, surrounding characters around `{|` and `|}` treats it as a well-escaped immutable string.

However, this checks badly:
```dedukti
symbol Typ : TYPE;
symbol eta : Typ → TYPE;

symbol id {|00|} (Hb: eta {|0|}) ≔ Hb;

symbol id2 {|00|} (Hb: eta {|00|}) ≔ Hb;
```

`{|00|}` becomes `{|0|}`.
This also happens when the number is prefixed by something (in my case, `{|:000|}` becomes `{|:0|}`)

fblanqui:


fblanqui:
@gabrielhdt Your review on this is welcome too ;-)

gabrielhdt:
It's ok for me, but this looks like a bug of bindlib that should be fixed upstream rather than here, don't you agree?

fblanqui:
It would be nice indeed. Thanks to open an issue on https://github.com/rlepigre/ocaml-bindlib. But it will probably take some time to be fixed. So, I propose to fix it here for the moment.

fblanqui:
Thanks @gabrielhdt to add here a link to the Bindlib issue when it will be open.

PR #626, opened 2021-03-30
Ensure protected numeric suffix truly is a suffix

### LINKED ISSUE #625: Issue with number suffix prevention
There must be a problem in the regex matching, as "000_" is rejected :
```ocaml
utop #   let digit = [%sedlex.regexp? '0' .. '9']
  let nonzero_nat = [%sedlex.regexp? '1' .. '9', Star digit]
  let nat = [%sedlex.regexp? '0' | nonzero_nat]
  let float = [%sedlex.regexp? nat, '.', Opt (nat)]
  let stringlit = [%sedlex.regexp? '"', Star (Compl ('"' | '\n')), '"']
  let comment = [%sedlex.regexp? "//", Star (Compl ('\n' | '\r'))]

  (* Identifiers are defined by what cannot appear in them. *)
  let forbidden_letter = [%sedlex.regexp? Chars " ,;\r\t\n(){}[]:.`\""]
  let regid = [%sedlex.regexp? Plus (Compl forbidden_letter)]

  (* Identifiers not compatible with Bindlib. *)
  let invalid_bindlib_id = [%sedlex.regexp? Star any, Plus '0', nat]
;;
utop #   let is_invalid_bindlib_id : string -> bool = fun s ->
    let lexbuf = Sedlexing.Utf8.from_string s in
    match%sedlex lexbuf with
    | invalid_bindlib_id -> true
    | _ -> false
;;
val is_invalid_bindlib_id : string -> bool = <fun>
utop # is_invalid_bindlib_id "000_";;
- : bool = true
```

yannl35133:
Not sure if this is the right way to fix #625 but it works on my side at least.

### DIFF
--- a/src/parsing/lpLexer.ml
+++ b/src/parsing/lpLexer.ml
@@ -153,7 +153,7 @@ end = struct
   let regid = [%sedlex.regexp? Plus (Compl forbidden_letter)]
 
   (* Identifiers not compatible with Bindlib. *)
-  let invalid_bindlib_id = [%sedlex.regexp? Star any, Plus '0', nat]
+  let invalid_bindlib_id = [%sedlex.regexp? Star any, Plus '0', nat, eof]
 
   (* Once unescaped, escaped identifiers must not be empty, as the empty
      string is used in the path of ghost signatures. *)


fblanqui:
Thank you @yannl35133 ! Following your proposal, I realize that the same trick needs to be applied in other functions, and that pp_uid is bugged.

PR #627, opened 2021-03-30
Move scope to parsing

gabrielhdt:
This pull requests moves `Scope` to the `Parsing` module. That way, the "core" of lambdapi does not depend on parsing. It specifies what the terms are, and the parser produces such terms.


src/core/libTerm.ml
```diff
@@ -5,6 +5,9 @@ open Term
 open Lplib.Base
 open Lplib.Extra
 
+(** Representation of a possibly qualified identifier. *)
+type qident = Common.Path.t * string
+
```
fblanqui, line 10: It would be better in term.ml.

src/core/print.ml
```diff
@@ -54,18 +54,35 @@ let pp_assoc : Pratter.associativity pp = fun ppf assoc ->
...
+let pp_notation : Sign.notation pp = fun ppf notation ->
   match notation with
   | Prefix(p) -> out ppf "prefix %f" p
   | Infix(a,p) -> out ppf "infix%a %f" pp_assoc a p
   | Zero -> out ppf "builtin \"0\""
   | Succ -> out ppf "builtin \"+1\""
   | Quant -> out ppf "quantifier"
 
-let pp_uid = LpLexer.pp_uid
+let pp_uid = Format.pp_print_string
```
fblanqui, line 65: This is problematic as pp_uid adds brakets around non-regular identifiers so that printed identifiers are parsable. However, we currently do this because surroundings brackets are stripped off from identifiers during lexing. We could as well decide not to do that but, then, we need to be careful every where we use identifiers:
- in filenames, brackets need to be removed
- when generating a new name from an existing one: if the existing one is escaped then its brackets needs to be removed first, and the new name needs to be bracketed -> we need to define and use add_prefix/add_suffix functions

src/core/sig_state.ml
```diff
@@ -119,16 +119,14 @@ let of_sign : Sign.t -> sig_state = fun signature ->
...
   fun ~prt ~prv st {elt = (mp, s); pos} ->
-  let pp_uid = Parsing.LpLexer.pp_uid in
-  let pp_path = Lplib.List.pp pp_uid "." in
   let s =
     match mp with
     | [] -> (* Symbol in scope. *)
         begin
           try fst (StrMap.find s st.in_scope)
-          with Not_found -> fatal pos "Unknown object %a." pp_uid s
+          with Not_found -> fatal pos "Unknown object %s." s
```
fblanqui, line 129: Ah, this reverting many changes I made recently in order to print identifiers correctly... Identifiers should not be printed as strings unless we decide to keep brackets in escaped identifier. See https://github.com/Deducteam/lambdapi/pull/627/files#r603836591.

gabrielhdt: Yeah, I realise that. It would motivate going back to the previous strategy and have strings with booleans to say whether it should be escaped or not.

src/core/term.ml
```diff
@@ -9,10 +9,38 @@
 
 open Timed
 
-module Tags = Parsing.Syntax.Tags
-
 (** {3 Term (and symbol) representation} *)
 
+(** Symbol modifiers and attributes. *)
+module Tags = struct
```
fblanqui, line 15: What's the point of this module? I would prefer to be able to access its types directly.

src/core/unif_rule.ml
```diff
@@ -16,20 +18,26 @@ open Parsing
...
 let sign : Sign.t =
   let dummy = Sign.dummy () in
-  let sign = {dummy with Sign.sign_path = LpLexer.unif_rule_path} in
-  Sign.loaded := Path.Map.add LpLexer.unif_rule_path sign !(Sign.loaded);
+  let sign = {dummy with Sign.sign_path = path} in
+  Sign.loaded := Path.Map.add path sign !(Sign.loaded);
   sign
 
+(** Parser representation of equiv symbol. *)
+let p_equiv = "≡"
```
fblanqui, line 26: Why exporting this constant? The name of equiv can be obtained by equiv.sym_name if needed.

src/core/sig_state.ml
fblanqui: No need for a boolean. Escape.is_escaped tells you whether an identifier is escaped or not.

gabrielhdt: Whether it is escaped, but not whether it should be escaped, right?

fblanqui: As explained in https://github.com/Deducteam/lambdapi/pull/627/files#r603836591, the only places where we have to take care of this is with filenames and when adding suffixes or prefixes to existing identifiers.

fblanqui: I can take care of this if you want.

src/core/print.ml
fblanqui: Another problem perhaps is that, when s is a regular identifier, then s and {|s|} are distinct, while they are identified currently.

src/core/sig_state.ml
gabrielhdt: You can :)

src/parsing/lpLexer.ml
```diff
@@ -103,44 +103,6 @@ type token =
...
-  (** [equiv] is the name of the symbol "≡" in unification rules. *)
-
-  val cons : string
-  (** [cons] is the name of the symbol ";" in unification rules. *)
-
-  val unif_rule_path : Path.t
-  (** [unif_rule_path] is the path for [equiv] and [cons].
-      It cannot be entered by a user. *)
-
-end = struct
```
gabrielhdt, line 143: I'd prefer to keep the signature, I think it really helps to know what can be used outside and what cannot.

fblanqui: It's not very nice to disconnect declarations, comments and regexp definitions. Probably one day we should add .mli files.

src/core/eval.ml
```diff
@@ -410,9 +410,21 @@ let rec hnf : ctxt -> term -> term = fun ctx t ->
...
 
+(** Type representing the different evaluation strategies. *)
+type strategy =
+  | WHNF (** Reduce to weak head-normal form. *)
+  | HNF  (** Reduce to head-normal form. *)
+  | SNF  (** Reduce to strong normal form. *)
+  | NONE (** Do nothing. *)
+
+(** Configuration for evaluation. *)
+type eval_config =
```
fblanqui, line 421: Just call it config?

src/core/sign.ml
```diff
@@ -374,3 +382,6 @@ let rec dependencies : t -> (Path.t * t) list = fun sign ->
     | d::deps -> minimize ((List.filter not_here d) :: acc) deps
   in
   List.concat (minimize [] deps)
+
+(** [ghost_path s] creates a module path that can be entered by a user. *)
```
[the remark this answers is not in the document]
gabrielhdt: Indeed, indeed

src/parsing/lpLexer.ml
gabrielhdt: We can already provide some interfaces right now.

CHANGES.md
```diff
@@ -1,5 +1,12 @@
 ### Unreleased
 
+#### Do not unescapde identifiers anymore (2021-03-30)
```
gabrielhdt: The main point was to move `scope.ml` from `Core` to `Parsing`

fblanqui:
This PR fixes #625 by including #626 as well.

PR #632, opened 2021-04-02
Fix lambdapi package install command

gabrielhdt:
Make install calls `dune build -p lambdapi @install` to build effectively the `lambdapi.install` file, and the emacs mode isn't installed by default.


fblanqui:
Thanks @gabrielhdt !

PR #636, opened 2021-04-05
Modularise some lambdapi functions

gabrielhdt:
This pull request proposes to split some functions to offer a richer API. Pragmatically, those functions are required by a prototype that uses branch #563 along with a coercion generation mechanism.

### DIFF
--- a/src/handle/command.ml
+++ b/src/handle/command.ml
@@ -37,6 +37,16 @@ let _ =
   in
   register "+1" expected_succ_type
 
+(** [rule_of_pre_rule r] converts a pre-rewrite rule into a rewrite rule. *)
+let rule_of_pre_rule : Scope.pre_rule -> Term.rule =
+  fun {pr_lhs; pr_vars; pr_rhs; pr_arities; pr_xvars_nb; _} ->
+  { lhs = pr_lhs
+  ; rhs = Bindlib.(unbox (bind_mvar pr_vars pr_rhs))
+  ; arity = List.length pr_lhs
+  ; arities = pr_arities
+  ; vars = pr_vars
+  ; xvars_nb = pr_xvars_nb }
+
 (** [handle_open ss p] handles the command [open p] with [ss] as the
    signature state. On success, an updated signature state is returned. *)
 let handle_open : sig_state -> p_path -> sig_state =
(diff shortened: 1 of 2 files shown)


src/handle/command.ml
```diff
@@ -37,6 +40,16 @@ let _ =
   in
   register "+1" expected_succ_type
 
+(** [rule_of_pre_rule r] converts a pre-rewrite rule into a rewrite rule. *)
+let rule_of_pre_rule : Scope.pre_rule -> Term.rule =
```
fblanqui, line 44: Wouldn't this function better in scope.ml?

src/handle/compile.ml
```diff
@@ -21,11 +21,15 @@ let parse_file : string -> Syntax.ast = fun fname ->
...
-    necessary (the corresponding object file does not exist,  must be updated,
-    or [force] is [true]).  In that case,  the produced signature is stored in
-    the corresponding object file. *)
-let rec compile : bool -> Path.t -> Sign.t = fun force mp ->
+(** [with_handle ~handle ~force mp] compiles the file corresponding to module
+    path [mp] using function [~handle] to process commands. Module [mp] is
+    process when it is necessary, i.e. the corresponding object file does not
+    exist, or it must be updated, or [~force] is [true]). In that case, the
+    produced signature is stored in the corresponding object file. *)
+let rec with_handle :
```
fblanqui, line 29: Why do you want to rename it into "with_handle"? Note that the output type is "compiler".

src/handle/command.ml
```diff
@@ -518,3 +513,24 @@ let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
   | Fatal(Some(None)   ,m)      -> fatal pos "Error on command.\n%s" m
   | e                           ->
       fatal pos "Uncaught exception: %s." (Printexc.to_string e)
+
+(** [with_proof compile_mod ss cmd] handles command [cmd] in signature state
+    [ss] (as defined by previously defined [handle]), and handles proof
+    scripts attached to symbols as well. The function [compile_mod] is used to
+    compile required modules recursively. *)
+let with_proofs : compiler -> Sig_state.t -> Syntax.p_command ->
+  Sig_state.t =
```
fblanqui, line 522: Could perhaps be on one line.

src/handle/command.ml
```diff
@@ -502,7 +497,7 @@ let too_long = Stdlib.ref infinity
     exception handling. In particular, the position of [cmd] is used on errors
     that lack a specific position. All exceptions except [Timeout] and [Fatal]
     are captured, although they should not occur. *)
-let handle : (Path.t -> Sign.t) -> sig_state -> p_command ->
+let handle : compiler -> sig_state -> p_command ->
```
fblanqui, line 500: Rename into proof_data_of_command ?

src/handle/compile.ml
fblanqui: I instead suggest the following renamings:
- handle -> proof_data_of_command
- with_proofs -> handle
- with_handle -> compile

gabrielhdt:
> Why do you want to rename it into "with_handle"? Note that the output type is
> "compiler".

Because I'd like to expose both "compile" (which remains the same as
before), and this one.

src/handle/compile.ml
gabrielhdt: I renamed `handle` into `get_proof_data` (which is a bit shorter, and as explicit provided we use `Command.get_proof_data`).

fblanqui: What about with_handle -> compile_with ?

fblanqui:
@gabrielhdt Do you want to update CHANGES before merging?

gabrielhdt:
> @gabrielhdt Do you want to update CHANGES before merging?

It's done, if you're happy with the PR, you can merge it.

fblanqui:
Thanks @gabrielhdt !

PR #639, opened 2021-04-07
add commutative and associative-commutative symbols

fblanqui:
by privatizing the type `term` and adding `term.mli`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,20 @@
 
 #### Add commutative and associative-commutative symbols (2021-04-07)
 
-- by privatizing the type `term` and adding `term.mli`
+- Add `term.mli` and turn the `term` type into a *private* type so that
+  term constructors are not exported anymore (they are available for
+  pattern-matching though). For constructing terms, one now needs to
+  use the provided construction functions `mk_Vari` for `Vari`,
+  `mk_Appl` for `Appl`, etc.
+
+- Move some functions `LibTerm` to `Term`, in particular `get_args`,
+  `add_args` and `cmp_term`.
+
+- Rename the field `sym_tree` into `sym_dtree`.
+
+- Redefine the type `rhs` as `(term_env, term) Bindlib.mbinder`
+  instead of `(term_env, term) Bindlib.mbinder * int` so that the old
+  `rhs` needs to be replaced by `rhs * int` in a few places.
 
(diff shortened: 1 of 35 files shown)


fblanqui:
@gabrielhdt Your comments are welcome as it may have some impact on your other branches. But the fix is easy:
- Some functions have been moved from LibTerm to Term -> remove the prefix "LibTerm.".
- Term constructors are not exported anymore except for matching (term is a private type) -> you need to use the provided term construction functions mk_Appl for Appl, mk_Vari for Vari, etc. tbox constructors are not affected as they already use the provided construction functions.
- I also renamed sym_tree into sym_dtree, and redefined rhs as `(term_env, term) Bindlib.mbinder` so that the old `rhs` needs to be replaced by `rhs * int` in a few places.

gabrielhdt:
All right, thanks for the summary, it looks good to me. Is there a reference describing the work you've implemented?

fblanqui:
On the implementation of construction functions for non-free concrete data types. F. Blanqui, T. Hardin and P. Weis. ESOP'07. LNCS 4421. http://dx.doi.org/10.1007/978-3-540-71316-6_8. It has been implemented in http://moca.inria.fr/ . ;-)

Abstract: Many algorithms use concrete data types with some additional invariants. The set of values satisfying the invariants is often a set of representatives for the equivalence classes of some equational theory. For instance, a sorted list is a particular representative wrt commutativity. Theories like associativity, neutral element, idempotence, etc. are also very common. Now, when one wants to combine various invariants, it may be difficult to find the suitable representatives and to efficiently implement the invariants. The preservation of invariants throughout the whole program is even more difficult and error prone. Classically, the programmer solves this problem using a combination of two techniques: the definition of appropriate construction functions for the representatives and the consistent usage of these functions ensured via compiler verifications. The common way of ensuring consistency is to use an abstract data type for the representatives; unfortunately, pattern matching on representatives is lost. A more appealing alternative is to define a concrete data type with private constructors so that both compiler verification and pattern matching on representatives are granted. In this paper, we detail the notion of private data type and study the existence of construction functions. We also describe a prototype, called Moca, that addresses the entire problem of defining concrete data types with invariants: it generates efficient construction functions for the combination of common invariants and builds representatives that belong to a concrete data type with private constructors.

PR #641, opened 2021-04-08
Rewrite engine testing, failure on product matching

gabrielhdt:
This pull request brings unit tests on the product matching, which is not tested into `.lp` files, since filtering on products is forbidden. These tests revealed a bug, rules of the form `rule C (Π _: _, A) ↪ Ok;` raise an error, because the rule is parsed as 
```
rule C (?0 → A) ↪ Ok;
```

Another bug has been found: the following file
```
symbol ι: TYPE;
symbol f: (ι → ι) → (ι → ι) → ι;
symbol z: ι;
symbol s : ι → ι;
rule f _ s ↪ z;
rule f s _ ↪ z;
rule f (λ x, x) s ↪ z;
```
produces a decision tree with non linearity checks and variable occurence tests, whereas it shouldn't be the case.

Here is the tree, obtained using `lambdapi decision-tree MOD.PATH.f`
out

The error came from the creation of new wildcards in the tree generation, the new generated tree is as follows
out


src/core/tree.ml
```diff
@@ -218,6 +196,13 @@ module CM = struct
     { arg_path : int list (** Reversed path to the subterm. *)
     ; arg_rank : int      (** Number of abstractions along the path. *) }
 
+  (** [pp_arg_path ppf pth] prints path [pth] as a dotted list of integers to
+      formatter [ppf]. *)
+  let pp_arg_path : int list pp = fun ppf pth ->
+    Format.(pp_print_list
+              ~pp_sep:(fun ppf () -> pp_print_string ppf ".")
```
fblanqui, line 203: You can use `Base.pp_sep "." ()` instead, or even `List.pp pp_print_int "."`.

src/core/tree.ml
```diff
@@ -213,10 +191,29 @@ end
...
+  (** [pp_head ppf t] prints head of term [t]. *)
+  let pp_head : term pp = fun ppf t ->
+    let open Format in
+    match get_args t with
+    | Symb s, _ -> pp_print_string ppf s.sym_name
+    | Patt _, _ -> pp_print_string ppf "$"
+    | Vari _, _ -> pp_print_string ppf "x"
+    | Abst _, _ -> pp_print_string ppf "λ"
+    | Prod _, _ -> pp_print_string ppf "Π"
+    | _ -> assert false (* Terms that souldn't appear in lhs *)
```
fblanqui, line 203: Perhaps you can factorize pp_print_string?
```
pp_print_string ppf
  (let h, _ = get_args t in
   match h with
   | Symb s -> s.sym_name
   | ...)
```

fblanqui:
In https://lambdapi.readthedocs.io/en/latest/dtrees.html, "var" should be replaced by "v". Perhaps, it would be nice to add some example in the doc?

fblanqui:
Why do you use ⊊ and not⊆?

gabrielhdt:
> Why do you use ⊊ and not⊆?

I don't think there's a good reason.

> What is the meaning of "@0" and "@2" in the graphs above? This does not seem to be explained in the doc.

Indeed it's not explained, I must have forgotten. "@n" means that we look at the nth term of the stack.

fblanqui:
@gabrielhdt Let me know when you want to merge.

gabrielhdt:
I think we ought to update the documentation of `term.mli` concerning `Patt` constructors: it says (as it used to be the case), that if a pattern is linear and not used in the RHS, then its first component is `None`. But it's not true anymore, scoping always create patterns of the form `Patt(Some(i), ...)`.

gabrielhdt:
Everything is ready, it seems. I update only the documentation in `term.mli` and removed the obsolete one from `term.ml`.

fblanqui:
@gabrielhdt the meaning of "@" is still not explained.

fblanqui:
In your example above, it would help to use different rhs, says a, b and c, instead of the same rhs z. Could you please then explain the tree that we get?

gabrielhdt:
> In your example above, it would help to use different rhs, says a, b and c, instead of the same rhs z. Could you please then explain the tree that we get?

Using
```
symbol ι: TYPE;
symbol f: (ι → ι) → (ι → ι) → ι;
symbol z: ι;
symbol s : ι → ι;
rule f _ s ↪ z;
rule f s _ ↪ s z;
rule f (λ x, x) s ↪ s (s z);
```
we obtain
out

gabrielhdt:
Given any term of the form `f t u`, the input stack is `[t; u]` at first. The first node says we look at term 0 of the stack, so term `t`. If `t` is `s`, we continue on the first arc. If `t` is of the form `\x: A, t'`, we continue on the middle arc, and the stack becomes `[A; t'; u]`. Otherwise, we continue on the third arc.

fblanqui:
Why do you have 2 square nodes?

fblanqui:
PS. Are you sure you have the right tree as there is no leaf with `s (s z)`?

gabrielhdt:
> Why do you have 2 square nodes?

The first square node is a consequence of the fact that scoping always generate patterns with an index, that is `Patt(Some(i), ...)` even if the pattern is not used in the rhs and is linear.

> PS. Are you sure you have the right tree as there is no leaf with s (s z)?

There is no leaf with `s (s z)` because a term of the form `f (\x, x) s` is also filtered by `f _ s`, and first rule supersedes the last one.

fblanqui:
@gabrielhdt For the record, could you please add an issue about this problem of always having an index?

fblanqui:
@gabrielhdt Don't you want to add something in CHANGES before merging?

fblanqui:
Thanks @gabrielhdt !

PR #651, opened 2021-04-28
Allow 0.001 float value

yannl35133:
Such float values with leading zeroes were rejected.

### DIFF
--- a/src/parsing/lpLexer.ml
+++ b/src/parsing/lpLexer.ml
@@ -108,7 +108,7 @@ exception SyntaxError of strloc
 let digit = [%sedlex.regexp? '0' .. '9']
 let nonzero_nat = [%sedlex.regexp? '1' .. '9', Star digit]
 let nat = [%sedlex.regexp? '0' | nonzero_nat]
-let float = [%sedlex.regexp? nat, '.', Opt (nat)]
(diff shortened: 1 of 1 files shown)


yannl35133:
#652

yannl35133:
Seems like why3 is not properly recognized, I doubt it has anything to do with my modifications and don't know how to fix it

src/parsing/lpLexer.ml
```diff
@@ -108,7 +108,7 @@ exception SyntaxError of strloc
 let digit = [%sedlex.regexp? '0' .. '9']
 let nonzero_nat = [%sedlex.regexp? '1' .. '9', Star digit]
 let nat = [%sedlex.regexp? '0' | nonzero_nat]
-let float = [%sedlex.regexp? nat, '.', Opt (nat)]
+let float = [%sedlex.regexp? Opt nat, '.', Opt (Star digit)]
```
fblanqui, line 111: Instead, I propose to use:
```
let float = [%sedlex.regexp? nat, '.', Plus digit]
```

fblanqui:
I seems that alt-ergo is not installed while it is needed for running tests. @gabrielhdt shouldn't tests/dune include alt-ergo as dependency?

Issue #430, opened 2020-06-30
Improve parsing of quantifiers


fblanqui:
```
set declared "∀"
symbol Prop:TYPE
symbol A:TYPE
symbol ∀:(A→Prop)→Prop
set quantifier ∀
symbol p:A→Prop
type ∀x,p x
symbol a:A
symbol imp:Prop→Prop→Prop
set infix left 5 "⇒" ≔ imp
type p a ⇒ (∀x, p x)
type p a ⇒ ∀x, p x
// [/home/blanqui/src/lambdapi/tmp/forall.lp, 12:13] Parse error.
```

fblanqui:
After https://github.com/Deducteam/lambdapi/pull/516, we have:
```
symbol Prop:TYPE;
symbol A:TYPE;
symbol ∀:(A → Prop) → Prop;
set quantifier ∀;
symbol p:A → Prop;
type `∀ x,p x;
symbol a:A;
symbol imp:Prop → Prop → Prop;
set infix left 5 "⇒" ≔ imp;
type p a ⇒ (`∀ x, p x);
type p a ⇒ `∀ x, p x;
// [11:11-12] Unexpected token [`].
```

fblanqui:
Current situation:
[`f x, t] indicates to the parser that the argument following the symbol f (even not declared as quantifier) is an abstraction without a lambda. It is equivalent to [f (λ x, t)].
When writing [a ⇒ `f x, t], note that ⇒ is not the builtin arrow type → but a user-defined symbol. So, before the application of symbol priorities, this is the application of a to ⇒ and [`f x, t]. But applications (sterm's) are defined as sequences of aterm's and [`f x, t] is not an aterm but a term, like abstractions and products. Term's can be injected into aterm's by enclosing them between parentheses. Having [`f x, t] be a term allows us to nest quantifiers and write [`f x, `f y, t] as the body of an abstraction is expected to be a term.

Note by the way that it would be also nice to be able to write [`f x y, t] instead of [`f x, `f y, t] but this is another story.
 
@yannl35133 asked: why aterm's and term's are not the same? This is a way to define a non-ambiguous LR(1) grammar for lambda-terms. But we could perhaps use another grammar or let the parser solve some conflicts in a specific way.

Proposals for changing the grammar and solve these problems are welcome.

PR #659, opened 2021-05-04
lpParser: better handling of quantifiers

fblanqui:
proposal for fixing #430

### DIFF
--- a/src/parsing/lpParser.mly
+++ b/src/parsing/lpParser.mly
@@ -126,26 +126,7 @@
 %start <Syntax.p_command> command
 %start <Syntax.p_qident> id
 
-%type <Syntax.p_ident> uid
-%type <Syntax.p_path> path
-%type <Syntax.p_ident option> patt_id
-%type <Syntax.p_params> params
-
-%type <Syntax.p_term> term
-%type <Syntax.p_term> aterm
-%type <Syntax.p_term> sterm
-
 %type <Sign.notation> notation
-%type <Syntax.p_rule> rule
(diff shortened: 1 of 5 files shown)


fblanqui:
@gabrielhdt @yannl35133 @QGarchery Thanks for your comments on this PR.
make doesn't print any conflict in the grammar. How to make sure that this is indeed the case?

fblanqui:
The file _build/default/src/parsing/lpParser.conflicts is empty.

QGarchery:
I don't know precisely how lambdapi is built but a call to `menhir --explain lpParser.conflict` in `src/parsing` does indeed produce an empty file `lpParser.conflicts` which indicates that there is no conflict.

Thank you for this PR !

src/parsing/lpParser.mly
```diff
@@ -325,13 +306,11 @@ aterm:
   | L_CU_BRACKET t=term R_CU_BRACKET { make_pos $sloc (P_Expl(t)) }
   | n=INT { make_pos $sloc (P_NLit(n)) }
 
-sterm:
-  | t=sterm u=aterm { make_pos $sloc (P_Appl(t,u)) }
+saterm:
```
gabrielhdt, line 309: Why `saterm`? I mean `sterm` and `aterm` refer to "atomic" terms and "symbolic" terms, is there a meaning being "saterm"?

fblanqui: I thought sterm = sequence of term's. So saterm = sequence of aterm's.

Issue #657, opened 2021-04-29
[FEATURE REQUEST] A syntax to distinguish a notation and its underlying value


yannl35133:
Having
```lambdapi
constant symbol Int: TYPE;
constant symbol + : Int → Int → Int;
notation + infix left 6;
constant notation - : Int → Int;
notation - prefix 7;
```
I'd like to have access to `(+)` and `(-)` without a λ-abstraction.
Another solution could be to define `plus` and then assign `notation + [...] infix left 6 ≔ plus;`

QGarchery:
If you are not worried about implicit arguments, you can use the `@` notation:

```
constant symbol Int: TYPE;
constant symbol + : Int → Int → Int;
notation + infix left 6;

symbol zero : Int;
symbol fortytwo : Int;
symbol iter (_n : Int) (_f : Int → Int) (_acc : Int) : Int;

symbol some_unknown_int ≔ @+ zero fortytwo;

symbol mul (m n : Int) ≔ iter n (@+ m) zero;
```

yannl35133:
I'd take it for now, but there are some irregularities which makes it unusable for me now: [see this example](https://github.com/Deducteam/lambdapi/issues/655#issuecomment-830160727)

fblanqui:
This was a duplicate of #359.

PR #661, opened 2021-05-04
fix parsing and printing of explicit symbols

### LINKED ISSUE #655: @id has the same notation properties as id
Considering `@id` doesn't have the same arity as `id`, I think it doesn't make sense for the notation properties to be conserved.
I think it should revert to "normal id" behaviour.

fblanqui:
fix #655 and #657

### DIFF
--- a/tests/OK/655b.lp
+++ b/tests/OK/655b.lp
@@ -3,9 +3,6 @@ symbol a:A;
 
 symbol +:A → A → A; notation + infix right 10;
 
-assert ⊢ a + + a a ≡ a + (a + a);
-assert ⊢ + a a + a ≡ (a + a) + a;
-
 assert ⊢ + : A → A → A;
 assert ⊢ + a : A → A;
 assert ⊢ @+ : A → A → A;
@@ -13,8 +10,6 @@ assert ⊢ @+ a : A → A;
 assert ⊢ (@+ a a) + a : A;
 
 symbol * {_:A} : A → A → A; notation * infix right 10;
-
-assert ⊢ (* a a) * a : A;
 
 assert ⊢ * : A → A → A;
 assert ⊢ * a : A → A;
--- a/tests/OK/655b.lp
+++ b/tests/OK/655b.lp
@@ -15,8 +15,14 @@ assert ⊢ (@+ a a) + a : A;
 symbol * {_:A} : A → A → A; notation * infix right 10;
 
 assert ⊢ (* a a) * a : A;
+
 assert ⊢ * : A → A → A;
 assert ⊢ * a : A → A;
 assert ⊢ @* : A → A → A → A;
 assert ⊢ @* a : A → A → A;
 assert ⊢ (@* _ a a) * a : A;
+
+symbol f : (A → A → A) → A;
+
+assert ⊢ f (+) : A;
+assert ⊢ f @+ : A;
--- a/tests/OK/655b.lp
+++ b/tests/OK/655b.lp
@@ -3,8 +3,9 @@ symbol a:A;
 
 symbol +:A → A → A; notation + infix right 10;
 
-assert ⊢ a + + a a : A;
-assert ⊢ (+ a a) + a : A;
+assert ⊢ a + + a a ≡ a + (a + a);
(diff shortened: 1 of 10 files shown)


fblanqui:
@gabrielhdt As you can see in tests/OK/655b.lp (in this PR), we can do funny things with infix operators already in the master branch (and this PR preserves this behavior) like:
```
symbol A:TYPE;
symbol a:A;
symbol +:A → A → A; notation + infix right 10;
// we can mix prefix and infix notations without using @
assert ⊢ a + + a a ≡ a + (a + a);
assert ⊢ + a a + a ≡ (a + a) + a;
```
I don't know if it is good or not, and I don't know how to avoid this.

fblanqui:
@yannl35133 Your comments concerning #655 and #657 are welcome. See the files tests/KO/655.lp, tests/KO/655b.lp and tests/OK/655.lp (in this PR). Does this solve your problems?

tests/OK/655b.lp
```diff
@@ -0,0 +1,22 @@
+symbol A:TYPE;
+symbol a:A;
+
+symbol +:A → A → A; notation + infix right 10;
+
+assert ⊢ a + + a a ≡ a + (a + a);
+assert ⊢ + a a + a ≡ (a + a) + a;
```
yannl35133, line 7: I don't understand how it works, nor why we want such behavior.
In my mind, notation and underlying definition should be separated, this seems really confusing.

yannl35133:
As I commented, I am not certain of this behavior for solving #657 but #655 should be solved

tests/OK/655b.lp
gabrielhdt: I agree that it's confusing, probably we want to raise an error when two infix operators follow each other.

yannl35133: In each of the example, it seems that one + is infix and the other is treated as a regular function.
I'd argue a symbol shouldn't be able to be treated as both infix and regular

fblanqui:
The above remark about the fact that we can mix prefix and infix notations is not related to your problem.
In #655 , you want `int @id int` to fail and `@id2 int int` to work. This PR fixes this.
In #657, you want something like `(+)` in OCaml. This PR should fix this too if you use @+ or (+).

tests/OK/655b.lp
gabrielhdt: > ```
> assert ⊢ a + + a a ≡ a + (a + a);
> assert ⊢ + a a + a ≡ (a + a) + a;
> ```

As we can see, the behaviour is the following: when a binary operator is called at the beginning of an expression, or after another binary operator, it is not considered as a binary operator. It looks surprising and _ad-hoc_, but in fact in these positions, operators are called without left context, so it's not a bug. Whether it's a desirable feature is another question.

It may be more harmful than useful though.

yannl35133:
Wait, `(+)` works?
If so, this should be the preferred way to denote the function, as `@+` also changes the behaviour with implicit arguments which we may not want.

In any case, `+ a a` should not be accepted and if not explicitly rejected, at least not endorsed in the tests.

tests/OK/655b.lp
fblanqui: An application is a sequence of terms. When we have a sequence of terms, LP reorders its arguments according to the symbols and their priorities that are in these arguments, using the Pratt algorithm (see ref given by Gabriel).
@gabrielhdt should be able to explain us why we get this parsing, and whether it is possible to fail instead.

gabrielhdt: It is indeed possible to fail. I'll update the parsing library.

fblanqui:
Yes, `(+)` works. See tests/OK/655b.lp. I added:
```
symbol f : (A → A → A) → A;

assert ⊢ f (+) : A;
assert ⊢ f @+ : A;

assert ⊢ f (*) : A;
assert ⊢ f (@* _) : A;
```
So, if you are happy with these changes, I will merge this PR. The problem with Pratt parsing can be solved later if needed.

tests/OK/655b.lp
fblanqui: I created a separate issue https://github.com/Deducteam/lambdapi/issues/662 for this.

PR #664, opened 2021-05-05
Operators parsing documentation

gabrielhdt:
The updated documentation says that any priority level can be provided: it is false with pratter 1.1 but true with pratter 1.2, so the assertion will become true when the version of pratter is updated.

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -267,7 +267,7 @@ Examples of patterns are available in ``tests/OK/patterns.lp``.
 ``notation``
 ----------------
 
-The ``notation`` command is used to specify a notation for a symbol.
+The ``notation`` command allows to change the behaviour of the parser.
 
 When declared as notations, identifiers then must be used at correct places
 and as such cannot make valid terms on their own anymore.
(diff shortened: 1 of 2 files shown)


docs/commands.rst
```diff
@@ -280,9 +280,9 @@ priority levels ``6`` and ``7`` respectively.
 
 The modifier ``infix``, ``infix right`` and ``infix left`` can be used
 to specify whether the defined symbol is non-associative, associative to
-the right, or associative to the left. The priority levels are floating
-point numbers, hence a priority can (almost) always be inserted between
-two different levels.
+the right, or associative to the left.
+Priority levels are floating point numbers (positive or negative), hence a
```
fblanqui, line 284: The grammar does not allow negative floats.

docs/commands.rst
```diff
@@ -291,6 +291,18 @@ negation with some priority level.
 
    notation ¬ prefix 5;
 
+*Remarks:*
+
+* Prefix and infix operators share the same levels of priority, hence depending
+  on the binding power, ``-x + z`` may be parsed ``(-x) + z`` or ``-(x + z)``.
+
+* Regular application has a higher binding power than operators application:
+  ``- f x`` is always parsed ``- (f x)``, whatever binding power is set for
```
fblanqui, line 300: prefix operator application

docs/commands.rst
gabrielhdt: Should we change the grammar to allow negative floats? 0.0 is allowed in Pratter 1.2.

docs/commands.rst
gabrielhdt: Prefix and infix, in fact, we have `f x + g y` = `(f x) + (g y)` as well.

docs/commands.rst
fblanqui: Why not indeed.

fblanqui:
@gabrielhdt It would be nice to update this PR and fix the above comments, so that we can merge it. Accepting negative floats can be done in a separate PR if you want.

docs/commands.rst
gabrielhdt: I suggest we leave that for another pull request.

gabrielhdt:
Should be OK for merging now.

fblanqui:
@gabrielhdt The branch is not up to date and CI failed.

docs/commands.rst
fblanqui: Ok. Thanks to add an issue for the record.

fblanqui: The doc still needs to be fixed: positive or negative -> non-negative.

fblanqui:
@yannl35133 Thanks to check the new proposed doc.

fblanqui:
Are you sure? I don't think that github does merges automatically.

gabrielhdt:
The error that popped up in the CI appeared when I merged the branches: the same modification was done in master and here, but at different lines, so there was no conflict, but the final code raised a warning. So I assume that when there's no merge conflict, github merges on its own for the CI.

yannl35133 (approved):

fblanqui:
Thanks @gabrielhdt and @yannl35133 !

PR #666, opened 2021-05-05
[parser/scope.ml] Remove problematic Pratt parser short-circuit

yannl35133:


gabrielhdt:
The errors in the CI are caused by the new version of pratter which rejects binary operators without left context. We obtain such terms because the pratt parser is called in some cases twice on the same terms (in `scope`). Assume the term `x + y * z` is parsed, then pratt parsing will provide `+ x (* y z)` and if pratt parsing is called again on `* y z`, an error is raised (because infix `*` is not in infix position).

If I'm not mistaken, arguments of applications are pratt parsed twice, once in `scope`, and again in `add_impl`.

fblanqui:
This means that we cannot use this new version of pratter unless it is modified so as to return `+ x (y * z)` instead of `+ x (* y z)`. Alternatively, can't we simply replace Pratt parsing by searching for the symbol of highest priority in the sequence of terms and build the corresponding application node without changing the arguments?

fblanqui:
I realize also that Pratt parsing maybe incompatible with explicit arguments `{ ... }`... What would we get if the user writes for instance `x = {N} y`?

gabrielhdt:
> I realize also that Pratt parsing maybe incompatible with explicit arguments `{ ... }`... What would we get if the user writes for instance `x = {N} y`?

I don't think we want infix operators to have more than two arguments, it's rather confusing.

> This means that we cannot use this new version of pratter unless it is modified so as to return + x (y * z) instead of + x (* y z).

I think we can use the new version, provided that we ensure that a term is pratt-parsed at most once.

fblanqui:
But this doesn't seem that easy as we don't know how deep the terms obtained after Pratt parsing have been reorganized. Perhaps, we need to introduce new p_term constructors. Meanwile, we need to stick to pratter 1.1 so that CI does not fail.

gabrielhdt:
> But this doesn't seem that easy as we don't know how deep the terms obtained after Pratt parsing have been reorganized. Perhaps, we need to introduce new p_term constructors. Meanwile, we need to stick to pratter 1.1 so that CI does not fail.

All terms reachable through applications ought to be processed by the pratt parser right? Menhir parsing produces exclusively terms of the form `P_Appl(P_Appl(....)...)`, so it should be feasible.

PR #668, opened 2021-05-12
Don't colourise newlines

yannl35133:
When a newline is colourised, especially if it is the last one before the prompt returns, the cursor keeps the colour on my console (KDE).

Not colouring the newlines shouldn't change anything else


fblanqui:
Thanks @yannl35133 !

PR #672, opened 2021-05-14
Record metavariable creation and instantiation during scoping, type inference and checking, and unification

### LINKED ISSUE #649: Performances with linear propositional problem
Consider the propositional problem with `n` boolean variables `p_1`, ..., `p_n` such that we have to prove `p_n` from `p_1`, `p_1 -> p_2`, ..., `p_{n-1} -> p_n`.

This problem grows linearly in size with `n` and also has an obvious linear solution.

The files are given with the `.txt` extension (so that I can upload them), to test it yourself you should rename them with the `.lp` extension. Encoding this problem into Lambdapi, using the first order logic encoding found in `tests/OK/firstOrder.dk`, we get:
* with 20 variables, the file [fo_20.txt](https://github.com/Deducteam/lambdapi/files/6357132/fo_20.txt) is checked in about 0.3 second on my machine
* with 27 variables, the file [fo_27.txt](https://github.com/Deducteam/lambdapi/files/6357135/fo_27.txt) is checked in about 17 seconds

fblanqui:
fix #649
fo_20.lp now takes 0,142s
fo_27.lp now takes 0,270s
fo_38.lp now takes 0,462s
fo_45.lp now takes 0,724s


tests/OK/bool.lp
```diff
@@ -32,6 +32,7 @@ with bool_and _     false ↪ false;
 
 // More connectives.
 
+//debug +huit;
```
gabrielhdt: You may want to remove this line.

tests/utils.ml
```diff
@@ -57,5 +57,5 @@ let _ =
     [ "hrs", [test_case "bool" `Quick test_hrs]
     ; "xtc", [test_case "bool" `Quick test_xtc]
     ; "dtree", [test_case "bool" `Quick test_dtree]
-    ; "dtree_ghost", [test_case "bool" `Quick test_dtree_ghost]
+                 (*; "dtree_ghost", [test_case "bool" `Quick test_dtree_ghost]*)
```
gabrielhdt: What happens with this test?

gabrielhdt:
Since meta variables are recorded, we may store their instantiation and types in the `problem` structure to simplify the `meta` data type so that only the key is recorded in terms. It may slightly improve performance.

fblanqui:
This is something to test perhaps, but in another PR.

src/core/term.ml
```diff
@@ -309,66 +296,30 @@ end
...
-
+  let compare m1 m2 = m2.meta_key - m1.meta_key
 end
 
-(** Sets and maps of metavariables. *)
 module MetaSet = Set.Make(Meta)
 module MetaMap = Map.Make(Meta)
 
+(** Representation of unification problems. *)
+type problem =
```
gabrielhdt, line 309: What would you think of abstracting this datatype, since references may well be error-prone?

fblanqui: What do you mean? Enforcing the use of update functions? Not sure it improves things very much. The `<-` are easy to track.

gabrielhdt: Indeed it's what I had in mind. Alright.

src/core/env.ml
```diff
@@ -78,15 +78,17 @@ let to_ctxt : env -> ctxt =
   List.map
     (fun (_,(v,a,t)) -> (v, Bindlib.unbox a, Option.map Bindlib.unbox t))
 
-(** [match_prod c t f] returns [f a b] if [t] matches [Prod(a,b)] possibly
-   after reduction. *)
-let match_prod : ctxt -> term -> (term -> tbinder -> 'a) -> 'a = fun c t f ->
+(** [match_prod s c t f] returns [f a b] if [t] matches [Prod(a,b)] possibly
+   after reduction.
+@raise [Invalid_argument s] if [t] is not a product. *)
+let match_prod : string -> ctxt -> term -> (term -> tbinder -> 'a) -> 'a =
```
gabrielhdt, line 84: What is the first argument used for?

src/core/eval.mli
```diff
@@ -0,0 +1,77 @@
...
+   [c]. It unfolds variables defined in the context [c]. *)
+val snf : ctxt -> term -> term
+
+(** [whnf t] computes a weak head-normal form of the term [t] wrt
+   beta-reduction, user-defined rewrite rules and variables defined in the
+   context [c]. *)
+val whnf : ctxt -> term -> term
+
+(** [simplify t] reduces simple redexes of [t]. *)
+val simplify : term -> term
```
gabrielhdt, line 41: What is a simple redex?

src/core/infer.mli
```diff
@@ -0,0 +1,19 @@
+open Term
+
+(** Given a meta [m] of type [Πx1:a1,..,Πxn:an,b], [set_to_prod p m] sets [m]
+   to a product term of the form [Πy:m1[x1;..;xn],m2[x1;..;xn;y]] with fresh
+   metavariables [m1] and [m2], and updates the metas of the problem [p]. *)
+val set_to_prod : problem -> meta -> unit
+
+(** [infer_noexn p ctx t] returns [None] if the type of [t] in context [ctx]
+   cannot be infered, or [Some a] where [a] is some type of [t] in the context
```
gabrielhdt, line 9: typo: inferred

src/core/libMeta.ml
```diff
@@ -0,0 +1,111 @@
...
+  (* Convert the context into a map. *)
+  let vm =
+    let f vm (x,_,v) =
+      match v with
+      | Some v -> VarMap.add x v vm
+      | None -> vm
+    in
+    Stdlib.ref (List.fold_left f VarMap.empty c)
+  in
+  (*TODO: make iter tail recursive. *)
```
gabrielhdt, line 84: Why do you want to make it tail recursive?

src/core/libMeta.ml
```diff
@@ -0,0 +1,111 @@
...
+let reset_meta_counter () = Stdlib.(meta_counter := -1)
+
+(** [fresh p ?name a n] creates a fresh metavariable of type [a] and arity [n]
+   with the optional name [name], and adds it to [p]. *)
+let fresh : problem -> ?name:string -> term -> int -> meta =
+  fun p ?name a n ->
+  let m = {meta_key = Stdlib.(incr meta_counter; !meta_counter);
+           meta_name = name; meta_type = ref a; meta_arity = n;
+           meta_value = ref None } in
+  p.metas <- MetaSet.add m p.metas; m
```
gabrielhdt, line 18: To avoid exposing the counter, the following can be written
```ocaml
let ((reset_meta_counter: unit -> unit), (fresh: problem -> ?name:string -> term -> int -> meta)) =
  let counter = Stdlib.ref (-1) in
  (fun () -> Stdlib.(counter := -1), fun p ?name a n -> ...)
```

src/core/print.ml
```diff
@@ -92,13 +92,13 @@ let pp_sym : sym pp = fun ppf s ->
   let n = s.sym_name in
   if StrMap.mem n !sig_state.in_scope then
     (* For printing Unif_rule.cons unescaped. *)
-    if s == Unif_rule.cons then out ppf "%s" n else pp_uid ppf n
+    (if s == Unif_rule.cons then out ppf "%s" n else pp_uid ppf n)
```
gabrielhdt, line 95: Are the parentheses needed?

src/core/unif.mli
```diff
@@ -0,0 +1,9 @@
+open Term
+
+(** [solve_noexn ~type_check p] tries to simplify the constraints of [p]. It
+   returns [false] if it finds a constraint that cannot be
+   satisfied. Otherwise, [p.to_solve] is empty but [p.unsolved] may still
+   contain constraints that could not be simplified. Metavariable
+   instantiations are type-checked only if the optional argument [~type_check]
+   is [true] (default). *)
```
gabrielhdt, line 8: I'm picky, but I think just writing `(default)` might be confusing when reading the comment for the first time, I'd be more explicit to say that the argument `~type_check` defaults to `true`.

src/core/env.ml
gabrielhdt: Isn't it only useful for debugging?

src/handle/proof.ml
```diff
@@ -42,13 +38,16 @@ module Goal = struct
...
+        let t, n = Ctxt.to_prod c mk_Type in
+        fst (Env.of_prod_nth __LOC__ c n t)
     | Typ gt -> gt.goal_hyps
 
   (** [of_meta m] creates a goal from the meta [m]. *)
   let of_meta : meta -> goal = fun m ->
     let goal_hyps, goal_type =
-      Env.of_prod_nth [] m.meta_arity !(m.meta_type) in
+      (*let s = Format.asprintf "%s, of_meta %a(%d):%a" __LOC__
+                pp_meta m m.meta_arity pp_term !(m.meta_type) in*)
```
gabrielhdt, line 49: You may want to remove these lines

src/core/env.ml
fblanqui: Right, it helps locate where the invalid argument is introduced.

src/core/libMeta.ml
fblanqui: Well, to make it more efficient.

src/core/eval.mli
fblanqui: Thanks. Will be fixed in next commit.

src/core/infer.mli
fblanqui: Thanks. Will be fixed in next commit.

src/core/libMeta.ml
fblanqui: I prefer to add a libMeta.mli file, but this is not urgent and could be made in another PR.

src/core/env.ml
gabrielhdt: IMO it's better to let people instrument the code when debugging, for instance using `Printexc` to obtain stack traces. I'm afraid adding extra arguments for debugging may be troubling for people discovering the  code.

src/core/libMeta.ml
gabrielhdt: I'm not sure it's always more efficient: in some cases it's better to keep not tail recursive functions when the data structures are small. But that's not important for now.

src/handle/proof.ml
fblanqui: No, it could be useful for debug later.

fblanqui:
Thanks for your comments @gabrielhdt !

src/core/env.ml
fblanqui: I don't know that. Could you please give more details?

src/core/libMeta.ml
fblanqui: I don't know either but this is something to test. This could be a way also to handle the different kinds of checking problems in a single function. We'll see this later.

src/core/env.ml
gabrielhdt: Looking at <https://ocaml.org/api/Printexc.html>, you can use `Printexc.record_backtrace true` to (obviously) record backtraces. Consider the following input (with ocaml toplevel)
```ocaml
# let f () = raise Not_found;;
val f : unit -> 'a = <fun>
# let g () = f ();;
val g : unit -> 'a = <fun>
# g ();;
Exception: Not_found.
```
then with backtrace recording:
```ocaml
# Printexc.record_backtrace true;;
- : unit = ()
# g ();;
Exception: Not_found.
Raised at f in file "//toplevel//", line 1, characters 11-26
Called from Toploop.load_lambda in file "toplevel/toploop.ml", line 212, characters 17-27
```

fblanqui: That's interesting but note that it works for ocaml >= 3.11.0 only.

fblanqui: However this is already used in ls_lsp.ml. So why CI does not fail?

gabrielhdt: Ocaml 3.11 is super old, we're at 4.11 now

fblanqui: Right, my mistake : - )

fblanqui: Fixed in last commit.

PR #673, opened 2021-05-17
Flush output on warning

gabrielhdt:
Fixes an issue where there would be no newline at the end of warning (can be reproduced running `dune exec -- lambdapi tests/OK/292.lp`).


fblanqui:
Thanks @gabrielhdt

PR #674, opened 2021-05-18
Pratter 1.2 and wrapped operators with explicit implicit arguments

yannl35133:
A bad solution to allow Pratter 1.2 and wrapped operators with explicit implicit arguments


gabrielhdt (changes requested):
Looks good to me. Could you update the "syntax" section of the documentation please?

gabrielhdt (dismissed):
Nice, thanks.

docs/commands.rst
```diff
@@ -301,6 +301,11 @@ negation with some priority level.
    type λ p, `∀ x, p; // quantifiers can be written as such
    type λ p, `f x, p; // works as well if f is any symbol
 
+When declared as notations, identifiers then must be used at correct places
+and as such cannot make valid terms on thier own anymore.
```
fblanqui, line 305: typo: thier

fblanqui: all this should come at line 271

src/parsing/pratt.ml
```diff
@@ -54,13 +54,17 @@ end = struct
...
-    let strm = Stream.of_list (h::args) in
+    let h, args = Syntax.p_get_args t in
+    let strm = Stream.of_list (h :: args) in
     let module Parse = Pratter.Make(Pratt_terms) in
-    try Parse.expression (st, env) strm with
-    | Parse.OpConflict(t, u) ->
+    try
+      Parse.expression (st, env) strm
+    with
+    | Parse.OpConflict (t, u) ->
```
fblanqui, line 63: thanks to minimize the diff here by not changing the previous indentation

src/parsing/scope.ml
```diff
@@ -125,13 +125,17 @@ let fresh_patt : lhs_data -> string option -> tbox array -> tbox =
    state [ss] is used to convert identifiers into symbols according to
    [find_qid]. *)
 let rec scope : mode -> sig_state -> env -> p_term -> tbox =
+  fun md ss env t -> scope_parsed md ss env (Pratt.parse ss env t)
+(** [scope_parsed md ss env t] turns a parser-level, Pratt-parsed term [t]
+   into an actual term. The variables of the environment [env] may appear
```
fblanqui, line 130: you can remove "The variables ..."

docs/syntax.bnf
```diff
@@ -14,6 +14,7 @@
 <path> ::= <qid>
 
 <term_id> ::= <id>
+            | (<id>)
```
fblanqui, line 17: This file is automatically generated by `make doc` from the Menhir grammar. So it should not be modified by hand. So thanks to revert your changes.

fblanqui:
Thanks @yannl35133 for your PR. A few things need to be fixed before merging.

docs/syntax.bnf
gabrielhdt: We ought to add a warning at the top of the file.

gabrielhdt: With something like this
```diff
diff --git a/docs/Makefile b/docs/Makefile
index 792f3cdb..b24e87b0 100644
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -21,7 +21,9 @@ help:

 .PHONY: syntax.bnf
 syntax.bnf:
-       echo "<qid> ::= [<uid>.]+<uid>\n" > $@
+       echo "; DO NOT EDIT THIS FILE MANUALLY" > $@
+       echo "; It is generated automatically with 'make syntax.bnf'." >> $@
+       echo "<qid> ::= [<uid>.]+<uid>\n" >> $@
        echo "<switch> ::= on\n           | off\n" >> $@
        echo "<assoc> ::= left\n          | right\n" >> $@
        $(BNF_GEN) ../src/parsing/lpParser.mly | sed \
@@ -116,7 +118,10 @@ syntax.bnf:

 .PHONY: syntax_dedukti.bnf
 syntax_dedukti.bnf:
-       echo "<qid> ::= [<uid>.]+<uid>\n" > $@
+       echo "; DO NOT EDIT THIS FILE MANUALLY" > $@
+       echo "; It is generated automatically with \
+'make syntax_dedukti.bnf'." >> $@
+       echo "<qid> ::= [<uid>.]+<uid>\n" >> $@
        echo "<assert> ::= #ASSERT\n           | #ASSERTNOT\n" >> $@
        $(BNF_GEN) ../src/parsing/dkParser.mly | sed \
        -e 's/ASSERT/<assert>/g' \
```

yannl35133: On my machine, echo was printing the \n literally; on the internet, people are mostly against using echo and backslash-escaped characters, so I replaced them with the more universal printf

docs/commands.rst
```diff
@@ -269,6 +269,12 @@ Examples of patterns are available in ``tests/OK/patterns.lp``.
 
 The ``notation`` command is used to specify a notation for a symbol.
 
+When declared as notations, identifiers then must be used at correct places
+and as such cannot make valid terms on their own anymore.
+To reaccess the value of the identifier without the notation properties,
+wrap it in parentheses.
+
```
fblanqui, line 276: @yannl35133 Would you mind to add an example on line 286?
```
// + is not a valid term since, as it is declared as infix,
// the system expects + to only appear in expressions of the form x+y
// you can get around this by using (+) instead
```

docs/terms.rst
```diff
@@ -23,6 +23,9 @@ An identifier can be:
 
 Remark: for any regular identifier ``i``, ``{|i|}`` is equivalent to ``i``.
 
+* an escaped notation identifier, that is, an identifier previously declared as notation,
+  wrapped in parentheses (this allows access to the notationless value of the identifier)
+
```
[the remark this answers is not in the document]
yannl35133: Even though the system understands it as a term, I think the user should understand it as an identifier with a decorator similar to `@`.
It is only understood as a term because if it were as an identifier there would be conflicts, but the behaviour pretty much tries to reduce it as just an identifier

fblanqui:
Thanks @yannl35133 to resolve the conflicts so that we can merge.

yannl35133:
Did I miss a conflict ?

fblanqui:
Thanks @yannl35133 !

PR #687, opened 2021-05-25
Changing problem datatype

gabrielhdt:
This pull request is preparatory for #563 (and #686). It changes the `problem` datatype to use `Timed` ref, because #563 needs to backtrack on failed unification, and hence needs to be able to restore problems.


fblanqui:
Ok. I can merge as soon as conflicts are resolved.

fblanqui:
in order to be able to backtrack on failed unification if needed

Issue #693, opened 2021-05-26
The assume tactic does too much reduction


QGarchery:
Very recent commits (last few days) seem to change how rewriting is handled. This cause various issues, but I believe they come from the same one. I think it is because some rewriting is done at some unexpected place (see `test_assume` below).

```
require open Stdlib.Set Stdlib.Prop Stdlib.Eq Stdlib.FOL Stdlib.Nat;

symbol fold_arrow {p1 p2 : Prop} : π (p1 ⇒ p2) → π p1 → π p2
≔ λ p1_p2 p1, p1_p2 p1;

symbol test_assume (a b : ℕ) f : π (a = b ⇒ f a ⇒ f b)
≔ begin
  assume a b f;

  assume eab;
  // assume rewrites π (f a ⇒ f b) into π (f a) → π (f b)

  // This should fix the issue but does nothing instead:
  // refine fold_arrow _;

  rewrite eab;
  // Goal is not of the correct form, how to fix ?

end;


symbol test_rew_lambda f : π (`∀ a, `∀ b, f (a + b) ⇒ f b)
≔ begin
  // Does not rewrite under binders
  rewrite add_com;
end;

// Following declaration does not type check
symbol test_type_check (a b : ℕ) f : π (f (a + b) ⇒ f (a + b));
```

fblanqui:
Thanks @QGarchery for your issues. These issues are not related and should be in separate issues but that's fine.

Issue 1 has been introduced in #681. Before, unification was in two steps: by first trying to unify without normalizing wrt user-defined rules and, in case of failure, trying to unify by normalizing wrt user-defined rules, to avoid normalization whenever possible. In #681, I removed the first step for simplifying the code. Unfortunately, our tests did not cover this case. I am going to revert this change.

Issue 2. You first need to do `assume f`. But, then this fails with `No subterm of [`∀ a, `∀ b, f (a + b) ⇒ f b] matches [x + y]`. Indeed, it is currently not possible to rewrite under binders. The PR #691 allowed to rewrite a closed term inside a binder. Here it is about rewriting a term containing bound variables. This requires some investigation. This issue may not be solved soon.

Issue 3. By default, the system assumes that implicit types are dependent. Hence, f gets the type `Π y:?1[a;b], ?2[a;b;y]` and, for the symbol type to be well typed, we need to solve the following constraint:
```
[...] Cannot solve ?2[a;b;b + a] ≡ Prop.
```
[...]

PR #696, opened 2021-05-29
New typechecker, refinement, placeholders, coercions

gabrielhdt:
This PR is equivalent to #563 but commits are separated in a more intelligible way:

1. Implementation of the basis of the new typechecker
2. Modifying signatures to make the type checker bidirectional (allowed to changed the type checked term), and subsequently modify calls to the typechecker
3. Introduce placeholders and generate meta-variables during type checking: placeholders of the form "?" are replaced by meta-variables during typechecking (or fully instantiated terms).

## Changelog

Introduce a new type checker which is able to refine (aka modify) terms. The algorithm is taken from <https://arxiv.org/abs/1202.4905>.

### Added

- Tems may now be _placeholders_. Placeholders are holes in the concrete syntax. They are refined into metavariables. *Placeholders cannot appear nonlinearly in terms*. From [A Bidirectional Refinement Algorithm...](https://arxiv.org/abs/1202.4905), p. 31,
  > Non linear placeholders are not allowed since two occurrences could be in contexts that bind different set of variables and instantiation with terms that live in one context would make no sense in the other one.

### Removed

- Unused variable warning: whether a variable is used or not cannot be decided while scoping anymore since placeholders that do not depend on variables may be refined later into metavariables that may depend on them.
- Placeholders do not depend on variables, hence the syntax `?M.[x;y]` is obsolete, only `?M` is valid

### Changed

- Naming metavariable is now useless since they must be linear.
- Because placeholders are simple holes, the term `_ → _` is scoped into a full dependent product `Π x: M, N` where `N` is a metavariable that depend on `x` (see file `tests/OK/767.lp`)
- Type checking takes longer because of refinement (not only the type must be destructured and rebuilt),
  |         | master | refiner |
  |---------|--------|---------|
  | holide  | 7:0    | 11:33   |
  | iprover | 5:58   | 6:50    |
- scoping term is more straightforward (no metavariable fetching), thus `scope_term` simplified

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,13 +17,40 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   two new builtins ("ex" and "all"). Codewise, it requires a new
   translation from encoded types to Why3 types.
 
+- Tems may be _placeholders_. Placeholders are holes in the
+  concrete syntax. They are refined into metavariables. *Placeholders
+  cannot appear nonlinearly in terms*. From [A Bidirectional Refinement
+  Algorithm...](https://arxiv.org/abs/1202.4905), p. 31,
+  > Non linear placeholders are not allowed since two occurrences could be
+  in contexts that bind different set of variables and instantiation with
+  terms that live in one context would make no sense in the other one.
+
 ### Changed
 
 - Moved the files tool/hrs.ml and tool/xtc.ml into the new export/ directory.
 
+- Because placeholders are simple holes, the term `_ → _` is scoped
+  into a full dependent product `Π x: M, N` where `N` is a metavariable
+  that depend on `x` (see file `tests/OK/767.lp`)
+
+- Type checking is slower following #696 because of refinement (not only
+  the type but also the term must be destructured and rebuilt),
+  |         | master | refiner |
+  |---------|--------|---------|
+  | holide  | 7:0    | 11:33   |
+  | iprover | 5:58   | 6:50    |
+
 ### Removed
 
 - The command beautify superseded by the new command export.
+
+- Unused variable warning: whether a variable is used or not cannot be
+  decided while scoping (following #696) since placeholders that do not
+  depend on variables may be refined later into metavariables that may
+  depend on them.
+
+- Metavariables cannot be referenced by their name anymore, hence the
+  syntax `?M.[x;y]` is obsolete, but `?0.[x;y]` isn't
 
 ## 2.0.0 (2021-12-15)
 
--- a/doc/terms.rst
+++ b/doc/terms.rst
@@ -49,7 +49,6 @@ A user-defined term can be either:
      current file or in some previously open module, possibly prefixed by ``@``
      to disallow implicit arguments
    * a bound variable
-   * a metavariable or goal when prefixed by ``?``
 
 * an anonymous function ``λ(x:A) y z,t`` mapping ``x``, ``y`` and ``z``
   (of type ``A`` for ``x``) to ``t``
@@ -64,9 +63,9 @@ A user-defined term can be either:
 * an application written by space-separated juxtaposition, except for
   symbol identifiers declared as infix (e.g. ``x + y``)
 
-* a meta-variable application ``?M.[t;u;v]``. ``?M`` alone, without arguments
-  between square brackets, is a shorthand for ``?M.[x1;..;xn]`` where
-  ``x1;..;xn`` are all the variables of the context.
+* a hole ``_`` to be filled automatically.
+
+* a metavariable ``?0.[x;y]`` that has been generated previously.
 
 * a pattern-variable application ``$P.[x;y]`` (in rules only). ``$P``
   alone, without arguments between square brackets, is a shorthand for
--- a/src/core/ctxt.ml
+++ b/src/core/ctxt.ml
@@ -33,6 +33,20 @@ let to_prod : ctxt -> term -> term * int = fun ctx t ->
   let t, c = List.fold_left f (lift t, 0) ctx in
   Bindlib.unbox t, c
 
+(** [to_prod_box bctx t] is similar to [to_prod bctx t] but operates on boxed
+    contexts and terms. *)
+let to_prod_box : bctxt -> tbox -> tbox * int = fun bctx t ->
+  let f (t, c) (x, a) =
+    let b = Bindlib.bind_var x t in
+    (_Prod a b, c + 1)
+  in
+  List.fold_left f (t, 0) bctx
+
+(** [box_context ctx] lifts context [ctx] to a boxed context. *)
+let box_context : ctxt -> bctxt =
+  List.filter_map
+    (fun (x, t, u) -> if u = None then Some (x, lift t) else None)
+
 (** [to_abst ctx t] builds a sequence of abstractions over the context [ctx],
     in the term [t]. *)
 let to_abst : ctxt -> term -> term = fun ctx t ->
--- a/src/core/ctxt.ml
+++ b/src/core/ctxt.ml
@@ -1,17 +1,8 @@
 (** Typing context. *)
 
 open Term
+open Lplib
 open Timed
-
-(** [unbind ctx a def b] returns a triple [(x,t,new_ctx)] such that [(x,t)] is
-   an unbinding of [b] (in the sense of [Bindlib.unbind]) and [new_ctx] is an
-   extension of the context [ctx] with the declaration that [x] has type [a]
-   (only if [x] occurs in [t]). If [def] is of the form [Some(u)], the context
-   also registers the term [u] as the definition of variable [x]. *)
-let unbind : ctxt -> term -> term option -> tbinder -> tvar * term * ctxt =
-  fun ctx a def b ->
-  let (x, t) = Bindlib.unbind b in
-  (x, t, if Bindlib.binder_occur b then (x, a, def) :: ctx else ctx)
 
 (** [type_of x ctx] returns the type of [x] in the context [ctx] when it
     appears in it, and
--- a/src/core/infer.ml
+++ b/src/core/infer.ml
@@ -9,15 +9,24 @@ open Timed
 let log = Logger.make 'i' "infr" "type inference/checking"
 let log = log.pp
 
+(* Optimised context *)
+type octxt = ctxt * bctxt
+let boxed = snd
+let classic = fst
+let extend (cctx, bctx) v ?def ty =
+  ((v, ty, def) :: cctx, if def <> None then bctx else (v, lift ty) :: bctx)
+
+let unbox = Bindlib.unbox
+
 (** Exception that may be raised by type inference. *)
 exception NotTypable
 
 (** [unif pb c a b] solves the unification problem [c ⊢ a ≡ b]. Current
     implementation collects constraints in {!val:constraints} then solves
     them at the end of type checking. *)
-let unif : problem -> ctxt -> term -> term -> unit =
+let unif : problem -> octxt -> term -> term -> unit =
  fun pb c a b ->
- if not (Eval.pure_eq_modulo c a b) then
+ if not (Eval.pure_eq_modulo (classic c) a b) then
  (* NOTE: eq_modulo is used because the unification algorithm counts on
     the fact that no constraint is added in some cases (see test
     "245b.lp"). We may however want to reduce the number of calls to
@@ -25,8 +34,8 @@ let unif : problem -> ctxt -> term -> term -> unit =
    begin
      if Logger.log_enabled () then
        log (Color.yel "add constraint %a") Print.pp_constr
-         (c, a, b);
-     pb := {!pb with to_solve = (c, a, b) :: !pb.to_solve}
+         (classic c, a, b);
+     pb := {!pb with to_solve = (classic c, a, b) :: !pb.to_solve}
    end
 
 (** {1 Handling coercions} *)
@@ -45,7 +54,7 @@ let coerce pb c t a b = unif pb c a b; (t, false)
 (** [type_enforce pb c a] returns a tuple [(a',s)] where [a'] is refined
     term [a] and [s] is a sort (Type or Kind) such that [a'] is of type
     [s]. *)
-let rec type_enforce : problem -> ctxt -> term -> term * term * bool =
+let rec type_enforce : problem -> octxt -> term -> term * term * bool =
  fun pb c a ->
   if Logger.log_enabled () then log "Type enforce [%a]" Print.pp_term a;
   let a, s, cui = infer pb c a in
@@ -63,21 +72,22 @@ let rec type_enforce : problem -> ctxt -> term -> term * term * bool =
 
 (** [force pb c t a] returns a term [t'] such that [t'] has type [a],
     and [t'] is the refinement of [t]. *)
-and force : problem -> ctxt -> term -> term -> term * bool =
+and force : problem -> octxt -> term -> term -> term * bool =
  fun pb c te ty ->
  if Logger.log_enabled () then
    log "Force [%a] of [%a]" Print.pp_term te Print.pp_term ty;
  match unfold te with
  | Plac true ->
      unif pb c ty mk_Type;
-     (LibMeta.make pb c mk_Type, true)
- | Plac false -> (LibMeta.make pb c ty, true)
+     (unbox (LibMeta.bmake pb (boxed c) _Type), true)
+ | Plac false ->
+     (unbox (LibMeta.bmake pb (boxed c) (lift ty)), true)
  | _ ->
      let (t, a, cui) = infer pb c te in
      let t, cu = coerce pb c t a ty in
      (t, cu || cui)
 
-and infer_aux : problem -> ctxt -> term -> term * term * bool =
+and infer_aux : problem -> octxt -> term -> term * term * bool =
  fun pb c t ->
   match unfold t with
   | Patt _ -> assert false
@@ -87,16 +97,16 @@ and infer_aux : problem -> ctxt -> term -> term * term * bool =
   | TRef _ -> assert false
   | Type -> (mk_Type, mk_Kind, false)
   | Vari x ->
-      let a = try Ctxt.type_of x c with Not_found -> assert false in
+      let a = try Ctxt.type_of x (classic c) with Not_found -> assert false in
       (t, a, false)
   | Symb s -> (t, !(s.sym_type), false)
   | Plac true ->
-      let m = LibMeta.make pb c mk_Type in
-      (m, mk_Type, true)
+      let m = LibMeta.bmake pb (boxed c) _Type in
+      (unbox m, mk_Type, true)
   | Plac false ->
-      let mt = LibMeta.make pb c mk_Type in
-      let m = LibMeta.make pb c mt in
-      (m, mt, true)
+      let mt = LibMeta.bmake pb (boxed c) _Type in
+      let m = LibMeta.bmake pb (boxed c) mt in
+      (unbox m, unbox mt, true)
   (* All metavariables inserted are typed. *)
   | (Meta (m, ts)) as t ->
       let cu = Stdlib.ref false in
@@ -124,7 +134,7 @@ and infer_aux : problem -> ctxt -> term -> term * term * bool =
       let t, cu_t = force pb c t t_ty in
       (* Unbind [u] and get new context with [x: t_ty ≔ t] *)
       let (x, u) = Bindlib.unbind u in
-      let c = (x, t_ty, Some t) :: c in
+      let c = extend c x ~def:t t_ty in
       (* Infer type of [u'] and refine it. *)
       let u, u_ty, cu_u = infer pb c u in
       ( match unfold u_ty with
@@ -148,7 +158,7 @@ and infer_aux : problem -> ctxt -> term -> term * term * bool =
       (* Domain must by of type Type (and not Kind) *)
       let dom, cu_dom = force pb c dom mk_Type in
       let (x, b) = Bindlib.unbind b in
-      let c = (x, dom, None) :: c in
+      let c = extend c x dom in
       let b, range, cu_b = infer pb c b in
       let range = Bindlib.(lift range |> bind_var x |> unbox) in
       let top_ty = mk_Prod (dom, range) in
@@ -164,7 +174,7 @@ and infer_aux : problem -> ctxt -> term -> term * term * bool =
       (* Domain must by of type Type (and not Kind) *)
       let dom, cu_dom = force pb c dom mk_Type in
       let (x, b) = Bindlib.unbind b in
-      let c = (x, dom, None) :: c in
+      let c = extend c x dom in
       let b, b_s, cu_b = type_enforce pb c b in
       let cu = cu_b || cu_dom in
       let top =
@@ -180,25 +190,30 @@ and infer_aux : problem -> ctxt -> term -> term * term * bool =
         let ty = Bindlib.subst range u and cu = cu_t || m in
         if cu then (mk_Appl (t, u), ty, cu) else (top, ty, cu)
       in
-      match Eval.whnf c t_ty with
+      match Eval.whnf (classic c) t_ty with
       | Prod (dom, range) ->
           if Logger.log_enabled () then log "Appl-prod arg [%a]" Print.pp_term u;
           let u, cu_u = force pb c u dom in
           return cu_u t u range
       | Meta (_, _) ->
           let u, u_ty, cu_u = infer pb c u in
-          let range = LibMeta.make_codomain pb c u_ty in
+          let range =
+            unbox (LibMeta.bmake_codomain pb (boxed c) (lift u_ty))
+          in
           unif pb c t_ty (mk_Prod (u_ty, range));
           return cu_u t u range
       | t_ty ->
-          let domain = LibMeta.make pb c mk_Type in
-          let range = LibMeta.make_codomain pb c domain in
+          let domain = LibMeta.bmake pb (boxed c) _Type in
+          let range = LibMeta.bmake_codomain pb (boxed c) domain in
+          let domain = unbox domain
+          and range = unbox range in
           let t, cu_t' = coerce pb c t t_ty (mk_Prod (domain, range)) in
           if Logger.log_enabled () then log "Appl-default arg [%a]" Print.pp_term u;
           let u, cu_u = force pb c u domain in
           return (cu_t' || cu_u) t u range )
 
-and infer : problem -> ctxt -> term -> term * term * bool = fun pb c t ->
+
+and infer : problem -> octxt -> term -> term * term * bool = fun pb c t ->
   if Logger.log_enabled () then log "Infer [%a]" Print.pp_term t;
   let t, t_ty, cu = infer_aux pb c t in
   if Logger.log_enabled () then
@@ -217,11 +232,11 @@ and infer : problem -> ctxt -> term -> term * term * bool = fun pb c t ->
     the call to [f] and [cs] is the list of constraints gathered by
     [f]. Function [f] may raise [NotTypable], in which case [None] is
     returned. *)
-let noexn : (problem -> ctxt -> 'a -> 'b) -> problem -> ctxt -> 'a ->
+let noexn : (problem -> octxt -> 'a -> 'b) -> problem -> ctxt -> 'a ->
   'b option =
   fun f pb c args ->
   try
-    Some (f pb c args)
+    Some (f pb (c, Ctxt.box_context c) args)
   with NotTypable -> None
 
 let infer_noexn pb c t : (term * term) option =
--- a/src/core/infer.ml
+++ b/src/core/infer.ml
@@ -32,17 +32,23 @@ let unif : problem -> ctxt -> term -> term -> unit =
 (** {1 Handling coercions} *)
 
 (* Simply unify types, no coercions yet. *)
-let coerce pb c t a b = unif pb c a b; t
+let coerce pb c t a b = unif pb c a b; (t, false)
 
 (** {1 Other rules} *)
+
+(** NOTE: functions {!val:type_enforce}, {!val:force} and {!val:infer}
+    return a boolean which is true iff the typechecked term has been
+    modified. It allows to bypass reconstruction of some Bindlib terms (which
+    call [lift |> bind_var x |> unbox]). It reduces the type checking time of
+    Holide by 21%. *)
 
 (** [type_enforce pb c a] returns a tuple [(a',s)] where [a'] is refined
     term [a] and [s] is a sort (Type or Kind) such that [a'] is of type
     [s]. *)
-let rec type_enforce : problem -> ctxt -> term -> term * term =
+let rec type_enforce : problem -> ctxt -> term -> term * term * bool =
  fun pb c a ->
   if Logger.log_enabled () then log "Type enforce [%a]" Print.pp_term a;
-  let a, s = infer pb c a in
+  let a, s, cui = infer pb c a in
   let sort =
     match unfold s with
     | Kind -> mk_Kind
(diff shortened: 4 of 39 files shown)


gabrielhdt:
Although the coercions are not yet implemented, the code can be integrated. It provides a new typechecking algorithm, and it adds placeholders so that meta variables are generated at type check time.

fblanqui:
Hello @gabrielhdt . There are a number of problems with this PR:
1) https://github.com/fblanqui/lib does not compile with it.
Remark: it's a good habit to check this library in addition to make tests. Perhaps we should integrate it in the CI.
2) Performances are degraded:
comparison between master and your PR on my machine:
make tests: 7.7s -> 9.2s = +20%
consider tests/OK/649_fo_27.lp and increase the number:
fo_45.lp: 0.643s -> 1.909s = +200%

src/core/ctxt.ml
```diff
@@ -1,17 +1,25 @@
...
-   also registers the term [u] as the definition of variable [x]. *)
+    an unbinding of [b] (in the sense of [Bindlib.unbind]) and [new_ctx] is an
+    extension of the context [ctx] with the declaration that [x] has type [a].
+    If [def] is of the form [Some(u)], the context also registers the term [u]
+    as the definition of variable [x]. *)
 let unbind : ctxt -> term -> term option -> tbinder -> tvar * term * ctxt =
   fun ctx a def b ->
   let (x, t) = Bindlib.unbind b in
-  (x, t, if Bindlib.binder_occur b then (x, a, def) :: ctx else ctx)
+  (x, t, (x, a, def) :: ctx)
```
fblanqui, line 15: Why do you want to extend the context if the variable is not used?
If you do that, then this function becomes useless and you can unfold it.

src/core/libMeta.ml
```diff
@@ -49,13 +50,14 @@ let of_name : string -> problem -> meta option = fun n p ->
...
+    name [?name] if provided in the context [ctx], and adds it to [p]. *)
+let make : problem -> ?name:string -> ctxt -> term -> term =
+  fun p ?name ctx a ->
   let a, k = Ctxt.to_prod ctx a in
-  let m = fresh p a k in
-  let get_var (x,_,_) = mk_Vari x in
-  mk_Meta(m, Array.of_list (List.rev_map get_var ctx))
+  let m = fresh ?name p a k in
+  let get_var (x,_,d) = if d = None then Some (mk_Vari x) else None in
+  mk_Meta(m, Array.of_list (List.(filter_map get_var ctx |> rev)))
```
fblanqui, line 60: Use List.filter_rev_map.

src/core/term.ml
```diff
@@ -61,7 +61,8 @@ type term =
   (** Pattern variable application (only used in rewriting rules LHS). *)
   | TEnv of term_env * term array
   (** Term environment (only used in rewriting rules RHS). *)
-  | Wild (** Wildcard (only used for surface matching, never in a LHS). *)
+  | Wild
+  | Plac of bool * string option
```
fblanqui, line 65: Meaning of the arguments?

gabrielhdt: I moved the comments to the interface file. Regarding the position of the comments, we must choose whether we duplicate them (and increase the risk of forgetting to update one of the copies), or we write them only in the interface files.

src/core/term.ml
```diff
@@ -232,11 +233,12 @@ let rec pp_term : term pp = let out = Format.fprintf in fun ppf t ->
   | Meta(m,ts) -> out ppf "?%a[%a]" pp_meta m pp_terms ts
   | Patt _ -> out ppf "<Patt>"
   | TEnv _ -> out ppf "<TEnv>"
-  | Wild -> out ppf "_"
+  | Wild   -> out ppf "_"
   | TRef _ -> out ppf "<TRef>"
   | LLet(a,t,u) ->
       let x, u = Bindlib.unbind u in
       out ppf "let %a : %a ≔ %a in %a" pp_var x pp_term a pp_term t pp_term u
+  | Plac _ -> out ppf "_"
```
fblanqui, line 241: Should be improved following #697 (see case of Patt). Term.pp_term is for debug and should print every thing useful for debug.

src/core/term.ml
```diff
@@ -402,9 +404,10 @@ let cmp : term cmp =
     | Meta _ -> 7
     | Patt _ -> 8
     | TEnv _ -> 9
-    | Wild -> 10
+    | Wild   -> 10
     | TRef _ -> 11
     | LLet _ -> 12
+    | Plac _ -> 13
```
fblanqui, line 410: We should simply remove this comment now I think.

src/core/term.ml
```diff
@@ -484,6 +488,7 @@ let mk_Meta (m,ts) = Meta (m,ts)
 let mk_Patt (i,s,ts) = Patt (i,s,ts)
 let mk_TEnv (te,ts) = TEnv (te,ts)
 let mk_Wild = Wild
+let mk_Plac (b: bool) (n: string option) = Plac (b, n)
```
fblanqui, line 491: Perhaps you do not need to repeat the types of arguments here.

src/parsing/scope.ml
```diff
@@ -66,12 +66,9 @@ type lhs_data =
...
   | M_Term of
-      { mutable m_term_new_metas : problem
-      (** Metavariables generated during scoping. *)
-      ; m_term_meta_of_name : string -> meta option
-      (** Returns the meta with the given name. *)
-      ; m_term_meta_of_key : int -> meta option
-      (** Returns the meta with the give key. *)
+      { mutable m_term_placeholders : meta_ident list
+      (** Named placeholders. Used to ensure they do not appear non
+          linearly. *)
```
fblanqui, line 71: Could you please recall why you need this linearity condition?

src/parsing/scope.ml
```diff
@@ -66,12 +66,9 @@ type lhs_data =
 
 type mode =
   | M_Term of
-      { mutable m_term_new_metas : problem
-      (** Metavariables generated during scoping. *)
-      ; m_term_meta_of_name : string -> meta option
-      (** Returns the meta with the given name. *)
-      ; m_term_meta_of_key : int -> meta option
-      (** Returns the meta with the give key. *)
```
fblanqui, line 74: Why do you remove m_term_meta_of_name and m_term_meta_of_key? How do you replace them?

gabrielhdt: Scoped terms do not contain meta variables, they only contain so-called placeholders that are turned into meta-variables during typechecking. This is why I asked https://github.com/Deducteam/lambdapi/discussions/667

src/parsing/scope.ml
gabrielhdt: The first possible answer is that it may be the case that non linear meta variables are useless. We can also say that it's dangerous to allow non linear meta variables because we don't know what happen if they are not in the same context, possibly the unification will just fail because it won't be able to instantiate the variable (it's just a guess).

src/core/infer.ml
```diff
@@ -1,216 +1,201 @@
...
-(** [infer p c t] tries to infer a type for the term [t] in context [c],
-   possibly adding new unification constraints in [p]. The set of
-   metavariables of [p] are updated if some metavariables are instantiated or
-   created. The returned type is well-sorted if the recorded constraints are
-   satisfied. [c] must be well sorted.
-@raise NotTypable when the term is not typable (when encountering an
-   abstraction over a kind). *)
-let rec infer : problem -> ctxt -> term -> term = fun p c t ->
-  if !log_enabled then log_infr "infer %a%a" pp_ctxt c pp_term t;
+(** [unif pb ctx a b] solves the unification problem [ctx ⊢ a ≡ b]. Current
```
fblanqui, line 13: I suggest to use c for the context, everywhere, as it is now done in master.

src/core/term.ml
fblanqui: I know that this is some extra work but I find really convenient to have the comments in the implementation files too, to avoid switching between the two files. Currently, the interface files are mainly here to define what is exported.

src/parsing/scope.ml
fblanqui: How in your code a placeholder gets replaced by an already existing meta referred to by the user?

src/parsing/scope.ml
fblanqui: I think I can live with this restriction. It's just that it would be good to add a comment here.

src/core/infer.ml
```diff
@@ -1,216 +1,201 @@
...
+              (* Meta type must be a product of arity greater or equal
+                 to the environment *)
+              assert false
+      in
+      let range = ref_esubst 0 !(m.meta_type) in
+      (t, range)
+  | LLet (t_ty, t, u) ->
+      (* Check that [a] is a type, and refine it. *)
+      let t_ty, _ = type_enforce pb ctx t_ty in
+      (* Check that [t] is of type [a'], and refine it *)
```
fblanqui, line 111: Typo: What is a' here?

src/core/infer.ml
```diff
@@ -1,216 +1,201 @@
...
+      let t = force pb ctx t t_ty in
+      (* Unbind [u] and get new context with [x: t_ty ≔ t] *)
+      let x, u, ctx = Ctxt.unbind ctx t_ty (Some t) u in
+      (* Infer type of [u'] and refine it. *)
+      let u, u_ty = infer pb ctx u in
+      let u_ty = Bindlib.(u_ty |> lift |> bind_var x |> unbox) in
+      let u_ty = Bindlib.subst u_ty t in
```
fblanqui, line 118: You could use a let instead, like in master, to try to keep some sharing.

src/core/infer.ml
```diff
@@ -1,216 +1,201 @@
...
+      let u = Bindlib.(u |> lift |> bind_var x |> unbox) in
+      (mk_LLet (t_ty, t, u), u_ty)
+  | Abst (dom, b) ->
+      (* Domain must by of type Type, we don’t use [type_enforce] *)
```
fblanqui, line 122: Typo: by -> be

fblanqui: Why don't you use type_enforce here? You do it in the let case.

src/core/infer.ml
```diff
@@ -1,216 +1,201 @@
...
+      let dom = force pb ctx dom mk_Type in
+      let x, b, ctx = Ctxt.unbind ctx dom None b in
+      let b, b_s = type_enforce pb ctx b in
+      let s =
+        match unfold b_s with
+        | Type -> mk_Type
+        | Kind -> mk_Kind
+        | b_s ->
+            Error.wrn None
+              "Type error, sort mismatch: there is no rule of the form \
```
fblanqui, line 140: precise: rule -> typing rule
But why do you speak of typing rules for products here? Isn't it just the case that the type of b is not a sort?

fblanqui:
@gabrielhdt It would be nice that you explained in more details what are the differences between the type inference/checking algorithm in master and the one in your PR. At first glance, I do not see much differences.
In fact, in your PR, I see two things:
1) At scoping time, replace underscores and implicit arguments by place holders. During type inference/checking, replace place holders by metavariables. You do this so that metavariables get well-typed contexts, as it may be necessary to add coercions in contexts too.
2) Perhaps some modifications in the type inference/checking algorithm. But this is not clear to me yet.
So, perhaps, this PR could be split in two. This would help understand the lost in performance.

src/core/infer.ml
gabrielhdt: `type_enforce` would allow `dom` to be of type `TYPE` or `KIND` so it would allow an expression `\s:TYPE, t`, which shouldn't be allowed because we don't have a product `(Kind, _, _)` in lambdapi, right?

fblanqui: I see. Perhaps just remove ", we don’t use [type_enforce]" to avoid any confusion.

src/parsing/scope.ml
gabrielhdt: As far as I understand, we don't need to do that, except for the "focus" tactic, in which case it's enough to just search for the name of the meta variable, right?

fblanqui: A user could use metas in many tactics: refine, apply, rewrite.

fblanqui: But we could decide to abandon this feature.

gabrielhdt:
Hi,
>     1. https://github.com/fblanqui/lib does not compile with it.
>        Remark: it's a good habit to check this library in addition to make tests. Perhaps we should integrate it in the CI.

This can be fixed by adding a final `reflexivity` to the proof script, line 224 of `Nat.lp`. I'll check whether this is justified or if it's a bug.

src/core/infer.ml
gabrielhdt: The type of `b` must always be a sort, doesn't it?

src/core/term.ml
gabrielhdt: I'm not sure to understand what you mean  here.

src/parsing/scope.ml
gabrielhdt: For all of these things, it would be nice to see what we lose, and we can deal with placeholders to achieve the same thing (hoping that we can).

src/core/term.ml
fblanqui: You can simply remove everything between lines 401 and 421.

src/parsing/scope.ml
fblanqui: If you do not start with [] but with the metas of the current typing goals:
https://github.com/Deducteam/lambdapi/blob/93082b2c04c0bb2ecb9dddf33a6c61970154bf7b/src/handle/tactic.ml#L183-L184

gabrielhdt:
> Hello @gabrielhdt . There are a number of problems with this PR:
>     2. Performances are degraded:
>        comparison between master and your PR on my machine:
>        make tests: 7.7s -> 9.2s = +20%
>        consider tests/OK/649_fo_27.lp and increase the number:
>        fo_45.lp: 0.643s -> 1.909s = +200%

Could you provide the file `fo_45.lp` please?

fblanqui:
[fo_45.txt](https://github.com/Deducteam/lambdapi/files/6578464/fo_45.txt)

gabrielhdt:
With first commit (so only new typechecker), `fo_45.lp` takes 1.1s (user time)

src/core/infer.ml
fblanqui: Right.

src/core/term.ml
```diff
@@ -434,6 +438,7 @@ let cmp : term cmp =
     | TEnv(e,ts), TEnv(e',ts') ->
         lex cmp_tenv (Array.cmp cmp) (e,ts) (e',ts')
     | TRef r, TRef r' -> Stdlib.compare r r'
+    | Plac _, Plac _ -> 0
```
fblanqui, line 441: I am not sure about that but I would prefer to have `Stdlib.compare t t'` instead of `0`.

src/handle/tactic.ml
```diff
@@ -195,26 +205,28 @@ let handle : Sig_state.t -> bool -> proof_state -> p_tactic -> proof_state =
   | P_tac_solve -> assert false (* done before *)
   | P_tac_apply pt ->
       let p = new_problem() in
```
fblanqui, line 207: What is this p for if scoping does not create metavariables anymore? Problem creation should be done as close as possible to their use.

fblanqui: It seems that problem creation could be moved to tac_refine.

gabrielhdt: Other tactics give problem with metavariables to `tac_refine`, it doesn't look obvious.

src/parsing/scope.ml
gabrielhdt: I was rather speaking at the level of Lambdapi files, can we do everything we want without creating and referencing explicitly meta-variables.

src/core/print.ml
```diff
@@ -234,6 +234,8 @@ and pp_term : term pp = fun ppf t ->
     | Kind        -> out ppf "KIND"
     | Symb(s)     -> pp_sym ppf s
     | Meta(m,e)   -> out ppf "%a%a" pp_meta m pp_env e
+    | Plac(_,None) -> out ppf "?"
```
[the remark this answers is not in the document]
gabrielhdt: I don't know, so I replaced with `_` which is.

gabrielhdt:
Regarding performance problem, I think they  are due to the fact that during typechecking, we traverse the context more often than before (twice as many times), since it is traversed to transform placeholders into meta variables and then to instantiate metavariables.

Things should be better if we change the type of meta variables, namely reduce the `meta` type to
```ocaml
type meta = { meta_key : int; meta_name : string option; meta_value : tmbinder ref }
```
and extend the type of problems to
```ocaml
type problem_aux =
  { ...
  ; metas : (ctxt * term) MetaMap.t }
```
so that the `arity` and `meta_type` field are moved to the context and type of the meta map, and that the creation of a meta variable boils down to the creation of a pointer that points to the current context. Inference of a meta type will still traverse the context, but this is already the case.

I think that this transformation should also allow to drop the `fresh` and `fresh_box` functions, so that we only work with the `make` functions. Consequently, environments `Env` should be used less as well, in favour of contexts `Ctxt`.

src/parsing/scope.ml
fblanqui: Difficult to answer such a question. Users always come up with uses you didn't think about.

fblanqui:
@gabrielhdt I see that you fixed the problem with https://github.com/fblanqui/lib but, now, there are a lot of new incorrect warnings like `[/home/blanqui/src/lambdapi/mylib/Eq.lp:26:15-17] Variable xy could be replaced by '_'.`

fblanqui:
Could you please give more details about that? Where the context is 
traversed?

Le 06/06/2021 à 14:30, Gabriel Hondet a écrit :
> Regarding performance problem, I think they are due to the fact that 
> during typechecking, we traverse the context more often than before 
> (twice as many times), since it is traversed to transform placeholders 
> into meta variables and then to instantiate metavariables.

gabrielhdt:
> @gabrielhdt I see that you fixed the problem with https://github.com/fblanqui/lib but, now, there are a lot of new incorrect warnings like `[/home/blanqui/src/lambdapi/mylib/Eq.lp:26:15-17] Variable xy could be replaced by '_'.`

Did I? Just by adding a `reflexivity` tactic you mean?

Warnings appear because the variables introduced by `assume` are not bound in the body. Before this PR, the variables were bound in the environment of the meta variables at scoping time. Now that we use placeholders, an `assume a` is scoped to a `refine \a, ?` where `?` is a placeholder, so `a` is indeed unbound. The old behaviour would do `refine \a, ?[a]`, so `a` would be bound.

gabrielhdt:
> Could you please give more details about that? Where the context is traversed? Le 06/06/2021 à 14:30, Gabriel Hondet a écrit :
> […](#)
> Regarding performance problem, I think they are due to the fact that during typechecking, we traverse the context more often than before (twice as many times), since it is traversed to transform placeholders into meta variables and then to instantiate metavariables.

Hmm it's not as obvious as I thought, but for instance when a placeholder for a value is encountered, it is traversed twice to generate two new meta variables (for the type and the meta itself).

fblanqui:
Le 07/06/2021 à 08:35, Gabriel Hondet a écrit :
>
>     @gabrielhdt <https://github.com/gabrielhdt> I see that you fixed
>     the problem with https://github.com/fblanqui/lib
>     <https://github.com/fblanqui/lib> but, now, there are a lot of new
>     incorrect warnings like
>     |[/home/blanqui/src/lambdapi/mylib/Eq.lp:26:15-17] Variable xy
>     could be replaced by '_'.|
>
> Did I? Just by adding a |reflexivity| tactic you mean?
>
No, I didn't change the files (and we a priori shouldn't anyway).


> Warnings appear because the variables introduced by |assume| are not 
> bound in the body. Before this PR, the variables were bound in the 
> environment of the meta variables at scoping time. Now that we use 
> placeholders, an |assume a| is scoped to a |refine \a, ?| where |?| is 
> a placeholder, so |a| is indeed unbound. The old behaviour would do 
> |refine \a, ?[a]|, so |a| would be bound.
>
How do you propose to solve this problem?

src/handle/tactic.ml
gabrielhdt: Tactics `induction` and `have` provide a non empty problem to `tac_refine`.

fblanqui:
Hi @gabrielhdt . Do you know why the CI continues to check for ocaml 4.07.1 ?

gabrielhdt:
> Hi @gabrielhdt . Do you know why the CI continues to check for ocaml 4.07.1 ?

No, I don't.

fblanqui:
@QGarchery Could you please check your developments with this PR?

QGarchery:
I have some trouble testing this PR. I believe this is due to commit 72d6e858877d9bdcd4d9966c4fcf0303abdc0fd0 where some additional reduction is done. The file `assume.txt` from #693 shows what the issue is.

gabrielhdt:
> I have some trouble testing this PR. I believe this is due to commit [72d6e85](https://github.com/Deducteam/lambdapi/commit/72d6e858877d9bdcd4d9966c4fcf0303abdc0fd0) where some additional reduction is done. The file `assume.txt` from #693 shows what the issue is.

Hi @QGarchery. After some bisection, it's the (not so) new type checking algorithm which causes your example to fail (commit 9ae0100d8a6346300f56988356fde83a54838685). I'm not surprised that this algorithm does a bit more reduction. I'll have a look.

fblanqui:
Hi @gabrielhdt. I do not understand what is the problem here. Did you solve it?

gabrielhdt:
> Hi @gabrielhdt. I do not understand what is the problem here. Did you solve it?

Not yet, but I think the problem is that this typechecker may perform reductions that the previous typechecker wouldn't perform. Some tactics seem to be a bit fragile, in the sense that they depend syntactically on the goal (and not modulo). It's a guess though. My idea is to try to not record reductions that are performed.

gabrielhdt:
> I have some trouble testing this PR. I believe this is due to commit [72d6e85](https://github.com/Deducteam/lambdapi/commit/72d6e858877d9bdcd4d9966c4fcf0303abdc0fd0) where some additional reduction is done. The file `assume.txt` from #693 shows what the issue is.

I forgot to use the `pure_eq_modulo`, it should be OK now.

fblanqui:
@QGarchery Thanks to confirm that this is fine for you now.

CHANGES.md
```diff
@@ -13,14 +13,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
...
+
 ## 2.0.0 (2021-12-15)
 
 ### Release of the VSCode extension on the Marketplace (2021-12-10)
 
 - Add `editors/vscode/CHANGES.md` and `editors/vscode/CONTRIBUTING.md`.
 - Update documentation and README.md files.
 
-### Structured proof scripts (2021-12-07)
+#### Structured proof scripts (2021-12-07)
```
fblanqui, line 44: Thanks to revert this modif.

doc/terms.rst
```diff
@@ -64,9 +64,7 @@ A user-defined term can be either:
 * an application written by space-separated juxtaposition, except for
   symbol identifiers declared as infix (e.g. ``x + y``)
 
-* a meta-variable application ``?M.[t;u;v]``. ``?M`` alone, without arguments
-  between square brackets, is a shorthand for ``?M.[x1;..;xn]`` where
-  ``x1;..;xn`` are all the variables of the context.
+* an named unknown term or meta-variable ``?M``.
```
fblanqui, line 67: typo: an named -> a named?

src/core/libMeta.ml
```diff
@@ -49,23 +50,37 @@ let of_name : string -> problem -> meta option = fun n p ->
...
+let make : problem -> ?name:string -> ctxt -> term -> term =
+  fun p ?name ctx a ->
   let a, k = Ctxt.to_prod ctx a in
-  let m = fresh p a k in
-  let get_var (x,_,_) = mk_Vari x in
-  mk_Meta(m, Array.of_list (List.rev_map get_var ctx))
+  let m = fresh ?name p a k in
+  let get_var (x,_,d) = if d = None then Some (mk_Vari x) else None in
+  mk_Meta(m, Array.of_list (List.filter_rev_map get_var ctx))
+
```
fblanqui, line 61: missing comment for bmake

src/core/unif.ml
```diff
@@ -32,13 +56,12 @@ let add_constr : problem -> constr -> unit = fun p c ->
    as well as the constraint [(c,a,b)] where [a] is the type of [t] and [b]
    the type of [u] if they can be infered. *)
 let add_unif_rule_constr : problem -> constr -> unit = fun p (c,t,u) ->
-  add_constr p (c,t,u);
   match Infer.infer_noexn p c t with
-  | None -> ignore (Infer.infer_noexn p c u)
-  | Some a ->
+  | None -> add_constr p (c, t, u)
```
fblanqui, line 60: Shouldn't you keep the test that u is typable?

src/core/unif.ml
```diff
@@ -32,13 +56,12 @@ let add_constr : problem -> constr -> unit = fun p c ->
...
+  | Some (t, a) ->
       match Infer.infer_noexn p c u with
-      | Some b when not (Eval.pure_eq_modulo c a b) -> add_constr p (c,a,b)
-      | _ -> ()
+      | None -> add_constr p (c, t, u)
+      | Some (u, b) -> add_constr p (c, t, u); add_constr p (c,a,b)
```
fblanqui, line 64: Why not keeping the test (Eval.pure_eq_modulo c a b)?

CHANGES.md
```diff
@@ -13,14 +13,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
...
+#### Type checker with refinement
+
+Implemented a refiner, a type checker which is able to edit the type checked
+term. The algorithm is taken from <https://arxiv.org/abs/1202.4905>.
+
+- New type checker (with refinement)
+- Added placeholders. Placeholders are the holes in the concrete syntax, they
+  are transformed into metavariables by the type checker. It simplifies some
+  code related to metavariables. *Placeholders cannot appear nonlinearly in a
+  term*.
```
fblanqui, line 25: Thanks to provide some justification with a reference.

tests/OK/767.lp
```diff
@@ -1,4 +1,11 @@
...
 constant symbol B : Set;
-type λ (f : _ → _ → τ B) (g : τ B → τ B) b1 b2, f (g b1) (g b2);
+type λ (f : _ → τ B) (b: τ B), f b;
+// type λ (f : _ → _ → τ B) (g : τ B → τ B) b1 b2, f (g b1) (g b2);
+// The command above fails because the instantiation of the second placeholder
+// triggers a unification problem τ B == ?1[g b1], which fails because `g b1`
+// is not a variable.
+// "arguments are not distinct variables: g b1"
+// It worked before this PR because the arrow notation would specify that the
+// second meta-variable cannot depend on the first argument.
```
fblanqui, line 11: That's annoying. This should be documented in CHANGES.

fblanqui (changes requested):

tests/OK/767.lp
fblanqui: Some idea: One way to recover this example and improve the efficiency of type inference also perhaps would be to add in the type term a new constructor Arrow for the non-dependent product. In type inference and unification, we could even use a new smart constructor mkArrow_or_Prod to convert a closed dependent product into a non-dependent product.

CHANGES.md
```diff
@@ -13,14 +13,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
...
+- No more unused variable warning (such variables may appear in the explicit
+  substitution of a metavariable).
+
```
fblanqui, line 28: There is still some code generating warnings:
```
16:24 ~/src/lambdapi (refiner-merge-steps) gl 'could be replaced'
./src/handle/command.ml:402:                    wrn pos "Variable [%a] could be replaced by [_]."
./src/parsing/scope.ml:307:            wrn pos "Variable %s could be replaced by '_'." id;
./src/parsing/scope.ml:381:              "Pattern [%a] could be replaced by [_]." Pretty.term t;
```

gabrielhdt: We keep warning for unused patterns

src/core/unif.ml
gabrielhdt: What does it mean when `t` is not typable? What should we do in that case?

src/core/unif.ml
gabrielhdt: Because even if `a` and `b` are equal modulo, term `u` may have been  refined. But what we may do is
```ocaml
| Some (u, b) -> add_constr p (c, t, u); if not (Eval.pure_eq_modulo c a b) then add_constr p (c, a, b)
```

src/core/unif.ml
fblanqui: Right. We should probably fail in this case.

src/core/unif.ml
fblanqui: Ok.

tests/OK/767.lp
gabrielhdt: I'll add an issue afteward

CHANGES.md
gabrielhdt: For the nonlinearity?

fblanqui: Yes, please. And for removing the input of metavariable applications by users too.

gabrielhdt: I've quoted the article in the changelog (which is the top comment of this PR for now)

fblanqui: Right but it would be nice to be more precise and refer to some specific places in the paper so that one does not have to read all the paper to understand why.

src/parsing/scope.ml
fblanqui: One way to simulate this is as follows. Imagine that you get two subgoals ?G1 and ?G2 with the same type T, and you need a long proof for proving one or the other. To avoid duplicating such a proof, you need to introduce a lemma before the generation of ?G1 and ?G2 with `have l:T {my long proof}`. Then, for proving ?G1 and ?G2, you can simply do `apply l`.

fblanqui:
Non-linearity is perhaps not so much a problem after all. We could improve the current code by recording the context in which a meta is generated for a given placeholder and use the same meta in case there is another occurrence of the placeholder in the same context, and fail otherwise. Indeed, in case you generate two different subgoals ?G1 and ?G2 with the same type T, you can always duplicate the proof of ?G1 to prove ?G2. If the proof is long, you can introduce a lemma with `have l:T {long proof}` before the generation of ?G1 and ?G2, and prove ?G1 and ?G2 by `apply l`.
Similarly for explicit metas. If you want a given placeholder generating the meta ?G be ?M[t], you need to duplicate the proof of ?M[t] or introduce a lemma before.

fblanqui:
You need to update syntax.bnf by doing make doc.

fblanqui:
Remark 1: Since placeholders are required to be linear, why keeping the notation "?M" for placeholders? It could be replaced by "_" which is semantically equivalent. At the price of losing user-defined names. But this could be recovered by using "_M" and have LpLexer.UNDERSCORE return Some "M".

Remark 2: Printed terms with metavariables (in goals, logs or error messages) are not parsable anymore. This is not very nice.
To solve this problem, we should keep the notation ?M[t] for already generated metas. Hence, distinguish between (linear) placeholders _M generating fresh metas, and already generates meta applications ?M[t]. To do this, you need to keep m_term_meta_of_name and m_term_meta_of_key.

doc/terms.rst
```diff
@@ -64,9 +63,9 @@ A user-defined term can be either:
 * an application written by space-separated juxtaposition, except for
   symbol identifiers declared as infix (e.g. ``x + y``)
 
-* a meta-variable application ``?M.[t;u;v]``. ``?M`` alone, without arguments
-  between square brackets, is a shorthand for ``?M.[x1;..;xn]`` where
-  ``x1;..;xn`` are all the variables of the context.
+* a hole ``_`` to be filled automatically.
+
+* a metavariable ``?0.[x;y]`` that have been generated previously.
```
fblanqui, line 68: typo: have -> has

src/tool/sr.ml
```diff
@@ -191,9 +191,8 @@ let check_rule : Scope.pre_rule Pos.loc -> rule = fun ({pos; elt} as pr) ->
     log_subj "LHS type: %a@.constraints: %a@.%a ↪ %a"
       pp_term ty_lhs pp_constrs lhs_constrs
       pp_term lhs_with_metas pp_term rhs_with_metas;
-  (* We build a map allowing to find a variable index from its name. *)
+  (* We build a map allowing to find a variable index from its key. *)
```
fblanqui, line 194: I propose to move these 2 lines to line 257.

fblanqui (commented):
Thank you Gabriel! You also need to update CHANGES.md.

gabrielhdt:
It should be OK for me, I've edited the changelog (following https://keepachangelog.com/en/1.0.0/) and slightly minimised the diff.

CHANGES.md
```diff
@@ -17,14 +17,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
...
 ### Removed
 
 - The command beautify superseded by the new command export.
 
+- Unused variable warning: whether a variable is used or not cannot be
+  decided while scoping (following #696) since placeholders that do not
+  depend on variables may be refined later into metavariables that may
+  depend on them.
+
+- Metavariables cannot be refenrenced by their name anymore, hence the
```
fblanqui, line 52: Typo: refenrenced -> referenced. I wanted to fix it myself but I cannot push to your branch... :-(

fblanqui (approved):

CHANGES.md
gabrielhdt: Ah you can edit now

gabrielhdt: I have fixed the typo

fblanqui:
Houra! Thank you @gabrielhdt !

PR #697, opened 2021-05-31
Improve logs

fblanqui:
- add Base.out = Format.fprintf
- uniformize printing code using Base.out
- rename oc arguments into ppf
- complete Term.pp_term
- improve some functions in Debug.D
- improve logging messages in Infer by adding a level argument

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,14 @@
 ### Unreleased
 
+#### Improve logs (2021-06-01)
+
+- add Base.out = Format.fprintf
+- uniformize printing code using Base.out
+- rename oc arguments into ppf
+- complete Term.pp_term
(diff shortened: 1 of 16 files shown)


yannl35133:
Pourquoi avoir retiré les noms des loggers dans les logs ?

fblanqui:
I was finding more disadvantages than advantages to keep it. Without, it is less verbose and it is easy to understand from which file it comes anyway.

yannl35133:
I don't find these 7 characters that invasive, and the context is not always sufficient to see easily where they come from.
Maybe the second point is what should be changed, I don't know

gabrielhdt:
I'd be in favour of keeping loggers' names.

PR #702, opened 2021-06-01
improve performances

fblanqui:
- substituting arguments in TEnv's at construction time (mk_TEnv)
- improving eq_modulo to avoid calling whnf when possible

This improves compilation time by 13% on make tests (6.7s instead of 7.7s).


fblanqui:
The final improvement is 8.0s -> 7.4s = -7,5%.

PR #704, opened 2021-06-03
Improve scoping of abstractions with implicit arguments

### LINKED ISSUE #698: Anonymous functions cannot take in implicit arguments
Currently, anonymous functions cannot take in implicit arguments : wrapping an argument in curly braces will have no effect and giving an explicit argument will raise an error

fblanqui:
Attempt to fix #698 . Can check `(λ {a}, λ x, f a x) z` but not `(λ {a}, λ x, f a x) {nat} z`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 ### Unreleased
 
+#### Improve evaluation and convertibility test (2021-06-02)
+
+- fix `_LLet` by calling mk_LLet
+- substitute arguments in TEnv's at construction time (mk_TEnv)
+- improve eq_modulo to avoid calling whnf when possible
+- use `Eval.pure_eq_modulo` in Infer and Unif (fix #693)
+
 #### Improve logs (2021-06-01)
 
 - add Base.out = Format.fprintf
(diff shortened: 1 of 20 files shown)


fblanqui:
The main change is in scope.ml: get_impl gets recursive and a case for abstraction is added.
@yannl35133 Your comments are welcome. It does not solve your issue completely but could already be merged.

fblanqui:
@yannl35133 Unfortunately, the PR fails on https://github.com/fblanqui/lib, file Eq.lp.

fblanqui:
@yannl35133 In the last commits, an abstraction gets implicit arguments only if it is applied. However, a symbol inherits the implicits of the abstractions used in its definition. Hence, `symbol id ≔ λ {a} (x:τ a), x` is equivalent to `symbol id {a} (x:τ a) :≔ x`. So top abstractions are not dealt with as inner abstractions. It is perhaps not completely satisfactory but, at least, it seems to solve all the issues. What do you think?

yannl35133:
It does look inelegant, even though it would work for what I need.

Do we need trailing implicit arguments ?
If not, we could make implicit arguments lazily inserted instead of greedily. This would also mean that implicit arguments information gets embedded in the type (and implicitness should become more robust ?)

fblanqui:
This is an interesting idea indeed but implies many smalls changes in many files. It seems better to study this in another PR.

PR #717, opened 2021-07-08
Structured proof scripts

Qzaac:
This PR makes the following modifications:

- Proof scripts must be structured into a tree:

A tactic replacing the current goal by n new goals must be followed by n proof scripts enclosed in curly brackets. For instance, instead of writing `induction; /* case 0 */ t1; ..; tm; /* case s */ q1; ..; qn`, we should now write `induction {t1; ..; tm} {q1; ..; qn}`.

Exception for tactics not really changing the current goal like "have": `/* proof of u */ have h: t; /* proof of t */ t1; ..; tm; /* proof of u continued */ q1; ..; qn` is now written `have h: t {t1; ..; tm}; q1; ..; qn`.

- Curly brackets are reserved for proof script structuration.
- Implicit arguments are must be declared using square brackets instead of curly brackets: we must write `[a:Set]` instead of `{a:Set}`.
- Term environments (in rules or in terms) and rewrite patterns (in the rewrite tactic) must be preceded by a dot: we must write `$f.[x]` instead of `$f[x]`.


src/parsing/syntax.ml
```diff
@@ -212,6 +213,18 @@ type p_tactic_aux =
 
 type p_tactic = p_tactic_aux loc
 
+type p_tactic_tree = Tactic of p_tactic option * p_tactic_tree list;;
+
+(**tree_to_list tactic_tree returns the list of tactics
+  described by the tree*)
```
fblanqui, line 219: Thanks to try to follow the style of comments used in the other files. You should instead write something like: `(** [tac_list_of_tact_tree t] converts the tactic tree [t] into a list of tactics by doing a depth-first search. *)`.

src/parsing/syntax.ml
```diff
@@ -212,6 +213,18 @@ type p_tactic_aux =
...
+let rec tree_to_list : p_tactic_tree -> p_tactic list = function
+  | Tactic (otac, l) -> match otac with
```
fblanqui, line 221: Thanks to try to follow the style of indentations used in other files. Here, match otac should be preceded by a newline.

src/parsing/syntax.ml
```diff
@@ -212,6 +213,18 @@ type p_tactic_aux =
...
+                        | None -> forest_to_list l
+                        | Some tac -> tac::(forest_to_list l)
+and forest_to_list : p_tactic_tree list -> p_tactic list = function
+  | [] -> []
+  | tree::f -> (tree_to_list tree) @ (forest_to_list f);;
```
fblanqui, line 226: Parentheses are useless here.

src/parsing/lpParser.mly
```diff
@@ -242,7 +250,13 @@ proof_end:
   | ADMITTED { make_pos $sloc Syntax.P_proof_admitted }
   | END { make_pos $sloc Syntax.P_proof_end }
 
-proof: BEGIN ts=terminated(tactic, SEMICOLON)* pe=proof_end { ts, pe }
+bracket_tactic :
+  L_CU_BRACKET tree_list=tree* R_CU_BRACKET { Tactic(None, tree_list) }
+
+tree :
```
fblanqui, line 256: tree is too generic for a name. Thanks to use tactic_tree for instance instead. (Note that we already have another notion of tree in LP, namely decision trees for matching.)

src/parsing/lpParser.mly
```diff
@@ -242,7 +250,13 @@ proof_end:
   | ADMITTED { make_pos $sloc Syntax.P_proof_admitted }
   | END { make_pos $sloc Syntax.P_proof_end }
 
-proof: BEGIN ts=terminated(tactic, SEMICOLON)* pe=proof_end { ts, pe }
+bracket_tactic :
+  L_CU_BRACKET tree_list=tree* R_CU_BRACKET { Tactic(None, tree_list) }
```
fblanqui, line 254: For arguments, you can use short names. For instance tree_list can be replaced by l or ts.
As mentioned in another comment, the first argument of Tactic should not be an option type.
So, we should rather have something like:
```
proof: BEGIN l=subproof* pe=proof_end { l, pe }
subproof: L_CU_BRACKET l=terminated(proof_step, SEMICOLON)* R_CU_BRACKET { l }
proof_step: t=tactic l=subproof* { Tactic(t, l) }
```
Indeed, if, after begin, there are several goals, there should be a subproof for each goal. It is like if begin would be a tactic. Hence, a proof for a theorem with 2 goals should look like:
```
opaque lemma: ... :=
begin
  { ... // proof of the first goal }
  { ... // proof of the second goal }
end;
```
Note that `t1; t2;` is then simply a shorthand for `t1 { t2 };`, when t1 does not generate any additional goals but simply transforms the current goal into a new one.

src/parsing/lpParser.mly
```diff
@@ -272,13 +286,20 @@ command:
     po=proof? SEMICOLON
     { let sym =
         {p_sym_mod=ms; p_sym_nam=s; p_sym_arg=al; p_sym_typ=Some(a);
-         p_sym_trm=None; p_sym_def=false; p_sym_prf=po}
+         p_sym_trm=None; p_sym_def=false;
+         p_sym_prf= match po with
+          | None -> None
+          | Some p -> Some (tree_to_list (fst p), snd p)}
```
fblanqui, line 292: This temporary transformation should be done in get_proof_data.

fblanqui (changes requested):
Thanks Aurélien for your PR. Here are some comments to fix.

src/parsing/syntax.ml
```diff
@@ -212,6 +213,18 @@ type p_tactic_aux =
 
 type p_tactic = p_tactic_aux loc
 
+type p_tactic_tree = Tactic of p_tactic option * p_tactic_tree list;;
```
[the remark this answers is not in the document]
Qzaac: It is indeed useless, thus the new type is 
`type p_proof_step = Tactic of p_tactic * p_subproof list
and p_subproof = p_proof_step list;;
type p_proof = p_subproof list;;`, as you advised.

src/parsing/lpParser.mly
Qzaac: Right, the new syntax is:
```ocaml
proof:
  (*in case there are more than one goal to prove in the beginning*)
  | BEGIN l=subproof+ pe=proof_end { l, pe }
  (*in case there is just one goal to prove in the beginning*)
  | BEGIN l=subproofbis pe=proof_end { [l], pe }

subproof: L_CU_BRACKET l=terminated(proof_step, SEMICOLON)* R_CU_BRACKET { l }

subproofbis: l=terminated(proof_step, SEMICOLON)* { l }

proof_step: t=tactic l=subproof* { Tactic(t, l) }```

**Comment by Qzaac** [comment]
Thank you Frédéric for your comments, the last commit should fix all of them and add an update to the beautifier.

File: src/parsing/lpParser.mly [hunk 8]
```diff
@@ -242,7 +242,17 @@ proof_end:
...
-proof: BEGIN ts=terminated(tactic, SEMICOLON)* pe=proof_end { ts, pe }
+proof:
+  (*in case there are more than one goal to prove in the beginning*)
+  | BEGIN l=subproof+ pe=proof_end { l, pe }
+  (*in case there is just one goal to prove in the beginning*)
+  | BEGIN l=subproofbis pe=proof_end { [l], pe }
+
+subproof: L_CU_BRACKET l=terminated(proof_step, SEMICOLON)* R_CU_BRACKET { l }
+
+subproofbis: l=terminated(proof_step, SEMICOLON)* { l }
```
fblanqui, line 253: We can get rid of the last semicolon and do some factorization by having something like:
```
proof:
  | BEGIN l=subproof+ pe=proof_end { l, pe }
  | BEGIN l=proof_steps pe=proof_end { [l], pe }
subproof: L_CU_BRACKET l=proof_steps R_CU_BRACKET { l }
proof_steps: l=separated_non_empty_list(SEMICOLON, proof_step) { l }
```

src/parsing/syntax.ml
```diff
@@ -212,7 +212,19 @@ type p_tactic_aux =
...
+
+type p_proof = p_subproof list;;
+
+(**[tac_list_of_pproof p] converts the p_proof [p] into a list of tactics*)
+let rec tac_list_of_pproof : p_proof -> p_tactic list = function
+| [] -> []
+| sp::spl ->
+  match sp with
+    | [] -> tac_list_of_pproof spl
+    | Tactic(t, splbis)::psl -> t::tac_list_of_pproof (splbis@[psl]@spl);;
```
fblanqui, line 226: Can't `[psl]@spl` be simplified into `(psl::spl)`?

src/parsing/syntax.ml
```diff
@@ -582,7 +604,7 @@ let fold_idents : ('a -> p_qident -> 'a) -> 'a -> p_command list -> 'a =
           (fold_term a
              (Pos.make pos
                 (P_LLet (p_sym_nam, p_sym_arg, p_sym_typ, t, d))))
-        p_sym_prf
+        (tac_list_option_of_pproof_option p_sym_prf)
```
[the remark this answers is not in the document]
fblanqui: Again, here, it would be nice not to call tac_list_of_pproof.

fblanqui (changes requested):

src/parsing/lpParser.mly
yannl35133: My suggestion :
```menhir
proof:
  (*in case there are more than one goal to prove in the beginning*)
  | BEGIN l=braced_proof_steps+ pe=proof_end { l, pe }
  (*in case there is just one goal to prove in the beginning*)
  | BEGIN l=separated_non_empty_list(SEMICOLON, proof_step) SEMICOLON? pe=proof_end { [l], pe }

braced_proof_steps: L_CU_BRACKET l=separated_non_empty_list(SEMICOLON, proof_step) SEMICOLON? R_CU_BRACKET { l }
proof_step: t=tactic l=braced_proof_steps* { Tactic(t, l) }
```

Last ";" in blocks would be optional (but I think they should be allowed, like in OCaml).
Maybe a dummy "split" tactic to introduce subproofs/braced_proof_steps at the toplevel would make the AST more simple ?
Also, I usually rename constructions like "separated_non_empty_list" to make them more usable, and use Menhir's string aliases for tokens so maybe my suggestion isn't the best here

src/parsing/syntax.ml
```diff
@@ -212,7 +212,19 @@ type p_tactic_aux =
...
 type p_tactic = p_tactic_aux loc
 
-(** Parser-level representation of a proof terminator. *)
+type p_proof_step = Tactic of p_tactic * p_subproof list
+and p_subproof = p_proof_step list;;
+
+type p_proof = p_subproof list;;
+
+(**[tac_list_of_pproof p] converts the p_proof [p] into a list of tactics*)
+let rec tac_list_of_pproof : p_proof -> p_tactic list = function
```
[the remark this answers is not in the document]
yannl35133: Why do we need to linearize the proof structure ? Based on the two current uses of the function, I think we can get away with keeping them in their tree structure

src/parsing/lpParser.mly
yannl35133: My suggestion doesn't work (shift/reduce conflict), so here goes one that works

```menhir
nonempty_sep_except_maybe_last(arg, sep):
  | a=arg { [a] }
  | a=arg sep { [a] }
  | a=arg sep l=nonempty_sep_except_maybe_last(arg, sep) { a :: l }


[...]

proof:
  (*in case there are more than one goal to prove in the beginning*)
  | BEGIN l=proof_block+ pe=proof_end { l, pe }
  (*in case there is just one goal to prove in the beginning*)
  | BEGIN l=nonempty_sep_except_maybe_last(proof_step, SEMICOLON) pe=proof_end { [l], pe }

proof_block: L_CU_BRACKET l=nonempty_sep_except_maybe_last(proof_step, SEMICOLON) R_CU_BRACKET { l }
proof_step: t=tactic l=proof_block* { Tactic(t, l) }
```

A conflict remains, because curly braces are also used for explicit arguments, and using TYPE_QUERY they can be mixed up

src/parsing/pretty.ml
```diff
@@ -245,6 +244,35 @@ let notation : Sign.notation pp = fun ppf n ->
   | Quant -> out ppf "quantifier"
   | _ -> ()
 
+(**[proof p ppf n] pretty prints the proof [p] using the formatter
+[ppf] at and identation level of [n] (initially 0)*)
```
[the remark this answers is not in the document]
yannl35133:
```ocaml
(** [proof ppf p] pretty prints proof [p] using formatter [ppf] *)
let rec proof : (p_proof * p_proof_end) pp =
  fun ppf (p, pe) ->
    out ppf "begin@[<hv2>@ %a@ @]%a"
      (fun ppf -> function
        (* No braces for a single toplevel block *)
      | [block] -> proof_steps ppf block
      | blocks -> proof_blocks ppf blocks) p
      proof_end pe
and proof_blocks : p_subproof list pp  =
  fun ppf blocks ->
    out ppf "@[<hv>%a@]"
      (pp_print_list ~pp_sep:pp_print_space proof_block) blocks
and proof_block : p_subproof pp =
  fun ppf block ->
    out ppf "{@[<hv2>@ %a@ @]}"
      proof_steps block
and proof_steps : p_proof_step list pp =
  fun ppf steps ->
    pp_print_list ~pp_sep:pp_print_space proof_step ppf steps
and proof_step : p_proof_step pp =
  fun ppf (Tactic (t, blocks)) ->
    out ppf "@[<hv2>%a@,%a;@];"
      tactic t
      proof_blocks blocks
```
Testing should be done to ensure I didn't do it wrong, but this should look decent.

All formatting should be redone so that no `\n` remain at all, they are fundamentally incompatible with the way `Forat` works

src/parsing/syntax.ml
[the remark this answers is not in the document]
fblanqui: @yannl35133 This is just temporary. This function will disappear in the end.

ejgallego:
This looks very interesting, sorry folks I still have a huge backlog of LP-related stuff.

A thing to consider indeed when equipping proof with structure is error recovery / nested state; usually, on the IDE-side you would like to handle something such as:
```
tac ; { foo; bar; { error } }
```
as an Adga-like hole, and be able to recover and leave it for the user to fill it. This is a nice way to add this kind of holes to proofs without a lot of works, and was tried before in Coq , and more formal treatments can be found in the work of Cyrus Omar.

But indeed, once you delinearize, it would be very convenient if the sub-proof state is local in a way that allows the proof engine to handle the missing part as a hole.

fblanqui:
@ejgallego Hi Emilio! The idea we try to implement here is that a tactic application should come with a subproof for each newly generated subgoal. For instance, assuming that you want to do induction on natural numbers, then you should write `induction { p1 } { p2 }` where, recursively, p1 is a proof for the first goal (case 0) and p2 is a proof for the second goal (case s). Writing only `induction` or `induction { p }` or `induction {p1} {p2} {p3}` would be an error. Moreover, in this approach, `p1; p2` is just syntactic sugar for `p1 { p2 }`, to avoid too many brackets, and thus can be used only when p1 generates only one goal.

fblanqui (approved):

fblanqui:
There are a number of problems to fix on the vscode and emacs interfaces.
Emacs: the electric mode does not handle "{".
Vscode: the goal printing is incorrect if "{" is not followed by a space.

PR #719, opened 2021-07-12
Printing decision tree of ghost symbols.

gabrielhdt:
Decision trees of ghost symbols can be printed with the new --ghost command line flag.

### DIFF
--- a/docs/dtrees.rst
+++ b/docs/dtrees.rst
@@ -3,7 +3,7 @@ Decision trees
 
 The pattern matching algorithm uses decision trees. These decision trees
 are attached to symbols and can be inspected for debugging purposes.
-To print the decision of a symbol ``s`` of a module whose *module path* is
+To print the decision tree of a symbol ``s`` of a module whose *module path* is
 ``M`` (see :doc:`Module path </module>`), its decision tree may be printed with
 
 ::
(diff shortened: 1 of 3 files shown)


fblanqui:
Thank you @gabrielhdt for your PR. There is something I do not fully understand from the doc though. Perhaps it needs some clarification.

The doc says: " To print the decision of a symbol s of a module M, the following command may be used if module M is in a package (the lambdapi.pkg file) placed _above_ the current working directory `lambdapi decision-tree M.s`".

However, if I have in a directory `tmp` with a package file with `package_name=tmp` and a symbol s in a file `Defs.lp`, I have to write `lambdapi decision-tree tmp.Defs.s`.

So shouldn't in the doc `M` be replaced by `package_name.file`?

In addition, I do not understand the distinction between above/below in the doc.

gabrielhdt:
`M` stands for a module path, so `M` is `tmp.Defs`. Perhaps the documentation isn't clear enough regarding terminology. _Above_ regarding the file system tree, with the root on top. It can be clarified to "_above_ (closer to the root)".

gabrielhdt:
I've updated the documentation, is it clearer now @fblanqui ?

fblanqui:
Thanks @gabrielhdt !

PR #720, opened 2021-07-13
Linking ghost signature while compiling

### LINKED ISSUE #718: Unification rules are not saved to lpo files
When importing a file that contains a `unif_rule`, this rule is only used if the file is loaded directly and not from an lpo file.

Here is a minimal example. Assume you have a file `Defs.lp` that contain :
```lambdapi
symbol T : TYPE;
symbol t : T;
symbol u : T;
symbol F : T → TYPE;
symbol a : F t;

unif_rule F $t1 ≡ F $t2 ↪ [ $t1 ≡ $t2 ];
```

And a file `Usage.lp` that contains: 
```lambdapi

require open Unif.Defs;

symbol f {x : T} : F x → T;
symbol g : T ≔ f a;
```

`lambdapi check Usage.lp` succeeds if there is no lpo file for `Defs.lp`, but `lambdapi check --gen-obj Defs.lp; lambdapi check Usage.lp` results in the error : 
```
Loading ".../Unif/Usage.lp" ...
[.../Unif/Usage.lp:5:0-19] Some metavariables could not be solved: a proof must be given

0. F t ≡ F ?1
1. ?1: T
2. ?1: T
[.../Unif/Usage.lp:5:5-19] The proof is not finished.


```

gabrielhdt:
Fixes #718

### DIFF
--- a/src/handle/compile.ml
+++ b/src/handle/compile.ml
@@ -88,6 +88,7 @@ let rec compile_with :
       Path.Map.iter compile !(sign.sign_deps);
       loaded := Path.Map.add mp sign !loaded;
       Sign.link sign;
+      Sign.link Unif_rule.sign;
       Console.out 2 "Loaded \"%s\"\n%!" obj; sign
     end
 


src/handle/compile.ml
```diff
@@ -88,6 +88,7 @@ let rec compile_with :
       Path.Map.iter compile !(sign.sign_deps);
       loaded := Path.Map.add mp sign !loaded;
       Sign.link sign;
+      Sign.link Unif_rule.sign;
```
fblanqui, line 91: Ah we just forgot to link the ghost signature! But why do we need to do this? And why each time we compile a file?

gabrielhdt:
On Tue, Jul 13, 2021 at 06:42:08AM -0700, Frédéric Blanqui wrote:
> In [1]src/handle/compile.ml:
> 
> > @@ -88,6 +88,7 @@ let rec compile_with :
>        Path.Map.iter compile !(sign.sign_deps);
>        loaded := Path.Map.add mp sign !loaded;
>        Sign.link sign;
> +      Sign.link Unif_rule.sign;
> 
> Ah we just forgot to link the ghost signature! But why do we need to do this?
The actual bug was that the decision tree of ghost symbol wasn't
updated, so in "Usage.lp", the decision tree of "\equiv" was only a
failure node.
> And why each time we compile a file?
If the file defines new unification rules, the tree has to be re
updated. Does it look good to you?

> References:
> 
> [1] https://github.com/Deducteam/lambdapi/pull/720#discussion_r668775456

fblanqui:
I see. So the problem is that decision trees are not updated when reading an lpo file. But this doesn't seem specific to ghost symbols, does it? Sign.link updates the decision tree of every symbol defined in the signature, but not of the symbols that have rules in sign_deps. Isn't this the problem actually? If so, then we should rather fix Sign.link.

gabrielhdt:
On Tue, Jul 13, 2021 at 07:18:18AM -0700, Frédéric Blanqui wrote:
> I see. So the problem is that decision trees are not updated when reading an
> lpo file. But this doesn't seem specific to ghost symbols, does it? Sign.link
> updates the decision tree of every symbol defined in the signature, but not of
> the symbols that have rules in sign_deps. Isn't this the problem actually? If
> so, then we should rather fix Sign.link.
It does not seem to be the problem: say we define three files
`Define.lp`, `Rules.lp` and `Use.lp` in a package `Rules` as follows

`Define.lp`:
```
symbol T: TYPE;
symbol foo: T → T;
symbol zero: T;
symbol succ: T → T;
```

`Rules.lp`:
```
require open Rules.Define;

rule foo (succ $x) ↪ $x;
```

and `Use.lp`
```
require open Rules.Define Rules.Rules;
compute foo (succ zero);
```

and we run
```sh
lambdapi check --gen-obj Define.lp
lambdapi check --gen-obj Rules.lp
lambdapi check Use.lp
```
we obtain
```
zero
```
so the rules are taken into account. I don't think we care with normal
symbols because the tree will be updated when the dependency is compiled
anyway. It's specific to ghost symbols because since we don't have
`require unif_rule`, we never call `compile unif_rule`, and hence never
link their symbols. Does it make sense?

fblanqui:
@gabrielhdt In fact, it seems enough to add in unif_rule.ml, line 38 after definition of cons:
```
(** Link symbols in [sign]. *)
let _ = Sign.link sign
```
But I do not fully understand what this changes to Unif_rule.sign that makes everything works.

fblanqui:
In master, rules in sign_deps are added in the sym_rules of the corresponding symbols, but decision trees are not updated:
https://github.com/Deducteam/lambdapi/blob/c4be48070fb3cf6da47c64fafdf0a34d991385dc/src/core/sign.ml#L136-L148
However linking a signature updates the decision trees of its symbols:
https://github.com/Deducteam/lambdapi/blob/c4be48070fb3cf6da47c64fafdf0a34d991385dc/src/core/sign.ml#L153
Hence, your solution, which consists in adding `link Unif_rule.sign` indeed includes the added rules in the decision of the equiv symbol. A simpler solution is to simply call `Tree.update_dtree Unif_rule.equiv`, but only if there are unification rules in dependencies.
Your example in https://github.com/Deducteam/lambdapi/pull/720#issuecomment-879267302 works in master because dependencies are loaded too:
https://github.com/Deducteam/lambdapi/blob/c4be48070fb3cf6da47c64fafdf0a34d991385dc/src/handle/compile.ml#L88
if they have not been loaded already:
https://github.com/Deducteam/lambdapi/blob/c4be48070fb3cf6da47c64fafdf0a34d991385dc/src/handle/compile.ml#L57-L60
But .unif_rule is set as already loaded:
https://github.com/Deducteam/lambdapi/blob/c4be48070fb3cf6da47c64fafdf0a34d991385dc/src/core/unif_rule.ml#L22
Hence, the decision tree of equiv is never updated.

fblanqui:
Improved in #721 .

PR #721, opened 2021-07-16
Sign.link: update decision trees of symbols in sign_deps

fblanqui:
Alternative to #720 for fixing #718 which optimizes decision tree updates for Unif_rule.equiv.
It also changes the type of sign_deps from `(string * rule) list Path.Map.t ref` to `rule list StrMap.t Path.Map.t ref`,
and does a few renamings in sign.ml.

### DIFF
--- a/src/handle/compile.ml
+++ b/src/handle/compile.ml
@@ -88,6 +88,8 @@ let rec compile_with :
       Path.Map.iter compile !(sign.sign_deps);
       loaded := Path.Map.add mp sign !loaded;
       Sign.link sign;
+      if Path.Map.mem Unif_rule.path !(sign.sign_deps) then
+        Tree.update_dtree Unif_rule.equiv;
       Console.out 2 "Loaded \"%s\"\n%!" obj; sign
     end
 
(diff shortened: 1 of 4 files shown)


src/handle/compile.ml
```diff
@@ -88,6 +88,8 @@ let rec compile_with :
       Path.Map.iter compile !(sign.sign_deps);
       loaded := Path.Map.add mp sign !loaded;
       Sign.link sign;
+      if Path.Map.mem Unif_rule.path !(sign.sign_deps) then
```
gabrielhdt, line 94: I think this should always be the case since ghost signatures are dependencies of all signatures.

fblanqui: Ah right. I should test whether there are rules added. Thanks!

fblanqui: Do you have other remarks?

gabrielhdt: You can systematically update the tree: since the operation is lazy, it shouldn't cost more.

fblanqui: See https://github.com/Deducteam/lambdapi/pull/722.

PR #731, opened 2021-09-29
fix printing of identifiers

### LINKED ISSUE #729: Identifiers not correctly qualified for printing
Let the following file `reqbool.lp` be in `tests/OK` (for convenience),
```
require tests.OK.bool as B;
symbol bool_neg: B.B → B.B;
symbol negate b ≔ B.bool_neg b;
```
then we have the following interaction
```
$ lambdapi check -v3 reqbool.lp
[...]
Checked "/home/hondet/workspace/deducteam/lambdapi/tests/OK/bool.lp"
(symb) add bool_neg : B.B → B.B
(symb) add negate : B.B → B.B ≔ λ b, bool_neg b
Checked "/home/hondet/workspace/deducteam/lambdapi/tests/OK/bool_req.lp"
```
and we see that `bool_neg` is not correctly qualified, it reads `bool_neg` instead of `B.bool_neg`.

This can be particularly confusing when developing generic libraries where, for instance, one defines the `cons` on lengthed lists, another one on lists, such that `LList.cons` and `List.cons` end up being printed as `cons`.

fblanqui:
fix #729

### DIFF
--- a/src/core/print.ml
+++ b/src/core/print.ml
@@ -86,17 +86,15 @@ let pp_match_strat : match_strat pp = fun ppf s ->
 
 let pp_sym : sym pp = fun ppf s ->
   if !print_implicits && s.sym_impl <> [] then out ppf "@";
-  let n = s.sym_name in
-  if StrMap.mem n !sig_state.in_scope then
-    (* For printing Unif_rule.cons unescaped. *)
-    if s == Unif_rule.cons then out ppf "%s" n else pp_uid ppf n
+  let ss = !sig_state and n = s.sym_name and p = s.sym_path in
+  if Path.Set.mem p ss.open_paths then pp_uid ppf n
   else
-    match Path.Map.find_opt s.sym_path (!sig_state).path_alias with
+    match Path.Map.find_opt p ss.path_alias with
     | None ->
         (* Hack for printing symbols replacing metavariables in infer.ml
-           unqualified and unescaped. *)
-        if n <> "" && let c = n.[0] in c = '$' || c = '?' then out ppf "%s" n
-        else out ppf "%a.%a" pp_path s.sym_path pp_uid n
+           unqualified. *)
+        if n <> "" && let c = n.[0] in c = '$' || c = '?' then pp_uid ppf n
+        else out ppf "%a.%a" pp_path p pp_uid n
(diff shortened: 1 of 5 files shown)


src/core/print.ml
```diff
@@ -86,17 +86,15 @@ let pp_match_strat : match_strat pp = fun ppf s ->
...
   else
-    match Path.Map.find_opt s.sym_path (!sig_state).path_alias with
+    match Path.Map.find_opt p ss.path_alias with
     | None ->
         (* Hack for printing symbols replacing metavariables in infer.ml
-           unqualified and unescaped. *)
-        if n <> "" && let c = n.[0] in c = '$' || c = '?' then out ppf "%s" n
-        else out ppf "%a.%a" pp_path s.sym_path pp_uid n
+           unqualified. *)
+        if n <> "" && let c = n.[0] in c = '$' || c = '?' then pp_uid ppf n
```
gabrielhdt, line 96: Just as a side remark, does the case `n = ""` happen?

tests/OK/729.lp
```diff
@@ -0,0 +1,6 @@
+require tests.OK.bool as B;
+symbol T : (B.B → B.B) → TYPE;
+symbol bool_neg: B.B → B.B;
+symbol foo : T B.bool_neg;
+symbol bar : T bool_neg;
+assertnot ⊢ T B.bool_neg ≡ T bool_neg;
```
gabrielhdt, line 6: Such tests should be performed on the output, with [CRAM tests](https://dune.readthedocs.io/en/stable/tests.html#cram-tests), but it requires some work.

src/core/print.ml
fblanqui: A priori not, although it was possible when identifiers were unescaped (take ```{||}```).

fblanqui:
@gabrielhdt Can this PR be merged?

gabrielhdt:
> @gabrielhdt Can this PR be merged?

Yes, I think so :)

src/core/print.ml
gabrielhdt: Should we put an `assert n <> ""` instead then?

fblanqui: No, I don't think so.

PR #752, opened 2021-11-25
fix make matita

### LINKED ISSUE #739: The matita library does not compile
I guess it has not been tested for quite some time. We should add dklib, focalide, matita and, perhaps, holide also, in the CI.
The problem is already in 2c30031f, 4 Feb 2021 (with why3 1.3.3).

fblanqui:
- rename identifiers to avoid problems with Bindlib (see https://github.com/rlepigre/ocaml-bindlib/issues/11)
- fix #739 by commenting the proof of le_fact_10


fblanqui:
@gabrielhdt Any comment on this patch?

PR #758, opened 2021-11-30
use hash table for LP keywords

fblanqui:


src/parsing/lpLexer.ml
```diff
@@ -403,8 +399,7 @@ let lexer buf =
   (* invalid token *)
 
   | _ ->
-      let loc = locate (lexing_positions buf) in
-      raise (SyntaxError(Pos.make (Some(loc)) (Utf8.lexeme buf)))
+      raise (SyntaxError(make_pos (lexing_positions buf) (Utf8.lexeme buf)))
```
gabrielhdt, line 402: Does this modificatiion make the lexing faster? Of how much?

fblanqui:
The use of a hash table does not seem to improve efficiency. On the other hand, it helps detect unused keywords. For instance, "set" was not used anymore but was still considered as a keyword.

PR #759, opened 2021-12-01
Improve and simplify LP lexer

### LINKED ISSUE #710: multi-lines comments cannot be nested

fblanqui:
fix #710

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,6 @@
 #### Improve and simplify LP lexer (2021-12-02)
 
 - allow nested comments (fix #710)
-- generate only one kind of identifiers
 - replace everywhere `%S` by `\"%s\"`
 - move checking compatibility with Bindlib of identifiers from lexer to scope
 - move `is_keyword` from `lexer` to `pretty`
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 ### Unreleased
 
-#### Improve and simplify LP lexer (2021-12-02)
+#### Improve and simplify LP lexer (2021-12-07)
 
 - allow nested comments (fix #710)
 - replace everywhere `%S` by `\"%s\"`
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 - generate only one kind of identifiers
 - replace everywhere `%S` by `\"%s\"`
 - move checking compatibility with Bindlib of identifiers from lexer to scope
+- move is_keyword from lexer to pretty
 
 #### Update dkParser to be in sync with dkcheck (2021-11-30)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,11 @@
 - generate only one kind of identifiers
 - replace everywhere `%S` by `\"%s\"`
 - move checking compatibility with Bindlib of identifiers from lexer to scope
-- move is_keyword from lexer to pretty
+- move `is_keyword` from `lexer` to `pretty`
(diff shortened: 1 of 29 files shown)


fblanqui:
@gabrielhdt Your comments are welcome!

src/common/debug.ml
```diff
@@ -17,7 +17,7 @@ module D = struct
 
   let int : int pp = fun ppf i -> out ppf "%d" i
 
-  let string : string pp = fun ppf s -> out ppf "%S" s
+  let string : string pp = fun ppf s -> out ppf "\"%s\"" s
```
gabrielhdt, line 20: Did `%S` escape things that shouldn't be?

CHANGES.md
```diff
@@ -1,5 +1,17 @@
 ### Unreleased
 
+#### Improve and simplify LP lexer (2021-12-02)
+
+- allow nested comments (fix #710)
+- generate only one kind of identifiers
+- replace everywhere `%S` by `\"%s\"`
```
gabrielhdt, line 7: Did `%S` escape things that shouldn't have been?

CHANGES.md
```diff
@@ -1,5 +1,17 @@
...
+- move checking compatibility with Bindlib of identifiers from lexer to scope
+- move `is_keyword` from `lexer` to `pretty`
+- move `package.ml` from `common/` to `parsing/`
+- change `Config.map_dir` field type to `(Path.t * string) list`,
+  `Library.add_mapping` type to `Path.t * string -> unit`
+  and Compile.compile argument `lm` type to `Path.t * string`
```
gabrielhdt, line 13: Why? What does this modification allow? What's the impact on the command line interface?

gabrielhdt (approved):

CHANGES.md
fblanqui: %S converts a string argument to OCaml syntax (double quotes, escapes). There is no reason for doing this in LP.

CHANGES.md
fblanqui: Before, the Path.t argument was a string and had to be parsed to convert it into a path. Doing this is cleaner and allows to simplify various things. In particular, we can now use escaped identifiers in package files and with the command decision-tree. package is moved to parsing since it is about package file parsing.

fblanqui: For instance, I can almost remove the file escape.ml (https://github.com/Deducteam/lambdapi/pull/759/commits/65d00e3c3ce0c6b16f20a122192c4e2ef1657685). I think I will do it in a next commit.

fblanqui:
@gabrielhdt Note also that I can now deal with paths/qualified identifiers directly in the parser instead of doing it in the lexer currently (not yet committed; I need to test it further before). Do you see any disadvantage a priori in doing this?

gabrielhdt:
> @gabrielhdt Note also that I can now deal with paths/qualified identifiers directly in the parser instead of doing it in the lexer currently (not yet committed; I need to test it further before). Do you see any disadvantage a priori in doing this?

So you mean that the lexer doesn't differentiate e.g. qualified from unqualified identifiers? No I don't see any disadvantage.

fblanqui:
I finally left this idea and came back to having several id constructors too as it is slightly more efficient. For instance, for parsing holide, we gain 3s (53.5s instead of 56.5s = -5%). Overall, the efficiency of parsing is improved wrt master as follows:

matita: 1.0s instead of 1.6s (-37%)
focalide: idem 2.3s
isabelle: 5.5s instead of 6.9s (-20%)
holide: 53.5s instead of 1m16.0s (-30%)

PR #763, opened 2021-12-07
assume tactic with no identifiers

gabrielhdt:
Assume as many hypotheses as can be.

### DIFF
--- a/docs/syntax.bnf
+++ b/docs/syntax.bnf
@@ -38,7 +38,7 @@
 <tactic> ::= <query>
            | admit
            | apply <term>
-           | assume <param>+
+           | assume <param>*
            | fail
            | focus <int>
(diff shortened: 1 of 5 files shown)


fblanqui:
Hi @gabrielhdt . The problem with this kind of tactics is that this is not very robust because it introduces many hypotheses not named by the user. This is not very good practice. So I would prefer not to add such a tactic.

gabrielhdt:
I see the point, but it's very handy for prototyping: propositions may almost be solved automatically using `assume; why3;`.

fblanqui:
assume shouldn't be needed before why3, that is, there should be no difference between assume;why3 and why3. This is perhaps not the case now but this should be fixed.

gabrielhdt:
Superseded by #765

PR #769, opened 2021-12-14
Binding quantifiers in the why3 tactic

gabrielhdt:
See the new goals proved in `tests/OK/why3.lp`: they involve quantifiers `ex` and `all`.

### DIFF
--- a/doc/tactics.rst
+++ b/doc/tactics.rst
@@ -176,6 +176,8 @@ The user should define those symbols using builtins as follows :
    builtin "and" ≔ … // : Prop → Prop → Prop
    builtin "or"  ≔ … // : Prop → Prop → Prop
    builtin "not" ≔ … // : Prop → Prop
+   builtin "all" ≔ … // : Π x: U, (T x → Prop) → Prop
+   builtin "ex"  ≔ … // : Π x: U, (T x → Prop) → Prop
 
 **Important note:** you must run ``why3 config detect`` to make
 Why3 know about the available provers.
--- a/src/handle/why3_tactic.ml
+++ b/src/handle/why3_tactic.ml
@@ -49,7 +49,9 @@ type config =
   ; symb_imp : sym (** Implication(⇒) symbol.    *)
   ; symb_bot : sym (** Bot(⊥) symbol.            *)
   ; symb_top : sym (** Top(⊤) symbol.            *)
-  ; symb_not : sym (** Not(¬) symbol.            *) }
+  ; symb_not : sym (** Not(¬) symbol.            *)
+  ; symb_ex  : sym (** Exists(∃) symbol.         *)
+  ; symb_all : sym (** Forall(∀) symbol.         *) }
 
 (** [get_config ss pos] build the configuration using [ss]. *)
 let get_config : Sig_state.t -> Pos.popt -> config = fun ss pos ->
@@ -61,10 +63,67 @@ let get_config : Sig_state.t -> Pos.popt -> config = fun ss pos ->
   ; symb_imp = builtin "imp"
   ; symb_bot = builtin "bot"
   ; symb_top = builtin "top"
-  ; symb_not = builtin "not" }
+  ; symb_not = builtin "not"
+  ; symb_ex  = builtin "ex"
+  ; symb_all = builtin "all" }
 
 (** A map from lambdapi terms to Why3 constants. *)
 type cnst_table = (term * Why3.Term.lsymbol) list
+
+(** Table of declared types, these types may be variables or constants. *)
+module TyTable : sig 
+  type t 
+
+  val empty : t 
+
+  val ty_of_term : t -> term -> t * Why3.Ty.ty
+  (** [ty_of_term tbl te] fetches the type associated to term [te] in
+      table [tbl] if it exists, or it creates such a type and return
+      the created type and the new table. *)
+
+  (** {b FIXME} the translation of types {!ty_of_term} is correct on
+      non empty types only. *)
+
+  val fold_left : 
+    do_var:('a -> term -> Why3.Ty.tvsymbol -> 'a) ->
+    do_sym:('a -> term -> Why3.Ty.tysymbol -> 'a) -> 'a -> t -> 'a
+  (** [fold_left ~do_var ~do_sym init tbl] folds over the table [tbl]
+      with initial value [init] using [do_var] to operate on type
+      variables and [do_sym] to operate on type symbols. *)
+end = struct
+  (* The module is created to hide the type [sym_or_var] *)
+  type sym_or_var = TySym of Why3.Ty.tysymbol | TyVar of Why3.Ty.tvsymbol
+
+  type t = (term * sym_or_var) list
+
+  let empty = []
+
+  let ty_of_term tbl te =
+    match List.assoc_eq (Eval.eq_modulo []) te tbl with
+    | TySym s -> (tbl, Why3.Ty.ty_app s [])
+    | TyVar v -> (tbl, Why3.Ty.ty_var v)
+    | exception Not_found -> (
+        (* FIXME generate a new goal to ensure that types translated as
+           fresh constants are inhabited. *)
+        match get_args te with
+        | Symb s, [] ->
+            let sym = Why3.(Ty.create_tysymbol (Ident.id_fresh s.sym_name) [] Ty.NoDef) in
+            ((te,TySym sym)::tbl, Why3.Ty.ty_app sym [])
+        | Vari x, [] ->
+            let sym = Why3.Ty.tv_of_string (Bindlib.name_of x) in
+            ((te,TyVar sym)::tbl, Why3.Ty.ty_var sym)
+        | _ -> 
+            let sym = Why3.(Ty.create_tysymbol (Ident.id_fresh "ty") [] Ty.NoDef) in
+            ((te,TySym sym)::tbl, Why3.Ty.ty_app sym []))
+
+  let fold_left ~do_var ~do_sym acc tbl =
+    let f acc (t,s_or_v) =
+      match s_or_v with
+      | TySym s -> do_sym acc t s
+      | TyVar v -> do_var acc t v
+    in
+    List.fold_left f acc tbl
+end
 
 (** [new_axiom_name ()] generates a fresh axiom name. *)
 let new_axiom_name : unit -> string =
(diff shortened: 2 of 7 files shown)


src/handle/why3_tactic.ml
```diff
@@ -49,7 +49,11 @@ type config =
   ; symb_imp : sym (** Implication(⇒) symbol.    *)
   ; symb_bot : sym (** Bot(⊥) symbol.            *)
   ; symb_top : sym (** Top(⊤) symbol.            *)
-  ; symb_not : sym (** Not(¬) symbol.            *) }
+  ; symb_not : sym (** Not(¬) symbol.            *)
+  ; symb_ex  : sym
+  ; symb_all : sym 
+  ; symb_codes : sym
```
fblanqui, line 55: What is sym_codes for? Isn't it the function of symb_T already?

src/handle/why3_tactic.ml
```diff
@@ -61,79 +65,130 @@ let get_config : Sig_state.t -> Pos.popt -> config = fun ss pos ->
...
+(** [translate_type tbl a] translate term [a] into a Why3 type using
+    constants in association list [tbl]. *)
+let rec translate_type : config -> ty_table -> term -> (ty_table * Why3.Ty.ty) option = 
+  fun cfg tbl ty -> 
+  match get_args ty with
+  | (Symb s, [e]) when s == cfg.symb_T ->
+      translate_typecode cfg tbl e
+  | _ -> None
+
+and translate_typecode : config -> ty_table -> term -> (ty_table * Why3.Ty.ty) option =
```
fblanqui, line 93: translate_typecode does not depend on translate_type.

src/handle/why3_tactic.ml
```diff
@@ -61,79 +65,130 @@ let get_config : Sig_state.t -> Pos.popt -> config = fun ss pos ->
...
-        (tbl, Why3.Term.t_not t)
+        let (tbl, ty_tbl, t) = translate_prop tbl ty_tbl t in
+        (tbl, ty_tbl, Why3.Term.t_not t)
     | (Symb(s), []      ) when s == cfg.symb_bot ->
-        (tbl, Why3.Term.t_false)
+        (tbl, ty_tbl, Why3.Term.t_false)
     | (Symb(s), []      ) when s == cfg.symb_top ->
-        (tbl, Why3.Term.t_true)
+        (tbl, ty_tbl, Why3.Term.t_true)
+    | (Symb(s), [a;Abst(_,t)]   ) when List.memq s [cfg.symb_ex; cfg.symb_all] ->
```
fblanqui, line 132: when clause can be simplified to `s == cfg.symb_e || s == cfg.symb_all`.

src/handle/why3_tactic.ml
```diff
@@ -61,79 +65,130 @@ let get_config : Sig_state.t -> Pos.popt -> config = fun ss pos ->
...
+        let x, t = Bindlib.unbind t in
+        let (tbl, ty_tbl ,t) = translate_prop tbl ty_tbl t in
+        let tquant = 
+          let vid = 
+            Why3.(Term.create_vsymbol @@ Ident.id_fresh (Bindlib.name_of x)) ty 
+          in
+          let close = 
+            if s == cfg.symb_ex then Why3.Term.t_exists_close
+            else if s == cfg.symb_all then Why3.Term.t_forall_close
+            else assert false (* Case avoided by pattern matching *)
```
fblanqui, line 147: can be simplified to `if s == cfg.symb_ex then Why3.Term.t_exists_close else Why3.Term.t_forall_close`

gabrielhdt:
The PR is ready for review. The modifications are:
- we introduce three builtin symbols: the type of type codes (called Set, should we call it U?), the existential quantifier and the universal quantifier,
- the tactic now involve a translation of types: since why3 uses some sort of first order with prenex polymorphism, type variables introduced by binders are translated by type variables while symbols are translated by constants,
- the proposition translator calls the type translator, which may fail, therefore the translation of propositions may fail, that's why it now returns an option type. Furthermore, the syntactic sugar `let*` to use conveniently options as monads has been introduced.

src/handle/why3_tactic.ml
```diff
@@ -42,30 +42,84 @@ let why3_env : Why3.Env.env =
...
+
+  type t = (term * sym_or_var) list
+
+  let empty = []
+
+  let ty_of_term tbl te =
+    match List.assoc_eq (Eval.eq_modulo []) te tbl with
+    | TySym s -> Some (tbl, Why3.Ty.ty_app s [])
+    | TyVar v -> Some (tbl, Why3.Ty.ty_var v)
+    | exception Not_found -> (
```
fblanqui, line 104: The parenthesis is not needed here, is it?

src/handle/why3_tactic.ml
```diff
@@ -42,30 +42,84 @@ let why3_env : Why3.Env.env =
...
+        match get_args te with
+        | Symb s, [] ->
+            let sym = Why3.(Ty.create_tysymbol (Ident.id_fresh s.sym_name) [] Ty.NoDef) in
+            Some ((te,TySym sym)::tbl, Why3.Ty.ty_app sym [])
+        | Vari x, [] ->
+            let sym = Why3.(Ty.tv_of_string (Bindlib.name_of x)) in
+            Some ((te,TyVar sym)::tbl, Why3.Ty.ty_var sym)
+        | _ -> None)
```
fblanqui, line 112: Why not returning a fresh type constant here, and make the function total (never returning None)?

src/handle/why3_tactic.ml
```diff
@@ -42,30 +42,84 @@ let why3_env : Why3.Env.env =
...
+  ; symb_imp   = builtin "imp"
+  ; symb_bot   = builtin "bot"
+  ; symb_top   = builtin "top"
+  ; symb_not   = builtin "not"
+  ; symb_ex    = builtin "ex"
+  ; symb_all   = builtin "all"
+  ; symb_codes = builtin "Set" }
 
 (** A map from lambdapi terms to Why3 constants. *)
 type cnst_table = (term * Why3.Term.lsymbol) list
```
fblanqui, line 73: We could use a more efficient data structure like Map using Term.cmp.

src/handle/why3_tactic.mli
```diff
@@ -0,0 +1,20 @@
...
+
+(** [default_prover] contains the name of the current prover. Note that it can
+    be changed by using the "set prover <string>" command. *)
+val default_prover : string ref
+
+(** [timeout] is the current time limit (in seconds) for a Why3 prover to find
+    a proof. It can be changed with "set prover <int>". *)
+val timeout : int ref
+
+(** [handle ss pos ps prover_name g] runs the Why3 prover corresponding to the
```
fblanqui, line 16: ps should be removed

src/lplib/option.ml
```diff
@@ -38,3 +38,9 @@ let eq : 'a eq -> 'a option eq = fun eq_elt o1 o2 ->
 
 let pp : 'a pp -> 'a option pp = fun pp_elt ppf o ->
   match o with None -> () | Some e -> pp_elt ppf e
+
+module Binding = struct
+  let ( let* ) : 'a t -> ('a -> 'b t) -> 'b t = fun o f -> 
+    match o with None -> None | Some x -> f x
+  let return x = Some x
+end
```
fblanqui, line 46: let* is bind with arguments permuted. It would be better to replace bind by let* and do not put let* and return in a module.

src/handle/why3_tactic.ml
```diff
@@ -42,30 +42,84 @@ let why3_env : Why3.Env.env =
...
+  ; symb_T     : sym (** Encoding of types.        *)
+  ; symb_or    : sym (** Disjunction(∨) symbol.    *)
+  ; symb_and   : sym (** Conjunction(∧) symbol.    *)
+  ; symb_imp   : sym (** Implication(⇒) symbol.    *)
+  ; symb_bot   : sym (** Bot(⊥) symbol.            *)
+  ; symb_top   : sym (** Top(⊤) symbol.            *)
+  ; symb_not   : sym (** Not(¬) symbol.            *)
+  ; symb_ex    : sym (** Exists(∃) symbol.         *)
+  ; symb_all   : sym (** Forall(∀) symbol.         *)
+  ; symb_codes : sym (** Type of encoded types.    *) }
```
fblanqui, line 55: sym_codes should be removed since it is used nowhere in the translation.

fblanqui (changes requested):

src/handle/why3_tactic.ml
gabrielhdt: Indeed they may not be necessary because it's the last clause, but it's a bit error prone because we may not distinguish the inner match from the outer.

src/handle/why3_tactic.ml
gabrielhdt: Right, I was not sure whether it was the right thing to do. I guess we want to have some conservativity property on the translation and that creating fresh constants does not break it.

src/lplib/option.ml
gabrielhdt: I propose to leave the `bind` for now: I agree it would be better to replace bind by `let*` (it's more ocamlish IMO), but I'd like it to be done in an other PR.

fblanqui: Ok but if you finally have a total translation function for types, there is no need for this new module.

gabrielhdt: Furthermore, indeed.

src/handle/why3_tactic.ml
fblanqui: This is what we already do for propositions. It seems ok to me. If I remember well, there is a justification in Yacine's PhD.

src/handle/why3_tactic.ml
gabrielhdt: For this we would have to ensure that `forall t, TMap.find t m = List.assoc_eq (Eval.eq_modulo []) t l`, and I don't think we need to be more efficient here

fblanqui: Right.

fblanqui:
It's fine for me. Let me know when you want me to merge it.

gabrielhdt:
I'll update the documentation, and then it'll be good.

gabrielhdt:
It's OK, it can be merged

fblanqui:
In fact, no. I realized that there is a problem when replacing unknown types by fresh constants. I found a small bug in the correctness proof of the encoding of first-order logic (see my mail to deducteam). It can however be fixed by making sure that all type constants are inhabited. In his PhD, @elhaddadyacine proved the correctness of propositional logic only. So there is no such problem in his PhD.

src/handle/why3_tactic.ml
fblanqui: This is not ok in fact: I found a small bug in the correctness proof of the encoding of first-order logic (see my mail to deducteam). It can however be fixed by making sure that all type constants are inhabited.
I see two possible solutions to fix this problem:
- Translate only types (including constants) that we know are inhabited, making the translation partial. However, in LP, type inhabitation is undecidable. So, we could only rely on some necessarily incomplete heuristic. Moreover, I don't know how to handle type variables. More theoretical work on this topic is needed, that is, a correctness proof of the encoding of polymorphic first-order logic.
- A nice alternative would be to generate a fresh meta for each translated type, keeping the translation total, and provide a tactic implementing the inhabitation heuristic.

gabrielhdt: Even without implementing talking about inhabitation heuristic, it would be already something to generate an inhabitation goal

fblanqui: See https://www21.in.tum.de/~blanchet/tff1spec.pdf, page 7, for the definition of the semantics of polymorphic FOL.

fblanqui: So what do you propose to do right now? To merge this PR as it is now but open an issue and fix it later in another PR?

gabrielhdt: Ah why not, indeed we can do that.

fblanqui: Or do you want to give it a try to generate metas?

gabrielhdt: I can do that in a separate PR anyway

fblanqui:
@gabrielhdt Thanks to update docs/tactics.md and CHANGES.md as well.

fblanqui:
Thanks also to put some FIXME comment in the code to indicate that the translation of quantifiers is correct on non-empty types only, and thus that we should generate a fresh meta for each type to make sure that it is not empty. A tactic could then be developed for proving this automatically in the simplest cases.

doc/tactics.rst
```diff
@@ -176,6 +176,8 @@ The user should define those symbols using builtins as follows :
    builtin "and" ≔ … // : Prop → Prop → Prop
    builtin "or"  ≔ … // : Prop → Prop → Prop
    builtin "not" ≔ … // : Prop → Prop
+   builtin "all" ≔ … // : Π x: U, (T U → Prop) → Prop
+   builtin "ex"  ≔ … // : Π x: U, (T U → Prop) → Prop
```
fblanqui, line 180: typo: T U -> T x

fblanqui:
@gabrielhdt One more time some sanity checks have not been done:
```
09:27 ~/src/lambdapi (master) make sanity_check
In src/handle/why3_tactic.ml, line 110 more than 78 characters...
In src/handle/why3_tactic.ml, line 116 more than 78 characters...
In src/handle/why3_tactic.ml, line 160 more than 78 characters...
In src/handle/why3_tactic.ml, line 166 more than 78 characters...
In src/handle/why3_tactic.ml, line 198 more than 78 characters...
In src/handle/why3_tactic.ml, line 74 has trailing spaces...
In src/handle/why3_tactic.ml, line 75 has trailing spaces...
In src/handle/why3_tactic.ml, line 77 has trailing spaces...
In src/handle/why3_tactic.ml, line 87 has trailing spaces...
In src/handle/why3_tactic.ml, line 115 has trailing spaces...
In src/handle/why3_tactic.ml, line 137 has trailing spaces...
In src/handle/why3_tactic.ml, line 164 has trailing spaces...
In src/handle/why3_tactic.ml, line 165 has trailing spaces...
In src/handle/why3_tactic.ml, line 166 has trailing spaces...
In src/handle/why3_tactic.ml, line 168 has trailing spaces...
In src/handle/why3_tactic.ml, line 170 has trailing spaces...
In src/handle/why3_tactic.ml, line 214 has trailing spaces...
In src/handle/why3_tactic.ml, line 217 has trailing spaces...
```

Issue #299, opened 2019-12-07
tools/latex_minted/dedukti.py needs to be updated with Lambdapi syntax


fblanqui:

francoisthire:
Maybe we should have two files, one for the old and one for the new syntax.

PR #775, opened 2021-12-15
remove misc/latex_minted directory

fblanqui:


fblanqui:
These files were for Dedukti and have never been adapted to Lambdapi (#299).

PR #779, opened 2021-12-17
Add regression tests for tools

gabrielhdt:
Test precisely the output to HRS, XTC and the decision tree export. It uses [this feature](https://dune.readthedocs.io/en/stable/tests.html#diffing-the-result) of dune. Mainly, for any file `test.ml`, if there is a `test.expected` file, then the output of `test.exe` is compared with `test.expected` (where `test.exe` is the compiled version of `test.ml`).


fblanqui:
Hi @gabrielhdt . That's a good idea but this is a little bit fragile because outputs are checked syntactically I guess. So it will break as soon as we change non-important things in the output. But we could improve this later I guess. Does dune allow to give options to diff to ignore spaces for instance? By the way, I think that the HRS and XTC outputs are partially buggy. They have to revised anyway. This is on my TODO list and will be part of our cooperation with Makoto.

gabrielhdt:
In a sense I don't think we care about it being fragile, because we want the output to be deterministic anyway. If we know the output is changed, we just have to promote the changes (using `dune --promote`, which automatically replaces the `<file>.expected` by the new output)

gabrielhdt:
And there should be no reason the output changes, if we don't intend to change it (and hence test and promote the new result).

fblanqui:
It's fine for me. Do you want to merge now?

gabrielhdt:
> It's fine for me. Do you want to merge now?

Yes we can

PR #784, opened 2021-12-22
scoping: add boolean argument to know if the term is a type

### LINKED ISSUE #767: Wrong meta type generation in scoping
example due to @amelieled
```
constant symbol Set : TYPE;
constant symbol B : Set;
type λ (f : _ → _ → τ B) (g : τ B → τ B) b1 b2, f (g b1) (g b2);
```

fblanqui:
fix #767

### DIFF
--- a/src/handle/command.ml
+++ b/src/handle/command.ml
@@ -162,7 +162,7 @@ let handle_inductive_symbol : sig_state -> expo -> prop -> match_strat
   (* We scope the type of the declaration. *)
   let p = new_problem() in
   let typ =
-    (if xs = [] then scope_term else scope_term_with_params)
+    (if xs = [] then scope_term ~typ:true else scope_term_with_params)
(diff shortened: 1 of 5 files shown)


fblanqui:
@gabrielhdt Your comments are welcome. I would prefer to merge in master before your PR, if you don't mind.

gabrielhdt (approved):
I don't know if it's a good idea to make the boolean default to `false`, I'm afraid it might be error prone, if developers forget to instantiate it. But it can be left  like that as well, it's no big deal. Otherwise LGTM.

fblanqui:
At the beginning, I started with a non-optional argument. But this has two disadvantages: this increases the diff and, in various cases where we don't know whether it's a type or not, it would be strange to write false or true. So I prefer to put nothing. Only in the cases where we are sure that it is a type, we should add something.

PR #798, opened 2022-01-03
Enable type inference in Menhir

### LINKED ISSUE #797: Enable type inference in Menhir
By @fpottier, see https://github.com/Deducteam/lambdapi/pull/794:

This change is required by Menhir 20211223.

As far as I can see, there are two more problems that need fixing:

```
File "src/parsing/lpParser.mly", line 311, characters 15-27:
Error: $syntaxerror is no longer supported.
```
Instead of using `$syntaxerr`, you should raise an exception of your own.

```
File "src/parsing/dkParser.ml", lines 1754-1826, characters 4-20:
1754 | ....match _tok with
1755 |     | DkTokens.ARROW ->
1756 |         "ARROW"
1757 |     | DkTokens.ASSERT _ ->
1758 |         "ASSERT"
...
1823 |     | DkTokens.TYPE _ ->
1824 |         "TYPE"
1825 |     | DkTokens.UNDERSCORE _ ->
1826 |         "UNDERSCORE"
Error (warning 8): this pattern-matching is not exhaustive.
(CCOLON|INT _)
```
To fix this, you could:
* either disable warning 8
* or declare the tokens `CCOLON` and `INT` in `dkParser.mly`;
* or just remove the tokens `CCOLON` and `INT` in `dkTokens.ml`.

The last problem also arises in the `Dedukti` repository.

fblanqui:
fix #797

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -8,7 +8,7 @@
 (maintainers "dedukti-dev@inria.fr")
 (license CECILL-2.1)
 
-(using menhir 1.0)
+(using menhir 2.0)
 
 (package
  (name lambdapi)
--- a/src/common/error.ml
+++ b/src/common/error.ml
@@ -3,10 +3,6 @@
 open Lplib
 open Base
 
-(** Exception raised by the parser when an unqualified identifier is given
-   instead of a qualified one. *)
-exception NotQualified
-
 (** [err_fmt] warning/error output formatter. *)
 let err_fmt = Stdlib.ref Format.err_formatter
 
--- a/src/parsing/dkTokens.ml
+++ b/src/parsing/dkTokens.ml
@@ -30,7 +30,6 @@ type token =
   | DEF
   | COMMA
   | COLON
-  | CCOLON
   | EQUAL
   | ARROW
   | EVAL of loc
@@ -42,4 +41,3 @@ type token =
   | PRINT of loc
   | GDT of loc
   | STRING of string
-  | INT of int
--- a/src/parsing/lpLexer.ml
+++ b/src/parsing/lpLexer.ml
@@ -16,8 +16,11 @@ let remove_ends : Sedlexing.lexbuf -> string = fun lb ->
 
 exception SyntaxError of strloc
 
+let syntax_error : Lexing.position * Lexing.position -> string -> 'a =
+  fun pos msg -> raise (SyntaxError (Pos.make_pos pos msg))
+
 let fail : Sedlexing.lexbuf -> string -> 'a = fun lb msg ->
-  raise (SyntaxError (make_pos (lexing_positions lb) msg))
+  syntax_error (Sedlexing.lexing_positions lb) msg
 
 let invalid_character : Sedlexing.lexbuf -> 'a = fun lb ->
   fail lb "Invalid character"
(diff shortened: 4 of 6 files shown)


fblanqui:
@gabrielhdt I get the following compilation error:
```
11:55 ~/src/lambdapi (menhir) make
Error: Module LpParser in directory _build/default/src/parsing depends on
Parsing.
This doesn't make sense to me.

Parsing is the main module of the library and is the only module exposed
outside of the library. Consequently, it should be the one depending on all
the other modules in the library.
make: *** [Makefile:11 : bin] Erreur 1

11:56 ~/src/lambdapi (menhir) ocamlc --version
4.13.1

11:57 ~/src/lambdapi (menhir) menhir --version
menhir, version 20211128
```
Do you have any idea how to fix it?

gabrielhdt:
> 

Is it the case that menhir uses internally a module named `Parsing` (in the generated code) which would conflict with our module `Parsing`?

fblanqui:
Perhaps it is https://ocaml.org/api/Parsing.html ?

gabrielhdt:
Ah indeed. So it looks like we have to change `Parsing` for something else.

fblanqui:
Hopefully not. See https://gitlab.inria.fr/fpottier/menhir/-/merge_requests/28.

fblanqui:
Finally, the problem may not be related to this. In `_build/default/src/parsing/lpParser.ml`, I get on line 521:
```
and ('s, 'r) _menhir_cell1_equation = 
  | MenhirCell1_equation of 's * ('s, 'r) _menhir_state * (Parsing.Syntax.p_term * Parsing.Syntax.p_term) * Lexing.position
```
Strangely, this is the only place where I have Syntax.p_term prefixed by Parsing.
I don't know why and do not see how to fix this.
So I opened an issue on Menhir: https://gitlab.inria.fr/fpottier/menhir/-/issues/61.

fpottier:
Yes, I am able to reproduce the problem. It is **not** caused by a name clash with the module `Parsing` in the standard library. Also, as you noted, the problem appears with Menhir 20211128 already, so not specific of the new code back-end. Apparently the problem is caused by the fact that `ocamlc` infers the type `Parsing.Syntax.p_term * Parsing.Syntax.p_term` for the symbol `equation`, whereas we would like it to infer `Syntax.p_term`.

fpottier:
Apparently, declaring `%type  <Syntax.p_term * Syntax.p_term> equation` in `lpParser.mly` fixes the problem.

fpottier:
With Menhir 20211230, I get a new error:
```
File "src/parsing/lpParser.mly", line 349, characters 35-40:
Error: Unbound constructor Infix
```
This is interesting. This use of `Infix` in a semantic action relies on type-directed disambiguation, and the manner in which Menhir 20211230 generates code does not allow type-directed disambiguation to take place. This will probably require a fix in Menhir.

fpottier:
Meanwhile, I would recommend replacing `Infix` with`Sign.Infix` in your code. This is enough to remove the problem.

fblanqui:
Thank you @fpottier for your help!

fpottier:
I have submitted a PR to `alt-ergo`, to achieve compatibility with Menhir 20211230, but it has not been reviewed yet.

PR #801, opened 2022-01-04
Add command export to translate a signature to the dk format

fblanqui:
- add command export with option -o/--output <FMT> where <FMT> can be dk or lp
- remove command beautify superseded by -o lp
- files tool/hrs.ml and tool/xtc.ml moved to directory export/


gabrielhdt:
This is very interesting. Could you detail briefly how you manage to export? Do you iter through the AST? Are tactic scripts converted to lambda terms?

fblanqui:
The dk export does not work on the ast but on the signature (I export complete terms). I use source file position information to order symbol and rule declarations.

src/export/dk.ml
```diff
@@ -0,0 +1,183 @@
...
+    (Array.pp tevar ", ") xs
+    path p ident n (List.pp (prefix " " term) "") r.lhs term rhs
+
+let decl : decl pp = fun ppf decl ->
+  match decl with
+  | Sym s -> sym_decl ppf s
+  | Rule r -> rule_decl ppf r
+
+(** [decls_of_sign sign] computes an ordered list of declarations from the
+   signature [sign]. *)
```
gabrielhdt, line 164: I think it would be wise to state in the comments that the list is ordered by the position of the declaration in the source file.

src/export/dk.ml
```diff
@@ -0,0 +1,183 @@
...
+    ; "#ASSERTNOT", ASSERTNOT loc
+    ; "#PRINT", PRINT loc
+    ; "#GDT", GDT loc]
+
+let is_ident : string -> bool = fun s ->
+  Parsing.DkLexer.is_ident (Lexing.from_string s)
+
+let escape : string pp = fun ppf s -> out ppf "{|%s|}" s
+
+let ident : string pp = fun ppf s ->
```
gabrielhdt, line 58: What happens when the symbol uses utf8? Does Dedukti handles utf8 in symbols?

src/export/dk.ml
```diff
@@ -0,0 +1,183 @@
+(** Export a Lambdapi signature to Dedukti. *)
+
+open Lplib open Base open Extra
+open Timed
+open Common
+open Core open Term
```
gabrielhdt, line 6: Newlines missing (probably)

src/cli/config.ml
```diff
@@ -12,6 +12,9 @@ module CLT = Cmdliner.Term
 
 (** {3 Configuration type for common values} *)
 
+(** Output formats for the export command. *)
+type output = Lp | Dk
```
gabrielhdt, line 16: Polymorphic variants may also be used, to avoid creating a new type (using `` `Lp`` instead of `Lp`)

gabrielhdt (commented):
It would be nice to format new PRs using ocamlformat, for this the new files to be formatted can be added to a file `.ocamlformat-enable` at  the root of the repository.

src/export/dk.ml
fblanqui: The symbol is escaped.

src/export/dk.ml
fblanqui: No.

src/cli/config.ml
fblanqui: What would be the advantage?

fblanqui:
The export can handle almost all files in tests/OK except a few ones because of restrictions in Dedukti (see tests/export.sh).
To check this, do `tests/export.sh`.

fblanqui:
@gabrielhdt Thank you for your comments. Let me know if you have more.

PR #806, opened 2022-01-07
Logic: add PTS encoding

fblanqui:


francoisthire:
FYI: The encoding of PTS proposed by Cousineau and Dowek only works for functional PTS.

fblanqui:
This is written in the README.md and the lp files as well.

fblanqui:
Other remarks @francoisthire and @thiagofelicissimo?

francoisthire:
My bad, I read the `arbitrary` and not the functional. By the way, the `arbitrary` could be contested too: the PTS relations Rules and Axioms need to be decidable. For axiom it means that given two sorts x and y, we can decide whether `(x,y)` is an axiom. Equality over sorts need to be decidable too. I tried to sum up those decidability conditions in the Chapter 1 of my Phd Thesis by the way.

Maybe it is a niptick though, but "arbitrary" is a bit vague IMHO.

thiagofelicissimo:
A not very important remark: strictly speaking, this is not Cousineau and Dowek's encoding, as you are building it over an encoding of sets. If you want an encoding which can be parametrized by the pts specification, you can use the ideas of Ali/Gaspard/François and directly encode a type of sorts. For instance, you could take the encoding from my draft and in https://github.com/thiagofelicissimo/examples-encodigs and replace abs and app by the rule El _ (prod s1 s2 _ A B) --> Pi x : El s1 A. El s2 (B x). However, then you would get an encoding that has not been formalized. If you want an encoding of PTSs not using abs and app, parametrized by sorts and that has been formalized and proven, then Ali and François prove in their thesis the soudness of encodings for CTSs, which use this idea of parametrisation.

fblanqui:
Thanks for your comments. @thiagofelicissimo note the small typo in your filename: encodigs -> encodings. In your encoding proposal, A and R are functions. I prefer to declare a and r as relations directly. Moreover, A and R are symbols and not parameters: you cannot have several PTSs in the same file.

PR #812, opened 2022-01-12
Fix meta printing

### LINKED ISSUE #811: Printed metavariables are not parsable
Metavariables are printed as `?X` or `?X[x;y;z]` which are both incorrect. To see this, one may add `flag "print_implicits" on` in `tests/OK/245.lp` and look at the logs of `dune exec -- lambdapi check tests/OK/245.lp --debug i`.

gabrielhdt:
Metavariables are always printed with their context. Fixes #811.


src/core/print.ml
```diff
@@ -231,7 +220,7 @@ and pp_term : term pp = fun ppf t ->
     | Type        -> out ppf "TYPE"
     | Kind        -> out ppf "KIND"
     | Symb(s)     -> pp_sym ppf s
-    | Meta(m,e)   -> out ppf "%a%a" pp_meta m pp_env e
+    | Meta(m,e)   -> out ppf "%a.%a" pp_meta m pp_env e
```
fblanqui, line 223: The dot should be included in pp_env since it is also needed for Patt. Then, dtrees.expected should be updated accordingly.

fblanqui (changes requested):

src/core/print.ml
fblanqui: Fixed in https://github.com/Deducteam/lambdapi/pull/812/commits/fecec3b56ce9160ac12f008e234cc6a69a4a1bb1.

PR #815, opened 2022-01-14
allow / and * in identifiers

### LINKED ISSUE #814: Identifiers and forbidden letters
https://github.com/Deducteam/lambdapi/blob/089f239ad951c8848e6b1394fa38dfe2eacd3a1c/src/parsing/lpLexer.ml#L156

Why are the characters `/`, `?`, `*`, `|` and `$` forbidden? As far as I understand, allowing them wouldn't clash with the language.

fblanqui:
fix #814

### DIFF
--- a/Logic/Makefile
+++ b/Logic/Makefile
@@ -1,7 +1,7 @@
 PKG := TFF U Zenon PTS
 
 .PHONY: default
-default:
+default: clean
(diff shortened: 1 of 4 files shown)


fblanqui:
@gabrielhdt Are you happy with this solution?

gabrielhdt:
I think the less constraints we have, the better it is, so this is already satisfying. Thank you.

PR #818, opened 2022-01-19
Extended identifiers

gabrielhdt:
Identifiers may contain $, ? or |, but not /. They may not begin with $ or ?. The identifier | is not valid (since it's the VBAR).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/),
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## Unreleased
+
+### Changed
+
+- Identifiers may contain `$` or `?` but may not begin with
+  them. They may contain `|` as well. Identifiers must not contain `/`.
+
 ## 2.1.0 (2022-01-17)
 
 ### Added
--- a/src/parsing/lpLexer.ml
+++ b/src/parsing/lpLexer.ml
@@ -153,8 +153,11 @@ let string = [%sedlex.regexp? '"', Star (Compl '"'), '"']
 
 (** Unqualified regular identifiers are any non-empty sequence of characters
    not among: *)
-let forbidden_letter = [%sedlex.regexp? Chars " ,;\r\t\n(){}[]:.`\"@$|/"]
-let regid = [%sedlex.regexp? '/' | '*' | Plus (Compl forbidden_letter)]
+let forbidden_letter = [%sedlex.regexp? Chars " ,;\r\t\n(){}[]:.`\"@/"]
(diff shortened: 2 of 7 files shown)


CHANGES.md
```diff
@@ -3,6 +3,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/),
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## Unreleased
+
+### Changed
+
+- Identifiers may contain `$` or `?` but may not begin with
+  them. They may contain `|` as well. Identifiers must not contain `/`.
```
fblanqui, line 11: If they can contain VBAR, this may conflict with constructor declarations in inductive type definitions.

fblanqui: Why not allowing `/` as (single-letter) identifier?

gabrielhdt: I don't think there would be a conflict with constructors: a constructor is declared with `| Foo` so you parse the vbar and `Foo`. We force that there is a space between the vbar and the identifier, which is rather natural since tokens are separated by spaces.

fblanqui: But what if the user forgets a space and write a|b ?

gabrielhdt: Well it's incorrect! Tokens are separated by spaces, so it's incorrect as much as `symbolfoo` is not `symbol foo`, isn't it?

gabrielhdt: Maybe we could try to follow standards and do like Rust https://doc.rust-lang.org/reference/identifiers.html, that is using standard sets of codepoints defined in https://www.unicode.org/reports/tr31/tr31-33.html.

gabrielhdt: Unfortunately, math symbols are not included in identifiers. But according to [this source](https://jonathanwarden.com/tag/unicode-identifier/), it is as good to keep the identifiers and maths symbols as separate classes. which we can do. We can say that an identifier is either `xid_start, Star xid_continue` or `Plus math`.

gabrielhdt: So it seems that either we want to work around something like this
```ocaml
let ident = [%sedlex.regexp? xid_start, Star (xid_continue | Chars "'")]
let regid = [%sedlex.regexp? ident | Plus math]
```
One of the main advantage is that people have taken care of removing ambiguous characters from these sets (such as the duplication between the Greek mu `μ` and the micro sign `μ`), and we can rely on the unicode standards to keep the set up to date, and our specification is handled by the standard (plus or minus some implementation specific details which shouldn't be that numerous).

amelieled:
Very fun things can appear with this signature: `type -> type -> type`.

gabrielhdt:
> Very fun things can appear with this signature: `type -> type -> type`.

What do you mean?

PR #819, opened 2022-01-20
Configuring reduction

gabrielhdt:
The goal of this pull request is to have a usable interface to parametrise the reduction engine. Additionally we'd like to be able to not rewrite some symbols.

This is preliminary work to #809.


CHANGES.md
```diff
@@ -3,6 +3,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/),
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## Unreleased
+
+### Changed
+
+- `whnf` function taks a problem as argument and a list of tags that configure
```
fblanqui, line 10: typo: taks -> takes

src/core/eval.ml
```diff
@@ -77,38 +79,55 @@ let snf : (term -> term) -> (term -> term) = fun whnf ->
...
 
-(** Configuration of the reduction engine. *)
-type config =
-  { context : ctxt (** Context of the reduction used for generating metas. *)
-  ; defmap : term VarMap.t (** Variable definitions. *)
-  ; rewrite : bool (** Use user-defined rewrite rules. *)
-  ; problem : problem (** Generated metavariables. *) }
-
-(*let pp_defmap = D.map VarMap.iter pp_var " ≔ " pp_term "; "*)
+type rw_tag = [ `NoBeta | `NoRw | `NoExpand ]
```
fblanqui, line 82: Why negative tags instead of positive tags? I would prefer positive tags (Beta, Rewrite, Def).

fblanqui: Why using backquotes?

src/core/eval.ml
```diff
@@ -77,38 +79,55 @@ let snf : (term -> term) -> (term -> term) = fun whnf ->
...
-      | Some v -> unfold v
-    end
-  | _ -> a
+(** Configuration of the reduction engine. *)
+module Config = struct
+
+  type t =
+    { context : ctxt
+    (** Context of the reduction used for generating metas. *)
+    ; defmap : term VarMap.t (** Variable definitions. *)
```
fblanqui, line 90: To avoid possible confusions, I suggest to rename defmap into varmap.

src/core/eval.ml
```diff
@@ -77,38 +79,55 @@ let snf : (term -> term) -> (term -> term) = fun whnf ->
...
+    let rewrite = not @@ List.mem `NoRw rewrite in
+    {context; defmap = Ctxt.to_map context; rewrite; expand_defs; beta; problem}
+
+  let unfold : t -> term -> term = fun c a ->
+    let a = unfold a in
+    match a with
+    | Vari x ->
+      begin match VarMap.find_opt x c.defmap with
+        | None -> a
+        | Some v -> unfold v
```
fblanqui, line 113: Shouldn't this function be recursive? `let rec unfold c a = let a = Term.unfold a in ...`

src/core/eval.ml
```diff
@@ -470,42 +496,49 @@ let hnf =
...
+    [c] with no side effects. *)
 let pure_eq_modulo : ctxt -> term -> term -> bool = fun c a b ->
   Timed.pure_test (fun (c,a,b) -> eq_modulo c a b) (c,a,b)
 
 (** [whnf c t] computes a whnf of [t], unfolding the variables defined in the
    context [c], and using user-defined rewrite rules if [~rewrite]. *)
-let whnf : ?rewrite:bool -> ctxt -> term -> term = fun ?(rewrite=true) c t ->
+let whnf :
+     ?problem:problem
+  -> ?rewrite:rw_tag list
```
fblanqui, line 516: rewrite -> tags

src/core/eval.ml
```diff
@@ -470,42 +496,49 @@ let hnf =
...
+  Debug.(
+    record_time Rewriting (fun () -> r := whnf ?problem ?rewrite c t));
+  !r
 
 (** [simplify t] computes a beta whnf of [t] belonging to the set S such that:
 - terms of S are in beta whnf normal format
 - if [t] is a product, then both its domain and codomain are in S. *)
 let rec simplify : term -> term = fun t ->
-  match get_args (whnf ~rewrite:false [] t) with
+  let rewrite = [`NoRw; `NoExpand ] in
```
fblanqui, line 535: extra space before ]

fblanqui (changes requested):
Seems fine; just a few thing to change.

fblanqui:
Do you want to replace rewrite:bool by rewrite:SymSet.t in this PR or in another PR?

src/core/eval.ml
gabrielhdt: I used negative tags because I assume that the default behaviour is to have the full rewrite engine. So default would be `[Beta; Rw; Expand]`, but we don't want to pass all these arguments each time. Backquotes allow to use again these constructors for other types, it's mainly for convenience (we can use ordinary variants).

gabrielhdt:
> Do you want to replace rewrite:bool by rewrite:SymSet.t in this PR or in another PR?

Regarding that, I can do it in this PR, it's quite lightweight, but I wondered whether we'd like to use such a set to not rewrite opaque symbols (that is, add opaque symbols to the set at some point), or if we want to keep both opacity tags and the SymSet.

src/core/eval.ml
gabrielhdt: Yes I think it should indeed.

src/core/eval.ml
gabrielhdt: Btw, this `defmap` is only optimisation, isn't it?

fblanqui: Probably.

fblanqui:
Do we have the need for this now?

src/core/eval.ml
gabrielhdt: Using backquotes also prevents from the need to qualify the identifiers, so in `unif.ml` we can write ``Eval.whnf [`NoRw]`` instead of ``Eval.whnf [Eval.NoRw]`` (and instead of ``Eval.whnf [Eval.Beta; Eval.Expand]`` if we take positive tags). It's all for convenience.

gabrielhdt:
> Do we have the need for this now?

Not yet. It can be done later on. With this api it's easier to change the rewriting engine anyway.

src/core/eval.ml
gabrielhdt: We should use automatic formatting for that kind of stuff

fblanqui:
Not rewriting pair to pair' is not compatible with the current implementation of eq_modulo for dealing with proof irrelevance. We perhaps need to introduce some new symbol modifier for declaring irrelevant arguments.

gabrielhdt:
> Not rewriting pair to pair' is not compatible with the current implementation of eq_modulo for dealing with proof irrelevance. We perhaps need to introduce some new symbol modifier for declaring irrelevant arguments.

Indeed, but this pull request does not deal with that (yet). It's just an opportunity to have a more extensible api.

fblanqui:
Please confirm that you want to merge it now as it is.
I don't mind to merge it. But I do not see why it is useful to do what you want to do on coercions.

gabrielhdt:
Indeed, I'm not sure either. But it might: tags such as opaque can be translated as a configuration option like, say, a function
``` ocaml
rewrite_sym: sym -> bool
```
embedded in the configuration, and the configuration may be accumulated in the signature state.

But anyway if we agree that the api proposed here is more convenient than the old one, I'm in favour of merging the PR right away.

PR #824, opened 2022-01-26
Add incremental local confluence checking

fblanqui:
### Added

- Incremental local confluence checking for non higher-order and non AC rules.
- Add options -o hrs and -o xtc to the export command.

### Changed

- Do not print empty term environments `.[]`.
- Allow users to use the pattern variables `$0`, `$1`, etc. and internally name pattern variables by their index.
- Fixed debug flag printing in Pretty.


fblanqui:
I think there was a bug in Sign.link:
https://github.com/Deducteam/lambdapi/blob/0e03d8a1f10ae49d0ee2ac67985a1276d49d4973/src/core/sign.ml#L124-L128
https://github.com/Deducteam/lambdapi/blob/0e03d8a1f10ae49d0ee2ac67985a1276d49d4973/src/core/sign.ml#L137
Decision trees of symbols in dependencies were not updated since sign.sign_symbols only contains the symbols of the current signature.

PR #826, opened 2022-01-28
fix #780

### LINKED ISSUE #780: Pure compilation is not that pure
Static symbols such as `Unif_rule.equiv` are modified by `Pure.compile`, in particular their rules. For instance, using `dune utop` (where the prompt is symbolised by the hash `#`)
```ocaml
# open Core
open Term
open Common
open Timed;;
# Library.set_lib_root None;;
- : unit = ()
# List.length @@ !(Unif_rule.equiv.sym_rules);;
- : int = 0
# ignore @@ Compile.Pure.compile_file "tests/OK/unif_hint.lp";;
Loading "/home/hondet/workspace/deducteam/lambdapi/tests/OK/unif_hint.lp" ...
[...]
- : unit = ()
# List.length @@ !(Unif_rule.equiv.sym_rules);;
- : int = 4
```

fblanqui:
fix #780

### DIFF
--- a/src/handle/compile.ml
+++ b/src/handle/compile.ml
@@ -111,8 +111,8 @@ let compile_file : string -> Sign.t = fun fname ->
 let pure_apply_cfg :
   ?lm:Path.t*string -> ?st:Console.State.t -> ('a -> 'b) -> 'a -> 'b =
   fun ?lm ?st f x ->
-  let libmap = !lib_mappings in
-  let unif_rules = !(Unif_rule.equiv.sym_rules)
+  let libmap = !lib_mappings
+  and unif_rules = !(Unif_rule.equiv.sym_rules)
   and unif_dtree = !(Unif_rule.equiv.sym_dtree) in
   Console.State.push ();
   Option.iter Library.add_mapping lm;
--- a/src/handle/compile.ml
+++ b/src/handle/compile.ml
@@ -4,7 +4,7 @@ open Lplib
 open Timed
 open Common open Error open Library
 open Parsing
-open Core open Sign
+open Core open Sign open Term
 
 (** [gen_obj] indicates whether we should generate object files when compiling
     source files. The default behaviour is not te generate them. *)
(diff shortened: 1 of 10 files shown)


src/handle/compile.mli
```diff
@@ -21,13 +21,11 @@ val compile_file : string -> Sign.t
...
-  val compile_file : ?lm:Path.t*string -> ?st:State.t -> string -> Sign.t
-end
+(** /!\ For testing only. Pure wrappers around compilation functions. The
+   provided functions perform the same computations as the ones defined above,
+   but restore the state when they have finished. An optional library mapping
+   or state can be passed as argument to change the settings. *)
+val pure_compile :
+  ?lm:Path.t*string -> ?st:Console.State.t -> bool -> Path.t -> Sign.t
+val pure_compile_file :
+  ?lm:Path.t*string -> ?st:Console.State.t -> string -> Sign.t
```
gabrielhdt, line 31: Were these modifications necessary? Changing the API shouldn't be done without concertation IMHO

src/handle/compile.ml
```diff
@@ -104,40 +104,39 @@ let compile_file : string -> Sign.t = fun fname ->
...
+    lib_mappings := libmap;
+    Unif_rule.equiv.sym_rules := unif_rules;
+    Unif_rule.equiv.sym_dtree := unif_dtree;
+    Console.State.pop ()
+  in
+  try let res = f x in restore (); res
+  with e -> restore (); raise e
 
-  let compile ?lm ?st force mp =
+(** /!\ For testing only. Pure wrappers around compilation
```
gabrielhdt, line 129: It's not for testing only: programs that use lambdapi through its API often want to compile potentially unrelated signatures.

src/handle/compile.mli
gabrielhdt: Furthemore, it's quite frustrating to see one's work be undone without any notice.

fblanqui: I removed Compile.Pure to avoid confusion with the Pure module.

src/handle/compile.ml
fblanqui: I see. Could you please provide me with some examples?
As said in a previous comment, I removed Compile.Pure to avoid confusion with Pure (in src/pure/ directory).
It is annoying to have two different interfaces.
We should clarify this.
There is currently no documentation (except the source code) explaining which one to use.

gabrielhdt: https://github.com/Deducteam/personoj/blob/c731827d3bdf238b7a70f5868f087c10eb846662/psnj_toolbox/bin/meta.ml#L14

gabrielhdt: IIRC, there was some comment by Emilio that said that ideally the `Pure` as in `src/pure` should be merged with the main code.

fblanqui: Sorry, I didn't know that you were using Compile.Pure, but this is easy to fix in your code.

fblanqui: Comment fixed in d903e61c.

PR #831, opened 2022-01-31
pure_compile is not that pure

fblanqui:
Rename pure_compile, following the discussion in #780.

### DIFF
--- a/src/handle/compile.ml
+++ b/src/handle/compile.ml
@@ -104,40 +104,38 @@ let compile_file : string -> Sign.t = fun fname ->
   (* Run compilation. *)
   compile Stdlib.(!recompile) mp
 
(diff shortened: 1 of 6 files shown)


gabrielhdt (approved):
I was wondering whether a function to reset the static symbols should be provided. In the case of normal symbols, it should be known that if A depends on B, then compiling A may modify B. But in the case of unification symbols, critical pairs &c., even if A does not depend on B, there can be interferences.

fblanqui:
But such a function would not be correct, as the correctness of A may depend on unification rules used.

gabrielhdt:
Well correctness depends on the proposition you consider. In a sense the compilation is already incorrect, as suggests the non termination problem raised in you PR. So I don't think that's an issue since the specification would simply be to reset the rewrite rules of ghost symbols. Then the process `compile; reset; compile` may have an undefined behaviour

PR #832, opened 2022-01-31
Fix admitted

### LINKED ISSUE #830: Assert failure on opaque definition containing metavariables
`lambdapi check --gen-obj FILE` crashes with `Uncaught ["Assert_failure src/core/sign.ml:175:22"].` on
```
symbol A: TYPE;
symbol a: A;

opaque symbol foo: A ≔
  let v ≔ _ in
  v
begin
admitted;
```

Probably there is a meta generated by the `let ... in` which is not caught by the refine because the definition is opaque.

fblanqui:
fix #830

### DIFF
--- a/src/handle/command.ml
+++ b/src/handle/command.ml
@@ -438,7 +438,7 @@ let get_proof_data : compiler -> sig_state -> p_command -> cmd_output =
               (* We admit all the remaining typing goals. *)
               let admit_goal ss g =
                 match g with
-                | Unif _ -> fatal pos "Remaining unification goals."
+                | Unif _ -> fatal pos "Cannot admit unification goals."
                 | Typ gt ->
                   let m = gt.goal_meta in
                   match !(m.meta_value) with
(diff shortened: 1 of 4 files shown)


src/handle/command.ml
```diff
@@ -435,21 +435,16 @@ let get_proof_data : compiler -> sig_state -> p_command -> cmd_output =
...
-                    | Typ gt ->
-                        let m = gt.goal_meta in
-                        match !(m.meta_value) with
-                        | None -> Tactic.admit_meta ss p_sym_nam.pos m
-                        | Some _ -> ss
-                  in List.fold_left admit_goal ss ps.proof_goals
+              (* We admit all the remaining typing goals. *)
+              let admit_goal ss g =
+                match g with
+                | Unif _ -> fatal pos "Remaining unification goals."
```
gabrielhdt, line 441: It's OK by me as it is, but aren't there cases where unification goals may be solved by user supplied unification rules in the proof?

gabrielhdt (dismissed):
It's not clear how to deal with unification goals, but in my opinion what you did is the safest.
**Edit:** I forgot it was for admitted goals, it's perfect this way.

src/handle/command.ml
gabrielhdt, line 441: I'd rather write "Cannot admit unification goals" (which explains a bit why it failed).

PR #849, opened 2022-02-22
do not type check terms twice in definitions

### LINKED ISSUE #842: For definitions with types, type-checking is done twice
With
```
symbol A:TYPE;
symbol a:A;
debug +i;
symbol b : A ≔ a;
```
the output is:
```
17:04 ~/src/lambdapi (db) lp tmp/a.lp 
Loading "/home/blanqui/src/lambdapi/tmp/a.lp" ...
debug +i
[infr] Top check sort A
[infr] Type enforce [A]
[infr] Infer [A]
[infr] Inferred [A: TYPE]
[infr] Top check a : A
[infr] Force [a] of [A]
[infr] Infer [a]
[infr] Inferred [a: A]
[infr] Top check a : A
[infr] Force [a] of [A]
[infr] Infer [a]
[infr] Inferred [a: A]
```

fblanqui:
fix #842

### DIFF
--- a/src/handle/tactic.ml
+++ b/src/handle/tactic.ml
@@ -110,7 +110,7 @@ let tac_refine : ?check:bool ->
     if check then
       match Infer.check_noexn p c t gt.goal_type with
       | None ->
-        fatal pos "%a@ does not have type@ %a."  term t  term gt.goal_type
+        fatal pos "%a@ does not have type@ %a." term t term gt.goal_type
       | Some t -> t
     else t
   in
--- a/src/handle/tactic.ml
+++ b/src/handle/tactic.ml
@@ -97,19 +97,23 @@ let tac_solve : popt -> proof_state -> proof_state = fun pos ps ->
 
 (** [tac_refine pos ps gt gs p t] refines the typing goal [gt] with [t]. [p]
    is the set of metavariables created by the scoping of [t]. *)
-let tac_refine :
+let tac_refine : ?check:bool ->
       popt -> proof_state -> goal_typ -> goal list -> problem -> term
       -> proof_state =
-  fun pos ps gt gs p t ->
+  fun ?(check=true) pos ps gt gs p t ->
   if Logger.log_enabled () then
     log_tact "@[tac_refine@ %a@]" term t;
   let c = Env.to_ctxt gt.goal_hyps in
   if LibMeta.occurs gt.goal_meta c t then fatal pos "Circular refinement.";
   (* Check that [t] has the required type. *)
-  match Infer.check_noexn p c t gt.goal_type with
-  | None ->
-      fatal pos "%a@ does not have type@ %a."  term t  term gt.goal_type
-  | Some t ->
+  let t =
+    if check then
+      match Infer.check_noexn p c t gt.goal_type with
+      | None ->
+        fatal pos "%a@ does not have type@ %a."  term t  term gt.goal_type
+      | Some t -> t
+    else t
+  in
   if Logger.log_enabled () then
     log_tact (Color.red "%a ≔ %a") meta gt.goal_meta term t;
   LibMeta.set p gt.goal_meta (binds (Env.vars gt.goal_hyps) lift t);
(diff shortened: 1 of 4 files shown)


fblanqui:
Hello @gabrielhdt . Could you please check that this modification is fine with personoj ?

fblanqui:
This PR improves efficiency quite a lot:

| library | master | PR | % |
| ------- | ------ | -- | --- |
| focalide | 25s | 20s | -20% |
| holide | 9m5s | 5m50s | -36% |
| matita | 7m15s | 4m12s | -42% |

fblanqui:
Concerning the difference of efficiency between dkcheck and lambdapi, note that:
- The files that are checked are dk files generated for dkcheck and not for lambdapi.
- Matita files contain no thm command: there is no distinction between definitions (that may be unfolded) and proofs (that usually do not need to be unfolded). So Lambdapi could perhaps get better results if the thm command was used.

gabrielhdt:
It makes `tests/KO/246.lp` to succeed, I'm looking at it.

gabrielhdt:
> So, is it ok if I merge this PR in master?

I guess, anyway there seem to be a lot of catch up to do between the master branch and the coercion branch.

src/handle/tactic.ml
```diff
@@ -97,19 +97,23 @@ let tac_solve : popt -> proof_state -> proof_state = fun pos ps ->
...
   let c = Env.to_ctxt gt.goal_hyps in
   if LibMeta.occurs gt.goal_meta c t then fatal pos "Circular refinement.";
   (* Check that [t] has the required type. *)
-  match Infer.check_noexn p c t gt.goal_type with
-  | None ->
-      fatal pos "%a@ does not have type@ %a."  term t  term gt.goal_type
-  | Some t ->
+  let t =
+    if check then
+      match Infer.check_noexn p c t gt.goal_type with
```
gabrielhdt, line 111: Shouldn't `Query.check` be used here?

fblanqui: I don't think so because we do not want to call solve at this stage yet. Solve is called after the refinement on all constraints generated by this check + the previously existing constraints, while Query.check tries to solve only the constraints of p which does not include the previously existing constraints (since we need to know what are the new ones). What do you think?

gabrielhdt: But isn't it the case that the problem `p` already contains constraints generated previously? How come `p` does not include previous constraints? There is only one `new_problem ()` in this section, isn't it?

fblanqui: p is given as argument to tac_refine. It is used to record new metas/goals and new constraints when applying a tactic. See for instance tac_induction. Hence also the call to add_goals_of_problem.

gabrielhdt: Sure, but the  `p` given to `tac_refine` contains everything collected so far, so refine must be able to say whether the unification problem is satisfiable or not. Or are there  cases where a unification problem is provably not satisfiable, but we don't want `tac_refine` to fail?

fblanqui: Exactly: unsolvable constraints are added as new goals.

gabrielhdt: Ah I'm afraid I was confused and thought of
https://github.com/Deducteam/lambdapi/blob/f89b2c32e397f0fc3860b48c6d4ad99b18c47fd7/src/handle/command.ml#L419

gabrielhdt: So it seems clear for this line and PR, no problem. For the one I just pasted, it may be discussed somewhere else.

fblanqui: In command.ml, we need to do something on t to replace place holders by metas.
But that's right that we should fail if check returns None here.

fblanqui: This is fixed in https://github.com/Deducteam/lambdapi/pull/849/commits/5f3fb142395150308aace5b64aeb5ad95606c2a9.
But we may wonder on the contrary why we could not replace the calls to Query.check_sort and Query.infer by calls to Infer.check_sort_noexn and Infer.check_noexn so as to be more general since the generated unsolved constraints could perhaps be solved by tactic applications. What do you think?

fblanqui: Well, currently, this would make no difference because we have no tactic other than solve for unification goals. But this could happen later.

fblanqui: Done in https://github.com/Deducteam/lambdapi/pull/849/commits/1050abd43ba0737a15ea096bd8cf9551fd0c971a. Your comments are welcome.

Issue #840, opened 2022-02-10
The option --no-colors doesn't work


fblanqui:
This problem has been introduced in #709. @yannl35133 could you please try to fix it?

PR #850, opened 2022-02-22
Fix --no-color

yannl35133:
Fixes https://github.com/Deducteam/lambdapi/issues/840


yannl35133:
NB: Stream is deprecated in 4.14 so dune doesn't compile if I don't suppress the alerts everywhere

fblanqui:
Thank you @yannl35133 for your PR! But I don't understand your comment on streams and dune.

yannl35133:
https://github.com/ocaml/ocaml/blob/4.14/stdlib/stdlib.mli#L1440-L1441
```ocaml
module Stream       = Stream
[@@deprecated "Use the camlp-streams library instead."]
```

PR #864, opened 2022-05-07
Upper bound on Why3 version

gabrielhdt:
Compilation fails with 1.5.0


dune-project
```diff
@@ -58,7 +58,7 @@ the Why3 platform.")
   (bindlib (>= 5.0.1))
   (timed (>= 1.0))
   (pratter (>= 1.2))
-  (why3 (>= 1.4.0))
+  (why3 (>= 1.4.0 & < 1.5.0))
```
fblanqui, line 61: Are you sure of the syntax?
```
File "dune-project", line 61, characters 20-21:
61 |   (why3 (>= 1.4.0 & < 1.5.0))
                         ^
Error: Too many argument for >=
```

gabrielhdt: Indeed I'm not

PR #867, opened 2022-05-19
fix code with why3 1.5.0

fblanqui:


fblanqui:
Related issue in Why3: https://gitlab.inria.fr/why3/why3/-/issues/647

PR #885, opened 2022-07-11
Coercions

gabrielhdt:
Simple coercions

TODO
- [x] use a better strategy than SNF to reduce coercions
- [x] ensure holes generated by coercion are properly transformed into existential variables

This pull request won't include unification during type checking, nor rewrite rules protection.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 - (API) the rewrite engine can match on the constant `TYPE`
+- Automatic coercion insertion mechanism.
+  For example, the command `coerce_rule coerce Int Float $x ↪ FloatOfInt $x;`
+  can be used to instruct Lambdapi to automatically coerce integers to floats
+  using the function `FloatOfInt`.
 
 ### Changed
 
--- a/doc/commands.rst
+++ b/doc/commands.rst
@@ -21,7 +21,7 @@ Multi-line comments are opened with ``/*`` and closed with ``*/``. They can be n
       ignored /* these ones too */ */
 
 .. _require:
-      
+
 ``require``
 -----------
 
@@ -417,6 +417,47 @@ transformed into ``?x ≡ bool``.
 
 *WARNING* This feature is experimental and there is no sanity check
 performed on the rules.
+
+.. _coerce_rule:
+
+``coerce_rule``
+---------------
+
+Lambdapi can be instructed to insert function applications into terms whenever
+needed for typability. These functions are called *coercions*. For instance,
+assuming we have a type ``Float``, a type ``Int`` and a function
+``FloatOfInt : Int → Float``, the latter function can be declared
+as a coercion from integers to floats with the declaration
+
+::
+
+    coerce_rule coerce Int Float $x ↪ FloatOfInt $x;
+
+Symbol ``coerce`` is a built-in function symbol that computes the coercion.
+Whenever a term ``t`` of type ``Int`` is found when Lambadpi expected a
+``Float``, ``t`` will be replaced by ``coerce Int Float t`` and reduced.
+The declared coercion will allow the latter term to be reduced to
+``FloatOfInt t``.
+
+Coercions can call the function ``coerce`` recursively,
+which allows to write, e.g.
+
+::
+
+    coerce_rule coerce (List $a) (List $b) $l ↪
+      map (λ e: El $a, coerce $a $b e) $l
+
+where ``Set: TYPE;``, ``List : Set → TYPE``, ``El : Set → TYPE`` and ``map`` is
+the usual mapping operator on lists such that ``map f (cons x l) ≡ cons (f x)
+(map l)``.
+
+*WARNING* Coercions are still experimental, and will not mix well is
+metavariables. Indeed, the term ``coerce ?1 Float t`` will not reduce to
+``FloatOfInt t`` even if the equation ``?1 ≡ Int`` has been registered during
+typing. Furthermore, for the moment, it is unsafe to have symbols that can be
+reduced to protected symbols in the right-hand side of coercions:
+reduction may occur during coercion elaboration,
+which may generate unsound protected symbols.
 
 .. _inductive:
 
--- a/doc/lambdapi.bnf
+++ b/doc/lambdapi.bnf
@@ -24,10 +24,11 @@ QID ::= [UID "."]+ UID
               <inductive>)* ";"
             | "rule" <rule> ("with" <rule>)* ";"
             | "builtin" <stringlit> "≔" <qid_or_nat> ";"
+            | "coerce_rule" <rule> ";"
             | "unif_rule" <unif_rule> ";"
             | "notation" <qid_or_nat> <notation> ";"
             | <query> ";"
-            
+
 
 <query> ::= <assert> <param_list>* "⊢" <term> ":" <term>
           | <assert> <param_list>* "⊢" <term> "≡" <term>
--- a/src/core/eval.ml
+++ b/src/core/eval.ml
@@ -67,7 +67,7 @@ let snf : (term -> term) -> (term -> term) = fun whnf ->
     | Appl(t,u)   -> mk_Appl(snf t, snf u)
     | Meta(m,ts)  -> mk_Meta(m, Array.map snf ts)
     | Patt(i,n,ts) -> mk_Patt(i,n,Array.map snf ts)
-    | Plac _      -> assert false
+    | Plac _      -> h (* may happen when reducing coercions *)
     | TEnv(_,_)   -> assert false
     | Wild        -> assert false
     | TRef(_)     -> assert false
@@ -377,7 +377,7 @@ and tree_walk : config -> dtree -> stack -> (term * stack) option =
             let stk = List.reconstruct left [] right in
             walk t stk cursor vars_id id_vars
           in
-          Option.bind fn default
+          Option.bind default fn
         else
           let s = Stdlib.(!steps) in
           let (t, args) = whnf_stk cfg examined [] in
@@ -402,7 +402,7 @@ and tree_walk : config -> dtree -> stack -> (term * stack) option =
               let stk = List.reconstruct left [] right in
               walk d stk cursor vars_id id_vars
             in
-            Option.bind fn default
+            Option.bind default fn
           in
           (* [walk_binder a  b  id tr]  matches  on  binder  [b]  of type  [a]
              introducing variable  [id] and branching  on tree [tr].  The type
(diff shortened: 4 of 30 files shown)


doc/commands.rst
```diff
@@ -418,6 +418,46 @@ transformed into ``?x ≡ bool``.
...
+typability. These functions are called *coercions*. For instance, assuming
+we have a type ``Float``, a type ``Int`` and a function
+``FloatOfInt : Int → Float``, the latter function can be declared
+as a coercion from integers to floats with the declaration
+
+::
+
+    coercion coerce Int Float $x ↪ FloatOfInt $x;
+
+Symbol ``coerce`` is a meta-function that computes the coercion.
```
fblanqui, line 436: meta-function -> builtin function symbol?

doc/commands.rst
```diff
@@ -418,6 +418,46 @@ transformed into ``?x ≡ bool``.
 *WARNING* This feature is experimental and there is no sanity check
 performed on the rules.
 
+.. _coercion:
+
+``coercion``
+------------
+
+Lambdapi can be instructed to insert functions into terms whenever needed for
```
fblanqui, line 426: insert functions -> insert function applications?

doc/commands.rst
```diff
@@ -418,6 +418,46 @@ transformed into ``?x ≡ bool``.
...
+Coercions can call the function ``coerce`` recursively,
+which allows to write, e.g.
+
+::
+
+    coercion coerce (List $a) (List $b) $l ↪
+      map (λ e: El $a, coerce $a $b e) $l
+
+where ``Set: TYPE;``, ``List : Set → TYPE``, ``El : Set → TYPE`` and ``map`` is
+the usual mapping operator on lists such that ``map f (cons x l) = cons (f x)
```
fblanqui, line 451: = -> \equiv

doc/commands.rst
```diff
@@ -418,6 +418,46 @@ transformed into ``?x ≡ bool``.
...
+(map l)``.
+
+*WARNING* Coercions are still experimental, and will not mix well is
+metavariables. Indeed, the term ``coerce ?1 Float t`` will not reduce to
+``FloatOfInt t`` even if the equation ``?1 = Int`` has been registered during
+typing. Furthermore, it is unsafe to use coercions that compute to symbol that
+can be reduced to protected symbols, for the reduction may generate the
+protected symbols.
```
fblanqui, line 459: Is it possible to have protected symbols? Isn't it the case that protected symbols are forbidden at scoping time?

doc/lambdapi.bnf
```diff
@@ -24,10 +24,11 @@ QID ::= [UID "."]+ UID
               <inductive>)* ";"
             | "rule" <rule> ("with" <rule>)* ";"
             | "builtin" <stringlit> "≔" <qid_or_nat> ";"
+            | "coercion" <rule> ";"
```
fblanqui, line 27: Why not coerce_rule to indicate that this is a rule, and follow the pattern we already have with unif_rule?

src/core/eval.ml
```diff
@@ -67,7 +67,9 @@ let snf : (term -> term) -> (term -> term) = fun whnf ->
     | Appl(t,u)   -> mk_Appl(snf t, snf u)
     | Meta(m,ts)  -> mk_Meta(m, Array.map snf ts)
     | Patt(i,n,ts) -> mk_Patt(i,n,Array.map snf ts)
-    | Plac _      -> assert false
+    | Plac _      -> h (* may happen when reducing coercions *)
+        (* TODO: the code for coercions should be updated to replace holes by
+           metavariables before reduction. *)
```
fblanqui, line 72: TODO element. When/where will it be done?

src/core/infer.ml
```diff
@@ -41,8 +41,72 @@ let unif : problem -> octxt -> term -> term -> unit =
...
+  (* Some bindings *)
+  let open Option.Monad in
+  (* Transform a list of options into [Some] list when there is no [None] in
+     the input. *)
+  let lift : 'a option list -> 'a list option = fun x ->
+    List.fold_right
+      (fun elt acc -> Option.Applicative.(pure List.cons <*> elt <*> acc))
+      x
+      (return [])
+  in
```
fblanqui, line 58: Move to option or list.

src/core/infer.ml
```diff
@@ -41,8 +41,72 @@ let unif : problem -> octxt -> term -> term -> unit =
 
 (** {1 Handling coercions} *)
 
-(* Simply unify types, no coercions yet. *)
-let coerce pb c t a b = unif pb c a b; (t, false)
+(** [simplify c t] tries to reduce coercions that are in term [t]. The
```
fblanqui, line 44: simplify -> reduce_coercions (to avoid possible confusion with other reduction funtions)

src/core/libTerm.ml
```diff
@@ -35,8 +35,8 @@ let remove_impl_args : sym -> term list -> term list = fun s ts ->
   remove s.sym_impl ts
 
 (** [iter f t] applies the function [f] to every node of the term [t] with
-   bound variables replaced by [Kind]. Note: [f] is called on already unfolded
-   terms only. *)
+    bound variables replaced by [Kind]. Note: [f] is called on already unfolded
+    terms only.*)
```
fblanqui, line 39: Useless change. It is better to remove it.

src/lplib/option.ml
```diff
@@ -8,15 +8,14 @@ let is_None : 'a t -> bool = fun o ->
 let get : 'a -> 'a option -> 'a = fun d o ->
   match o with None -> d | Some e -> e
 
+let bind : 'a t -> ('a -> 'b t) -> 'b t = Stdlib.Option.bind
+
```
fblanqui, line 12: Can't we remove it completely if it is in stdlib? (Perhaps only after some ocaml version we have in CI?)

src/lplib/option.ml
```diff
@@ -38,3 +37,16 @@ let eq : 'a eq -> 'a option eq = fun eq_elt o1 o2 ->
 
 let pp : 'a pp -> 'a option pp = fun elt ppf o ->
   match o with None -> () | Some e -> elt ppf e
+
+module Monad = struct
+  let ( let* ) = Stdlib.Option.bind
```
fblanqui, line 42: Could you please add a comment explaining how to use this notation. "let* x = t in u" is the same as "bind ...".

doc/lambdapi.bnf
fblanqui: No need to change coercion everywhere, just in the concrete syntax.

fblanqui (changes requested):
Thank you Gabriel!

src/core/eval.ml
gabrielhdt: This note is not relevant anymore: rewrite rules must generate holes rather than existential variables. What could be done is to develop an SR-check for coercions, so that implicit arguments may instantiated automatically in coercion rules, like they are in rewrite rules.

doc/commands.rst
fblanqui: Sorry but I don't understand. This is your comment that is talking about protected symbols. So why are you talking about ghost symbols here. I am confused. What do you want to tell with your remark "it is unsafe to use coercions that compute to symbol that can be reduced to protected symbols"? Would you have an example? pair can reduce to the protected symbol pair'. Do you mean that we shouldn't use pair in a coercion rule? I guess no. So could you please clarify all this?

fblanqui:
Hi Gabriel. You need to update the doc and the bnf to use coerce_rule instead of coercion.

doc/commands.rst
gabrielhdt: I mentioned *ghost* symbols because they are the ones that are forbidden at scoping time, that's why I suspected a confusion. In that comment, I say that if a coercion rule contains a symbol that can reduce to a protected variant, then the protection can be violated.

For instance, let's take the `pair`, `pairp` protection scheme. In presence of the rule `coerce_rule coerce (El $a) (El (psub $a $p)) $e --> pair $a $p $e _`,
everything seems ok because we reduce to `pair`. However, in the function `reduce_coercions` of `infer.ml`, many subterms are reduced to `whnf`, so the reduction `pair $a $p $e _ --> pairp $a $p $e` may be triggered during the coercion elaboration.

fblanqui: It seems related to some discussion we had before about limiting applicable rules when reducing a coercion. Is it right? Do you plan to do something about that?
Anyway isn't it problematic to say that this is unsafe but still have to use it in some developments? Perhaps you could instead say that the *current* implementation may be unsafe but we will try to do something about it.

doc/commands.rst
```diff
@@ -418,6 +418,46 @@ transformed into ``?x ≡ bool``.
 *WARNING* This feature is experimental and there is no sanity check
 performed on the rules.
 
+.. _coercion:
+
+``coercion``
```
fblanqui, line 423: -> coerce_rule

gabrielhdt: Ah right, right

CHANGES.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 - Propagate recompile flag to dependencies.
 - Postfix operators with the `notation <op> postfix <priority>;`
+- Coercions
```
fblanqui, line 24: This should be on line 11. And more details are welcome.

doc/commands.rst
gabrielhdt: > It seems related to some discussion we had before about limiting applicable rules when reducing a coercion. Is it right?

It is, indeed.

> Anyway isn't it problematic to say that this is unsafe but still have to use it in some developments?

It is, but I don't use that version of lambdapi for PVS related things, I use another branch with more reduction options for symbols (I have added a tag for this).

> Perhaps you could instead say that the current implementation may be unsafe but we will try to do something about it.

Sounds good to me.

> Do you plan to do something about that?

I can't pledge to fix it, but I may be able to try something sometime.

fblanqui: Aren't you going to include those "reduction options for symbols" in master?

fblanqui: What are they?

gabrielhdt: The mechanism I used is not very elegant, I added the tag `redopaque` for the symbol `pair`, and the reduction for coercion would reduce all but `redopaque` symbols.

gabrielhdt: I can at least post an issue on the bugtracker.

fblanqui: Can't we imagine to mark every symbol having a rule RHS containing a protected symbol as "progen" (add field sym_progen : bool ref) and do not reduce those symbols when reducing coercions?

gabrielhdt: Yes, it seems it's what I implemented. It feels clunky, but it works.

fblanqui: You prefer to add this in a separate PR?

gabrielhdt: Yes I'd prefer, this one allows already interesting features (such as setting `El` as a coercion).

PR #889, opened 2022-07-14
Matching and rewriting constant TYPE

gabrielhdt:
This pull request allows the rewriting engine to match on constant `TYPE`. This will allow to create coercions of the form
```
coercion coerce Set $x TYPE --> El $x;
```


fblanqui:
It is fine for me. Just update CHANGES. And in which order do you prefer me to merge your PRs?

PR #898, opened 2022-07-28
make install command compatible with macos

### LINKED ISSUE #896: "lambdapi install" on macOS
The command `lambdapi install ` makes system calls with invalid parameters on macOS. For example,

- `mkdir --parents ...` should be replaced by `mkdir -p ...`
- `cp --force ...`  should be replaced by `cp -f ...`
- ...

fblanqui:
fix #896

### DIFF
--- a/src/cli/install.ml
+++ b/src/cli/install.ml
@@ -42,14 +42,17 @@ let run_install : Config.t -> bool -> string list -> unit =
       (* Create directories as needed for [dest]. *)
       let cmd =
         let dest_dir = Filename.dirname dest in
-        String.concat " " ["mkdir"; "--parents"; dest_dir]
+        String.concat " " ["mkdir"; "-p"; dest_dir]
+        (* use short option for compatibility with MacOS *)
       in
       run_command dry_run cmd;
       (* Copy files. *)
       let cmd =
         let file = Filename.quote file in
         String.concat " "
-          ["cp"; "--preserve"; "--no-target-directory"; "--force"; file; dest]
+          ["cp"; "-p"; "-f"; (*"--no-target-directory";*) file; dest]
+          (* use short option for compatibility with MacOS.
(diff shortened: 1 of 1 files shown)


gabrielhdt (approved):
LGTM, these options are POSIX (see https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cp.html and https://pubs.opengroup.org/onlinepubs/9699919799/utilities/mkdir.html#tag_20_79)

src/cli/install.ml
```diff
@@ -42,14 +42,17 @@ let run_install : Config.t -> bool -> string list -> unit =
       (* Create directories as needed for [dest]. *)
       let cmd =
         let dest_dir = Filename.dirname dest in
-        String.concat " " ["mkdir"; "--parents"; dest_dir]
+        String.concat " " ["mkdir"; "-p"; dest_dir]
+        (* use short option for compatibility with MacOS *)
```
gabrielhdt, line 46: It's for posix compliance rather than macos compatibility

src/cli/install.ml
```diff
@@ -42,14 +42,17 @@ let run_install : Config.t -> bool -> string list -> unit =
...
       in
       run_command dry_run cmd;
       (* Copy files. *)
       let cmd =
         let file = Filename.quote file in
         String.concat " "
-          ["cp"; "--preserve"; "--no-target-directory"; "--force"; file; dest]
+          ["cp"; "-p"; "-f"; (*"--no-target-directory";*) file; dest]
+          (* use short option for compatibility with MacOS.
+             There seems to be no option --no-target-directory in MacOS. *)
```
gabrielhdt, line 55: There is indeed no long option in posix' cp https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cp.html

gabrielhdt (commented):
LGTM, I'd just make the comment more posix oriented

PR #910, opened 2022-09-10
Fix Dedukti export

gabrielhdt:
Properly brace-quote (`{|...|}`) module identifiers that cannot be parsed by Dedukti


gabrielhdt:
It seems that Dedukti does not handle brace-quoting of module identifiers, at least in `#REQUIRE` directives.

fblanqui:
Thanks Gabriel!

PR #948, opened 2023-01-28
Document interaction prefix with →

gabrielhdt:
Document how operators mix with arrows at the level of types.


fblanqui (approved):

fblanqui:
Thanks @gabrielhdt !

PR #987, opened 2023-05-03
Fix highlighting of multi-line comments in vim

bpandreotti:
The vim syntax file provided in the repo did not correctly detect multi-line comments.


fblanqui:
Thank you for your PR!

fblanqui:
It main be the case that there are other incompatibilities with the current version of Lambdapi because the vim mode has not been maintained properly along the changes in Lambdapi. I will check it during the EPN meeting in Liège on May 11-13. But don't hesitate to propose other PRs to improve this very minimal mode.

PR #997, opened 2023-07-02
fix order of parameters in the type of generated induction principles

### LINKED ISSUE #989: Error when my inductive type have more than one parameter
Hi!
When I try to add a second parameter to the inductive type 𝕃, I get the error:

`Cannot solve $130 □ ≡ $130 □`

Here is a snippet of code to reproduce the problem:

```
symbol Prop : TYPE;
injective symbol Prf : Prop → TYPE;
builtin "Prop" ≔ Prop;
builtin "P" ≔ Prf;

symbol Set : TYPE; 

(a:Set) (b:Set) inductive 𝕃:TYPE ≔
| □  : 𝕃 a b
| ⸬   : τ a → τ b → 𝕃 a b → 𝕃 a b;
```

Thank you in advance!

fblanqui:
fix #989

### DIFF
--- a/src/handle/inductive.ml
+++ b/src/handle/inductive.ml
@@ -209,7 +209,7 @@ let fold_cons_type
     | _ -> fatal pos "The type of %a is not supported" sym cons_sym
   in
   let _, t = Env.of_prod_using [] vs !(cons_sym.sym_type) in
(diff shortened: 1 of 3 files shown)


fblanqui:
There is currently a problem with gitlab.inria.fr: `[ERROR] Failed to get sources of menhir.20230415: Curl failed`.

PR #1016, opened 2023-11-01
WIP: Add opaque command

NotBad4U:
This PR add a new command to declare a symbol as opaque after it has been defined. It also fixes the function is_injective so that opaque symbols are considered as injective since they can never be unfolded.

TODO
- [x] update is_injective
- [x] update CHANGES.md
- [x] update doc

-------------
I was able to modify `sym_opaq` without making a pointer to it (`sym_opaq  : bool ref`).

@fblanqui , setting the `sym_opaq` at true seems to be not enough.
In the test that I added _OK/opaque.lp_, when I call the tactic `simplify;`  in `∨ᶜᵢ₁'`
the symbol `πᶜ` is correctly Not unfold, but when I do `apply ∨ᶜᵢ₁` (for testing purpose),
it crashed with this error:
```
[/Users/alessiocoltellacci/Junk/lambdapi/tests/OK/opaque.lp:46:4-14] Missing subproofs (0 subproofs for 5 subgoals):
p: Prop
q: Prop
Hp: πᶜ p

--------------------------------------------------------------------------------
0. πᶜ (p ∨ᶜ q) ≡ π ⊥
1. ?10: Prop
2. ?11: Prop
3. ?12: πᶜ ?10.[p;q;Hp]
4. ?13: π (¬ (?10.[p;q;Hp] ∨ᶜ ?11.[p;q;Hp]))
```
I am mostly surprised by the two subgoals
- `0. πᶜ (p ∨ᶜ q) ≡ π ⊥`
- `4. ?13: π (¬ (?10.[p;q;Hp] ∨ᶜ ?11.[p;q;Hp]))`
It keeps using the definition of the symbol made opaque.
I tried to set `sym_def` at `None` when the opaque command is used but I got the same error.
Any idea where I can have a look?


src/handle/command.ml
```diff
@@ -204,6 +204,13 @@ let get_proof_data : compiler -> sig_state -> p_command -> cmd_output =
   Console.out 3 (Color.cya "%a") Pos.pp pos;
   Console.out 4 "%a" Pretty.command cmd;
   match elt with
+  | P_opaque {elt = (_, id); _} ->
+    let ss = {
+      ss with in_scope =  StrMap.update id (fun s ->  match s with
+      | Some (si) -> Some ({si with sym_opaq = true ; sym_def = ref None })
+      | _ -> fatal pos "Can not make opaque %s because it does not exists" id  ) ss.in_scope
```
fblanqui, line 211: English typo: exists -> exist

src/parsing/syntax.ml
```diff
@@ -289,7 +289,7 @@ type p_command_aux =
   | P_unif_rule of p_rule
   | P_coercion of p_rule
   | P_query of p_query
-
+  | P_opaque of p_qident
```
fblanqui, line 292: blank line missing after

src/parsing/syntax.ml
```diff
@@ -613,6 +613,7 @@ let fold_idents : ('a -> p_qident -> 'a) -> 'a -> p_command list -> 'a =
 
   let fold_command : 'a -> p_command -> 'a = fun a {elt;pos} ->
     match elt with
+    | P_opaque _ -> a
```
fblanqui, line 616: please remove `-> a` here

fblanqui (changes requested):

tests/OK/opaque.lp
```diff
@@ -0,0 +1,48 @@
...
+rule π ($p ⇒ $q) ↪ π $p → π $q;
+
+symbol ¬ p ≔ p ⇒ ⊥; notation ¬ prefix 35;
+
+constant symbol ∨ : Prop → Prop → Prop; notation ∨ infix left 6; // \/ or \vee
+constant symbol ∨ᵢ₁ [p q] : π p → π (p ∨ q);
+
+symbol πᶜ p ≔ π (¬ ¬ p);
+
+symbol ∨ᶜ p q ≔ (¬ (¬ p)) ∨ (¬ (¬ q)); notation ∨ᶜ infix right 20;
```
fblanqui, line 22: remark: the prefix notation should allow you to instead write:
```
symbol ∨ᶜ p q ≔ ¬ ¬ p ∨ ¬ ¬ q;
```

fblanqui:
The absence of definition at a given moment is not sufficient to consider a symbol as injective: rules can be added later.
You need to update the function is_injective as well: opaque symbols should be considered as constants and thus injective.

NotBad4U:
> The absence of definition at a given moment is not sufficient to consider a symbol as injective: rules can be added later. You need to update the function is_injective as well: opaque symbols should be considered as constants and thus injective.

It seems to not be enough because I have the same behavior.

```
p: Prop
q: Prop
Hp: πᶜ p
--------------------------------------------------------------------------------
0. πᶜ (p ∨ᶜ q) ≡ π ⊥
1. ?10: Prop
2. ?11: Prop
3. ?12: πᶜ ?10.[p;q;Hp]
4. ?13: π (¬ (?10.[p;q;Hp] ∨ᶜ ?11.[p;q;Hp]))
```

fblanqui:
Other problems to fix:
```
File "src/parsing/pretty.ml", lines 310-344, characters 2-5:
310 | ..begin match elt with
311 |   | P_builtin (s, qid) -> out ppf "@[builtin \"%s\"@ ≔ %a@]" s qident qid
312 |   | P_inductive (_, _, []) -> assert false (* not possible *)
313 |   | P_inductive (ms, xs, i :: il) ->
314 |     let with_ind ppf i = out ppf "@,%a" (inductive "with") i in
...
341 |     end
342 |   | P_unif_rule ur -> out ppf "unif_rule %a" unif_rule ur
343 |   | P_coercion c -> out ppf "%a" (rule "coerce_rule") c
344 |   end.
Warning 8 [partial-match]: this pattern-matching is not exhaustive.
P_opaque _
File "src/parsing/pretty.ml", lines 310-344, characters 2-5:
310 | ..begin match elt with
311 |   | P_builtin (s, qid) -> out ppf "@[builtin \"%s\"@ ≔ %a@]" s qident qid
312 |   | P_inductive (_, _, []) -> assert false (* not possible *)
313 |   | P_inductive (ms, xs, i :: il) ->
314 |     let with_ind ppf i = out ppf "@,%a" (inductive "with") i in
...
341 |     end
342 |   | P_unif_rule ur -> out ppf "unif_rule %a" unif_rule ur
343 |   | P_coercion c -> out ppf "%a" (rule "coerce_rule") c
344 |   end.
Warning 8 [partial-match]: this pattern-matching is not exhaustive.
P_opaque _
Warning: one state has shift/reduce conflicts.
Warning: 2 shift/reduce conflicts were arbitrarily resolved.
```

src/core/term.ml
```diff
@@ -374,7 +374,7 @@ let is_constant : sym -> bool = fun s -> s.sym_prop = Const
 (** [is_injective s] tells whether [s] is injective, which is in partiular the
    case if [s] is constant. *)
 let is_injective : sym -> bool = fun s ->
-  match s.sym_prop with Const | Injec -> true | _ -> false
+  s.sym_opaq || (match s.sym_prop with Const | Injec -> true | _ -> false)
```
fblanqui, line 377: could be simplified to:
```
match s.sym_prop with Const | Injec -> true | _ -> s.sym_opaq
```

fblanqui:
There is something wrong in your code. The type of the field sym_opaq of the record type sym must be changed to bool ref.

NotBad4U:
The test passed now by changing the type of `sym_opaq ` to `bool ref`! However, I am not sure to understand what is the necessity to have a pointer to make it work. I updated the doc, but it is you who will update `CHANGES.md` right?

fblanqui:
No, please update CHANGES.md yourself.
Why a ref is need, this is because sym's are shared: there is only one sym in memory for a given symbol.

src/parsing/scope.ml
```diff
@@ -647,7 +647,7 @@ let scope_rule :
     fatal p_lhs.pos
       "Symbol %s has been declared constant, it cannot be used as the \
        head of a rewrite rule LHS." pr_sym.sym_name;
-  if pr_sym.sym_opaq || Timed.(!(pr_sym.sym_def)) <> None then
+  if Timed.(!(pr_sym.sym_opaq)) || Timed.(!(pr_sym.sym_def)) <> None then
     fatal p_lhs.pos "No rewriting rule can be given on a defined symbol.";
```
fblanqui, line 651: "No rewriting rule can be added on an opaque symbol or a symbol already defined with ≔'

src/parsing/syntax.ml
```diff
@@ -289,8 +289,9 @@ type p_command_aux =
   | P_unif_rule of p_rule
   | P_coercion of p_rule
   | P_query of p_query
+  | P_opaque of p_qident
 
-(** Parser-level representation of a single (located) command. *)
+  (** Parser-level representation of a single (located) command. *)
```
fblanqui, line 294: useless diff

src/parsing/syntax.ml
```diff
@@ -613,6 +614,7 @@ let fold_idents : ('a -> p_qident -> 'a) -> 'a -> p_command list -> 'a =
 
   let fold_command : 'a -> p_command -> 'a = fun a {elt;pos} ->
     match elt with
+    | P_opaque _
```
fblanqui, line 617: This function aims at going though all identifiers. The argument of P_opaque is an identifier. So we should do something with it. P_opaque should work like
```
    | P_opaque qid
    | P_builtin (_, qid)
    | P_notation (qid, _) -> f a qid
```
below

doc/lambdapi.bnf
```diff
@@ -33,6 +33,7 @@ QID ::= [UID "."]+ UID
             | "coerce_rule" <rule> ";"
             | "unif_rule" <unif_rule> ";"
             | "notation" <qid_or_nat> <notation> ";"
+            | "opaque" <qid_or_nat> ";"
```
fblanqui, line 36: Just to be sure: did you generate this file with `make bnf`?

src/parsing/lpParser.mly
```diff
@@ -147,6 +147,7 @@ search_query_alone:
     { q }
 
 command:
+  | OPAQUE i=qid_or_nat SEMICOLON { make_pos $sloc (P_opaque i) }
```
fblanqui, line 150: There is still a problem with this grammar:
```
14:59 ~/lambdapi (add_opaque_cmd) dune build
Warning: one state has shift/reduce conflicts.
Warning: 2 shift/reduce conflicts were arbitrarily resolved.
```
There should be no conflict.

fblanqui (changes requested):

fblanqui:
CI currently fails because of a setup-ocaml failure.

src/parsing/lpParser.mly
fblanqui: If we write `OPAQUE SYMBOL ...` we get:
```
Warning: one state has shift/reduce conflicts.
Warning: one shift/reduce conflict was arbitrarily resolved.
```

doc/lambdapi.bnf
NotBad4U: Yes I did, I just did not push the new .bnf

src/parsing/lpParser.mly
fblanqui: Yes

src/parsing/lpParser.mly
```diff
@@ -167,9 +168,9 @@ command:
         {p_sym_mod=ms; p_sym_nam=s; p_sym_arg=al; p_sym_typ=ao;
          p_sym_trm=fst tp; p_sym_prf=snd tp; p_sym_def=true}
       in make_pos $sloc (P_symbol(sym)) }
-  | ms=modifier* xs=param_list* INDUCTIVE
+  | xs=param_list* INDUCTIVE
     is=separated_nonempty_list(WITH, inductive) SEMICOLON
-      { make_pos $sloc (P_inductive(ms,xs,is)) }
+      { make_pos $sloc (P_inductive([],xs,is)) }
```
fblanqui, line 173: I propose to remove the first argument of P_inductive everywhere.

fblanqui (changes requested):

fblanqui:
Thanks Alessio. I propose you a few cosmetic changes before merging.

src/parsing/lpParser.mly
NotBad4U: Sorry, I made a little mistake. After reviewing the code and documentation, `inductive` type could not use _property_ modifier but they can use _exposition_ modifiers (`private`, `protected`). So I created a subrule `exposition` in the grammar and used it in `modifier` rule and `inductive` case in `command` rule.
I keep a modifier List for `P_inductive` to avoid more modification in this PR, but maybe we can refactor the code in another PR. Moreover, I was not sure that the `constant` modifier was added later in the code to the `ms`, which further encourages keeping a modifier list.

NotBad4U:
If you agree with my last modification, then I think we can merge this PR.

fblanqui:
Thanks Alessio!

PR #1017, opened 2023-11-05
Add upper bound on pratter

gabrielhdt:
Pratter is going to be [updated](https://github.com/ocaml/opam-repository/pull/24746) and its interface will change. To avoid conflicts, make lambdapi depend on the interface of version 2.


fblanqui:
Thanks Gabriel!

fblanqui:
Will you make a PR on Lambdapi once Pratter 3 will be available on Opam?

PR #1036, opened 2024-01-27
Fix compiling pack of rules

### LINKED ISSUE #1033: Wrong evaluation order with sequential modifier
Hi everyone!

When the modifier `sequential` is used on a symbol, the order of execution is reversed.
The last case is evaluated first (`"toto (⊤ ∧ ⊤) ↪ ⊥"` in the below example).

```
sequential symbol toto: Prop →  Prop;
rule toto (⊤ ∧ ⊤) ↪ ⊤
with toto (⊤ ∧ ⊤) ↪ ⊥;

compute toto (⊤ ∧ ⊤);   // compute ⊥ 
```

gabrielhdt:
Rules declared in group appeared to be reverted when handled, causing the "sequential" strategy to not behave as intended (the rule with highest priority was the last to appear in the source file). Fixes #1033


fblanqui:
Thanks Gabriel!

PR #1059, opened 2024-02-28
Documentation : Fix the documentation generation pipeline with ReadTheDoc

Alidra:
This PR fixes the ReadTheDoc build pipeline by adding the .readthedocs.yaml file to the root of the repo where the build options are specified.
This file is required in the recent versions of the readTheDoc tool.


fblanqui:
Thank you @Alidra !

PR #1074, opened 2024-04-04
coq: export opaque definition as opaque definition instead of as Lemma

fblanqui:


fblanqui:
Won't merge finally. This is a bad idea wrt the way Require's are handled in Coq: https://github.com/coq/coq/issues/18891 .

PR #1075, opened 2024-04-04
Fix/logs when missing pkg

### LINKED ISSUE #1069: Error message should be displayed when no lambdapi.pkg is found
As documented in https://lambdapi.readthedocs.io/en/latest/module.html, a lambdapi.pkg must be placed at the root of the source tree of a library under development. When used with the command line, the absence of this file rises an error. However, when used with editors (both Vscode and Emacs), no error is displayed.

Thus, it is important that whenever the lambdapi.pkg file is missing, the editors should display an error message when navigating the proofs.

Alidra:
As documented in https://lambdapi.readthedocs.io/en/latest/module.html, a lambdapi.pkg must be placed at the root of the source tree of a library under development. When used with the command line, the absence of this file rises an error. However, when used with editors (both Vscode and Emacs), no error is displayed.

This PR fixes #1069

### DIFF
--- a/src/lsp/lp_doc.ml
+++ b/src/lsp/lp_doc.ml
@@ -128,7 +128,7 @@ let new_doc ~uri ~version ~text =
       (* We remove the ["file://"] prefix. *)
       assert(String.is_prefix "file://" uri);
       let path = String.sub uri 7 (String.length uri - 7) in
-      Some(Pure.initial_state path), [] 
+      Some(Pure.initial_state path), []
     with Error.Fatal(_pos, msg) ->
       let loc : Pos.pos =
         {
@@ -137,7 +137,7 @@ let new_doc ~uri ~version ~text =
           start_col  = 0;
           end_line = 0;
           end_col = 0
-        } in 
+        } in
       (None, [(1, msg), Some(loc)])
   in
   { uri;
@@ -164,11 +164,11 @@ let dummy_loc =
 
 let check_text ~doc =
   let uri, version = doc.uri, doc.version in
-  let root = 
+  let root =
     match doc.root with
     | Some(ss) -> ss
-    | None -> 
-      raise(Error.fatal_no_pos "Root signature is missing 
+    | None ->
+      raise(Error.fatal_no_pos "Root signature is missing
       probably because new_doc has raised exception")
   in
   try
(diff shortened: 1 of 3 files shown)


src/lsp/lp_doc.ml
```diff
@@ -153,20 +164,27 @@ let dummy_loc =
 
 let check_text ~doc =
   let uri, version = doc.uri, doc.version in
+  let root =
+    match doc.root with
+    | Some(ss) -> ss
+    | None ->
+      raise(Error.fatal_no_pos "Root signature is missing
```
fblanqui, line 171: signature -> state

src/lsp/lp_doc.ml
```diff
@@ -153,20 +164,27 @@ let dummy_loc =
...
+      probably because new_doc has raised exception")
```
fblanqui, line 172: Just to know: will this error message be printed somewhere in the editor?

src/lsp/lp_lsp.ml
```diff
@@ -382,36 +406,39 @@ let hover_symInfo ofmt ~id params =
...
-    (Range.column start - 1),
-    (Range.line finish - 1),
-    (Range.column finish - 1)
-  in
+  try
+    (* The information about the tokens is stored *)
+    let sym =
+      match doc.final with
+      | Some ss -> Pure.get_symbols ss
+      | None    -> raise (Error.fatal_no_pos("Horror")) in
```
fblanqui, line 414: Just to know: will this error message be printed somewhere in the editor?

fblanqui (changes requested):

src/lsp/lp_doc.ml
Alidra: Done.

src/lsp/lp_doc.ml
Alidra: Not displayed in the editor but will be written in the /tmp/lambdapi_lsp_log.txt file

src/lsp/lp_lsp.ml
Alidra: No. The exception will be caught at the end of the function (line 456) and the message discarded. I changed the message though.

PR #1081, opened 2024-04-17
Show the Goal panel if it has lost focus

Alidra:
In Vscode, if the Goal panel looses focus (typically, because user selected another panel) Goals are updated in this panel when the user navigates them in the lp file but he can't see them unless clicking on this panel to bring it in front again.

This PR, gives back focus to the Goals panel to bring it back as soon as the user navigates goals without having to clicking the panel himself.

TODO:
- [ ] test with emacs


Alidra:
Tests with Vscode has been successfully performed. Tests with Emacs still need to be performed to unsure that sending null goals does not break the Emacs client.

Issue #1049, opened 2024-02-20
Displaying subgoals on VS Code


thomastraversie:
On VS Code, when writing a proof, subgoals are correctly displayed on the Goals window. But once the proof is written, the subgoals do not appear anymore if we move the green cursor. For instance, with
```
constant symbol Prop : TYPE;
injective symbol Prf : Prop → TYPE;

constant symbol ∧ : Prop → Prop → Prop; notation ∧ infix right 30;
symbol and_i : Π p : Prop, Prf p → Π q : Prop, Prf q → Prf (p ∧ q);

opaque symbol prop_ad : Π A : Prop, Prf A → Π B : Prop, Prf B → Prf (A ∧ B) ≔ 
begin
    assume A pA B pB;
    refine and_i A _ B _ 
    {}
    {}
end;
```
the 2 subgoals are correctly displayed when the cursor is at the end of the line `refine and_i A _ B _`. Once we have the complete proof:
```
opaque symbol prop_ad : Π A : Prop, Prf A → Π B : Prop, Prf B → Prf (A ∧ B) ≔ 
begin
    assume A pA B pB;
    refine and_i A _ B _ 
    {refine pA} 
    {refine pB}
end;
```
if we move the cursor one step further the line `assume A pA B pB;`, the cursor goes at the first bracket (instead of being at the end of the line `refine and_i A _ B _`) and only the second subgoal is displayed.

Alidra:
Hi @thomastraversie ,
With @fblanqui, we tried to reproduce the issue with the code you provided above but we couldn't. 
The subgoals seem to be displayed correctly. Would you please check again and update the code if needed?

thomastraversie:
Hi @Alidra, I can reproduce the issue (on version 2.5.0 of Lambdapi and v0.2.1 of the VSCode extension). 
When the green cursor is at the end of line `assume A pA B pB;` the goal is correctly `Prf (A ∧ B)`. At the next step the cursor is at the first bracket of line `{refine pA}` and the goal is `Prf B`, so the step where we have the two goals `Prf A` and `Prf B` is missing.

Alidra:
Hi @Thomas,
On my side, I just upgraded lambda to version 2.5.0 and tried again. It seems to work (see attached screenshot)
![image](https://github.com/Deducteam/lambdapi/assets/40537601/630676f1-cb3c-4b27-9974-1397aa89b86a)
Can you attach a screenshot with the problem please.

[...]

PR #1086, opened 2024-04-18
fix parsing error : return commands even when the end of the file contains parsing error

Alidra:
When working with Vscode or Emacs, if a Lambdapi file contains parsing errors, the parsing phase should return the commands that have been successfully parsed (this is already the case with the command line Lambdapi).

This PR fixes this issue.

Additionally, this PR changes the way navigating the proofs is done in Vscode with the `navigate until cursor`.
Instead of stopping the green zone at the beginning of current the line, the green zone is expended to the end of the current command.

Finally, this PR fixes stops the green zone before the `{` delimiter (and incidentally `begin` delimiter) to avoid that the LSP server answers with subgoals of the next tactic missing the current one as described in issue [#1049](https://github.com/Deducteam/lambdapi/issues/1049)


Alidra:
@fblanqui , Can you check if the behavior described in this PR suites you?

Alidra:
@fblanqui, as discussed earlier today, the modifications in this PR work fine with both Vscode and Emacs.
We still need to fix the syntax error message in Emacs terminal but it is better to do it in a separate PR.

Thus, I believe the PR can be merged.

fblanqui:
Thanks Abdelghani!

Issue #706, opened 2021-06-07
Support for negative priorities


gabrielhdt:
See #664.

The Pratt parsing library <https://forge.tedomum.net/koizel/pratter> allows to use negative priorities for operators, but the parser doesn't allow to input negative floats,
```
notation o infix -3;
```
is not allowed.

PR #1093, opened 2024-04-24
Parse integers and negative floats for notations

gabrielhdt:
Fixes issue #706.


fblanqui:
for CI to work, you need to exclude your example in tests/export_raw_dk.sh

PR #1095, opened 2024-04-24
Fix #554

### LINKED ISSUE #554: Remove Parser.lp_lexbuf_fixup ?
introduced in #552

gabrielhdt:
Bump sedlex to 3.2 which fixes its issue #96
https://github.com/ocaml-community/sedlex/issues/96.

I'm not sure the problem is fixed though: all I could see is that running
```command
$ lambdapi check tests/OK/natproof.lp --verbose 4
```
resulted in the same output before and after the commit in this pull request.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -52,7 +52,7 @@ the Why3 platform.")
  (depends
   (ocaml (>= 4.08.0))
   (menhir (>= 20200624))
-  (sedlex (>= 2.2))
+  (sedlex (>= 3.2))
   (alcotest :with-test)
   (dedukti (and :with-test (>= 2.7)))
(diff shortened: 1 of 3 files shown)


fblanqui:
For the record, this is related to https://github.com/Deducteam/lambdapi/issues/549 .

fblanqui:
We should check that this does not affect IDEs (vscode and emacs) as they use from_string.

fblanqui:
@gabrielhdt I was already using sedlex 3.2 on my machine.
@Alidra Note that this affect the PR you are doing on position/green zone.

Alidra (approved):
Just tested with Both Vscode and Emacs and seems to work fine.

PR #1098, opened 2024-04-24
Fix green cursor position in the presence of opening bracket

Alidra:
When Proofs are navigated in Vscode (specifically with the step forward and backward commands) and in the presence of an opening bracket, the green zone is delimited by the opening bracket included. However, this results in erroneous subgoals displaying as described in [this issue](https://github.com/Deducteam/lambdapi/issues/1049#issuecomment-2072085474)

To fix that, this PR make the green zone end just before the opening bracket (and the `begin` keyword) so that the Lsp server correctly answers with the subgoals following the green zone.


Alidra:
I will abandon this PR and move the changes to [this one](https://github.com/Deducteam/lambdapi/pull/1086) because they are related

PR #1112, opened 2024-05-13
Vscode : release version 0.2.2 of the vscode extension

Alidra:
This version implements the following changes/fixes :

- code refactoring of the client for maintenability.
- fix the bug that causes the proof navigation to malfunction when the `Goals` panel is closed by the user. Now the panel is recreated whenever needed. If focus is taken away frol the `Goals` panel, focus is given back to it when user starts navigating proofs again.
- fix bug related to navigating sub-goals : navigating next subgoal stops before `{` instead of after it so that next subgoal is correctly shown in the `Goals` panel.
- change navigation with ``navigate until cursor`` : Navigation includes the command if the cursor is whithin its range instead of the line above. 
- first command is no more systematically navigated. If the current command is the first one, navigating the previous command results in no command being navigated.

Note that the second point in the above list requires the last modifications in the Lambdapi server (currently in master) to work properly. However, the modifications do not seem to break anything if used with a previous version of Lambdapi (At least not the last released one from the market)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,12 +182,22 @@ JQ = $(shell which jq)
 ifneq ($(VSCE),)
 ifneq ($(JQ),)
 EXT = $(shell vsce show --json deducteam.lambdapi 2>/dev/null | jq '.versions[0]' | jq '.version')
+
 .PHONY: publish-vscode-extension
 publish-vscode-extension:
 ifeq ($(EXT), $(shell cat editors/vscode/package.json | jq '.version'))
 	echo "extension already exists. Skip"
 else
+ifeq ($(PAT),)
+	echo "The \"PAT\" secret is not set. Please add a secret named \"PAT\" to be able to publish the Vscode extension."
+else
+ifeq ('master', $(GITHUB_REF_NAME))
 	cd editors/vscode && vsce publish -p ${PAT}
+else
+	echo "Env Variable \"GITHUB_REF_NAME\" is set to \"$(GITHUB_REF_NAME)\". Please set it to \"master\" to force publishing the extension"
 endif
+endif
+endif
+
 endif
 endif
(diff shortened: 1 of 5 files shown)


fblanqui:
Hi Abdelghani. Thank you for your PR. Note that CI/build_vscode_extension failed.

editors/vscode/CHANGES.md
```diff
@@ -3,9 +3,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/),
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
-## [Unreleased]
+## [0.2.2]
```
fblanqui, line 6: Shouldn't it be 0.1.3 instead?

Alidra:
> Hi Abdelghani. Thank you for your PR. Note that CI/build_vscode_extension failed.

Indeed. The secret name is incorrect.

Alidra:
@fblanqui. 
I corrected the name of the secret in the workflow file. I also added an additional condition to unsure that the extension is published only when pushing to master branch (and not to feature branches). I added some comments too in the Makefile to explain when the extension will be published.

Makefile
```diff
@@ -182,12 +182,27 @@ JQ = $(shell which jq)
 ifneq ($(VSCE),)
 ifneq ($(JQ),)
 EXT = $(shell vsce show --json deducteam.lambdapi 2>/dev/null | jq '.versions[0]' | jq '.version')
+
 .PHONY: publish-vscode-extension
 publish-vscode-extension:
 ifeq ($(EXT), $(shell cat editors/vscode/package.json | jq '.version'))
 	echo "extension already exists. Skip"
 else
+ifeq ($(PAT),)
```
fblanqui, line 191: shouldn't it be VSCODE_PAT as mentioned in RELEASE.md ?

Alidra: VSCODE_PAT is the name of the secret in gitHub. PAT is the name of the env variable (as defined in the workflow file line 55)

PR #1119, opened 2024-06-20
fix vsce@2.15.0 warning

Alidra:
Replace `npm install vsce` by `npm install vscode/vsce` in the Makefile to get ride of the warnings displayed at install.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,7 +48,7 @@ jobs:
       - name: generate-vscode-extension
         run: |
           npm install -g @types/vscode
-          npm install -g vsce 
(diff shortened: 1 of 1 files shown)


Alidra:
This PR fixes the first warning related to use of `vsce` instead of `npm install vscode/vsce` so long.
It is not yet clear if it is possible the two other warnings can be avoided. 
If so, this PR will be merged once a solution is implemented.

Alidra:
I believe this PR can be merged as the remaining warnings related to deprecated packages will only be fixed by a new release of vsce that uses more up-to-date dependencies.

PR #1126, opened 2024-07-18
Fix/vscode plugin

Alidra:
Salwa has recently noticed some issues when working with the Vscode plugin of Lambdapi (some intempestive message and the focus on current document being lost each time the proof is navigated).

This PR fixes these issues. Please note that other issues related to updating the console are related to the released version of Lambdapi that should be updated on the Opam packages repository.


Alidra:
The pipeline fails because of what seems to be a big in Vsce package (It did succeed yesterday). I will investigate further.

PR #1129, opened 2024-07-31
new parser for lp files

fblanqui:
This PR provides a new parser for lp files not based on Menhir but written by hand. It is more efficient, provides better error messages and is easier to change later if needed.

Performance of parsing on the translation to lp of the HOL-Light base file hol.ml (1.2 Go): 1m22s with the master branch, and 1m8s with the PR: -16%.


src/parsing/lpParser.mly
```diff
@@ -210,7 +210,7 @@ query:
   | TYPE_QUERY t=term
     { make_pos $sloc (P_query_infer(t, {strategy=NONE; steps=None}))}
   | SEARCH s=STRINGLIT
-    { make_pos $sloc (P_query_search s) }
+    { make_pos $sloc (P_query_search(QBase(QName s))) }
```
fblanqui, line 213: looks wrong

src/parsing/pretty.ml
```diff
@@ -258,7 +282,7 @@ let query : p_query pp = fun ppf { elt; _ } ->
   | P_query_print(Some qid) -> out ppf "print %a" qident qid
   | P_query_proofterm -> out ppf "proofterm"
   | P_query_verbose i -> out ppf "verbose %s" i
-  | P_query_search s -> out ppf "search \"%s\"" s
+  | P_query_search q -> out ppf "search \"%a\"" Search.query q
```
fblanqui, line 285: There should be no \" anymore ?

src/parsing/lpParser.mly
fblanqui: this is a temporary fix for the old parser to compile

sacerdot:
Since you are working on the parser, the old parser did not accept binders of the form   "id+ : term", it only accepted "id+" or "(id+ : term)" (and many more variations). This was a limitation of the Menhir technology (a shift/reduce conflict). In the Rocq parser me and Abdelghani have just written in Bologna, we managed to accept that combination (that is valid in Rocq) by having Menhir parse a larger grammar and then check if the input made sense in the OCaml code. I think that it would make sense to also accept the missing case in your new parser.

Issue #1144, opened 2024-10-11
Emacs/VSCode: error messages and goals are not printed as expected


fblanqui:
```
require open Stdlib.Set Stdlib.Prop Stdlib.Nat Stdlib.List;

symbol sample_1 (n:ℕ) : π ⊤ ≔
begin
  induction
    { } // cursor here
end;
```

*Goals* buffer: "No goals", while it should be:
```
?27: ℕ → π ⊤
```
If there is an error, the *Goals* buffer should be left unchanged.

*lp-logs* buffer: no error is printed, so we don't know what is the problem. We can see it by hover though but a new user may not know this.

fblanqui:
#1145 fixes one half of the problem: the printing of goals.
Remains the second half: printing error messages in the lp-logs buffer.

fblanqui:
fixed in #1150

PR #1145, opened 2024-10-24
Return proof state in case of error related to number of sub proofs

Alidra:
In case of error related to number of sub proofs, the Goals panel should show the remaining sub-proofs despite the error as described in issue #1144. To this end, the Tactic.handle function should return the computed proof state. In this PR, a specific exception is introduced with this information.

### DIFF
--- a/src/handle/tactic.ml
+++ b/src/handle/tactic.ml
@@ -145,6 +145,7 @@ let ind_data : popt -> Env.t -> term -> Sign.ind_data = fun pos env a ->
       begin
         try
           let ind = SymMap.find s !(sign.sign_ind) in
+          let _, ts = List.cut ts ind.ind_nb_params (*remove parameters*) in
           let ctxt = Env.to_ctxt env in
           if LibTerm.distinct_vars ctxt (Array.of_list ts) = None
           then fatal pos "%a is not applied to distinct variables." sym s
--- a/src/lsp/lp_doc.ml
+++ b/src/lsp/lp_doc.ml
@@ -65,7 +65,8 @@ let process_pstep (pstate,diags,logs) tac nb_subproofs =
     pstate, (tac_loc, 4, qres, goals) :: diags, logs
   | Tac_Error(loc,msg) ->
     let loc = option_default loc tac_loc in
-    pstate, (loc, 1, msg, None) :: diags, ((1, msg), loc) :: logs
+    let goals = Some (current_goals pstate) in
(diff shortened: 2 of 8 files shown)


src/lsp/lp_doc.ml
```diff
@@ -63,11 +63,15 @@ let process_pstep (pstate,diags,logs) tac nb_subproofs =
...
-let process_proof pstate tacs logs =
+  | Tac_Error(a, loc,msg) ->
+    match a with 
+    | None -> let loc = option_default loc tac_loc in
+      pstate, (loc, 1, msg, None) :: diags, ((1, msg), loc) :: logs
+    | Some pstate -> 
+      let goals = Some (current_goals pstate) in
+      pstate, (tac_loc, 4, msg, goals) :: diags, logs
+      
+      let process_proof pstate tacs logs =
```
fblanqui, line 74: wrong indentation, should not be in the diff as it is unchanged

fblanqui:
Hi Abdelghani. Thanks for your PR. I need some time to think about it but I won't have it before November 4 probably.

src/handle/tactic.ml
```diff
@@ -391,6 +391,11 @@ let rec handle :
 (** Representation of a tactic output. *)
 type tac_output = proof_state * Query.result
 
+exception Tactic_error of tac_output * Pos.popt option * string
```
fblanqui, line 394: Pos.popt is already an option type. Do you really need another option layer?

fblanqui: This is explained in error.ml.

src/lsp/lp_doc.ml
```diff
@@ -63,11 +63,15 @@ let process_pstep (pstate,diags,logs) tac nb_subproofs =
     let goals = Some (current_goals pstate) in
     let qres = match qres with None -> "OK" | Some x -> x in
     pstate, (tac_loc, 4, qres, goals) :: diags, logs
-  | Tac_Error(loc,msg) ->
-    let loc = option_default loc tac_loc in
-    pstate, (loc, 1, msg, None) :: diags, ((1, msg), loc) :: logs
```
fblanqui, line 68: Why not simply returning here `Some (current_goals pstate)` instead of `None`?

fblanqui:
Hi. The modifications that you propose look complicated to me, and a little bit strange: why creating a new proof state in case of error? Perhaps it is enough to, in lp_deoc.ml line 68, to replace None (no goal) by `Some (current_goals pstate)`, list of current goals?

PR #1150, opened 2024-11-15
Fix the list of messages to be displayed in terminal due to position error

Alidra:
When executing get_logs, the use of `<`instead of `<=` prevents the last message to be displayed in the terminal (this is partially described in issue #1144 )

Besides, the calculation of position in `do_goals` is moved to `get_goals` because it is redundant with the ones in `get_logs`.

Note one could as well remove this code from `get_logs` and keep it in `do_goals` for efficiency. However, it is safer and **probably** better to call it in the functions actually using the position (namely `get_logs` and `get_goals`)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ocaml-version: [5.2.0, 5.1.1, 5.0.0, 4.14.1, 4.13.1, 4.12.1, 4.11.2, 4.10.2, 4.09.1, 4.08.1]
+        ocaml-version: [5.2.0, 5.1.1, 5.0.0, 4.14.2, 4.13.1, 4.12.1, 4.11.2, 4.10.2, 4.09.1, 4.08.1]
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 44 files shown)


fblanqui:
I am wondering: shouldn't we print logs in reverse order so that the last log comes first?

Alidra:
> I am wondering: shouldn't we print logs in reverse order so that the last log comes first? For this, we just need to replace in get_logs, filter_map by filter_rev_map.

Why not? I can try to print logs in reverse order and see how suitable it is for the user.

fblanqui:
@melanie-taprogge @NotBad4U What do you think of changing the order of logs?

NotBad4U:
I think vscode print shows you the last lines in the terminal tab. I believe that changing the order will need me to scroll up to view the last message (that needs to be tested). Maybe the vscode extension can re-reverse the order?

melanie-taprogge:
> @melanie-taprogge @NotBad4U What do you think of changing the order of logs?

Like @NotBad4U, I am also using vscode and am happy with the order the logs are displayed there currently!

fblanqui:
Ok, so it seems that this a problem with emacs only.

PR #1170, opened 2024-12-30
Remove outdated editors/emacs/lambdapi-mode-pkg.el

tarsius:
The information in `<name>-pkg.el` did not agree with the information in `<name>.el`.  This pull-requests addresses that be removing the outdated `<name>-pkg.el`.

While the end-user package manager `package.el` expects a file `<name>-pkg.el`, this should only be generate by the package archive (such as GNU ELPA and MELPA), instead of being tracked in the upstream repository.

The tools used maintain the various *ELPA, do *not* use `<name>-pkg.el` as a data *source*, they only generate it.

- `elpa-admin.el`, the tool used for GNU ELPA and NonGNU ELPA, does *not* use `<name>-pkg.el` as a data source and it never has.

- `package-build.el`, the tool used for Melpa, prefers `<name>.el` but *currently* falls back to get information missing from there from `<name>-pkg.el` instead. I am goint to change that; soon `<name>-pkg.el` will be ignored as a data source by this tool too.

### DIFF
--- a/editors/emacs/lambdapi-mode-pkg.el
+++ b/editors/emacs/lambdapi-mode-pkg.el
@@ -1,5 +1,0 @@
-(define-package "lambdapi-mode" "1.1.0"
-  "A major mode to edit Lambdapi files."
-  '((eglot "1.6")
-    (math-symbol-lists "1.2.1")
-    (highlight "20190710.1527")))


fblanqui:
Thank you very much @tarsius for your PR. We're going to look after this as soon as possible.

tarsius:
Thanks!

Alidra:
Thank you @tarsius for the PR.
If I understand it right, you suggest to remove `<name>-pkg.el` because in our particular case it does not hold any information that is not already in the `<name>.el` file and in all cases, the use of `<name>-pkg.el` is deprecated and will be ignored in the future.

Is that right?

However, I have the following error when trying to install the package locally (with `"M-x package-install-file` ) :
`Wrong type argument: arrayp, nil`

Please note, it works when I install from a tar file that contains the `<name>-pkg.el` file.

Alidra:
Also, please update the editors/emacs/Makefile by removing line 16:  `SRC += lambdapi-mode-pkg.el`

tarsius:
> Is that right?

Yes, except ...

> install the package locally (with `"M-x package-install-file` )

... this is the one use case that *does* require `<name>-pkg.el`.

I don't think many people do that, but since you do, I recommend you keep `lambdapi-mode-pkg.el` and instead duplicate the information in `lambdapi-mode.el`'s header, while making sure that you never forget to update the information in both places. (Since you are used to editing the `lambdapi-mode-pkg.el` file,  you are unlikely to forget about that, so I recommend adding a `;; Do not forget to also update headers in lambdapi-mode.el!` comment to that file.)

Alidra:
Thanks again @tarsius,

Indeed, we use this method to test the plugin locally before releasing it. If you know any better way to do so, please let us know.
Also, we will keep the `<name>-pkg.el` file and make sure it is sync with `<name>.el` for now. In the future, we will generate it locally the first from for testing purposes.

tarsius:
> If you know any better way to do so, please let us know.

Unfortunately I do not prepared any easy to use steps for that. Also see https://github.com/okamsn/loopy/pull/224#issuecomment-2595564466.

Alidra:
This PR has been re submitted in PR#1188 with necessary modifications on the Makefile

fblanqui:
Subsumed by #1188 . Thank you @tarsius .

tarsius:
Thanks to @Alidra and you too!

PR #1174, opened 2025-01-02
Fix why3 tactic (WIP)

### LINKED ISSUE #1165: The why3 encoding is wrong for the quantifier and predicates
Hi everyone,
I know that the why3 support is quite experimental. I am playing with it because I want to reconstruct some proof steps where I do not have a trace. During my experimentation, I noticed that the why3 encoding does not use the quantified variable in its formula.
Consider the example in the test `why3_quantifier.lp`:

```
opaque symbol thm_ex : P (ex o (λ a: T o, imp a a))
≔ begin 
  why3; 
end;
```

It is translated into 
```
theory Task
 predicate p
 type o
 goal main_goal : exists a:o. p -> p
end
```

but the correct translation should be:

```
theory Task
 type o
 goal main_goal : exists a:o. a -> a  (* use the quantified value a and do not generate a predicate p *)
end
```

I also face a problem with predicates. Consider the Lambdapi goal below:
```
0 : πᶜ ((`∀ᶜ x1, ¬ (a <=> x1) ∨ᶜ p x1) = (¬ (a <=> a) ∨ᶜ p a))
```

the why3 encoding output this Task:

```
 theory Task
         
         predicate p
         
         predicate p1
         
         predicate p2
         
         predicate p3
         
         type U
         
         goal main_goal :
           (forall x1:U. not (p3 <-> p2) \/ p1) <-> not (p3 <-> p3) \/ p
         
       end
 ```
First, the encoding should use the bounded variable `x1` and not create a predicate `p3`. Second, it should not create a predicate `p1`, but a predicate that can take a `bool` as argument to encode `p x1` and `p a`.

I am interested to contribute to fix this issue whereas I am lacking of time. Anyone who wants to help me fix this issue will be welcome.

---------
If you want to print a task in Lambdapi, this is my (partial) `git diff` of [why3_tactic.ml](src/handle/why3_tactic.ml)  for printing the Task with `debug +y` activated:
```bash
@@ -138,14 +146,14 @@ let translate_term : config -> cnst_table -> TyTable.t -> term ->
           close [vid] [] t
         in
         (tbl, ty_tbl, tquant)
-    | _ ->
+    | _a, _args ->
         (* If the term [p] is mapped in [tbl] then use it. *)
         try
           let sym = List.assoc_eq (Eval.eq_modulo []) t tbl in
           (tbl, ty_tbl, Why3.Term.ps_app sym [])
         with Not_found ->
           (* Otherwise generate a new constant in why3. *)
-          let sym = Why3.Term.create_psymbol (Why3.Ident.id_fresh "P") [] in
+          let sym = Why3.Term.create_psymbol (Why3.Ident.id_fresh "p") [] in
           ((t, sym)::tbl, ty_tbl, Why3.Term.ps_app sym [])
   in
   match get_args t with
@@ -191,7 +199,9 @@ let encode : Sig_state.t -> Pos.popt -> Env.env -> term -> Why3.Task.task =
   (* Add the goal itself. *)
   let goal = Why3.Decl.create_prsymbol (Why3.Ident.id_fresh "main_goal") in
   (* Return the task that contains the encoded lambdapi formula in Why3. *)
-  Why3.Task.add_prop_decl tsk Why3.Decl.Pgoal goal why3_term
+  let tsk = Why3.Task.add_prop_decl tsk Why3.Decl.Pgoal goal why3_term in
+  log_why3 "@[task 1 is:@\n%a@]@." Why3.Pretty.print_task tsk;
+  tsk
```

fblanqui:
- why3_tactic.ml: many fixes (fix #1165) and changes
- fol.ml: add fields symb_Prop and sym_Set (domains of symb_P and symb_T), and symb_eqv
- tests: merged why3_quantifiers.lp into why3.lp
- rewrite.ml: make get_eq_data return a list instead of an array, and rename local function check_t_or_p into check_codomain_is_Type
- tactic.ml: introduce the functions get_prod_ids and gen_valid_idopts, and improve assume so that it uses the user-defined variable names instead of h1, h2, etc.
- base.ml: add char
- list: add assoc_eq_opt

As before, the why3 tactic starts by doing some assume. Then, the tactic tries to translate the environment as much as possible, and then the goal. The tactic fails if the goal cannot be translated.

A term [t:Set] is translated to a Why3 type if it is algebraic (i.e. built from function applications and variables only) and its variables are in the environment.

A term [t:T _] is translated to a Why3 term if it is algebraic and its variables are in the environment or are quantified variables.

A term [t:P _] is translated to a Why3 formula if it has the form of a formula in predicate calculus. Subterms that are not in predicate calculus are replaced by fresh propositions.

TODO:
- [x] remove debug flags in test files
- [ ] when abstracting a subformula, compute its free variables, check that they are quantified, and replace the subterm by the application of a fresh predicate applied to those variables. Example:
```opaque symbol test11 a p x q : π ((`∀ x:τ a, p x ⇒ (λ y, q y) x) ⇒ p x ⇒ (λ y, q y) x) ≔ begin why3 end;```
- [ ] make translate_term abstract non-Why3 subterms too. Example:
```opaque symbol test12 a p x q : π ((`∀ x:τ a, p x ⇒ q ((λ y,y)x)) ⇒ p x ⇒ q ((λ y,y)x)) ≔ begin why3 end;```
- [ ] take into account rule on T and P; make the translation invariant by reduction

### CODE REVIEW

**Comment by NotBad4U** [comment]
Awesome! Please ping me when I can try it.

**Comment by fblanqui** [comment]
Hi @NotBad4U . I already made quite some improvements in the why3 tactic. But it may not be sufficient for what you want though. Indeed it still does not handle quantifications on propositions, which is out of the scope of first-order logic. However, could you please try it to give me your feedback.

**Comment by NotBad4U** [comment]
I can not compile your branch. I got this error:

```
File "src/handle/why3_tactic.ml", line 418, characters 16-46:
418 |   and limits = {Why3.Call_provers.empty_limits
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound value Why3.Call_provers.empty_limits
Hint: Did you mean empty_limit?
```

I have the version `1.7.2` of the `why3` library.

**Comment by fblanqui** [comment]
You now need why3 1.8.0 (https://github.com/Deducteam/lambdapi/pull/1176).

**Comment by NotBad4U** [comment]
I did some tests and it returned nice first results.
My encoding with Alethe works with propositional extensionality, and I do not use `eqv` but `=` instead.
It would be great if we could support `=`. For example, it will help to prove:
```
symbol S : U;
symbol a: T S;
symbol p: T S → Prop;
opaque symbol test : P (all S (λ x, (or (not (a = x)) (p x)) ) = (or (not (a = a)) (p a))) ≔ begin
  why3
end;
```
I can convert `=` to `eqv` with the tactic rewrite before calling`why3`, but we still need to add the encoding for a term `x = y`.

In this test, the first one failed, but the second passed when `g` was passed as a parameter.
```
symbol ⤳ : U → U → U; notation ⤳ infix right 20;
rule T ($x ⤳ $y) ↪ T $x → T $y;
symbol g: T (S ⤳ o);

opaque symbol test_hol [x] : P (eqv (ex S (λ x, g x)) (not (all S (λ x, not (g x)))))
≔ begin
  why3 // FAIL
end; 

opaque symbol test_hol2 [x] [g: T (S ⤳ o)] : P (eqv (ex S (λ x, g x)) (not (all S (λ x, not (g x)))))
≔ begin
  why3 // PASS
end; 
```

fblanqui:
The failure is due to the fact that, currently, I take into account predicate symbol of type [T a1 -> .. -> T an -> Prop] only.

PR #1182, opened 2025-01-24
extend decimal notation to integers

fblanqui:
- change builtins necessary for parsing decimal numbers:
  ask for 0, 1, .., 10, +, * instead of 0 and +1 only
  a string of digit d1..dn is then interpreted in a more efficient way as dn+10*(dn-1+..)
- this extends to integers if the builtin - is provided
- remove the limit on integers that can be parsed

TODO:
- [x] add builtins for printing binary numbers in decimal notation
- [x] update bnf
- [ ] allow the mapping of decimal notations to different types at the same time ?


NotBad4U:
When we use builtin notation `+1` in two modules, such as `Stdlib.Nat`and `Stdlib.Z`, the last module open is considered.
Should we also add a feature to support number notation with scope e.g.  "1%Nat" and 1%Z"?

fblanqui:
That could be interesting. This is the topic of the last TODO item but I'm not going to do it in this PR.

PR #1188, opened 2025-01-31
Emacs/del outdated files

Alidra:
This PR subsumes #1170 by @tarsius.

The information in <name>-pkg.el did not agree with the information in <name>.el. This pull-requests addresses that be removing the outdated <name>-pkg.el.

However, since <name>-pkg.el is necessary to install the package locally for debugging, an entry is added to the make file to generate it and include it to the `tar` file for local testing only.

### DIFF
--- a/editors/emacs/Makefile
+++ b/editors/emacs/Makefile
@@ -22,7 +22,6 @@ $(NAME)-$(VERSION).tar: $(SRC)
 	cp *.el "$(NAME)-$(VERSION)"
 	tar -cf "$(NAME)-$(VERSION)".tar "$(NAME)-$(VERSION)"
 
-.PHONY: lambdapi-mode-pkg.el
 lambdapi-mode-pkg.el: lambdapi-mode.el
 	$(eval description := "A major mode for editing Lambdapi source code.")
 	$(eval requirements := $(shell grep ";; Package-Requires: " lambdapi-mode.el | sed 's/"/\\"/g' | cut -d' ' -f5-))
--- a/editors/emacs/Makefile
+++ b/editors/emacs/Makefile
@@ -1,7 +1,7 @@
 .POSIX:
 LAMBDAPI_FILE = lambdapi-mode.el
 VERSION = $(shell cat $(LAMBDAPI_FILE) | grep ";; Version:" | sed "s/;; Version: \(.*\)/\1/")
-NAME = $(shell cat $(LAMBDAPI_FILE) | grep "provide " | grep -v legacy | sed "s/(provide '\(.*\))/\1/")
+NAME = $(shell cat $(LAMBDAPI_FILE) | grep "^(provide" | grep ")" | grep -v "legacy" | sed "s/(provide '\(.*\))/\1/")
 # The path to lambdapi built by dune
 LAMBDAPI = ../../_build/install/default/bin/lambdapi
 
@@ -22,8 +22,6 @@ $(NAME)-$(VERSION).tar: $(SRC)
 	mkdir -p "$(NAME)-$(VERSION)"
 	cp *.el "$(NAME)-$(VERSION)"
 	tar -cf "$(NAME)-$(VERSION)".tar "$(NAME)-$(VERSION)"
-
-FILENAME = $(shell head @<)
 
 .PHONY: lambdapi-mode-pkg.el
 lambdapi-mode-pkg.el: $(LAMBDAPI_FILE)
--- a/editors/emacs/Makefile
+++ b/editors/emacs/Makefile
@@ -42,6 +42,4 @@ check: dist
 
 .PHONY: clean
 clean:
-	rm -f "$(NAME)-$(VERSION)".tar
-	rm -rf "$(NAME)-$(VERSION)"
-	rm -f lambdapi-mode-pkg.el
+	rm -fr "$(NAME)-$(VERSION)".tar "$(NAME)-$(VERSION)" lambdapi-mode-pkg.el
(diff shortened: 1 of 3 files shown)


editors/emacs/Makefile
```diff
@@ -23,6 +23,14 @@ $(NAME)-$(VERSION).tar: $(SRC)
 	cp *.el "$(NAME)-$(VERSION)"
 	tar -cf "$(NAME)-$(VERSION)".tar "$(NAME)-$(VERSION)"
 
+.PHONY: lambdapi-mode-pkg.el
```
fblanqui, line 26: Why do you make it phony?

editors/emacs/Makefile
```diff
@@ -1,7 +1,7 @@
 .POSIX:
-
-VERSION = 1.1.0
-NAME = lambdapi-mode
+LAMBDAPI_FILE = lambdapi-mode.el
+VERSION = $(shell cat $(LAMBDAPI_FILE) | grep ";; Version:" | sed "s/;; Version: \(.*\)/\1/")
```
fblanqui, line 3: This can be simplified.
You can replace [cat file | grep string] by  [grep string file].
Idem below.
Alternative for this line: ```awk '/^;; Version: /{print $3;exit}' $(LAMBDAPI_FILE)```

editors/emacs/Makefile
```diff
@@ -1,7 +1,7 @@
...
+NAME = $(shell cat $(LAMBDAPI_FILE) | grep "^(provide" | grep ")" | grep -v "legacy" | sed "s/(provide '\(.*\))/\1/")
```
fblanqui, line 4: Is it worth doing this for NAME which is not going to change?

editors/emacs/Makefile
```diff
@@ -23,6 +23,14 @@ $(NAME)-$(VERSION).tar: $(SRC)
 	cp *.el "$(NAME)-$(VERSION)"
 	tar -cf "$(NAME)-$(VERSION)".tar "$(NAME)-$(VERSION)"
 
+.PHONY: lambdapi-mode-pkg.el
+lambdapi-mode-pkg.el: $(LAMBDAPI_FILE)
+	$(eval description := $(shell cat $< | grep ";;; $(LAMBDAPI_FILE) ---" | sed "s/;;; $(LAMBDAPI_FILE) --- \(.*\)-\*- lexical-binding:.*/\1/"))
```
fblanqui, line 28: Can be simplified:
```head -n1 $< | sed "s/;;; $(LAMBDAPI_FILE) --- \(.*\)-\*- lexical-binding:.*/\1/"```
But is it worth doing this as the description is not going to change?
All the more so since it is used for local testing only.

File: editors/emacs/Makefile [hunk 5]
```diff
@@ -23,6 +23,14 @@ $(NAME)-$(VERSION).tar: $(SRC)
...
+	$(eval requirements := $(shell cat $(LAMBDAPI_FILE) | grep ";; Package-Requires: " | sed "s/;; Package-Requires: \(.*\)/\1/" | sed 's/"/\\"/g' | cut -d' ' -f3-100))
```
fblanqui, line 29: Can be simplified:
```grep ";; Package-Requires: " $(LAMBDAPI_FILE) | sed 's/"/\\"/g' | cut -d' ' -f5-```

File: editors/emacs/Makefile [hunk 6]
```diff
@@ -1,7 +1,7 @@
 .POSIX:
-
-VERSION = 1.1.0
-NAME = lambdapi-mode
+LAMBDAPI_FILE = lambdapi-mode.el
```
fblanqui, line 2: Is it worth introducing this variable as it is not going to change?

fblanqui (changes requested):
Thanks Abdelghani for your PR. There are several things that can be simplified. Could you please add a file editors/emacs/README.md explaining what to do to test the emacs package locally, and remove the install target in Makefile?

editors/emacs/Makefile
Alidra: Could be useful if lambdapi-mode-pkg.el is accidentally altered between two generations. I can remove it if not really useful which would allow one to change the lambdapi-mode-pkg.el file manually before installation.

editors/emacs/Makefile
Alidra: The "problem" is that the plugin will not install if the name is different from the one in `lambdapi-mode.el`. So if it changes there the installation will fail (same for version). So ultimately we would generate every thing statically and it would also work for local debugging but it won't be flexible to any future change. At the same time it does not look too costly for me to do ti this way. 

What do you thing?

editors/emacs/Makefile
fblanqui: I do not understand. PHONY means that this is a target that do not correspond to an actual file, doesn't it?

editors/emacs/Makefile
fblanqui: I understand for VERSION but not so much for NAME.

editors/emacs/Makefile
```diff
@@ -47,3 +44,4 @@ check: dist
 clean:
 	rm -f "$(NAME)-$(VERSION)".tar
 	rm -rf "$(NAME)-$(VERSION)"
+	rm -f lambdapi-mode-pkg.el
```
fblanqui, line 47: remark: you can do it with only one call to rm:
```rm -rf "$(NAME)-$(VERSION)" "$(NAME)-$(VERSION)".tar lambdapi-mode-pkg.el```

PR #1196, opened 2025-02-12
Update to pratter 5

gabrielhdt:
From pratter 4, it's possible to interpret a single token in different ways. This allows for instance to consider the symbol `-` as both the unary and binary minus.


01mf02:
Is the bumping of the minimal OCaml version really required here? Perhaps this is now not necessary anymore after our monad omission ...? :)

01mf02:
I guess you could then very easily replace `Seq.uncons` by something equivalent in Pratter. It's currently used only here:

~~~ ocaml
let tok : ('tok, 'tok option) parser =
 fun inp ->
  match Seq.uncons inp with
  | None -> Ok (None, Seq.empty)
  | Some (t, rest) -> Ok (Some t, rest)
~~~

Unfolding [`Seq.uncons`](https://ocaml.github.io/odoc/ocaml-base-compiler/src/stdlib/seq.ml.html#L91), this could easily be something like (modulo importing Nil and Cons):

~~~ ocaml
let tok : ('tok, 'tok option) parser =
 fun inp ->
  match inp () with
  | Nil -> Ok (None, Seq.empty)
  | Cons (t, rest) -> Ok (Some t, rest)
~~~

Not tested, but if `uncons` is the only reason for bumping the MSOV (minimal supported OCaml version), it might be worth to reconsider that. ;)

fblanqui:
Hi @gabrielhdt . Michael's suggestion looks nice and easy to implement. Do you want to try it? I will then wait for the version 5.1 of pratter and an update of this PR.

fblanqui:
Hi @gabrielhdt . I see that pratter 5.0.1 is now available on opam. Could you please update your PR so that I can merge it?

PR #1198, opened 2025-02-12
Cleanup syntax highlighting file for vscode

Alidra:
The syntax highlighting file for vscode (`editors/vscode/syntaxes/lp.tmLanguage.json`) contains information for the tokenization and syntax coloring of Lambdapi files under Vscode. 
This file contains the keywords of the Lambdapi language and their "syntactic categories" which are matched to "coloring categories".
Currently, some keywords appear in different categories at the same time and one category (`commands`) duplicates all the keywords that are in other categories which is confusing and can be source of errors/inconsistencies in the future.
Thus, this PR cleans up this file.

### DIFF
--- a/editors/vscode/syntaxes/lp.tmLanguage.json
+++ b/editors/vscode/syntaxes/lp.tmLanguage.json
@@ -40,7 +40,7 @@
     },
 
     "tactics": {
-      "match": "\\b(apply|assume|fail|focus|generalize|have|induction|orelse|refine|reflexivity|remove|repeat|rewrite|set|simplify|solve|symmetry|try|why3)\\b",
+      "match": "\\b(apply|assume|focus|generalize|have|induction|orelse|refine|reflexivity|remove|repeat|rewrite|set|simplify|solve|symmetry|try|why3)\\b",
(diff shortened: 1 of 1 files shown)


editors/vscode/syntaxes/lp.tmLanguage.json
```diff
@@ -44,7 +40,7 @@
     },
 
     "tactics": {
-      "match": "\\b(apply|assume|fail|focus|generalize|have|induction|orelse|refine|reflexivity|remove|repeat|rewrite|set|simplify|solve|symmetry|try|why3)\\b",
+      "match": "\\b(apply|assume|focus|generalize|have|induction|orelse|refine|reflexivity|remove|repeat|rewrite|set|simplify|solve|symmetry|try|why3)\\b",
```
fblanqui, line 43: fail is a tactic (https://github.com/Deducteam/lambdapi/blob/00c63619580c2bf679ec6ee85ba5161da8cd2cb7/src/handle/tactic.ml#L458); we should keep it here I guess

fblanqui (changes requested):
Thank you Abdelghani. There is only a small change to do I think.

PR #1204, opened 2025-02-17
add builtin strings

fblanqui:
The user needs to declare a type for strings, and map it to the builtin "String".
Then string literals between double quotes are considered as constant symbols of this type.


fblanqui:
@NotBad4U another PR that should be of interest to you ;-)

NotBad4U:
Sorry, we do not support `String` in Alethe. That was a motivation for Eunonia to add the features to write proof about `String` and `Bitvector`.

fblanqui:
Anyway, one cannot do much with this builtin type. But it will be useful to allow in eval tactics taking identifiers as argument (assume, have, set) (https://github.com/Deducteam/lambdapi/pull/1202).

PR #843, opened 2022-02-14
Replace Bindlib by de Bruijn indices

fblanqui:

PR #1205, opened 2025-02-19
Replace Bindlib by de Bruijn and closures

barras:
An implementation of binders that could replace Bindlib without losing efficiency. Here's a quick benchmark (timings on my machine). Here, bindlib is the master branch, closure is the one i propose in this PR and db is the previous attempt, using dB indices

|	        | tests	   | perf_rw_engine	    | matita  	    	| holide |
| -----------| -------------| ------------------------------| ------------ | ------- |
|bindlib:	| 12'	   | 1'47 (8.6Gb RAM)	   | 1'20 (1.6Gb)	| 9'25 (1.4Gb) |
|closure:	| 6'29	   | 29" (2Gb RAM)	   | 0'47 (2.3Gb)	| 7'30 (0.7Gb) |
|db:		| 6'45	   | 28" (2Gb RAM)	   | 3'39 (8.7Gb)	| 8'45 (0.6Gb) |

Note that the time difference between Bindlib and closure in make tests only comes from perf_rw_engine, so it means there's a huge inefficiency in the interpretation of rewrite rules with Bindlib.

This PR is built on top of #843 . The diff is https://github.com/fblanqui/lambdapi/compare/db..barras:lambdapi:db . The modifications are mainly in term.ml + the renaming of Db into Bvar.

Other interesting data: checking the translation in Lambdapi of the HOL-Light base library hol.ml takes 7m14s with master and 4m30s with this PR (on a machine different from the above). Checking the translation in Coq takes about 50 minutes...


src/core/term.ml
```diff
@@ -65,60 +107,67 @@ type 'a notation =
...
       true if the placeholder stands for a type. *)
-  | TRef of term option ref (** Reference cell (used in surface matching). *)
-  | LLet of term * term * tbinder
+  | TRef of term option Timed.ref (** Reference cell (used in surface matching,
+                                      and evaluation with sharing). *)
+  | LLet of term * term * binder
   (** [LLet(a, t, u)] is [let x : a ≔ t in u] (with [x] bound in [u]). *)
 
+(** Type for binders, implemented as closures. The closure term must have
+    all its bound variables declared eitherde Bruijn indices bound either by the binder itself or by
```
fblanqui, line 138: Hi @barras. Thank you so much for this great PR! Perhaps you want to fix this comment?

barras: comment fixed

sacerdot:
Are you sure about this comment in term.ml: (????)

For instance, the term [λx. λy. x y] is represented as
    [Abst(_,(_,Abst(_,(_,Appl(Bvar(InSub 0)|Bvar(InEnv 0)),
                       [|Bvar(InSub 0)|])),
             [||]))]

barras:
@sacerdot ok i mixed up x and y.... x is InEnv 0 and y is InSub 0 so it should read Appl(Bvar(InEnv 0), Bvar(InSub 0)).
I'll fix the comment.

sacerdot:
Quick question/observation: I have just had a cursory look at the data structure and it is exactly the one we (= me, Beniamino, Guerrieri and Gicka and some others from Huawei) studied in a paper that we are trying to publish (rejected once).  The Huawei guys used this representation in their implementation of a programming language. Did you get the idea from some paper/previous work/etc.?

src/core/term.ml
```diff
@@ -633,6 +556,314 @@ let mk_Appl : term * term -> term = fun (t, u) ->
...
+   application is built as a left or right comb depending on the associativity
+   of the symbol, and arguments are ordered in increasing order wrt [cmp].
+
+- In [LLet(_,_,b)], [binder_occur b = true] (useless let's are erased). *)
+let mk_Vari x = Vari x
+let mk_Type = Type
+let mk_Kind = Kind
+let mk_Symb x = Symb x
+let mk_Prod (a,b) = Prod (a,b)
+let mk_Arro (a,b) = let x = new_var "_" in Prod(a, bind_var x b)
```
fblanqui, line 858: @barras Can't avoid the call to bind_var here?

barras:
@sacerdot the type of closure is quite standard: a term with an environment. There is also the boolean information indicating whether the bound variable(s) actually occur, this comes from Bindlib. Maybe you are more specifically referring to the type of bound variables (bvar) which distinguishes references to the substitution or refs to the environment. This is inspired from the Zinc machine (X. Leroy) where you have ACC bytecode to access the stack (= substitution here) and ENVACC to access the environment.

src/core/term.ml
barras: > @barras Can't avoid the call to bind_var here?

I think it's quite possible. We could create a binder without occurrence in constant time. The speedup might be noticed if the rest of code tracked whether a Prod is actually dependent more often. We could have unbind return a var option (None when the variable does not occur) so rebuilding the binder would be in constant time. In the end, iterating under a non-occurring binder would have no penalty (asymptotically at least). I'll check.

Issue #1200, opened 2025-02-13
The modifier associative commutative can ble slow for large term


NotBad4U:
Hi everyone,

I am trying to use the modifier `associative commutative` to run reflective proof similar to `lia` and `lia` from Coq.
I would like to use the `associative commutative` for the normalization of my term.

Here is a gist that contains my experimentation:
https://gist.github.com/NotBad4U/bb86fb2c18f9e98b5a82a9c5c6f0d9b1

NotBad4U:
Maybe @fblanqui  and @barras, you can add more details about this issue?

PR #1211, opened 2025-02-27
Reimplement conversion modulo AC

fblanqui:
This PR reimplements conversion modulo AC without enforcing terms to be in AC-canonical form.
This is much more efficient but may not work with any rewrite system.
It seems to work well with:
- the theory ACI = AC + Idempotence used in max-suc algebra for universe levels,
- the theory of commutative groups AG = AC + Inverse + Neutral used in linear arithmetic.
This fixes the issue #1200.

Main changes in term.ml and eval.ml:
- do not declare the type term as private and remove the mk_* construction functions, except mk_Arrow
 (it reduces checking time of holide by 7%)
- sym type: add field sym_rstrat: red_strat = Outermost (default) | Innermost (for symbols defined by matching on AC symbols); with Innermost, in a function symbol application, arguments are normalized before trying to apply a rule
- add type side = Left | Right for the rewrite tactic and AC symbols (bool arguments are replaced by arguments of type side)

Other changes:
- Term: add type side = Left | Right, used in properties and rewrite tactic
- Term: remove property Assoc
- when declaring a symbol as associative commutative, a side must be given now
- LibTerm: add contains_ac_sym
- List: add rev_append2
- Logger.set_debug_in: permute arguments and use string argument instead of char

TODO:
- [x] do not declare term as private and remove functions mk_
- [x] handle commutativity alone, fix tests/OK/ac.lp
- [x] fix compute
- [x] in applications, compare arguments from left to right
- [x] try improve eq_modulo further
- [x] update CHANGES
- [x] update doc, add links to examples
- [ ] Logger.set_debug: permute arguments?

### DIFF
--- a/src/handle/tactic.ml
+++ b/src/handle/tactic.ml
@@ -352,7 +352,7 @@ let p_tactic (ss:Sig_state.t) (pos:popt) :term -> p_tactic =
             | T_repeat, _ -> assert false
             | T_rewrite, [_;t] -> P_tac_rewrite(Right,None,p_term pos t)
             | T_rewrite, _ -> assert false
-            | T_rewrite_left, [_;t] -> P_tac_rewrite(false,None,p_term pos t)
+            | T_rewrite_left, [_;t] -> P_tac_rewrite(Left,None,p_term pos t)
             | T_rewrite_left, _ -> assert false
             | T_set, [t1;_;t2] ->
                 P_tac_set(p_ident_of_sym pos t1,p_term pos t2)


fblanqui:
@NotBad4U Could you please try this PR on some other examples for linear arithmetic ?

fblanqui:
Be careful that I changed the ordering: arguments are now compared from left to right. So, you need to permute the arguments of var.

NotBad4U:
I experimented the new `AC`  on my previous test [here](https://gist.github.com/NotBad4U/1638cd3787b8e5284c7030ca8a8534ec). It is very fast now! Before we could only compute AC with ~20 variables, and I remember it took >5min for 23 variables. In the test, I can compute with 40 variables. I will try this feature on an SMT solver bench.

NotBad4U:
It also does not work with my proof by reflexivity on reordering the pivot.
In the code below, the algebra that represents `Clause` does not compute an AC form.
The constructor `P` has an arity of `1`. Does this PR normalize only symbols with an arity of 2?
```
symbol ℂ: TYPE;
symbol P: Prop → ℂ;
right associative commutative symbol ⊔: ℂ → ℂ → ℂ; notation ⊔ infix right 2;
symbol ⬓: ℂ;

symbol F: Clause → ℂ;

rule F ($x ⟇ $xs) ↪ (P $x) ⊔ (F $xs)
with F ▩ ↪ ⬓
;

sequential symbol  ⩦: ℂ → ℂ → 𝔹; notation ⩦ infix 11;
rule (P $x) ⩦ (P $x) ↪ true
with (P $x) ⩦ (P $y) ↪ false
with ($x ⊔ $xs) ⩦ ($y ⊔ $ys) ↪ ($x ⩦ $y) Stdlib.Bool.and ($xs ⩦ $ys)
with ⬓ ⩦ ⬓ ↪ true
with ⬓ ⩦ (P _) ↪ false
with (P _) ⩦ ⬓ ↪ false
with (_ ⊔ _) ⩦ ⬓ ↪ false
with ⬓ ⩦ (_ ⊔ _) ↪ false
with (P _) ⩦ (_ ⊔ _) ↪ false
with (_ ⊔ _) ⩦ (P _) ↪ false;

compute ((P b ⊔ P a ⊔ ⬓)  ⩦  (P a ⊔ P b ⊔ ⬓)); // No AC computed here
```

fblanqui:
@NotBad4U https://gist.github.com/NotBad4U/1638cd3787b8e5284c7030ca8a8534ec now works with the last commits. Could you please provide me with a complete file to test the example with Clause too?

NotBad4U:
Thanks a lot for your work! I will give you an example as soon as possible!

PR #1213, opened 2025-03-04
Vscode/fix follow cursor

Alidra:
This PR fixes the problem of the cursor mode where proofs are navigated following the cursor position. The problem is that this mode checks proofs even when non Lambdapi files are navigated resulting in highlighting the text (mostly in red) in every file that is open in Vscode.

Besides, this PR adds a rule in the vscode/Makefile to install the lambdapi Vscode extension from local code essentially for debugging and beta testing the extension and it updates the package-lock.json file in editors/vscode.

Finally, the extension is released to version 0.2.4

### DIFF
--- a/editors/vscode/Makefile
+++ b/editors/vscode/Makefile
@@ -19,7 +19,7 @@ endif
 .PHONY: install_local
 install_local: compile
 ifeq ($(VERSION),)
-	@echo 'jq not found. Please install with sudo apt install jq or export version before calling install_local (PLEASE use same as in package.json): VERSION=?.?.? make install_local'
+	@echo 'jq not found. Please install with sudo apt install jq or export version before calling install_local (PLEASE use same as in package.json): VERSION=X.Y.Z make install_local'
 else
 	vsce package
 	code --install-extension lambdapi-$(VERSION).vsix
--- a/editors/vscode/INSTALL.md
+++ b/editors/vscode/INSTALL.md
@@ -38,6 +38,22 @@ npm install -g @types/vscode
 npm install -g @vscode/vsce # to create VSCE packages
 ```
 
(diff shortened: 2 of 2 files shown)


editors/vscode/Makefile
```diff
@@ -8,3 +8,24 @@ node_modules:
...
+ifeq ($(VERSION),)
+	@echo "jq not found. Please install with sudo apt install jq or export version before calling install_local (PLEASE use same as in package.json) :VERSION=?.?.? make install_local"
+else
+	vsce package
+	code --install-extension lambdapi-$(VERSION).vsix
+endif
+
+else
+install_local:
+	@echo "vsce not found. Please install with npm install -g @vscode/vsce or export in the PATH. Details can be found on INSTALL.md"
```
fblanqui, line 30: on --> in

editors/vscode/src/client.ts
```diff
@@ -293,7 +293,7 @@ export function activateClientLSP(context: ExtensionContext,
     window.onDidChangeTextEditorSelection(e => {
 
         const cursorMode: boolean = context.workspaceState.get('cursorMode') ?? false;
-        if (!cursorMode)
+        if (!cursorMode || !e.textEditor.document.fileName.endsWith(".lp"))
```
fblanqui, line 296: allow ".dk" too

fblanqui: Yes, there is no proofs in dk file, only declarations. Still it indicates if there is an error somewhere. I think that it might be useful. What do you think?

Alidra: It makes sens indeed. Do we need to fix navigation using shortcuts in .dk files?

PR #1225, opened 2025-03-26
Emacs : fix version of Emacs in requirements

Alidra:
This PR uses version 27.1 instead of version 28.1 of Emacs in the Lambdapi extension requirements since this extension has been tested successfully with version 27.1 of Emacs.


fblanqui:
Thanks @Alidra . In the future, could you please try to set up some CI script to test various Emacs versions?

PR #1227, opened 2025-03-28
fix printing of positive integers in print.ml

ciaran-matthew-dunne:
fixing bugged printing of positive integers due to of weird tail-recursion in `pos_of_term`.
e.g., currently `compute (I (O H))` incorrectly prints as `6` rather than `5`.

now we just use recursion which is maybe less efficient????


fblanqui:
Thank you!

PR #1233, opened 2025-04-08
Customize websearch server to work online

Alidra:
This PR modifies the websearch engine of Lambdapi so that it can be easily deployed and customized for online use. This includes the following points :
- Make the server listen to the all interfaces instead of localhost only.
- Read the header of the Lambdapi websearch webpage from an external file so that it can be customized to specific libraries (for instance, to include description of the HOL-Light library). This may include specific CCS styles also.
- Enhance the formatting of the search results.

### DIFF
--- a/src/cli/lambdapi.ml
+++ b/src/cli/lambdapi.ml
@@ -38,12 +38,37 @@ let search_cmd cfg rules require s =
    (Tool.Indexing.search_cmd_txt ss s) in
  Error.handle_exceptions run
 
-let websearch_cmd cfg rules port require =
+let read_description filename =
+  let ic = open_in filename in
+  let rec read_lines acc =
+    try
+      let line = input_line ic in
+      read_lines (acc ^ line ^ "\n")
+    with End_of_file ->
+      close_in ic;
+      acc
+  in
+read_lines ""
+
+let websearch_cmd cfg rules port require description_file=
  Config.init cfg;
  let run () =
   Tool.Indexing.load_rewriting_rules rules ;
   let ss = sig_state_of_require require in
-  Tool.Websearch.start ss ~port () in
+  let description = "
+    <h1><a href=\"https://github.com/Deducteam/lambdapi\">LambdaPi</a>
+     Search Engine</h1>
+
+    <p>
+    The <b>search</b> button answers the query. Read the <a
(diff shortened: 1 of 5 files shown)


fblanqui (changes requested):
Thank you Abdelghani for your PR. Here are some remarks and improvement suggestions:

src/cli/lambdapi.ml
```diff
@@ -38,12 +38,38 @@ let search_cmd cfg rules require s =
    (Tool.Indexing.search_cmd_txt ss s) in
  Error.handle_exceptions run
 
-let websearch_cmd cfg rules port require =
+(* [string_of_file f] puts the contents of file [f] in a string. *)
+let string_of_file f =
```
fblanqui, line 42: We should move this function to lplib/string.ml.

src/common/pos.ml
```diff
@@ -99,11 +109,17 @@ let to_string : ?print_fname:bool -> pos -> string =
...
+  fun ?(print_dirname=true) ?(print_fname=true) pop ->
   match pop with
     | None -> "Unknown location "
-    | Some (p) -> to_string ~print_fname p ^ " "
+    | Some (p) -> to_string ~print_dirname ~print_fname p ^ " "
+
+(** [basename p] returns a string representing the position [p]
+ with the filename reduced to its basename. *)
+let basename = fun p ->
+  (popt_to_string ~print_dirname:false p)
```
fblanqui, line 122: I propose that you remove this definition and unfold it where it is used.

src/common/pos.ml
```diff
@@ -84,13 +84,23 @@ let before = shift (-1)
...
+    | Some(n) ->
+      let filename =
+        if print_dirname then n
+        else
+          try
+            let last_slash_index = String.rindex n '/' in
+            (String.sub n (last_slash_index + 1)
+            (String.length n - last_slash_index - 1))
+          with Not_found -> n
+      in filename ^ ":"
```
fblanqui, line 103: For a better portability (Windows), use the OCaml library Filename:
```
    | Some n -> (if print_dirname then n else Filename.basename n) ^ ":"
```

fblanqui (changes requested):

Issue #1236, opened 2025-04-10
Include requests/responses in the logs of the websearch server


Alidra:
Currently, logs of the websearch server include information about date and time of requests, browser and OS from which requests are sent, response time, ...

It would be interesting, especially when using the search tool online, to log the text of the requests and their corresponding responses for statistics and debugging.

Alidra:
This is issue has been addressed by PR #1244

PR #1244, opened 2025-05-02
Add logs, transform ASCII to unicode in requests and listen to custom URLs

Alidra:
This PR  :

- fixes the following issue https://github.com/Deducteam/lambdapi/issues/1236
- transforms `forall` and `->` to unicode caracters `Π` and `→` in search queries
- Added the `url` flag to the `websearch` command to accept request on URLs with custom paths (for instance `localhost:8080/<STRING>`)
- Adds a button to show hide the description in the header of the web page of the server.

### DIFF
--- a/tests/dune
+++ b/tests/dune
@@ -8,7 +8,7 @@
   (glob_files KO/*.dk)
   lambdapi.pkg)
  (libraries lambdapi.core lambdapi.parsing lambdapi.handle lambdapi.tool
-            alcotest lambdapi.common)
+            alcotest)
  (flags -w +3))
 
 ;(rule


doc/options.rst
```diff
@@ -178,6 +178,8 @@ websearch
 
 * ``--header <FILE.html>`` uses <FILE.html> as header of the search engine web page.
 
+* ``--url <STRING>`` tells the server to accept requests only on URLs using path ``<STRING>`` (for instance ``localhost:8080/<STRING>``). Use ``**`` to accept requests with any path.
```
fblanqui, line 181: I suggest to name the option to `--restrict-url` which is more informative.
I don't fully understand what this option does actually. Could you clarify the explanation?
In particular the occurrences 2 and 3 of `<STRING>` seems to represent two different variables.
What is the point of `**` if you can get the same without using the option?
Why not interpreting `<STRING>` as  regular expression?
In this case, you could call the option: `--match-url` tells the server to accept requests only on URLs matching `<STRING>` interpreted as a regular expression. Does the url always start with `localhost:8080/`?

src/cli/lambdapi.ml
```diff
@@ -63,7 +63,11 @@ let websearch_cmd cfg rules port require header_file dbpath_opt =
         query language specification</a> to learn about the query language.
         <br>The query language also uses the <a
         href=\"https://lambdapi.readthedocs.io/en/latest/terms.html\">
-        Lambdapi terms syntax</a>.<br>
+        Lambdapi terms syntax</a>.<br> with the possibility to use,
+        for commodity,
+        \"forall\" and \"->\"  instead of \"Π\" and \"→\" respectively
+        (note,results are displayed with the Unicode symbols
```
fblanqui, line 69: remove "note,"

src/cli/lambdapi.ml
```diff
@@ -63,7 +63,11 @@ let websearch_cmd cfg rules port require header_file dbpath_opt =
...
+        \"Π\" and \"→\").
```
fblanqui, line 70: Why not replacing Π and → by forall and -> in the output too?
Actually, there should be an option `--coq-syntax` for enabling this replacement.

CHANGES.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Added tactic eval which evaluates a term and interprets it as a tactic. This allows one to define tactics using rewriting.
 - Added builtin String for string literals between double quotes.
 - Added the `--header` flag to the `websearch` command
+- Added the `url` flag to the `websearch` command
```
fblanqui, line 15: flag -> option

src/cli/lambdapi.ml
fblanqui: This replacement should be applied to the header as well.

fblanqui (changes requested):

src/tool/websearch.eml.ml
```diff
@@ -40,28 +40,32 @@ let show_form ~from ?(message="") ?output request =
...
-let start ~header ss ~port ~dbpath () =
+let start ~header ss ~port ~dbpath ~path_in_url () =
   (*Common.Logger.set_debug true "e" ;*)
   let interface = "0.0.0.0" in
   Dream.run ~port ~interface
   @@ Dream.logger
   @@ Dream.memory_sessions
   @@ Dream.router [
 
-    Dream.get  "/"
```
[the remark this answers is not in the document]
Alidra: As discussed offline, the option we will keep for now is to use the last option and document it more.

doc/options.rst
fblanqui: What about `--accept-url` ?
Another thing not clear: are all the occurrences of <STRING> the same in your explanation of the option?
```
* ``--url <STRING1>`` tells the server to accept requests only on URLs using path ``<STRING2>`` (for instance ``localhost:8080/<STRING3>``). Use ``**`` to accept requests with any path.
```
It seems that we have STRING2=STRING1 but what is the relation between STRING3 and STRING1? Where "localhost:8080/" does it come from?

fblanqui: What is the default behavior when we do not use that option?

doc/query_language.rst
```diff
@@ -42,6 +42,8 @@ The semantics of the query language is the following:
   * ``=`` the pattern must match the whole position
   * ``>`` the pattern must match a strict subterm of the position
 
+Note that for commodity, ``forall`` and ``->`` can be used inside the query instead of unicode caracteres ``Π`` and ``→`` respectively.
```
fblanqui, line 45: instead of the Unicode characters

src/tool/indexing.ml
```diff
@@ -656,7 +658,8 @@ module UserLevelQueries = struct
    let resultsno = List.length items in
    let _,items = Lplib.List.cut items from in
    let items,_ = Lplib.List.cut items how_many in
-   Format.asprintf "<h1>Number of results: %d</h1>%a@."
+   Format.asprintf
+    "<h1>Number of results: %d</h1>%a@."
```
fblanqui, line 662: This change does not seem to be needed.

fblanqui (changes requested):

doc/options.rst
fblanqui: I believe that I now better understand the meaning and all this and propose:
```
* ``--url <STRING>`` tells the server to only accept requests on URLs which path is ``<STRING>``. By default, the server only accepts requests on URLs which path is empty. Use ``**`` to accept requests with any path.
```

doc/options.rst
```diff
@@ -220,3 +220,5 @@ websearch
 * ``--db <FILE.db>`` tells the server to use ``<FILE.db>`` instead of ``~/.LPSearch.db``.
 
 * ``--header <FILE.html>`` uses <FILE.html> as header of the search engine web page.
+
+* ``--url <STRING>`` tells the server to only accept requests on URLs which path is ``<STRING>``. By default, the server only accepts requests on URLs which path is empty. Use ``**`` to accept requests with any path. It is possible to express more complex constraints using the `Dream framework <https://aantron.github.io/dream/#routing>`__.
```
Alidra, line 224: @blanqui, do we keep `--url` or do we use another option?

tests/test_utils.ml
```diff
@@ -0,0 +1,32 @@
...
+  let open Alcotest in
+  run "Test_tests" [
+    ("TransformArrow", [ test_case "Arrow" `Quick test_transform_Aarrow_to_Uarrow
+                       ; test_case "forall with spaces" `Quick test_transform_Forall_to_Pi_withSpaces
+                       ; test_case "forall at begining" `Quick test_transform_Forall_to_Pi_at_begining
+                       ; test_case "forall with with a dot" `Quick test_transform_Forall_to_Pi_with_dot
+                       ; test_case "forall with parathensis" `Quick test_transform_Forall_to_Pi_with_parathensis
+    ])
+
+  ]
```
fblanqui, line 32: That's nice to add some unit tests but, in this case, I suggest to avoid alcotest and do something simpler, and keep unit tests where the function is defined. See for instance unit tests in lplib/list.ml.

fblanqui (changes requested):

tests/dune
```diff
@@ -8,7 +8,7 @@
   (glob_files KO/*.dk)
   lambdapi.pkg)
  (libraries lambdapi.core lambdapi.parsing lambdapi.handle lambdapi.tool
-            alcotest)
+            alcotest lambdapi.common)
```
fblanqui, line 11: Hi. Do you still need to add this dependency?

Alidra: No, indeed. I removed it.

PR #1245, opened 2025-05-03
reify patterns

### LINKED ISSUE #1240: Adding pattern to the builtin  #rewrite
Currently, the built-in rewrite is defined as follows:
```
builtin "rewrite" ≔ …; // : Π [a], Prf a → T
```
Making it impossible to pass an optional rewrite pattern like the original `rewrite` tactic.
Since the built-in type `string` has been introduced, could we add the pattern as a string parameter?

fblanqui:
fix #1240 

- change the type of the builtin for "refine": it now takes a String which is parse as a term; this allows to use terms with underscores at the time of evaluation
- change the type of the builtin for "rewrite": it now takes two additional String arguments at the beginning, first the side ("" or "left"), second a pattern ".[...]"


fblanqui:
@Alidra do you have any idea why CI fails for vscode? I didn't change anything in this PR. Perhaps some package version needs to be updated? Could you please check this?

fblanqui:
It seems to use Node.js v24.0.0. On my machine I have node 22.15.0 and it works well. 24.0.0 is the latest version but our scripts are probably not adapted to this last version. The CI script should use a matrix with various node versions. Could you please take care of this in a new PR?

PR #1263, opened 2025-06-11
Remove html code from the results of terminal search queries

Alidra:
This PR addresses issue#1254 so that html code is removed from the results of search queries when search is performed from the command line or from an editor


fblanqui:
Thanks @Alidra

PR #1269, opened 2025-06-26
Websearch fix csrf and monitoring

Alidra:
This PR fixes web searcher :
- fixes csrf token to avoid the problem when deploying the server on the lambdapi server on a sub path.
- adds a favicon and a response to HEAD request (useful among others when monitoring the websearver with uptimerobot.com)


fblanqui:
By the way, what is the branch https://github.com/Deducteam/lambdapi/tree/websearch_fix_csrf_and_monitoring for? PRs should be done from cloned repo.

PR #1277, opened 2025-07-01
Websearch fix csrf and monitoring

Alidra:
This PR fixes web searcher :

    fixes csrf token to avoid the problem when deploying the server on the lambdapi server on a sub path.
    adds a favicon and a response to HEAD request (useful among others when monitoring the websearver with uptimerobot.com)


fblanqui:
For the ico file, could you please use "λΠ" instead of "π"?

PR #1279, opened 2025-07-02
improve applicability of the apply tactic

fblanqui:
In various cases, apply fails and one needs to instead use refine with various underscores. This PR tries to make apply works more often.


fblanqui:
In various cases, apply fails and you need to instead use refine with various underscores. I'm trying to make apply works more often.

PR #1301, opened 2025-09-19
Emacs test pipeline

Alidra:
This PR create a workflow to test the Emacs package with different version of Eglot, Math-symbol-lists and Highlight.
Note the script used for this purpose hacks the installation process of Elpa to install specific versions of the dependencies because Elpa does not allow as far as I know to install specific (previous) versions of packages.

Todo before merging 
 - Keep only the versions with whome Lambdapi-mode works.


Alidra:
For the record, currently the Lambdapi mode works with the following versions :
        eglot-version: [1.9, 0]
        math-symbol-list-version: [1.2.1, 1.3, 0]
        highlight-version: [20190710.1727, 20210318.2248, 20220320.2141, 20250722.0617, 20250724.2154, 20250815.1830, 0]

Issue #1299, opened 2025-09-11
Wrong error location when reporting errors in another file


nicomarg:
When requiring another file which contains errors, (at least on emacs) the error isn't reported at the import but at its position in the required file.

MWE: consider file1 which throws an error on line 3, (the assertnot fails):
```
symbol A : TYPE;

assertnot ⊢ A : TYPE ;
```
When importing file1 into another file, the error is reported on line 3, and on the characters corresponding to the length of the assertnot (see pictures below with examples for reporting above and below the require statement).
Moreover, this has weird properties on interactive mode (notably, it can prevent it from running properly on perfectly valid beginnings of files).

I did not test this in VSCode.

<img width="877" height="234" alt="Image" src="https://github.com/user-attachments/assets/7f425442-11f4-4505-a3d5-bd17b1cee278" />
<img width="880" height="239" alt="Image" src="https://github.com/user-attachments/assets/826b483e-63a5-4798-9aa8-edc65031cec9" />

PR #1310, opened 2025-10-08
Fix location when reporting errors in another file

Alidra:
This PR fixes the issue #1299 .
The new solution checks the error is located in the same file.
If not, the error is reported on the import command displaying the location in the imported file


fblanqui:
@Alidra is this PR ready to be merged?

PR #1326, opened 2025-11-26
Allow decimal notation for multiple types by scoping builtin map…

### LINKED ISSUE #1268: Decimal notation not possible for two types simultaneously

NotBad4U:
Previously, Lambdapi only allowed decimal notation (e.g., "0", "+1") to map to a single type, preventing its use across multiple types simultaneously. This change introduces scoped `builtin` mappings, enabling users to define decimal notation for multiple types in the same context. This PR, when finished, will close #1268.

TODO:
- [ ] revert changes on sig_state type definition
- [ ] change Builtin.get following https://github.com/Deducteam/lambdapi/pull/1326#pullrequestreview-3511736145
- [ ] in notation data type, add path information on constructors used for decimal notation
- [ ] fix print.ml accordingly to prefix decimal numbers by the appropriate path whenever it is not in scope

### DIFF
--- a/src/parsing/lpParser.mly
+++ b/src/parsing/lpParser.mly
@@ -89,8 +89,7 @@
 // other tokens
 
 %token <bool * string> DEBUG_FLAGS
-%token <string> INT
-%token <Path.t * string> PINT
+%token <Path.t * string> INT
 %token <string> FLOAT
 %token <Pratter.associativity> SIDE
 %token <string> STRINGLIT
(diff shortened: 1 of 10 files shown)


NotBad4U:
Now all the tests pass, so I think we can start the review @fblanqui. I removed the tag WIP.

src/parsing/lpParser.mly
```diff
@@ -90,6 +90,7 @@
 
 %token <bool * string> DEBUG_FLAGS
 %token <string> INT
+%token <Path.t * string> PINT
```
fblanqui: Instead of introducing a new token, I suggest to change the type of INT by adding a path.

src/export/rawdk.ml
```diff
@@ -25,7 +25,7 @@ let rec term : p_term pp = fun ppf t ->
   | P_Expl u -> out ppf "(%a)" term u
   | P_Type -> out ppf "Type"
   | P_Wild -> out ppf "_"
-  | P_NLit i -> string ppf i
+  | P_NLit (path, i) -> out ppf "%a%s" Dk.path path i
```
fblanqui: Integer literals are undefined in Dedukti. Actually we should fail (already in master) like for string literals.

fblanqui:
Yes. Actually, it will depend on the order of require's. But it will always be possible to write Nat.1. Isn't it what we want?

PR #1357, opened 2026-03-07
fix(lsp): comply with LSP spec for positions and go-to-definition

ciaran-matthew-dunne:
## Summary

- **mk_range**: clamp line/character to `max 0` so negative values are never sent. LSP spec requires `uinteger` for positions, but error fallback positions with `start_line=0` would produce `line=-1` after the 1-based to 0-based conversion.
- **Error fallback positions**: use `start_line=1`/`end_line=1` instead of `0` in `new_doc` and `do_check_text` error handlers, so `mk_range` produces valid `line=0` (beginning of file).
- **Go-to-definition**: prefix URI with `file://` in `mk_definfo`. The LSP spec requires `DocumentUri` ([RFC 3986](https://www.rfc-editor.org/rfc/rfc3986)), but bare file paths were being sent, causing editors to fail to parse the `GotoDefinitionResponse`.

These issues were discovered while developing the [Zed editor extension for Lambdapi](https://github.com/ciaran-matthew-dunne/zed-lambdapi). Zed's strict LSP deserialization surfaced all three problems:
- `Error("invalid value: integer -1, expected u32")` from the position issue
- `data did not match any variant of untagged enum GotoDefinitionResponse` from the missing `file://` prefix

## Test plan

- [x] Verified go-to-definition works in Zed after the `file://` fix
- [x] Verified no more `-1` position errors in Zed logs
- [x] `dune build` passes
- [ ] Existing test suite (`make test`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/lsp/lp_doc.ml
+++ b/src/lsp/lp_doc.ml
@@ -151,10 +151,10 @@ let new_doc ~uri ~version ~text =
       let loc : Pos.pos =
         {
           fname = Some(uri);
-          start_line = 0;
+          start_line = 1;
           start_col  = 0;
           start_offset  = 0;
-          end_line = 0;
+          end_line = 1;
           end_col = 0;
           end_offset  = 0
         } in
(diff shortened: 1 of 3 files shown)


fblanqui:
Why this PR has been closed? If you change things in the LSP server, you also need to check and possibly fix the code in vscode and emacs.

PR #1366, opened 2026-04-20
fix(lsp): go-to-definition, hover, and stale session state

ciaran-matthew-dunne:
## 1. Fix go-to-definition and hover

Four bugs in `do_definition`/`hover_symInfo`. Three share a root cause: the handlers carried hand-rolled conversions (URI vs path, short-name vs qualified lookup, 1-based vs 0-based columns) instead of reusing the helpers the rest of the codebase uses.

### 1a. Go-to-definition on an external symbol lands at line 1 of the target file

- **Issue:** Go-to-definition on any symbol defined in another module always points to line 1 of the file.
- **Cause:** `do_definition` compared the symbol's `sym_pos.fname` (a raw filesystem path) against `doc.uri` (a `file://` URI). The comparison never succeeds for external symbols, so the code fell back to a synthesized line-1 position.
- **Fix:** use `sym_pos` directly when present; `mk_definfo` takes the filesystem path (the only thing its caller ever passes) and builds the URI from it.

### 1b. Qualified identifiers resolve to the wrong symbol; aliases don't resolve at all

- **Issue:** 
(i). go-to-definition/hover on `M.foo` while a local symbol is also named `foo` targets the local `foo`. 
(ii). go-to-definition/hover on a qualified symbol (i.e.,  `A.foo` given `require M as A;`) gives no result.
- **Cause:** the handlers looked the name up in the flat in-scope short-name map, trying the unqualified name first, with no knowledge of module aliases.
- **Fix:** new `Pure.find_sym` wrapping `Sig_state.find_sym` as used by scoper, with correct shadowing precedence and alias support. Both handlers now pass the full qualified identifier through it.

### 1c. Hover highlights a range shifted one column to the left

- **Cause:** `hover_symInfo` was the only handler still assembling its range JSON by hand, and its column arithmetic had drifted: a stale `column - 1` from when columns were 1-based.
- **Fix:** new helper `LSP.mk_range_of_interval`; hover now uses the same `Range.t` → JSON conversion that diagnostics and go-to-definition go through.

### 1d. Illegal positions reach the client

- **Issue:**  LSP positions must be `uinteger` with `start <= end`.
- **Fix:** `json_of_range` sanitizes every outgoing range: coordinates floored at 0, inverted ranges collapsed to a caret at their start.

## 2. Requests on one document run against another document's state

- **Issue:** open document A, then open document B from a different package; any go-to-definition / hover / documentSymbol request on A afterwards resolves paths against B's configuration. 
e.g., with `Foo.lp` open, opening `Nat.lp` from Stdlib made subsequent queries in `Foo.lp` produce URIs relative to `<lib_root>/Stdlib/...` rather than that of the package. 
- **Cause:** `Pure.initial_state` does `Time.restore t0` and re-applies the opened file's package config, which installs that file's library mappings and *wipes the mappings of every other open document* (they live in timed state).
- **Fix:** expose `Pure.restore_time` and call it at the start of `do_definition`, `hover_symInfo` and `do_symbols`, so each request restores the timed state saved with its own document before touching timed globals like `Library.lib_mappings`.

### DIFF
--- a/src/lsp/lp_lsp.ml
+++ b/src/lsp/lp_lsp.ml
@@ -10,7 +10,7 @@
 (* Status: Very Experimental                                            *)
 (************************************************************************)
 
-open Lplib open Extra
+open Lplib
 open Common
 open Core
 
@@ -59,17 +59,8 @@ let do_check_text ofmt ~doc =
     try
       Lp_doc.check_text ~doc
     with Common.Error.Fatal(_pos, msg, err_desc) ->
-      let loc : Pos.pos =
-        {
-          fname = Some(doc.uri);
-          start_line = 0;
-          start_col  = 0;
-          start_offset  = 0;
-          end_line = 0;
-          end_col = 0;
-          end_offset  = 0;
-        } in
-      (doc, Lp_doc.mk_error ~doc loc (msg ^ "\n" ^ err_desc))
+      (doc, Lp_doc.mk_error ~doc (Pos.file_start doc.uri)
+              (msg ^ "\n" ^ err_desc))
   in
   Hashtbl.replace doc_table doc.uri doc;
   Hashtbl.replace completed_table doc.uri doc;
@@ -131,9 +122,11 @@ let mk_syminfo file (name, _path, kind, pos) : J.t =
                   ]
   ]
 
+(* [file] is a plain filesystem path (as produced by
+   [Library.file_of_path]), never a URI. *)
 let mk_definfo file pos =
   `Assoc [
-      "uri", `String file
+      "uri", `String ("file://" ^ file)
     ; "range", LSP.mk_range pos
         ]
 
@@ -157,6 +150,7 @@ let do_symbols ofmt ~id params =
     let msg = LSP.mk_reply ~id ~result:`Null in
     LIO.send_json ofmt msg
   | Some ss ->
+    Pure.restore_time ss;
     let sym = Pure.get_symbols ss in
     let sym =
       Extra.StrMap.fold
@@ -230,11 +224,7 @@ let get_node_at_pos doc line pos =
   let open Lp_doc in
   List.find_opt (fun { ast; _ } ->
       let loc = Pure.Command.get_pos ast in
-      let res = in_range ?loc (line,pos) in
-      let ls = Format.asprintf "%B l:%d p:%d / %a"
-                 res line pos Pos.pp loc in
-      LIO.log_error "get_node_at_pos" ("call: "^ls);
-      res
+      in_range ?loc (line,pos)
     ) doc.Lp_doc.nodes
 
 (** [get_first_error doc] returns the first error inferred from doc.logs *)
@@ -272,24 +262,6 @@ let rec get_goals ~doc ~line ~pos =
     | Some (v,_) -> Some v
 
 let get_logs ~doc ~line ~pos : string =
-  (* DEBUG LOG START *)
-  LIO.log_error "get_logs"
-    (Printf.sprintf "%s:%d,%d" doc.Lp_doc.uri line pos);
-  let log_to_str ((sev, log), posopt) =
-    let pos_str =
-      match posopt with
-      | None -> "None"
-      | Some Pos.{start_line; start_col; _} ->
-          Printf.sprintf "(%d, %d)" start_line start_col
-    in
-    let log_str =
-      let len = String.length log in
-      Printf.sprintf "length: %d | %s" len (String.sub log 0 (min 30 len))in
-    Format.asprintf "element(severity:%d): %s -> %s\n " sev pos_str log_str
-  in
-  Lsp_io.log_error "get_logs"
-    (List.fold_left (^) "\n" (List.map log_to_str doc.Lp_doc.logs));
-  (* DEBUG LOG END *)
   let line = line+1 in
   let end_limit =
     match get_first_error doc with
@@ -318,20 +290,15 @@ let do_goals ofmt ~id params =
   let msg = LSP.mk_reply ~id ~result in
   LIO.send_json ofmt msg
 
-let msg_fail hdr msg =
-  LIO.log_error hdr msg;
-  failwith msg
 
-let get_symbol : Range.point ->
-('a * 'b) RangeMap.t -> ('b * Range.t) option
+let get_symbol : Range.point -> 'a RangeMap.t -> ('a * Range.t) option
 = fun pos doc ->
 
   let open RangeMap in
 
   match (find pos doc) with
   | None -> None
-  | Some(interval, (_, token)) -> Some (token, interval)
-
+  | Some(interval, value) -> Some (value, interval)
 
 let do_definition ofmt ~id params =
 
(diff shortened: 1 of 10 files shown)


fblanqui:
Hi @ciaran-matthew-dunne . Thanks for your PR! Let me know when it is ready for review.

ciaran-matthew-dunne:
@Alidra !

src/pure/pure.ml
```diff
@@ -15,6 +15,26 @@ module Util = struct
 
 end
 
+(** Make a position spanning [len] characters from the start of [p]. *)
```
ciaran-matthew-dunne: position for `symbol` and `inductive` commands should be on the keyword itself, rather than the symbol that's being declared

src/pure/pure.ml
```diff
@@ -15,6 +15,26 @@ module Util = struct
...
+let keyword_pos (p : Pos.pos) (len : int) : Pos.pos =
```
ciaran-matthew-dunne: move `keyword_pos` and other stuff to a more sensible place

src/lsp/lp_lsp.ml
```diff
@@ -101,7 +101,8 @@ let do_change ofmt params =
 let do_close _ofmt params =
   let document = dict_field "textDocument" params in
   let doc_file = string_field "uri" document in
-  Hashtbl.remove doc_table doc_file
+  Hashtbl.remove doc_table doc_file;
+  Hashtbl.remove completed_table doc_file
```
ciaran-matthew-dunne: questionable that this actually solves any 'memory leak' issue. needs deeper investigation/profiling. assigned to @ciaran-matthew-dunne

src/lsp/lp_lsp.ml
```diff
@@ -151,6 +151,7 @@ let do_symbols ofmt ~id params =
     let msg = LSP.mk_reply ~id ~result:`Null in
     LIO.send_json ofmt msg
   | Some ss ->
+    Pure.restore_time ss;
```
ciaran-matthew-dunne, line 153: questionable that this actually solves the 'mapping issue'. needs some testing to demonstrate. assigned to @ciaran-matthew-dunne

src/lsp/lp_lsp.ml
```diff
@@ -132,8 +122,12 @@ let mk_syminfo file (name, _path, kind, pos) : J.t =
   ]
 
 let mk_definfo file pos =
+  let uri =
```
ciaran-matthew-dunne: why do we branch on `String.is_prefix` here? are there cases in which the given `file` is already a URI?

src/lsp/lp_lsp.ml
```diff
@@ -230,11 +224,7 @@ let get_node_at_pos doc line pos =
   let open Lp_doc in
   List.find_opt (fun { ast; _ } ->
       let loc = Pure.Command.get_pos ast in
-      let res = in_range ?loc (line,pos) in
```
ciaran-matthew-dunne, line 233: why are we removing logging?

src/lsp/lp_lsp.ml
```diff
@@ -272,24 +262,6 @@ let rec get_goals ~doc ~line ~pos =
     | Some (v,_) -> Some v
 
 let get_logs ~doc ~line ~pos : string =
-  (* DEBUG LOG START *)
```
ciaran-matthew-dunne, line 275: once again, why are we removing logging?

src/lsp/lp_lsp.ml
```diff
@@ -343,120 +310,51 @@ let do_definition ofmt ~id params =
...
+      | Some (qid, _) ->
+        match Pure.find_sym ss (Pos.none qid) with
         | None -> `Null
-        | Some pos ->
-        (* A JSON with the path towards the definition of the term
-          and its position is returned
-          /!\ : extension is fixed, only works for .lp files *)
-          mk_definfo
-            Library.(file_of_path s.Term.sym_path ^ lp_src_extension) pos
+        | Some s when s.Term.sym_path = Sign.Ghost.path -> `Null
```
ciaran-matthew-dunne, line 327: under what conditions does `s.Term.sym_path = Sign.Ghost.path` occur?

src/lsp/lp_lsp.ml
```diff
@@ -343,120 +310,51 @@ let do_definition ofmt ~id params =
...
-      `Assoc [ "contents", `String sym_type; "range", range ] in
-    let msg = LSP.mk_reply ~id ~result in
-    LIO.send_json ofmt msg
-
-  with _ ->
-    let msg = LSP.mk_reply ~id ~result:`Null in
-    LIO.send_json ofmt msg
+    let ss = match doc.final with
+      | Some ss -> ss
+      | None -> raise (Error.fatal_no_pos "Root state is missing") in
```
ciaran-matthew-dunne: weird error message. why "root state" rather than "final state"?

src/lsp/lsp_base.ml
```diff
@@ -31,13 +32,27 @@ let mk_reply ~id ~result =
 let mk_event m p   =
   `Assoc [ "jsonrpc", `String "2.0"; "method", `String m; "params", `Assoc p ]
 
+(* LSP positions are uinteger; clamp defensively to avoid rejection by
```
ciaran-matthew-dunne: why do we need this? when do we pass `Pos.pos` with negative or zero line/col values? then why does `file_start` have a col values `0`? is there a difference between client-side/server-side line/col position indexing? does this differ between vscode/zed/emacs/vim???

src/pure/pure.ml
```diff
@@ -168,7 +168,7 @@ let handle_tactic : proof_state -> Tactic.t -> int -> tactic_result =
     let qres = Option.map (fun f -> f ()) qres in
     Tac_OK((Time.save (), ss, ps, finalize, prv, sym_pos), qres)
   with Fatal(Some p,m) ->
-    Tac_Error(Some p, Pos.popt_to_string p ^ " " ^ m)
+    Tac_Error(Some p, m)
```
ciaran-matthew-dunne: what failures actually trigger this branch? @Alidra couldn't repro.

src/lsp/lp_lsp.ml
fblanqui: The ghost signature is a builtin signature that is implicitly loaded and open in every module. It contains special symbols used for unification (and coercion?) rules, and all the string literals input by users.

src/lsp/lsp_base.ml
fblanqui: It may be the case that symbols generated by the inductive commands internally have negative positions (to be checked). I don't know whether this can affect LSP which normally only sees user declared symbols (which have >=0 positions).

Emacs counts columns from 0.

src/pure/pure.ml
fblanqui: I don't understand your question here. This change seems to be a good one anyway: the position should not be duplicated in the error message as it is already passed as first argument.

Alidra: I beleive this is because this issue has already been resolved in a previous PR : (https://github.com/Deducteamlambdapi/pull/1322)

Alidra:
Thank you @ciaran-matthew-dunne for the PR.
It would be nice to have different PRs for different features. That would help a lot with the management of changes and review (for instance the fix related to redundant position prefix in tactic error messages seems to be outdated)

src/lsp/lsp_base.ml
ciaran-matthew-dunne: Thanks. The `inductive N : TYPE ≔ ...` command generates an illegal position for `ind_N` where the end col is before the start col. I won't touch this since a comment in `term.ml` says:  "These virtual positions are however important for exporting lambdapi files to other formats. "

`json_of_range` now handles the sanitization.

src/pure/pure.ml
```diff
@@ -182,6 +182,18 @@ let end_proof : proof_state -> command_result =
 let get_symbols : state -> Term.sym Extra.StrMap.t =
   fun (_, ss) -> ss.in_scope
 
+let find_sym : state -> Term.qident Pos.loc -> Term.sym option =
```
fblanqui: Pure.find_sym is always called with (Pos.none qid) and Sig_state.find_sym uses the position in Fatal only. Therefore, Pure.find_sym could take a Term.qident as argument (with no position).

fblanqui (changes requested):
Thanks Ciaran for this PR! You removed quite some logging code. It is not useful anymore, or do you have other ways to get these information?

src/lsp/lsp_base.ml
fblanqui: With this PR, do you get meaningful positions for these generated symbols?

fblanqui:
@ciaran-matthew-dunne Please add your name in AUTHORS.md and update CHANGES.md.

ciaran-matthew-dunne:
> Thanks Ciaran for this PR! You removed quite some logging code. It is not useful anymore, or do you have other ways to get these information?

I removed it because it seemed somewhat excessive and could degrade performance (e.g., the logging commands within `get_node_at_pos` log are inside the function passed to `find_opt` which runs over every node in the document).   

~~I've restored it now and made it a bit leaner in case somebody currently relies on this info.~~

src/lsp/lsp_base.ml
ciaran-matthew-dunne: Yes, doing go-to-definition on `ind_T` now points to the end of the `inductive` command that defined `T`.

PR #1369, opened 2026-04-21
rewrite emacs SMIE grammar

fblanqui:


fblanqui:
@Alidra: could you please check if, in emacs, indentation is done correctly when using this PR?
For SMIE doc, see https://www.gnu.org/software/emacs/manual/html_node/elisp/SMIE.html
and https://www.iro.umontreal.ca/~monnier/smie.pdf.

PR #1371, opened 2026-04-22
LSP: completion, document outline, documentation hovers, tests

ciaran-matthew-dunne:
**Stacked on #1444** — its 3 commits appear here until it merges.

LSP UX additions, split out of the old bundled branch per review; the bug fixes went to #1443 and #1366 (both merged) and #1444.

### Completion
- in-scope symbols (ghost symbols excluded); no completion kind is assigned — the LSP kinds don't match lambdapi's notions
- keyword completions are grammar-driven: the parser reports the tokens it accepts at the cursor (new `Parser.Lp.expected_tokens`, see below), and exactly those keywords are offered — `open`/`private`/`as` after `require`, the notation kinds after `notation f`, `left`/`right` after `infix` or `associative`, `on`/`off` after `flag "…"`, `rule` after `simplify`, `begin` after a statement's type, tactic keywords, queries and proof enders at proof-step positions, command keywords and modifiers between commands, …
- symbols are only offered where the grammar accepts an identifier reference — never in binder positions (after `assume`, `require M as`, a `symbol` name, …)
- context-aware items with `.` as a trigger character: module paths after `require`/`open` (as a `textEdit`, since `.` is not a word character), qualified identifiers after `M.` (the non-private symbols of the required module, resolving `require as` aliases), flag names inside `flag "`, and hypothesis-first ranking in the argument of `apply`, `rewrite`, etc.
- keyword docs and snippets come from the same tables as before; snippets no longer presume one alternative (e.g. `require $0;` — `open`, `private` or a path completes after the keyword)
- when the text before the cursor does not parse (mid-edit), keyword completions fall back to the context-blind tables and the proof context of the last successful parse
- type details are attached lazily on `completionItem/resolve`; docs are sent as markdown `MarkupContent` when the client supports it

### Hover
- documentation for tactic keywords (inside proofs) and for command and query keywords and modifiers (`symbol`, `inductive`, `rule`, `constant`, `injective`, `flag`, …), sourced from `doc/*.rst`
- types of local symbols introduced by `assume`, inside proofs; types are printed with the document's own printer state; names in a tactic's arguments resolve in the state *before* the tactic runs (goal snapshots are post-tactic, so a tactic that closes its subgoal or the whole proof used to lose its own arguments' hypotheses)
- falls back to the token under the cursor looked up in the in-scope symbol table when the checked-AST lookup misses — hover works in mid-edit text and past parse errors

### Go-to-definition
- on `require`/`open` module paths (jumps to the start of that module's file)
- same in-scope fallback as hover

### Document outline
- hierarchical `DocumentSymbol[]` when the client supports it (constructors as children of their inductive type); the flat `SymbolInformation[]` list is kept as fallback; symbols are not classified into LSP kinds

### Parser
- new entry point `Parser.Lp.expected_tokens` (exposed as `Pure.expected_tokens`): parses a source prefix and returns the tokens acceptable at its end — the follow set the completion handler uses
- the expected-token recordings behind `Expected: …` error messages now accumulate until the next consumed token instead of overwriting one another, and a finished application records that the term could continue; syntax errors list more of the true alternatives, with a full set of term starters rendered as one word: `Expected: "begin", "≔", ";", "→", a term.`

### Protocol robustness
- `initialize` honours `rootUri` and `workspaceFolders` (deduplicated — clients typically send both for the same directory; package-config failures are logged, never fatal) and reads client capabilities
- unknown request methods get a JSON-RPC `MethodNotFound` error reply instead of silence; notifications stay unanswered per spec; request ids are echoed back verbatim, including 0 (Zed's first request) and strings
- message framing parses header lines up to the blank separator, accepting `Content-Length` in any position

### Tests
- Python integration test suite: `make test_lsp` (also part of `make test`), 143 tests speaking JSON-RPC over stdio to a `lambdapi lsp` subprocess — lifecycle, diagnostics, symbols, hover, definition, completion (including the grammar follow-set contexts), goals, and cross-cutting invariants (range validity, repeated-request stability, boundary positions, both protocol modes)
- the tactic and query completion lists are checked against the names documented in `doc/*.rst`
- Python ≥ 3.8, standard library only; stdlib-dependent tests skip when no lambdapi stdlib is installed

An earlier `didChange`-debouncing scheme was dropped to keep the diff reviewable.

### DIFF
--- a/tests/lsp/README.md
+++ b/tests/lsp/README.md
@@ -1,0 +1,42 @@
+# LSP integration tests
+
+End-to-end tests for the lambdapi LSP server. Each test spawns a
+`lambdapi lsp` subprocess, speaks JSON-RPC over stdio, and asserts on the
+responses.
+
+## Running
+
+From the repository root:
+
+```
+make test_lsp                         # runs as part of `make test` too
+python3 -m tests.lsp                  # equivalent, more verbose
+python3 -m unittest tests.lsp.test_hover -v   # single file
+```
+
+The harness prefers a locally-built binary (`_build/install/default/bin/lambdapi`)
+over whatever is on `PATH`, so you always test the current tree.
+
+## Environment variables
+
+| Name | Purpose |
+|---|---|
+| `LAMBDAPI_LIB_ROOT` | override stdlib location (defaults to the opam install) |
+| `LSP_TEST_LOG` | when set, pass `--log-file=$LSP_TEST_LOG` to the server |
+
+## Layout
+
+- `__init__.py`    — empty; marks `tests/lsp` as a regular Python package
+  so the modules can import each other (`from .base import …`) and the
+  suite runs as `python3 -m tests.lsp` on every supported Python
+- `__main__.py`    — entry point for `python3 -m tests.lsp`: discovers
+  and runs the whole suite
+- `base.py`        — `LSPTestCase` with per-test server + fixture helpers
+- `client.py`      — JSON-RPC / stdio client (`LSPServer` class)
+- `source.py`      — pattern-based position finder for fixtures
+- `fixtures/*.lp`  — small focused test documents
+- `test_*.py`      — one file per LSP concern (lifecycle, diagnostics, …)
+
+## Requirements
+
+Python ≥ 3.8, stdlib only. No pip packages.
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ libraries/zenon_modulo
 .DS_Store
 *.map
 *.log
+__pycache__/
+*.pyc
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -52,6 +52,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   duplicated directory component.
 - LSP server: go-to-definition and hover on qualified identifiers, and session
   state leaking between open documents.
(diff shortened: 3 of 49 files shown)


tests/lsp/README.md
```diff
@@ -0,0 +1,37 @@
...
+| `LAMBDAPI_LIB_ROOT` | override stdlib location (defaults to the opam install) |
+| `LSP_TEST_LOG` | when set, pass `--log-file=$LSP_TEST_LOG` to the server |
+
+## Layout
+
+- `client.py`      — JSON-RPC / stdio client (`LSPServer` class)
+- `source.py`      — pattern-based position finder for fixtures
+- `base.py`        — `LSPTestCase` with per-test server + fixture helpers
+- `fixtures/*.lp`  — small focused test documents
+- `test_*.py`      — one file per LSP concern (lifecycle, diagnostics, …)
```
fblanqui, line 38: Not all py files are listed here.

fblanqui (commented):
Thanks for your PR!

What is the empty file tests/lsp/__init__.py for?

ciaran-matthew-dunne:
Still working on this now. Give me 30min.

ciaran-matthew-dunne:
More or less ready for review. May change some things tomorrow morning.
Removed changes to the stdin buffer that were previously included in this PR.

PR #1376, opened 2026-05-22
Eval tactic: weak head normalize argument step by step

### LINKED ISSUE #1374: #rewrite tactic appears to unfold symbols in transparent rewrite lemmas before matching
I found a case where manual `rewrite` succeeds, but the same rewrite theorem used through a tactic with `#rewrite` fails if the theorem is transparent/non-opaque.

In the attached minima [example.zip](https://github.com/user-attachments/files/27963000/example.zip), the opaque stdlib rewrite lemmas work as expected when combined into a tactic:

  `symbol rem¬⊤ ≔ #and (#rewrite "" "" ¬⊤) (#rewrite "" "" ⊥∨);`

proves:

````
  `π ((¬ ⊤ ∨ x) = x)`
````

However, if I restate the same lemmas without `opaque`:

````
  symbol ¬⊤_noOp : π (¬ ⊤ = ⊥) ≔ ...
  symbol ⊥∨_noOp x : π ((⊥ ∨ x) = x) ≔ ...
````

then manual rewriting still works:

````
  rewrite ¬⊤_noOp;
  rewrite ⊥∨_noOp;
  reflexivity
````

but the tactic version fails:

````
  symbol rem¬⊤_noOP ≔ #and (#rewrite "" "" ¬⊤_noOp) (#rewrite "" "" ⊥∨_noOp);

  eval rem¬⊤_noOP;
````

with:

````
  No subterm of [(¬ ⊤ ∨ x) = x] matches [⊤ ⇒ ⊥].
````

This is unexpected because the theorem used by `#rewrite` has statement `π (¬ ⊤ = ⊥)`, and the goal clearly contains `¬ ⊤`. It looks as if `#rewrite` unfolds defined symbols in the transparent theorem/its left-hand side before matching, turning `¬ ⊤` into like `⊤ ⇒ ⊥`, while ordinary `rewrite` does not have this problem. 

Is this expected behaviour? It becomes problematic for proof export pipelines where opaque modifiers are removed before translation, because tactics that worked with opaque rewrite lemmas can fail once the same lemmas become transparent.

fblanqui:
- fix #1374 
- replace full snf reduction of tactic terms by a more incremental reduction strategy using whnf and reducing only subterms that are tactic terms
- infer the type and solve unification constraints of tactic terms before reduction and interpretation as tactics
- simplify the interface of compilation functions
- fix loading of (ghost) string symbols (their type was not set to the String builtin)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 - `simplify` now fails if the goal cannot be simplified.
 - Position of the error is removed from diagnostics when the error occurs in the file currently open in the editor.
-- Change lp parser based on menhir by one written by hand to provide more helpful error messages while being equally efficient.
+- Change lp parser based on Menhir by one written by hand to provide more helpful error messages while being equally efficient.
+- Replace full SNF reduction of tactic terms by a more incremental reduction strategy using WHNF and recursively reducing only subterms that are tactic terms.
+- Infer the type and solve unification constraints of tactic terms before reduction and interpretation as tactics.
 
 ### Fixed
 
--- a/doc/tacticals.rst
+++ b/doc/tacticals.rst
@@ -64,14 +64,14 @@ An example of use is given in `Tactic.lp <https://github.com/Deducteam/lambdapi/
 ``orelse``
 ----------
 
-``orelse T1 T2`` applies ``T1``. If ``T1`` succeeds, then ``orelse T1 T2`` stops. Otherwise, ``orelse T1 T2`` applied ``T2``.
+``orelse T1 T2`` applies ``T1``. If ``T1`` succeeds, then ``orelse T1 T2`` stops. Otherwise, ``orelse T1 T2`` applies ``T2``.
 
 .. _repeat:
 
 ``repeat``
 ----------
 
-``repeat T`` applies ``T`` on the first goals as long as possible, unless the number of goals decreases, in which case the tactic stops.
+``repeat T`` applies ``T`` on the first goal until the number of goals decreases.
 
 .. _try:
 
--- a/doc/tacticals.rst
+++ b/doc/tacticals.rst
@@ -35,7 +35,7 @@ The BNF grammar of tactics is in `lambdapi.bnf <https://raw.githubusercontent.co
    builtin "try" ≔ …; // : T → T
    builtin "why3" ≔ …; // : T
 
-The tactics taking a string as argument need the ``"String"`` :ref:`builtin` to be set. The string argument of ``refine`` is parsed as a term, and thus can contain underscores.
+The tactics taking a string as argument need the ``"String"`` :ref:`builtin` to be set. The string argument of ``refine`` is parsed as a term, and thus can contain underscores. If the builtin ``and`` is mapped to some symbol, say ``&``, then ``t & u`` is interpreted as follows: the tactic ``t`` is applied and, in case of success, the tactic ``u`` is applied. All other symbols are interpreted by the corresponding tactics.
 
 An example of use is given in `Tactic.lp <https://github.com/Deducteam/lambdapi/blob/master/tests/OK/Tactic.lp>`__:
 
--- a/src/cli/lambdapi.ml
+++ b/src/cli/lambdapi.ml
@@ -27,7 +27,8 @@ let sig_state_of_require =
     (* Search for a package from the current working directory. *)
     Package.apply_config (Filename.concat (Sys.getcwd()) ".") ;
     Core.Sig_state.of_sign
-     (Compile.compile (Parsing.Parser.path_of_string req))
+      (Compile.compile Core.Sig_state.dummy
+         (Parsing.Parser.path_of_string req))
 
 let search_cmd cfg rules require s dbpath_opt =
  Config.init cfg;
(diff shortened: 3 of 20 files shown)


fblanqui:
@melanie-taprogge @bodeveix @NotBad4U Could you please check how this PR impacts your LP devs?

melanie-taprogge:
Thank you for looking into this right away!
Unfortunately, while the PR does fix my initial issue, I encountered some new problems when trying out the PR on my examples. One problem I identified is that now tactics that are defined inline seem to cause problems in certain constellations (I experienced issues when both #repeat and #rewrite were used, while using them individually was okay.)
I attached a new [example](https://github.com/user-attachments/files/28188786/example.zip) to this comment, the interesting file is `newMiniExample.lp`, here you can see that in the first two example, Lambdapi checks the proofs, however, when I uncomment the last one, I get the error 

````
Uncaught [File "src/core/term.ml", line 926, characters 14-20: Assertion failed].
Raised at Core__Term.cleanup in file "src/core/term.ml", line 926, characters 14-26
Called from Core__Print.term_in in file "src/core/print.ml", line 350, characters 41-52
Called from Stdlib__Format.output_acc in file "format.ml", line 1307, characters 32-48
Called from Stdlib__Format.kfprintf.(fun) in file "format.ml", line 1356, characters 16-34
Called from Handle__Tactic.p_term.term in file "src/handle/tactic.ml", line 258, characters 38-56
Called from Handle__Tactic.p_term.term_aux in file "src/handle/tactic.ml", line 268, characters 40-52
Called from Handle__Tactic.p_term.term in file "src/handle/tactic.ml", line 258, characters 38-56
Called from Handle__Tactic.p_term.term_aux in file "src/handle/tactic.ml", line 268, characters 26-38
Called from Handle__Tactic.p_term.term in file "src/handle/tactic.ml", line 258, characters 38-56
Called from Handle__Tactic.p_tactic.tac_eval in file "src/handle/tactic.ml", line 360, characters 44-54
Called from Handle__Tactic.p_tactic.tac in file "src/handle/tactic.ml", line 394, characters 43-55
Called from Handle__Tactic.handle in file "src/handle/tactic.ml", line 698, characters 31-56
Called from Handle__Tactic.handle in file "src/handle/tactic.ml", line 718, characters 4-32
Called from Parsing__Syntax.fold_proof.proofstep in file "src/parsing/syntax.ml", line 527, characters 28-53
Called from Stdlib__List.fold_left in file "list.ml", line 121, characters 24-34
Called from Stdlib__List.fold_left in file "list.ml", line 121, characters 24-34
Called from Handle__Command.handle in file "src/handle/command.ml", line 642, characters 6-107
Called from Handle__Compile.compile_with.consume in file "src/handle/compile.ml", line 63, characters 38-60
Called from Common__Debug.stream_iter.(fun).do_rec in file "src/common/debug.ml", line 175, characters 40-45
Called from Handle__Compile.compile_with in file "src/handle/compile.ml", line 64, characters 6-55
Called from Dune__exe__Lambdapi.check_cmd.run.handle in file "src/cli/lambdapi.ml", line 130, characters 21-46
Called from Stdlib__List.iter in file "list.ml", line 110, characters 12-15
Called from Common__Error.handle_exceptions in file "src/common/error.ml", line 72, characters 6-10
````

fblanqui:
This problem should be solved now. It were coming from the fact that tactic terms were not type-checked.

fblanqui:
CI checked in https://github.com/Deducteam/lambdapi/pull/1380

melanie-taprogge:
I encountered two more problems, which are demonstrated in the attached 
[example.zip](https://github.com/user-attachments/files/28301959/example.zip)

1. **Lambdapi incorrectly gives "KIND is not unifiable with String" errors when compiled files already contain a `#rewrite "" "" ....` tactic.**
  You will only see this once you compile the package with `make`. Then, `UserTactic`, which contains a tactic defined with `#rewrite "" "" ...` is compiled first and then the file `more_tactics` is handled. The tactic there also relies on `#rewrite "" "" ...` and does not cause any issues before. It seems to me like compiling the first file somehow causes Lambdapi to assert that "" is a `KIND` rather than a `String`.

2. **Using a protected symbol useful for defining some tactics is no longer possible**
  In order to define tactics rewriting with theorems depending on a variable of type `Set`, we introduced 
  ```
   protected symbol set0 : Set;
   rule τ set0 ↪ Set; 
   ``` 
   (see file `UserTactic.lp` and the tactic `norm_lit`)
   We defined this symbol as a protected symbol, as adding it to the theory or proofs themselves breaks theoretical meta-properties (... right?). However, the new type checking of tactics reveals this symbol in some cases, causing `Protected symbol not allowed here.` Errors. I am aware that this is not really a bug, but it breaks some of my tactics...

fblanqui:
The first problem is fixed by the last commit: it revealed a bug in the loading of string symbols.
As for the 2nd problem, could you please provide me with an example?

melanie-taprogge:
> The first problem is fixed by the last commit: it revealed a bug in the loading of string symbols. As for the 2nd problem, could you please provide me with an example?

Great, thank you for looking into it!
There should be an example for the second problem in the file I posted! The proof `example3` in the file `example.lp` fails with he `Protected symbol not allowed here` error when you uncomment the `eval norm_lit;` line.

fblanqui:
Is it ok if I merge the PR as it is now without solving the problem with protected symbols? This problem can be avoided by not declaring set0 as protected. This is not completely satisfactory but this is orthogonal to the initial issue and requires more thinking.

melanie-taprogge:
Yes, of course. I made the symbol non-protected and checked again, my entire benchmark checks now. So from my end, it seems like there are not any other issues with this PR. Thank you again for looking into it!

PR #418, opened 2020-06-09
[POC] run Elpi code in lambdapi

gares:
*do not merge* this is a POC to be discussed at a seminar on the 18th

TODO (hopefully for the 18th):
- [ ] support `Terms.Meta` in the HOAS encoding using Elpi's unification variables
- [ ] from `Terms.term Conversion.t` to `(Terms.term,Env.t,_) ContextualConversion.t`
- [ ] bind `Typing.infer(_constr)` and `Eval.eq_modulo`

PR #1378, opened 2026-05-23
Tools to use Elpi in Lambdapi + type classes

agontard:
Up-to-date version of https://github.com/Deducteam/lambdapi/pull/418, with important help from Enrico Tassi.
Draft adding tools to convert Lambdapi terms to and from Elpi, used to write a typeclass solver in Elpi, which is now called in tac_solve to help with inference.
### How to use
- opam install . (new dependency: Elpi) 
### Added syntax
- Symbol declaration modifiers: "typeclass" and "instance".
- Declaring existing symbols as tc or tc instance: `typeclass s` or `instance s`
- Note that currently, a class must have type `A → TYPE` for some `A`
### Example
An example of use is in tests/OK/elpi_isa_test.lp, which mimics the construction of groups from the Isabelle Group.thy file from the HOL session. It looks a bit overcomplicated without proper syntax for record types, I did not even use inductive syntax.
### TODO
When changing the syntax of Lambdapi, make sure to update the
following files:
- [x] `doc/lambdapi.bnf`
- [x] `src/core/lpLexer.ml`
- [x] `src/core/lpParser.ml`
- [x] `src/core/pretty.ml`
- [x] `src/core/print.ml`
- [x] `editors/vim/syntax/lambdapi.vim`
- [x] `editors/emacs/lambdapi-vars.el` (syntax coloring)
- [x] `editors/emacs/lambdapi-smie.el` (grammar and indentation)
- [x] `editors/vscode/lp.configuration.json` (comments configuration),
- [x] `editors/vscode/syntaxes/lp.tmLanguage.json` (syntax highlighting),
- [x] `misc/lambdapi.tex`
- [x] the User Manual files in the `doc/` repository (do `make doc` to generate and check the Sphynx documentation).


agontard:
@fblanqui are the following two files contradicting each other? Or perhaps it would be clearer if the second one explained how the file should be updated? 

https://github.com/Deducteam/lambdapi/blob/1e3b2ba3326d4a33afc7a5377e28d613c64c91dd/doc/README.md?plain=1#L58-L60

https://github.com/Deducteam/lambdapi/blob/1e3b2ba3326d4a33afc7a5377e28d613c64c91dd/CONTRIBUTING.md?plain=1#L67-L69

fblanqui:
Fixed in https://github.com/Deducteam/lambdapi/commit/7c36a3f7b30d3268754e0bf925cedd0f14e6e058 . The BNF grammar is not generated anymore. It is written by hand.

agontard:
Report:
- Just merged with master
- I reevaluated loading times now that lambdapi is faster: doing lambdapi check on an empty file takes 0.045s with master and 0.08s with this PR (presumably due to loading elpi). This is satisfying to me, but I believe I could try improving if necessary by using the Marshall module.
- Two errors appear in tests:
  + I'm using a function that does not exist in OCaml 4.14.4, which I shall fix in the next commit.
  + It seems that dune build / tests are different in Ocaml 5.5.0, where the test directory is not at the same place, and therefore my ad hoc way of locating files I need fails. I don't know yet of a better solution to avoid this kind of problems happening.

PR #1380, opened 2026-05-26
Eval tactic: weak head normalize argument step by step

### LINKED ISSUE #1374: #rewrite tactic appears to unfold symbols in transparent rewrite lemmas before matching

Alidra:
fix https://github.com/Deducteam/lambdapi/issues/1374
replace full snf reduction of tactic terms by a more incremental reduction strategy using whnf and reducing only subterms that are tactic terms
infer the type and solve unification constraints of tactic terms before reduction and interpretation as tactics
simplify the interface of compilation functions


fblanqui:
CI works again on #1376

PR #1397, opened 2026-06-08
coq export: check that identifiers are valid in rocq

### LINKED ISSUE #1391: Feature request: sanitize DK identifiers that are illegal in Rocq
## Summary

The DK-to-Rocq exporter can emit Rocq declarations whose names are not legal
Rocq identifiers.

In [minimal-example.zip](https://github.com/user-attachments/files/28643350/minimal-example.zip), DK accepts a name such as `1_p0`, but Rocq cannot parse an identifier starting with a digit.

## Steps to reproduce

```sh
cd minimal-example
./run.sh
```

The DK input is:

```dk
1_p0 : Type.
```

## Expected behavior

The Rocq exporter should produce a legal fresh Rocq identifier, for example:

```coq
Axiom lp_1_p0 : Type.
```

The exact sanitization scheme is not important, but the generated file should
be accepted by Rocq and should avoid collisions.

## Actual behavior

The generated Rocq is:

```coq
Axiom 1_p0 : Type.
```

`coqc` rejects this file with a syntax error.

## Request

Please consider adding Rocq identifier sanitization to the DK-to-Rocq exporter
for names that are legal in Lambdapi/DK but illegal in Rocq.

## Observed version

Observed with Lambdapi `3.0.0-72-g1e3b2ba`.

fblanqui:
This PR fixes #1391. The translation of an identifier is as follows:
- We first check whether it must be renamed using the maps given with the options --renaming and --mapping. If so, we just return the string given by the renaming maps without checking it further.
- Otherwise, we check that it is an ASCII identifier valid in Rocq. If not, we fail.

Hence, for instance, ε will fail, unless it is renamed to itself in the --renaming map (this works because we do not check the validity of strings to which identifiers are renamed).

Recognizing the unicode identifiers accepted by Rocq is non trivial (see https://github.com/rocq-prover/rocq/blob/master/clib/unicode.ml and https://github.com/rocq-prover/rocq/blob/master/parsing/cLexer.ml). As files do not usually contain so many unicode identifiers, asking users to declare them in --renaming seems like a good compromise.

Impact on performances:
- on holide library (626 Mb of dk files): the PR takes 33.3s while master takes 32.3s (+3%)

Other changes:
- We check that identifiers declared with the options --renaming and --mapping are not renamed twice.

TODO:
- [x] measure impact of identifier validity checking on performances
- [ ] add option to deactivate validity checking ?
- [ ] port these changes to lean export (#1197)

### DIFF
--- a/src/export/coq.ml
+++ b/src/export/coq.ml
@@ -23,7 +23,7 @@ open Core
 let log = Logger.make 'x' "xprt" "export"
 let log = log.pp
 
-(** Valid Rocq identifiers. *)
+(** Valid Rocq identifiers (excluding unicode characters). *)
 
 let is_valid_first_letter =
   function 'a'..'z' | 'A'..'Z' | '_' -> true | _ -> false
@@ -81,7 +81,7 @@ let sym b = builtin.(index_of_builtin b)
 
 let rmap = ref StrMap.empty
 
-let add_renaming id1 id2 = check id2; rmap := StrMap.add id1 id2.elt !rmap
+let add_renaming id1 id2 = rmap := StrMap.add id1 id2.elt !rmap
 
 let set_renaming : string -> unit = fun f ->
   let consume = function
--- a/src/export/coq.ml
+++ b/src/export/coq.ml
@@ -80,8 +80,11 @@ let sym b = builtin.(index_of_builtin b)
 (** Set renaming map from file. *)
 
 let rmap = ref StrMap.empty
+let codom = ref StrSet.empty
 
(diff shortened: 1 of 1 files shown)


fblanqui:
@melanie-taprogge @agontard @navimath Do you agree with those changes?

melanie-taprogge:
> @melanie-taprogge @agontard @navimath Do you agree with those changes?

I agree with the changes. You are of course right that automatic renaming could introduce collisions and make names harder to trace. Fail-fast behaviour together with explicit renaming is surely the better solution. I can sure work with this solution in my translation pipeline. Thank you for looking into it Frédéric.

PR #1399, opened 2026-06-09
lexer: fail when escaping a regular id

fblanqui:
- lexing: fail when the user escape a regular identifier
- dk export: escape only identifiers that are not regular in Dedukti, fail if a module identifier is not valid in Dedukti


fblanqui:
Abandoned and replaced by #1401 .

PR #1408, opened 2026-06-13
Replace list with nelist for token parsing

Alidra:
This PR updates the list parser handling by replacing some calls to the `list` function with the non-empty list (`nelist`) function. The previous implementation allowed an empty token list, which could lead to invalid parser states. These changes enforce that the parser always receives at least one token, preventing the empty-token case.

### DIFF
--- a/src/parsing/lpParser.ml
+++ b/src/parsing/lpParser.ml
@@ -578,7 +578,7 @@ and inductive (lb:lexbuf): p_inductive =
         let l = c::cs in
         extend_pos (*__FUNCTION__*) pos0 (i,t,l)
     | VBAR ->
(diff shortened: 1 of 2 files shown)


fblanqui:
Hi Abdelghani. I don't know why CI fails. It works on my machine...

fblanqui:
I have the same problem in https://github.com/Deducteam/lambdapi/pull/1403.

fblanqui:
CI now fails because the Tedomum forge hosting Pratter is down. I sent a message to @gabrielhdt . It would be good to migrate Pratter to github/Deducteam to ensure its long-term availability and maintenance.

PR #1416, opened 2026-06-22
add (i)assumption tactics

bodeveix:
This PR add the tactic: assumption. assumption proves the current goal if it is an instance of one of the hypotheses.

TODO:
- [x] check CONTRIBUTING.md
- [x] make sanity_check succeeds
- [x] update CHANGES.md
- [x] update the above message: this is what will be added as commit (note that all commits are squashed into a single one when the PR is merged)

### DIFF
--- a/doc/tactics.rst
+++ b/doc/tactics.rst
@@ -36,7 +36,7 @@ If the focused typing goal is of the form ``Π x₁ … xₙ,T``, then
 ``assumption``
 --------------
 
-Proves the current goal if it is (an instance of) an hypothesis
+Proves the current goal if it is (an instance of) an hypothesis.
 
 .. _fail:
 
--- a/doc/tactics.rst
+++ b/doc/tactics.rst
@@ -36,7 +36,7 @@ If the focused typing goal is of the form ``Π x₁ … xₙ,T``, then
 ``assumption``
 --------------
 
-Proves the current goal if it is present in the hypotheses
+Proves the current goal if it is (an instance of) an hypothesis
 
 .. _fail:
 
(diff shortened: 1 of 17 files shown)


doc/tactics.rst
```diff
@@ -31,6 +31,13 @@ If the focused typing goal is of the form ``Π x₁ … xₙ,T``, then
 ``assume h₁ … hₙ`` replaces it by ``T`` with each ``xᵢ`` replaced by
 ``hᵢ``.
 
+.. _assumption:
+
+``assumption``
+--------------
+
+Proves the current goal if it is (an instance of) an hypothesis
```
fblanqui: dot missing at the end

src/handle/tactic.ml
```diff
@@ -551,6 +555,24 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
...
       end
+  | P_tac_assumption ->
+     let rec find_assumption = function
+         [] -> fatal pos "no matching assumption for %a" term gt.goal_type
+       | (_,(v,_t,_))::al ->
+(*
+          let idmap = get_names g in
+          let v0 = mk_Vari v in
+          let v1 = p_term pos idmap v0 in
+*)
```
fblanqui: Please remove dead code/comments.

src/handle/tactic.ml
```diff
@@ -551,6 +555,24 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
...
+          let v = P.var v in
```
fblanqui: use instead: Pos.make pos (mk_var v)

src/handle/tactic.ml
```diff
@@ -551,6 +555,24 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
         let u = mk_Meta (m2, Array.append ts [|mk_Meta (m1, ts)|]) in
         tac_refine pos ps gt gs p u
       end
+  | P_tac_assumption ->
+     let rec find_assumption = function
+         [] -> fatal pos "no matching assumption for %a" term gt.goal_type
```
fblanqui: please write "| []" instead so that each branch is preceded by "|"

fblanqui (changes requested):
Thanks JP for your PR! Only a few small changes are needed.

fblanqui:
You also need to make the CI succeed. You need to exclude your new test file from test/export_raw_dk.sh.

PR #1417, opened 2026-06-23
add focus tactic

bodeveix:
Add a focus tactic that allows changing the currently focused goal.

### DIFF
--- a/doc/lambdapi.bnf
+++ b/doc/lambdapi.bnf
@@ -101,7 +101,7 @@
            | "change" <term>
            | "eval" <term>
            | "fail"
-           | "focus" <int>
+           | "focus" <integer>
            | "generalize" <uid>
            | "have" <uid> ":" <term>
            | "induction"
--- a/doc/lambdapi.bnf
+++ b/doc/lambdapi.bnf
@@ -101,6 +101,7 @@
            | "change" <term>
            | "eval" <term>
            | "fail"
+           | "focus" <int>
            | "generalize" <uid>
            | "have" <uid> ":" <term>
            | "induction"
--- a/doc/tactics.rst
+++ b/doc/tactics.rst
@@ -40,6 +40,14 @@ Proves the current goal if it is (an instance of) an hypothesis.
 
 .. _fail:
 
+
+.. _focus:
+
+``focus <n>``
+-------------
+
+Makes the nth goal become the current goal.
+
 ``change``
 ----------
 
--- a/doc/tactics.rst
+++ b/doc/tactics.rst
@@ -46,7 +46,7 @@ Proves the current goal if it is (an instance of) an hypothesis.
 ``focus <n>``
 -------------
 
-Makes the nth goal become the current goal.
+Makes the nth goal (starting from 1) become the current goal.
 
 ``change``
 ----------
--- a/editors/emacs/lambdapi-smie.el
+++ b/editors/emacs/lambdapi-smie.el
@@ -102,7 +102,7 @@ Indent by `lambdapi-indent-basic' in proofs, and 0 otherwise."
               ("change" sterm)
               ("eval" sterm)
               ("fail")
-              ("focus" int)
+              ("focus" sterm)
               ("generalize" ident)
               ("have" ident ":" sterm)
               ("induction")
--- a/editors/emacs/lambdapi-smie.el
+++ b/editors/emacs/lambdapi-smie.el
@@ -102,7 +102,7 @@ Indent by `lambdapi-indent-basic' in proofs, and 0 otherwise."
               ("change" sterm)
               ("eval" sterm)
               ("fail")
-              ("focus" ident)
+              ("focus" int)
               ("generalize" ident)
               ("have" ident ":" sterm)
               ("induction")
--- a/src/handle/tactic.ml
+++ b/src/handle/tactic.ml
@@ -203,6 +203,7 @@ type tactic =
   | T_assumption
   | T_change
   | T_fail
+  | T_focus
   | T_generalize
   | T_have
   | T_induction
@@ -236,6 +237,7 @@ let get_config (ss:Sig_state.t) (pos:Pos.popt) : config =
   add "assumption" T_assumption;
   add "change" T_change;
   add "fail" T_fail;
+  add "focus" T_focus;
   add "generalize" T_generalize;
   add "have" T_have;
   add "induction" T_induction;
@@ -340,6 +342,18 @@ let p_rwpatt_of_string (pos:popt) (t:term): p_rwpatt option =
            Some (Parsing.Parser.Lp.parse_rwpatt_string p string)
   | _ -> fatal pos "not a string literal"
 
+(** [p_int_of_string pos t] returns the int contained in a string literal
+    term [t]. *)
+let p_int_of_string : popt -> term -> int = fun pos t ->
+  match t with
+  | Symb s when String.is_string_literal s.sym_name ->
+      begin
+        let string = String.trim (remove_quotes s.sym_name) in
+        try int_of_string string
+        with _ -> fatal pos "string should contain an int"
+      end
+  | _ -> fatal pos "not a string literal"
+
 let is_right (pos:popt) (t:term): bool =
   match t with
   | Symb s when String.is_string_literal s.sym_name ->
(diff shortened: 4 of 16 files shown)


src/handle/tactic.ml
```diff
@@ -340,6 +342,18 @@ let p_rwpatt_of_string (pos:popt) (t:term): p_rwpatt option =
            Some (Parsing.Parser.Lp.parse_rwpatt_string p string)
   | _ -> fatal pos "not a string literal"
 
+(** [p_int_of_string pos t] returns the int contained in a string literal
+    term [t]. *)
+let p_int_of_string : popt -> term -> int = fun pos t ->
+  match t with
+  | Symb s when String.is_string_literal s.sym_name ->
+      begin
+        let string = String.trim (remove_quotes s.sym_name) in
```
fblanqui: Why using String.trim? s.sym_name cannot contain spaces.

src/handle/tactic.ml
```diff
@@ -340,6 +342,18 @@ let p_rwpatt_of_string (pos:popt) (t:term): p_rwpatt option =
...
+        try int_of_string string
+        with _ -> fatal pos "string should contain an int"
```
fblanqui: explicit exception name like in https://github.com/Deducteam/lambdapi/blob/f3286759f3b8a1ec3aaa55706a67dd36eb020de5/src/handle/query.ml#L92-L94

src/handle/tactic.ml
```diff
@@ -340,6 +342,18 @@ let p_rwpatt_of_string (pos:popt) (t:term): p_rwpatt option =
            Some (Parsing.Parser.Lp.parse_rwpatt_string p string)
   | _ -> fatal pos "not a string literal"
 
+(** [p_int_of_string pos t] returns the int contained in a string literal
+    term [t]. *)
+let p_int_of_string : popt -> term -> int = fun pos t ->
```
fblanqui: rename into int_of_term

src/handle/tactic.ml
```diff
@@ -512,6 +528,18 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
       let id =  Pos.make pos (P_Abst(vparam,idbody)) in
       let pt = Pos.make pos (P_Appl(id,Pos.make pos P_Wild)) in
       tac_refine pos ps gt gs (new_problem()) (scope pt)
+  | P_tac_focus n -> begin
```
fblanqui: I think that this should be moved after https://github.com/Deducteam/lambdapi/blob/f3286759f3b8a1ec3aaa55706a67dd36eb020de5/src/handle/tactic.ml#L455 .

src/handle/tactic.ml
```diff
@@ -512,6 +528,18 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
...
+      let rec extract n l =
+        match l with
+        | [] -> raise Not_found
+        | x::l' -> if n = 0 then x,l'
+                   else let (y,r) = extract (n-1) l' in (y,x::r) in
+      try
+        let (x,gs) = extract n ps.proof_goals in
+        {ps with proof_goals = x::gs}
+      with
+        Not_found -> ps
```
fblanqui: a fatal should be raised if the focus is out of bound

editors/emacs/lambdapi-smie.el
```diff
@@ -102,7 +102,7 @@ Indent by `lambdapi-indent-basic' in proofs, and 0 otherwise."
               ("change" sterm)
               ("eval" sterm)
               ("fail")
-              ("focus" ident)
+              ("focus" int)
```
fblanqui: I don't think that int is defined, is it? just leave ident otherwise

doc/tactics.rst
```diff
@@ -40,6 +40,14 @@ Proves the current goal if it is (an instance of) an hypothesis.
 
 .. _fail:
 
+
+.. _focus:
+
+``focus <n>``
+-------------
+
+Makes the nth goal become the current goal.
```
fblanqui: nth -> n-th + add " (n≥2)" at the end (focus 1 is useless)

doc/lambdapi.bnf
```diff
@@ -101,6 +101,7 @@
            | "change" <term>
            | "eval" <term>
            | "fail"
+           | "focus" <int>
```
fblanqui: `<integer>`

src/parsing/lpParser.ml
```diff
@@ -1014,6 +1017,16 @@ and tactic (lb:lexbuf): p_tactic =
...
       consume_token lb;
       make_pos pos1 P_tac_fail
+  | FOCUS ->
+      let pos1 = current_pos() in
+      consume_token lb;
+      begin
+        match current_token() with
+        | INT n ->
+            consume_token lb;
+            extend_pos (*__FUNCTION__*) pos1 (P_tac_focus(int_of_string n))
```
fblanqui: P_tac_focus should take the string n as argument. The conversion to int should be done in tactic.ml.

src/parsing/pretty.ml
```diff
@@ -315,6 +316,7 @@ let rec tactic : p_tactic pp = fun ppf { elt;  _ } ->
   | P_tac_change t -> out ppf "change %a" term t
   | P_tac_eval t -> out ppf "eval %a" term t
   | P_tac_fail -> out ppf "fail"
+  | P_tac_focus n -> out ppf "focus %d" n
```
fblanqui: %s

src/parsing/syntax.ml
```diff
@@ -272,6 +272,7 @@ type p_tactic_aux =
   | P_tac_change of p_term
   | P_tac_eval of p_term
   | P_tac_fail
+  | P_tac_focus of int
```
fblanqui: of string

src/handle/tactic.ml
```diff
@@ -512,6 +528,18 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
...
+    end
```
fblanqui: Can't we code this with a tail-recursive function?
For instance, something like (in pseudo code):
```
let move_nth (n:int) (l:'a list): 'a list =
  let rec move (acc:'a list) n l =
    match l with
     | [] -> invalid_arg "remove_nth"
     | x::l -> if n <= 0 then x :: L.rev_append acc l else move (x:acc) (n-1) l
  in move []
```
to be added in lplib/list.ml
then the tactic code would be something like:
```
  let n = try ... int_of_string ... in
  if n < 2 || n > 1 + List.length gs then fatal pos ...
  else {ps with proof_goals = move_nth (n-1) (g::gs)}
```

fblanqui (changes requested):
Thanks JP for your PR. Here are some comments:

src/handle/tactic.ml
bodeveix: If #focus is used inside eval, it is declared as having the type String -> Tactic. Managing an Int would be too complex.

src/handle/tactic.ml
bodeveix: OK I use fatal ...

src/handle/tactic.ml

doc/lambdapi.bnf
bodeveix: OK

src/parsing/lpParser.ml
```diff
@@ -1014,6 +1017,16 @@ and tactic (lb:lexbuf): p_tactic =
...
       make_pos pos1 P_tac_fail
+  | FOCUS ->
+      let pos1 = current_pos() in
+      consume_token lb;
+      begin
+        match current_token() with
+        | INT n ->
+            consume_token lb;
+            extend_pos (*__FUNCTION__*) pos1 (P_tac_focus n)
+        | _ -> expected "" [INT "<int>"]
```
fblanqui: should be `INT ""` (the argument of INT is ignored)

src/lplib/list.ml
```diff
@@ -315,6 +315,19 @@ let pos : ('a -> bool) -> 'a list -> int = fun f ->
...
+let move_nth (n:int) (l:'a list): 'a list =
+  let rec move (acc:'a list) n l =
+    match l with
+     | [] -> invalid_arg "remove_nth"
+     | x::l -> if n <= 0 then x :: rev_append acc l
+               else move (x::acc) (n-1) l
+  in move [] n l
+let _ =
+  let r = move_nth 2 [0;3;2;1;4] in
+  assert (r = [2;0;3;1;4])
```
fblanqui: please inline r in case we want to add more unit tests
please add a few more unit tests (corner cases when n is the first or the last element)

src/lplib/list.ml
```diff
@@ -315,6 +315,19 @@ let pos : ('a -> bool) -> 'a list -> int = fun f ->
...
+(** [move_nth n l] moves the nth element of l (counting from 0) to its front,
+    or raises [Invalid_argument]. *)
+let move_nth (n:int) (l:'a list): 'a list =
+  let rec move (acc:'a list) n l =
+    match l with
+     | [] -> invalid_arg "remove_nth"
+     | x::l -> if n <= 0 then x :: rev_append acc l
+               else move (x::acc) (n-1) l
+  in move [] n l
+let _ =
```
fblanqui, line 329: please add a new and a comment "(* unit tests *)" before

fblanqui (commented):
I have a few more change requests.
Please fix also editors/emacs/lambdapi-smie.el.

PR #1422, opened 2026-06-24
Add #print + change interface of #assume

bodeveix:
Adds a #print term tactic 
Updates the #assume term tactic


fblanqui:
It would have been better to do 2 different PRs: one for print, and another one for #assume.

src/handle/tactic.ml
```diff
@@ -374,7 +376,16 @@ let p_tactic (ss:Sig_state.t) (g:goal) (env:Env.t) (pos:Pos.popt) (t:term)
             | T_and, _ -> assert false
             | T_apply, [_;t] -> P_tac_apply (p_term t)
             | T_apply, _ -> assert false
-            | T_assume, [t] -> P_tac_assume [Some(p_ident_of_sym pos t)]
+            | T_assume, [u;_;Abst (_, bi)] ->
+               let uname = string_of_term pos u in
+               let nv = new_var uname in
+               let uname = uniq_name nv in
+               let nv = new_var uname in
```
fblanqui: This line can be removed.

src/handle/tactic.ml
```diff
@@ -374,7 +376,16 @@ let p_tactic (ss:Sig_state.t) (g:goal) (env:Env.t) (pos:Pos.popt) (t:term)
...
+               let uname = Pos.make pos uname in
+               let uvar = mk_Vari nv in
+               let t' = subst bi uvar in
```
fblanqui: Please try to not use so many let's.

fblanqui (changes requested):
Thanks JP. But I don't understand why you add P_tac_print since we already have P_query_print. I would prefer to have #print interpreted as follows:
- if the argument of #print is "" then call P_query_print None
- if the argument is some symbol s then call P_query_print(Some s)
- otherwise fail

src/handle/tactic.ml
bodeveix: I have reduced the number of lets

PR #1425, opened 2026-06-25
add tactic first_hyp

bodeveix:
Tactic first_hyp and tactic term #first_hyp iterate a parameterised tactic on all hypotheses until success.


fblanqui:
Hi. Please update your branch wrt master.

doc/tactics.rst
```diff
@@ -40,6 +40,12 @@ Proves the current goal if it is (an instance of) an hypothesis.
 
 .. _fail:
 
+.. _first_hyp:
+
+``first_hyp <t>``
+-------------
+
+Iterates the application of parameterized term tactic t on all hypotheses until success.
```
fblanqui: Please specify the order, and say in which case it fails:

Starting from the first (or last?) hypothesis, evaluates the tactic (t h) on every hypothesis h until one succeeds, or fail.

src/handle/tactic.ml
```diff
@@ -538,6 +542,21 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
           tac_refine pos ps gt gs p t
         with Not_found -> fatal idpos "Unknown hypothesis %a" uid id;
       end
+  | P_tac_first_hyp tac ->
+      let tac = scope tac in
```
fblanqui: Please rename tac into t as it is a term.

src/handle/tactic.ml
```diff
@@ -538,6 +542,21 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
...
+      let rec find_assumption = function
+          [] -> fatal pos "tactic %a fails on all assumptions" term tac
```
fblanqui: write [%a] as t could be big

src/handle/tactic.ml
```diff
@@ -538,6 +542,21 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
...
+       | (_,(v,p,_))::al ->
+          let v = mk_Vari v in
+          try
+            let t = mk_Appl (mk_Appl (tac, p), v) in
+            log "FIRST_HYP %a\n" term t;
```
fblanqui: every call to log must be prefixed by "if Logger.enabled() then "

src/handle/tactic.ml
```diff
@@ -538,6 +542,21 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
           tac_refine pos ps gt gs p t
         with Not_found -> fatal idpos "Unknown hypothesis %a" uid id;
       end
+  | P_tac_first_hyp tac ->
+      let tac = scope tac in
+      let rec find_assumption = function
+          [] -> fatal pos "tactic %a fails on all assumptions" term tac
+       | (_,(v,p,_))::al ->
```
fblanqui: please rename al into env

bodeveix:
C’est fait pour l'itérateur first_hyp

tests/OK/Tactic.lp
```diff
@@ -31,6 +31,9 @@ builtin "fail" ≔ #fail;
 constant symbol #focus : String → Tactic;
 builtin "focus" ≔ #focus;
 
+constant symbol #first_hyp : (Π P, π P → Tactic) → Tactic;
```
fblanqui: please rename P into a (we reserve names starting with a capital letter for types)

fblanqui (changes requested):

tests/OK/Tactic.lp
bodeveix: OK. In fact, it would be interesting yo have both Set and Prop with a different name...

fblanqui: I would like to avoid that. There are already too many names used as keywords. We need to change that also. We may also wonder whether it is really useful to duplicate every tactic with tac and #tac...

PR #1426, opened 2026-06-25
add tactic all_hyps

bodeveix:
The tactic `all_hyps t` and the tactic term `#all_hyps t` runs the parameterised tactic t  on all the hypotheses, ignoring failing calls.


src/handle/tactic.ml
```diff
@@ -488,6 +492,21 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
...
+          match ps.proof_goals with
+            | [] -> ps
+            | g :: _ ->
+                let v = mk_Vari v in
+                let t = mk_Appl (mk_Appl (tac, p), v) in
+                log "ALL_HYPS on %a\n" term v;
+                let h = try handle ps (p_tactic ss g env pos t)
+                        with Fatal _ -> ps in
+                all_assumptions h al
+     in all_assumptions ps gt.goal_hyps
```
fblanqui: same remarks as previous PR: tac -> t, al -> env, p -> a, log, etc.
you can also unfold v

could you please implement that function with List.fold_left?

and make the tactic fail in case all calls failed

fblanqui (changes requested):

src/handle/tactic.ml
bodeveix: ok...

PR #1430, opened 2026-06-26
#with_goal tactic

bodeveix:
The tactic term with_goal (only available in eval) calls its parameter tactic with the current goal seen as a Prop.


src/handle/tactic.ml
```diff
@@ -194,6 +194,56 @@ let get_prod_ids env =
...
+  in
+  let is_set g =
+    match get_args g with
+    | t, (u::_) when is_symb cfg.symb_T t -> Some u
+    | _ -> None
+  in
+  let rec as_prop g =
+    match is_prf g with
+    | Some u -> u
+    | None -> match g with
```
fblanqui: matching on terms must always be of the form `match unfold t with ...`

src/handle/tactic.ml
```diff
@@ -194,6 +194,56 @@ let get_prod_ids env =
...
+              | Prod(p,bi) when binder_name bi = "_" ->
```
fblanqui: testing binder_names is not sound; you should use binder_occur instead

tests/OK/with_goal.lp
```diff
@@ -0,0 +1,24 @@
+require open tests.OK.Tactic;
+
+injective symbol σ: Prop → Set;
+rule π $P ↪ τ (σ $P);
```
fblanqui, line 4: Why do you need this?

CHANGES.md
```diff
@@ -32,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - LSP server: position of the error is removed from diagnostics when the error occurs in the file currently open in the editor.
 - Syntax of search query is modified as follows : `in` is used instead of `|` (filtering). `with` is used instead of `,` (conjunction). `|` is used instead of `;` (disjunction).
 - Type of `#assume` in order to generate a new symbol and use it inside a tactic term.
+- Type of `#refine' is in fact `String -> Tactic`
```
fblanqui: Why this change in this PR?

fblanqui (changes requested):
Hi Jean-Paul.

1) If there are problems with assume, please open an issue with an example, and propose a separate PR to solve that problem.

2) A more general approach to with_goal, not requiring gconf.ml, would be to generate a unification problem goal == Prf ?M, solve it to instantiate ?M and, in case of success, handle the tactic (t ?M).

Yes but I suppose this would need another tactic to solve the unification problem automatically.

fblanqui:
I see the problem with assume. I'll fix it this afternoon.

src/handle/tactic.ml
bodeveix: ok

src/handle/tactic.ml
bodeveix: ok

tests/OK/with_goal.lp
bodeveix: Using this, a unique profile for #assume can be used for both Set and Prop. The idea is to say that each object of type Prop can itself be seen as a type, which is given by the function σ.

CHANGES.md
bodeveix: because I saw it accidentally...

bodeveix:
The goal is now matched with (Prf _) through LP standard unification. The example uses the new encoding of Set and Prop declared in Univ. The corresponding unification rules replace what was done in OCaml.

src/core/builtin.ml
```diff
@@ -240,6 +240,7 @@ let _ =
   register_typ "solve" tac;
   register_typ "symmetry" tac;
   register_typ "try" (arr tac tac);
-  register_typ "why3" tac
+  register_typ "why3" tac;
+  register_typ "with_goal" (arr (arr prop tac) tac);
```
fblanqui: Why restricting goals to Prop? It could be of type `Π l:L, U l → Tactic`.

src/core/sign.ml
```diff
@@ -112,7 +112,7 @@ let link : t -> unit = fun sign ->
               StrMap.add s.sym_name s !(Ghost.sign.sign_symbols);
             s
           end
-        else assert false
+        else begin failwith s.sym_name end
```
fblanqui: please use the qualified name:
`Console.out 0 "link: didn't find %a" Raw.sym s`
and add `let qsym = qsym let _ = qsym` in Core.Print.Raw

src/handle/tactic.ml
```diff
@@ -805,7 +824,7 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
           fatal pt.pos "Cannot infer the type of [%a]" term t
       | Some(t,_) ->
         if Unif.solve_noexn p then
-          let ps, t = p_tactic ps g env pos t in handle ps t
+          let ps,t = p_tactic ps g env pos t in handle ps t
```
fblanqui: unnecessary change

CHANGES.md
```diff
@@ -37,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Type of `#assume` in order to generate a new symbol and use it inside a tactic term.
 - Errors occurring while a proof is in progress now report the proof state: the goals a failing tactic was applied to, the goals before and after the tactic for a subproof-count mismatch, and the remaining goals when a proof is unfinished at `end`. The state is printed after the error message, which stays unchanged. The LSP server does not attach the proof state to tactic failures since editors display it themselves.
 - Lambdapi does not use Cmdliner anymore.
+>>>>>>> dk/master
```
fblanqui: to be removed

CHANGES.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Tactic `all_hyps t` calls parameterized tactic term t on all hypotheses ignoring failing calls.
 - Extend `print` query to the following arguments: `verbose`, `debug`, `flag`, `builtin`, `prover`, `prover_timeout`.
 - add a version number to the header of the index db file to prevent crash of LP when the structure of db changes.
+- Tactic `#with_goal t` which calls term tactic t with current goal of type Prop as parameter.
```
fblanqui: applies the tactic producing term t to the current goal

doc/structure.rst
```diff
@@ -87,6 +87,7 @@ Overview of directories and files
 
   * ``command.ml``: command handling
   * ``compile.ml``: file parsing and compiling (.lpo files)
+  * ``gconf.ml``: builtins needed to build current goal (tactic #with_goal)
```
fblanqui: to be removed

fblanqui (changes requested):
Hi JP. Here are some comments.

src/core/builtin.ml
bodeveix: OK, I have changed to this profile: Π [l], (U l → Tactic) → Tactic. In general the parameter is not polymorphic.

src/core/sign.ml
bodeveix: OK

src/handle/tactic.ml

CHANGES.md
bodeveix: OK

CHANGES.md
bodeveix: I have changed the comment: Tactic `#with_goal t` applies the tactic t with current goal as parameter.

doc/structure.rst
bodeveix: OK

PR #1440, opened 2026-07-03
Fix lean export of unmapped types

navimath:


fblanqui:
Hi Ivan. There are several problems with this PR:
- The title is incorrect.
- There is no description.
- CI fails.
- It includes many unrelated commits. This means that you didn't start from an up-to-date master branch.
- The diff is not minimal.
Please fix all this.

PR #1441, opened 2026-07-03
Fix parser errors

sacerdot:
Work in progress, but already better.

This commit makes the parser reject syntax that shall be rejected. Moreover it improves on the error messages.

Bugs still known:
a) an error message following a "production" that end with *, + or ? should list as expected also the set of symbols accepted in head position by the preceding production. A possible implementation consists in adding to the lexbuf data structure such a set
b) some error messages (marked with CSC) do not use the list of tokens expected, but an hand written string. Feature? Sloppiness? Some even pass both parameters: in that case the list of tokens is currently ignored.


sacerdot:
a [continued]) The case of * and + is now fixed by letting the zipper lexbuf remember a list of tokens expected at the end. The one of ? not yet (since it has no combinator and it is a bit all over the grammar) but it is now easy to fix.
c) the lexer and parser share the same exception. However, the nature of lexing errors is different: a lexing error is unrecoverable and it occurs during consume_token, but on the *next* token to be consumed. I used a boolean to differentiate and do not backtrack any more on lexing exceptions. The error is now reported in the correct position, but not using "expected" as usual. Therefore the quality of lexer errors is not on par with the one of parsing errors. I do not see an easy fix and I won't fix this. A potential fix could be to produce an ERROR token in the lexer holding a string explaining the error and make the expected function recognize it and include it in the expected error message. I leave this to @fblanqui

sacerdot:
a [continue]) completed: also the case ? is done. In practice, sometimes it is possible to use the opt combinator; in all remaining cases, if a catchall |_ -> is not followed by an expected, then it must invoke set_expected_tokens

sacerdot:
d) The parser is now fully implemented as an LL(1) parser, further improving the error messages. In particular the zipperTokenLexbuf.ml{,i} files are now renamed to .unused so that they can be committed to GitHub before removing them from GitHub. @fblanqui do that so that in case one needs/wants later to allow backtracking productions/non LL(1) grammars he can do that without having to reimplement it again. Two new files oneLookaheadCellLexbuf.ml{,i} are now used, holding a 1-token lookahead cell only, plus additional fields for error management.


@fblanqui this was my last commit on this pull request. I won't have time to be active on this in the next period. I suggest you review and merge it and then you review every (* CSC: *) comment to decide what error message to produce. As far as I know it passes all tests and it produces correct and well-localized error messages on the few (!) tests I made.
@Alidra it is probably a good idea to change the current OK/KO testing framework to be able to inspect the output in case of KO. Then you should implement negative tests for the various parsing errors you can get

fblanqui:
Thank you very much @sacerdot ! Is point c) "the quality of lexer errors is not on par with the one of parsing errors. I do not see an easy fix and I won't fix this. A potential fix could be to produce an ERROR token in the lexer holding a string explaining the error and make the expected function recognize it and include it in the expected error message" still an issue after your last changes? If so, could you please elaborate, give an example?

fblanqui:
Concerning "b) some error messages (marked with CSC) do not use the list of tokens expected, but an hand written string. Feature? Sloppiness? Some even pass both parameters: in that case the list of tokens is currently ignored.": in those cases, I prefer to provide a short message (e.g. "term expected") instead of a long list of possible tokens.

sacerdot:
Yes, c) still holds. Take a file and add an error somewhere. If it is a parsing error you will see the message produced by expected that lists the tokens expected. However, if it is a lexing error (e.g. you put a @ followed by a space), you get a simple error message saying that something found is wrong instead of seeing the list of expected tokens.

The point is that the lexing error occurs in the code that consumes the token, during the lookahead that fetches the next token. Therefore we are too low level to emit a reasonable "contextual" error message.

fblanqui:
I see. We also have the problem that the list of expected symbols depends on whether we parse rocq or lp syntax, but allow_rocq_syntax is available in the lexer only. Shouldn't we record allow_rocq_syntax in the lexbuf data structure?

sacerdot:
You are right, I did not think about it. Yes, a boolean should go in the lexbuf to be used to select the right expected tokens list!

src/parsing/lpParser.ml
```diff
@@ -133,7 +133,7 @@ let match_token tok patt =
   | SIDE _, SIDE _
   | STRINGLIT _, STRINGLIT _
   | SWITCH false, SWITCH false
-  | SWITCH true, SWITCH _
+  | SWITCH true, SWITCH true
```
sacerdot: Nope, the underscore was not a mistake! Have a look at the function at the beginning of the file that is used to print the tokens that occur in the expected list. You will see that "SWITCH true" is used as a generic pattern, while "SWITCH false" is used to match exactly false. This is because in the grammar there is an occurrence where you can either omit the token or set it to "SWITCH false" only. Revert this change.

sacerdot (commented):
Comment to highlight an introduced error.

fblanqui:
I am going to merge this PR now as it is and fix the expected tokens with rocq syntax in a separate PR.

PR #1444, opened 2026-07-06
fix(lsp): show success hints on the introducing keyword of commands and tactics

ciaran-matthew-dunne:
**Stacked on #1450** (bare `open` parsing fix): the first commit in the diff belongs to that PR; review the last three here.

The LSP server emits severity-4 "OK" diagnostics for every successfully checked command and tactic. These previously spanned the whole command or tactic, so the success underline covered symbol bodies, entire proofs and rule right-hand sides. With this PR the hint is shown on the introducing keyword only (`symbol`, `rule`, `require`, …; `assume`, `rewrite`, …).

Notes:

- For most commands the keyword position is computed as a prefix of the command's position (`Pos.prefix`). This is not possible for `symbol` (modifiers may precede the keyword), `inductive` (modifiers and parameters may) and `open` (a `private` modifier may), so for these three constructors the parser records the keyword position in the AST. `Syntax.command_keyword_pos` handles all cases, and falls back to the whole command's position when no keyword position was recorded (Dedukti input). `require` needs no recorded position: with #1450 it is always the first keyword of its command.
- Tactic hints keep the start of their position unchanged: the same (position, goals) records anchor the goals panel, and its lookup (`closest_before`) only reads position starts. Only the end of the range shrinks, so `proof/goals` answers are unchanged. The keyword is used rather than the tactic's end because the end is ill-defined once a tactic has subproofs.
- Symbol declarations always create proof data, so their "OK" goes through the proof path. The initial goals stay anchored at the symbol name (the goals panel shows the initial proof state there); the visible hint goes on the `symbol` keyword.
- The keyword tables (`tactic_keyword`, `query_keyword`) have no wildcard case, so adding a constructor is a compile error until they are updated.
- Error diagnostics are unchanged: they still span the whole command or tactic.
- Also renames the `lp_doc` node field `ast` to `cmd`.

### DIFF
--- a/src/lsp/lp_doc.ml
+++ b/src/lsp/lp_doc.ml
@@ -54,17 +54,6 @@ let buf_get_and_clear buf =
   let res = Buffer.contents buf in
   Buffer.clear buf; res
 
-(* [at_keyword p kw] is the sub-position of [p] covering only its leading
-   keyword [kw] (keywords are ASCII, so their string length is their width
-   both in columns and in character offsets). *)
-let at_keyword : Pos.popt -> string -> Pos.popt = fun p kw ->
-  match p with
-  | None -> None
-  | Some p ->
-      Pos.(Some { p with end_line = p.start_line
-                       ; end_col = p.start_col + String.length kw
-                       ; end_offset = p.start_offset + String.length kw })
-
 let process_pstep (pstate,diags,logs) tac nb_subproofs =
   let open Pure in
   let tac_loc = Tactic.get_pos tac in
@@ -77,11 +66,7 @@ let process_pstep (pstate,diags,logs) tac nb_subproofs =
     (* Show the success hint on the tactic's leading keyword only. This
        tuple also anchors the goals panel (see [get_goals]), whose lookup
        reads the start of the position: the start must not move. *)
-    let hint_loc = match Tactic.keyword tac with
-      | Some kw -> at_keyword tac_loc kw
-      | None -> tac_loc
-    in
-    pstate, (hint_loc, 4, qres, goals) :: diags, logs
+    pstate, (Tactic.keyword_pos tac, 4, qres, goals) :: diags, logs
   | Tac_Error(loc,msg) ->
     let loc = option_default loc tac_loc in
     let goals = Some (current_goals pstate) in
@@ -101,31 +86,6 @@ let get_goals dg_proof =
   in get_goals_aux [] dg_proof
 (* XXX: Imperative problem *)
 
-(* [at_semicolon p] is the position of the ";" terminating a command whose
-   position is [p]. Success ("OK") hints are shown there so that they do
-   not underline the whole command. Position ends are exclusive (as in the
-   LSP ranges they are mapped to), so the position of the one-character
-   ";" ends one column after it. The parser's [extend_pos] ends [p] one
-   character before the ";", except when the ";" is at column 0, where it
-   does not back up over the newline: a ";" at column 0 and one at column
-   1 both yield [end_col = 0] (with [end_offset] the offset of column 0),
-   so in that case the returned position covers both columns. *)
-let at_semicolon : Pos.popt -> Pos.popt = function
-  | None -> None
-  | Some p ->
-      Pos.(
-        if p.end_col = 0 then
-          Some { p with start_line = p.end_line; start_col = 0
-                      ; start_offset = p.end_offset
-                      ; end_col = 2
-                      ; end_offset = p.end_offset + 2 }
-        else
-          Some { p with start_line = p.end_line
-                      ; start_col = p.end_col + 1
-                      ; start_offset = p.end_offset + 1
-                      ; end_col = p.end_col + 2
-                      ; end_offset = p.end_offset + 2 })
-
 let process_cmd _file (nodes,st,dg,logs) cmd =
   let open Pure in
   (* let open Timed in *)
@@ -139,8 +99,8 @@ let process_cmd _file (nodes,st,dg,logs) cmd =
   | Cmd_OK (st, qres) ->
     let qres = match qres with None -> "OK" | Some x -> x in
     let nodes = { cmd; exec = true; goals = [] } :: nodes in
-    (* Show the success hint on the terminating ";", not the whole command. *)
-    let ok_diag = at_semicolon cmd_loc, 4, qres, None in
+    (* Show the success hint on the command's introducing keyword only. *)
+    let ok_diag = Command.keyword_pos cmd, 4, qres, None in
     nodes, st, ok_diag :: dg, logs
   | Cmd_Proof (pst, tlist, thm_loc, qed_loc) ->
     let start_goals = current_goals pst in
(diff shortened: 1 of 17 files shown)


ciaran-matthew-dunne (commented):
Looks good to me.

src/pure/pure.ml
```diff
@@ -48,6 +61,13 @@ module Tactic = struct
   let equal = Syntax.eq_p_tactic
   let get_pos t = Pos.(t.pos)
   let print = Util.located Pretty.tactic
+
+  (** Focused position for OK diagnostics: the tactic's leading keyword,
+      or the full tactic span for infix tacticals. *)
+  let get_focus_pos (t : t) : Pos.popt =
```
fblanqui: Please rename to keyword_pos as focus as some other meaning (focused goal and focus tactic).

src/pure/pure.ml
```diff
@@ -23,6 +23,19 @@ module Command = struct
...
   let get_pos c = Pos.(c.pos)
   let print = Util.located Pretty.command
+
+  (** Focused position for OK diagnostics: the command's keyword rather
+      than its whole span, so that a success hint does not underline entire
+      declarations and proofs. The position of a command starts at its
+      keyword (modifiers are not included in it), so taking a prefix of the
+      command's position suffices, except for inductive definitions with
+      leading parameters, whose keyword position is recorded in the AST. *)
+  let get_focus_pos (c : t) : Pos.popt =
```
fblanqui: Please rename to keyword_pos as focus as some other meaning (focused goal and focus tactic).

src/lsp/lp_doc.ml
```diff
@@ -54,7 +54,7 @@ let buf_get_and_clear buf =
   let res = Buffer.contents buf in
   Buffer.clear buf; res
 
-let process_pstep (pstate,diags,logs) tac nb_subproofs =
+let process_pstep (pstate, diags, logs) tac nb_subproofs =
```
fblanqui: please minimize diff

src/lsp/lp_doc.ml
```diff
@@ -96,12 +97,13 @@ let process_cmd _file (nodes,st,dg,logs) ast =
   | Cmd_OK (st, qres) ->
     let qres = match qres with None -> "OK" | Some x -> x in
     let nodes = { ast; exec = true; goals = [] } :: nodes in
-    let ok_diag = cmd_loc, 4, qres, None in
+    let ok_diag = Command.get_focus_pos ast, 4, qres, None in
```
fblanqui: ast is not an appropriate name in fact; thanks to take advantage of this PR to rename it to cmd

fblanqui (changes requested):

src/lsp/lp_doc.ml
```diff
@@ -83,27 +100,41 @@ let get_goals dg_proof =
...
-let process_cmd _file (nodes,st,dg,logs) ast =
+(* A command's recorded position ends one character before its terminating
+   ";" (the parser's [extend_pos] stops at the start of the following token).
+   [at_semicolon p] returns the position of that ";", where success ("OK")
+   hints are shown so that they do not underline the whole command. *)
+let at_semicolon : Pos.popt -> Pos.popt = function
+  | None -> None
+  | Some p ->
+      Pos.(Some { p with start_line = p.end_line; start_col = p.end_col + 1
+                       ; end_line = p.end_line;   end_col = p.end_col + 2 })
```
fblanqui: There are several problems here:

- start_offset and end_offset should be set properly as well

- Why not taking end_col = p.end_col + 1 (";" is one character long only)?

- But this position is wrong if the semicolon is at the beginning of the line as in:
```
command ...
;
```
In this case, we should use {p with start_line = p.end_line; start_col = 0; end_col = 0} instead.
Problem:
```
command ...
 ; // semicolon with one space before
```
gives the same position as the previous case because of the current definition of extend_pos which extends a position until the previous column of the next token if it is >0.

So we should instead do:
```
let at_semicolon (pos:Pos.popt): Pos.popt =
  match pos with
  | None -> None
  | Some p ->
    if p.end_col = 0 then {p with start_offset=p.end_offset; start_line=p.end_line; start_col=0;
                                                        end_offset=p.end_offset+1; end_col=1}
      (* in this case we don't know whether ";" is at position 0 or 1, so we include both columns *)
    else {p with start_offset=p.end_offset+1; start_line=p.end_line; start_col=p.end_col+1;
                            end_offset=P.end_offset+1; end_col=p.end_col+1}
```

To be fully precise and not include a useless space, we need to include ";" in the position of a command.

src/lsp/lp_doc.ml
```diff
@@ -54,6 +54,16 @@ let buf_get_and_clear buf =
...
 
+(* [at_keyword p kw] is the sub-position of [p] covering only its leading
+   keyword [kw] (keywords are ASCII, so their column width is their string
+   length). *)
+let at_keyword : Pos.popt -> string -> Pos.popt = fun p kw ->
+  match p with
+  | None -> None
+  | Some p ->
+      Pos.(Some { p with end_line = p.start_line
+                       ; end_col = p.start_col + String.length kw })
```
fblanqui: end_offset should be set properly as well, I believe to the same value as for end_col

fblanqui (changes requested):

src/lsp/lp_doc.ml
sacerdot: Wait. The problem with the position of the trailing ";" comes from line 1719 of lpParser.ml. Instead of
   | SEMICOLON -> c
one should do
   | SEMICOLON -> extend_pos lb ... c
   
 I think that that should fix the issue(*)
 
 (*) In truth, the fact that the "command" parser is not allowed to consume the terminator is just plain weird and a symptom of bad coding elsewhere.  The point is that in pure.ml (if I remember correctly) every time a new command is to be parsed, the lexbuf at the char Stream.t level is passed again. Instead the lexbuf at the lookahead level (the one I implemented) should be passed around.   Right now the main parser functions exported by parser take in input the lexbuf at char level and build on-the-fly a new lexbuf at the backtracking level. Instead one should export a function to turn a lexbuf into an higher level lexbuf and then pass that again and again. Or, a less impacting alternative on the code, the parse_lexbuf function in parsing should take the low level lexbuf and return a closure that goes from unit to command, remembering the high level lexbuf inside the closure. Then pure.ml could just iterate this closure.

ciaran-matthew-dunne: Oh well, I decided that the diagnostic should sit on the keyword of the command anyway.

e.g., the diagnostic for the command `constant symbol foo : TYPE;` is placed on `symbol`,
and similarly the diagnostic for `[a : Set] inductive T : TYPE ≔ ...` is placed on `inductive`.

Frédéric talked to me this morning about the possibility of retiring the use of the semicolon in `lambdapi`. So hopefully these changes in this PR are compatible with any future changes to the parser.

sacerdot: The fix at line 1719 I proposed should be implemented anyway since it is nevertheless a bug IMHO. About the idea of retiring the ; terminator: in that case it will be MANDATORY to also apply the long explanation improvement I described, because otherwise with no semicolon to stop at one character will be definitely lost at the next call.

src/lsp/lp_doc.ml
fblanqui: This has been fixed in https://github.com/Deducteam/lambdapi/pull/1448

PR #1445, opened 2026-07-06
Export to Lean: Fix on non-aligned types and warning-styles

navimath:


src/export/lean.ml
```diff
@@ -138,6 +138,61 @@ let open_mod oc p = string oc "open "; path oc p; string oc "\n"
 
 let openings = ref []
 
+let export_types oc p_sym_nam t =
```
fblanqui, line 141: Please add comments before function definitions explaining what they should do (specification).
This includes here: export_types and set_to_type.

PR #1448, opened 2026-07-09
fix virtual positions for symbols and rules generated by inductive

fblanqui:
- fix Pos.pos_end by setting start_offset properly
- remove Pos.before (unused)
- unfold Pos.after to Pos.shift 1 to easily localize all the uses of shift
- Tactic.p_term_of_string_term: fix initial buffer position using pos_of_string
- rename Lcr.shift into Lcr.reindex to avoid confusion with Pos.shift

### DIFF
--- a/src/common/pos.ml
+++ b/src/common/pos.ml
@@ -101,7 +101,8 @@ let equal : popt -> popt -> bool = fun p1 p2 ->
 let pos_end : popt -> popt = fun po ->
   match po with
   | None -> None
-  | Some p -> Some {p with start_line = p.end_line; start_col = p.end_col}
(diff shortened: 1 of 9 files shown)


fblanqui:
@sacerdot Is it all right in index_sym to use sym_pos instead of sym_decl ?

sacerdot:
No, I don't think so. The position corresponds to what it is fetched back and printed in the list of results. We want to print the whole symbol declaration, not just the name.

fblanqui:
I see. I'm going to revert the change on indexing.ml. But, then, we need to use something better than virtual positions to order declarations.

PR #1450, opened 2026-07-09
fix parsing of bare open: do not load modules as require open does

ciaran-matthew-dunne:
Since the LL(1) parser rework (#1441, 078dba47), a bare `[private] open M;` is parsed as `P_require(Some private?, ...)` — the same AST as `require [private] open M;`. Consequently:

- `open M;` silently requires `M` when it is not yet loaded, instead of failing with `Module "M" needs to be required first.` — `Handle.Command.handle_open` became unreachable, since no parser constructed `P_open` anymore;
- the lp export prints `open M;` back as `require open M;`.

This PR restores `P_open` for the bare forms (`open`, `private open`), keeps `P_require` for the `require` forms, and adds a KO test (`open` of a module that was not required). No change to the grammar itself.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,9 +36,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
-- A bare `open` (not preceded by `require`) was parsed as `require open`
-  since the parser rework, silently loading not yet required modules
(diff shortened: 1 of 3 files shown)


sacerdot:
Yes, my bad!

PR #1290, opened 2025-07-17
Improvements and new features for indexing

sacerdot:
Bugs to be fixed:

- [x]  "lambdapi websearch --require HOLLight.theory_hol" triggers an assert failure due to changes to recursive require/open

Todo:

- [x] docs and changelog need to be updated
- [x] test if search in vscode still works with latest version
- [x] use lplib/color.ml in place of our color management of the output [ but do not lose our code to check if we are in lsp_mode or targeting a tty ]
- [x] restore the black color after the error message in case of Fatal exception
- [x] commit the very nice new look&feel of websearch

New features (to be documented too):

1. new command "lambdapi deindex"
2. the current development (i.e. the set of currently required files) is now indexed and deindexed automatically
3. new filter   | "regexp"  to restrict the set of results using a regexp over the full qid
4. "lambdapi websearch" now uses a rocqParser that accepts (also) the syntax of Rocq for fun/forall/exists; it also remaps /\ \/ ~ to the corresponding LP tokens for logical connectives
5. new flag "--source" to load a map from LP identifiers to Rocq sources (file + line); the Rocq sources are also shown by websearch
6. the "--require" flag can now be passed multiple times (no longer necessary since Frederic has implemented recursive open, but nice to have anyway; see bug to be fixed above though)
7. when an identifier is ambiguous, but after normalization it is no more, the query is now considered non ambiguous
8. "lambdapi search" now streams the result
9. Plac in rewriting rule is now supported (handled like a variable... hoping that it makes sense)
10. bug fix: many spurious justifications were included
11. bug fix: avoid Stack_overflow by rewriting a few functions using tail recursion

PR #1453, opened 2026-07-10
Report the proof state when an error occurs during a proof

ciaran-matthew-dunne:
Errors raised during a proof now report the proof state. An unfinished proof at `end` and most subproof-count mismatches already print the goals, but a failing tactic prints nothing:

```
[foo.lp:12:2-6] Call to tactic "fail".
```

The state is now attached to every such error, in the `err_desc` field of `Fatal` (from #1290): the message is unchanged, and the state is printed uncolored after it.

```
[foo.lp:12:2-6] Call to tactic "fail".
Proof state before tactic application:
a: π A
b: π B
------------------------------------------------------------------------------
0. ?4: π A
```

A subproof-count mismatch is only detected after the tactic ran, so it reports the state before and after it:

```
[foo.lp:11:2-11] Missing subproofs (0 subproofs for 2 subgoals).
Proof state before tactic application:
0. ?1: Π n:N, π (Q n)
Proof state after tactic application:
0. ?5: π (Q z)
1. ?7: Π x0:N, π (Q x0) → π (Q (succ x0))
```

An unfinished proof at `end` reports the remaining goals under a `Proof state:` heading.

The state is attached in the outer `Tactic.handle`, below which tacticals catch their own errors, so failures swallowed by `try`/`orelse` are unaffected. The LSP server disables the attachment for tactic failures (`Proof.state_on_error`) since editors display the goals themselves. Tests: `tests/proof_state.ml` on three new KO fixtures.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - LSP server: position of the error is removed from diagnostics when the error occurs in the file currently open in the editor.
 - Syntax of search query is modified as follows : `in` is used instead of `|` (filtering). `with` is used instead of `,` (conjunction). `|` is used instead of `;` (disjunction).
 - Type of `#assume` in order to generate a new symbol and use it inside a tactic term.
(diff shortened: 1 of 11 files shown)


fblanqui:
Thanks @ciaran-matthew-dunne !

PR #1467, opened 2026-07-21
emacs: update package versions + clean installAndLaunch.sh

fblanqui:


editors/emacs/installAndLaunch.sh
```diff
@@ -41,25 +41,26 @@ cat <<'EOF' > ~/.emacs.d/init.el
...
- mkdir -p ~/.emacs.d/elpa/
+echo "Creating ~/.emacs.d/elpa/ ..."
+mkdir -p ~/.emacs.d/elpa/
 
-echo "cloning dependencies repos"
+echo "Cloning dependencies repos"
 if [ ! -d ~/.emacs.d/elpa/eglot ]; then
   if [[ ${EGLOT_V} == "0" ]]; then  # ignore branch
     git clone --depth 1 https://github.com/joaotavora/eglot.git ~/.emacs.d/elpa/eglot
-    EGLOT_V="1.9"
```
fblanqui, line 51: What's the point of the comment line 49 "ignore branch"?

editors/emacs/installAndLaunch.sh
```diff
@@ -41,25 +41,26 @@ cat <<'EOF' > ~/.emacs.d/init.el
...
+    EGLOT_V=1.9
```
fblanqui: Why EGLOT_V is reset to 1.9?

editors/emacs/installAndLaunch.sh
```diff
@@ -41,25 +41,26 @@ cat <<'EOF' > ~/.emacs.d/init.el
...
   echo "Eglot cloned to " ~/.emacs.d/elpa/eglot
 else
   echo "Eglot is already cloned. Skipping"
 fi
+
 if [ ! -d ~/.emacs.d/elpa/math-symbol-lists ]; then
   if [[ ${MATH_SYMB_V} == "0" ]]; then # ignore branch
     git clone --depth 1 https://github.com/vspinu/math-symbol-lists.git ~/.emacs.d/elpa/math-symbol-lists
-    MATH_SYMB_V=1.2.1
+    MATH_SYMB_V=1.3
```
fblanqui: Why MATH_SYMB is reset to 1.2.1?

editors/emacs/installAndLaunch.sh
fblanqui: Why HIGHLIGHT_V is reset to 20250815.1830?

editors/emacs/installAndLaunch.sh
```diff
@@ -91,59 +92,10 @@ echo "(define-package \"math-symbol-lists\" \"${MATH_SYMB_V}\")" > ~/.emacs.d/el
...
-# echo
-# PATH="$BIN:$PATH" emacs \
-#   --batch \
-#   -l ~/.emacs.d/init.el \
-#   --eval="(package-refresh-contents) "
-
-# echo "\
-# " >> ~/.emacs.d/init.el
-
-
```
fblanqui, line 149: Is it ok to remove all the previous comments?

fblanqui (commented):
Hi @Alidra . Please review and fix this PR.

editors/emacs/lambdapi-mode.el
```diff
@@ -9,7 +9,7 @@
 ;; Homepage: https://github.com/Deducteam/lambdapi
 ;; Keywords: languages
 ;; Compatibility: GNU Emacs 27.1
```
fblanqui, line 11: line 11: upgrade Emacs version to 30.2?

fblanqui:
Other problem to solve: CI runs main.yml and ocaml.yml while it should not.

editors/emacs/installAndLaunch.sh
```diff
@@ -41,47 +35,42 @@ cat <<'EOF' > ~/.emacs.d/init.el
...
+}
+
+clone $EGLOT_V https://github.com/joaotavora/eglot.git
+if [[ $EGLOT_V == "0" ]]; then EGLOT_V=`version eglot`; fi
+
+clone $MATH_SYMB_V https://github.com/vspinu/math-symbol-lists.git
+if [[ $MATH_SYMB_V == "0" ]]; then MATH_SYMB_V=`version math-symbol-lists`; fi
+
+clone $HIGHLIGHT_V https://github.com/emacsmirror/highlight.git
+if [[ $HIGHLIGHT_V == "0" ]]; then HIGHLIGHT_V=`version highlight`; fi
```
fblanqui: What is the effect of the lines below when the variables have been reset?

Alidra:
Thanks @fblanqui for this PR. Following are my comments

.github/workflows/emacs.yml
```diff
@@ -14,9 +14,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        eglot-version: [1.9, 0]
-        math-symbol-list-version: [1.2.1, 1.3, 0]
-        highlight-version: [20250724.2154, 20250815.1830, 0]
+        eglot-version: [0]
```
Alidra: Do we drop checking lambdapi-mode with older versions and test only with latest ones? 
In this case, we can simply install the dependencies from the ELPA repository and simplify further the script.

fblanqui:
Hi. I do not see any comment yet.

editors/emacs/installAndLaunch.sh
```diff
@@ -1,35 +1,29 @@
...
-
-# For instance, one can run the scrupt with ./installAndLaunch.sh lambdapi-mode 1.1.0 /snap/bin/emacs 1.5 1.3 20210318.2248 (Does not work because Eglot.V.1.5 is too old)
-# ./installAndLaunch.sh lambdapi-mode 1.1.0 /snap/bin/emacs 1.18 1.3 20210318.2248 Does not work because Eglot repo stoped using tags since 2012.
-# For using the latest commit of a library just use version 0. For instance : ./installAndLaunch.sh lambdapi-mode 1.1.0 /snap/bin/emacs 0 1.3 0
+NAME="$1" # lambdapi
+VERSION="$2" # defined in lambdapi-mode.el
+BIN="$3" # emacs binary, e.g. /snap/bin/emacs
+EGLOT_V="$4" # "0" to use the latest version
+MATH_SYMB_V="$5" # "0" to use the latest version
+HIGHLIGHT_V="$6" # "0" to use the latest version
```
Alidra: Maybe add : "20210318.2248" to use version corresponding to commit dated 03/18/2021

editors/emacs/installAndLaunch.sh
```diff
@@ -1,35 +1,29 @@
...
 
 convertVersionToCommitDate() {
```
Alidra: This function is no more used and useless if we want to drop checking previous versions of highlight specifically.

editors/emacs/installAndLaunch.sh
Alidra: It can be removed

editors/emacs/installAndLaunch.sh
```diff
@@ -41,47 +35,42 @@ cat <<'EOF' > ~/.emacs.d/init.el
...
+    sed -n -E "/;; Package-Requires:/ s/.*\(\b${pkg}\b +\"([^\"]+)\"\).*/\1/p" lambdapi-mode.el
+}
+
+clone $EGLOT_V https://github.com/joaotavora/eglot.git
+if [[ $EGLOT_V == "0" ]]; then EGLOT_V=`version eglot`; fi
+
+clone $MATH_SYMB_V https://github.com/vspinu/math-symbol-lists.git
+if [[ $MATH_SYMB_V == "0" ]]; then MATH_SYMB_V=`version math-symbol-lists`; fi
+
+clone $HIGHLIGHT_V https://github.com/emacsmirror/highlight.git
```
Alidra: Please note Highlight is checkedout from a commit and not a branch (only one branch, master. Versions in Emacs correspond to commit dates)

editors/emacs/installAndLaunch.sh
Alidra: following lines are intend to "inform" ELPA that specific versions of the dependency packages have been installed in order to fulfill Lambdapi-mode requirements

editors/emacs/installAndLaunch.sh
Alidra: OK

Alidra (changes requested):

editors/emacs/lambdapi-mode.el
```diff
@@ -9,7 +9,7 @@
...
-;; Package-Requires: ((emacs "27.1") (eglot "1.6") (math-symbol-lists "1.2.1") (highlight "20190710.1527"))
+;; Package-Requires: ((emacs "30.2") (eglot "1.9") (math-symbol-lists "1.3.0") (highlight "20250815.1830"))
```
Alidra: Please note `lambdapi-mode` **is** compatible with with previous versions. Changing to latest version of `Eglot`, `math-symbol-list` and `highlight` will prevent installing it with old configurations in Emacs (especially that Highlight latest version available in Elpa is 20210318.2248)

editors/emacs/lambdapi-mode.el
Alidra: I tested with Emacs 28.2 and `lambdapi-mode` seems to work correctly (provided that constraint on `Eglot` version is relaxed to version `1.24`)

Alidra (changes requested):

.github/workflows/emacs.yml
fblanqui: We can keep this option in case we need it later.

editors/emacs/installAndLaunch.sh
fblanqui: I don't understand that comment at this line.

editors/emacs/lambdapi-mode.el
fblanqui: What the comment `Compatibility: GNU Emacs 27.1` actually means? Is it just a comment or does it have some effect?

editors/emacs/lambdapi-mode.el
fblanqui: I see. So these requirements are the minimum ones? Does that mean that any combination of newer versions should work?

Alidra:
> Other problem to solve: CI runs main.yml and ocaml.yml while it should not.

Yes. Actually, this is because the `paths` only applies to the `push` section and not to the `pull_request` one. section `pahts` should be replicated in both sections

editors/emacs/lambdapi-mode.el
Alidra: `Compatibility: GNU Emacs 27.1` is a compatibility declaration used by the Emacs package ecosystem. It indicates that the package is intended to be compatible with GNU Emacs 27.1 and later.

By itself, it does not enforce anything at runtime.
It is therefore mainly metadata/documentation, whereas:

```;; Package-Requires: ((emacs "30.2"))```
is an actual package dependency declaration used by the package manager.

editors/emacs/lambdapi-mode.el
Alidra: Yes. I believe so

fblanqui:
> > Other problem to solve: CI runs main.yml and ocaml.yml while it should not.
> 
> Yes. Actually, this is because the `paths` only applies to the `push` section and not to the `pull_request` one. section `pahts` should be replicated in both sections

As shown by the last commit, this does not seem to be sufficient. Could you please fix this in a separate PR?

Alidra:
Weird

> > > Other problem to solve: CI runs main.yml and ocaml.yml while it should not.
> > 
> > 
> > Yes. Actually, this is because the `paths` only applies to the `push` section and not to the `pull_request` one. section `pahts` should be replicated in both sections
> 
> As shown by the last commit, this does not seem to be sufficient. Could you please fix this in a separate PR?

Weird indeed. I tested on my fork and it seems the workflows keeps be triggered with previous commits!!

editors/emacs/installAndLaunch.sh
fblanqui: The problem is that the version numbers are set to wrong number when we test the last versions with 0. How to get the last version numbers from the git repos?

fblanqui:
Another problem is that the script does not test different versions of emacs. We have no idea whether we actually support all the versions greater or equal to the ones indicated in the lambdapi-mode.el file.

Alidra:
> > 
> > As shown by the last commit, this does not seem to be sufficient. Could you please fix this in a separate PR?
> 
> Weird indeed. I tested on my fork and it seems the workflows keeps be triggered with previous commits!!
Actually, the pipelines are triggered by past commits in the PR even though theses commits have already triggered the same workflows. Then, workflows are re-run each time the PR is updated (In fact; one would expect that only workflows that are affected by the last commit be triggered but this does not seem to be the case)

editors/emacs/installAndLaunch.sh
fblanqui: Fixed in the last commits.

fblanqui:
@Alidra I made further fixes and simplifications. Thanks to review the PR again.

Issue #1498, opened 2026-09-01
Error not shown on LP console when occurs in a different file


Alidra:
When an `LP` file `A` `require open` another file `B` and that `B` contains an error, the location of that error in the logs is the one in file `B`. The consequence is that when file A is excuted in VScode for instance, the error doesn't appear in the consol if its position in file `B` is superior to the current location even though the error is related to the `require open` command which maybe at the top of the file `A`

## To reproduce
  open file `tests/OK/natural.lp` and add an error, for instance `rrrrr` at line 7. open file `tests/OK/215.lp` which opens `tests/OK/natural` at line 1.
Then a diagnostic is assiciated to line 1 with the correct error message. However, the error message related to `tests/OK/natural.lp` is not shown.

fblanqui:
fixed in #1499

Issue #1456, opened 2026-07-10
Missing position in error message


fblanqui:
With the example in https://github.com/Deducteam/lambdapi/issues/1436 (tests/OK/apply.lp with "require open Stdlib.Prop;" added at the beginning, the displayed error message in VSCode and Emacs is:
```
file:///home/blanqui/src/lambdapi/tests/OK/apply.lp:1:0-23"/home/blanqui/.opam/5.5.0/.opam-switch/build/lambdapi-stdlib.1.4.0/Prop.lp" cannot be mapped under the library root.
Consider adding a package file under your source tree, or use the [--map-dir] option.
```
instead of:
```
file:///home/blanqui/src/lambdapi/tests/OK/apply.lp:1:0-23
"/home/blanqui/.opam/5.5.0/.opam-switch/build/lambdapi-stdlib.1.4.0/Prop.lp" cannot be mapped under the library root.
Consider adding a package file under your source tree, or use the [--map-dir] option.
```
or perhaps:
```
tests/OK/apply.lp:1:0-23
"/home/blanqui/.opam/5.5.0/.opam-switch/build/lambdapi-stdlib.1.4.0/Prop.lp" cannot be mapped under the library root.
Consider adding a package file under your source tree, or use the [--map-dir] option.
```

Now, if I remove "require open Stdlib.Prop;" and replace "constant symbol A : TYPE;" by "constant symbol 0 : TYPE;", I get:
```
Syntax error. Expected: "non-qualified identifier".
```
with no position information.

Alidra:
This PR seems outdated especially after the latest fixes on the parser

fblanqui:
The first problem is now fixed indeed thanks to #1443 . It remains that the second problem that the position is not reported in Emacs nor in VSCode.

fblanqui:
fixed in #1499

Issue #1405, opened 2026-06-10
error message not correctly printed in console


fblanqui:
Take any file with an error (but not a syntax error). For instance:
```
symbol a:TYPE;
symbol b:TYPE;
symbol x:a;
symbol y:b;
rule x ↪ y;
```
In Emacs or VSCode, what is printed in the console is:
```
file:///home/blanqui/src/lambdapi/tmp/zo.lp:5:5-9Cannot solve a ≡ b
Unable to prove type preservation.
```
with no color and no brackets around the position,
while it is correctly printed when running lambdapi check in a terminal.

fblanqui:
fixed in #1499

PR #1499, opened 2026-09-01
Fix error location reporting for required files in LP console

Alidra:
## Summary
This PR fixes :
-1 The location is displayed in the LP console when a syntax error occurs (issue #1456)
-2 Errors are displayed in Red color and location is put between bracket (issue #1405)
-3 When an LP file requires another file and the error originates in the required file, the location of the `require open` command is used so that when the "requiring" file is executed the error is displayed in the terminal when the user reaches the require command (issue    #1498)

## Changes
- keep the precise  from the originating error when an error is reported from a required module
- prefer the real error location when it points to a different file than the current command
- add a space separator between the file location and the message in the console output
- ensure logs include the position in a readable format for debugging and traceability

### DIFF
--- a/src/lsp/lp_doc.ml
+++ b/src/lsp/lp_doc.ml
@@ -133,6 +133,7 @@ let process_cmd _file (nodes,st,dg,logs) cmd =
   | Cmd_Error(err_loc, err_msg) ->
     let nodes = { cmd; exec = false; goals = [] } :: nodes in
     let loc, diag_msg, log_msg = match cmd_loc, err_loc with
+    | _, Some None -> cmd_loc, err_msg, err_msg
     | Some l, Some Some l' ->
         if l.fname = l'.fname then
           (* if the error is in the same file as the command,
@@ -155,8 +156,9 @@ let process_cmd _file (nodes,st,dg,logs) cmd =
           cmd_loc,
           Pos.popt_to_string (Some l') ^ "\n" ^ err_msg,
           Pos.popt_to_string (Some l') ^ "\n" ^ err_msg
-    | _, Some l' -> cmd_loc, err_msg, Pos.popt_to_string (l') ^ "\n" ^ err_msg
-    | _, None -> cmd_loc, err_msg, err_msg in
+    | None, _ -> assert false
+    | _, None -> assert false
+    in
     nodes, st, (loc, 1, diag_msg, None) :: dg, ((1, log_msg), loc) :: logs
 
 let new_doc ~uri ~version ~text =
--- a/src/lsp/lp_doc.ml
+++ b/src/lsp/lp_doc.ml
@@ -132,20 +132,20 @@ let process_cmd _file (nodes,st,dg,logs) cmd =
 
   | Cmd_Error(loc, msg) ->
     let nodes = { cmd; exec = false; goals = [] } :: nodes in
-    let cmd_loc, loc, diag, log = match cmd_loc, loc with
+    let cmd_loc, diag, log = match cmd_loc, loc with
     | Some l, Some Some l' ->
         if l.fname = l'.fname then
           (* if error in the same file, use the precise location *)
-          Some l', Some l', msg, Pos.popt_to_string (Some l') ^ msg
+          Some l', msg, Pos.popt_to_string (Some l') ^ msg
         else
           (* else, use the location of the command *)
-          cmd_loc, Some l', Pos.popt_to_string (Some l') ^ "\n" ^ msg
+          cmd_loc, Pos.popt_to_string (Some l') ^ "\n" ^ msg
             , Pos.popt_to_string (Some l') ^ "\n" ^ msg
     (* Otherwise,
       cmd_loc doesn't change and loc is : option_default loc cmd_loc *)
-    | _, Some l' -> cmd_loc, l', msg, Pos.popt_to_string (l') ^ "\n" ^ msg
-    | _, None -> cmd_loc, cmd_loc, msg, msg in
-    nodes, st, (cmd_loc, 1, diag, None) :: dg, ((1, log), loc) :: logs
+    | _, Some l' -> cmd_loc, msg, Pos.popt_to_string (l') ^ "\n" ^ msg
+    | _, None -> cmd_loc, msg, msg in
+    nodes, st, (cmd_loc, 1, diag, None) :: dg, ((1, log), cmd_loc) :: logs
 
 let new_doc ~uri ~version ~text =
   let root, logs =
--- a/src/lsp/lp_doc.ml
+++ b/src/lsp/lp_doc.ml
@@ -212,7 +212,13 @@ let check_text ~doc =
     match error with
     | None -> logs, diags
     | Some(pos,msg) ->
-      logs @ [((1, (Pos.popt_to_string (Some pos) ^ ": " ^ msg)), Some pos)],
+      logs @ [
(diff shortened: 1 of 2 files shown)


src/lsp/lp_doc.ml
```diff
@@ -132,20 +132,26 @@ let process_cmd _file (nodes,st,dg,logs) cmd =
 
   | Cmd_Error(loc, msg) ->
     let nodes = { cmd; exec = false; goals = [] } :: nodes in
-    let cmd_loc, loc, diag, log = match cmd_loc, loc with
+    let cmd_loc, diag, log = match cmd_loc, loc with
```
fblanqui: diag should be renamed into diag_msg as it is not a diagnostic but a message.
Idem for log which should be renamed into log_msg.
In addition, to better understand the code, I also propose to rename cmd_loc into loc:
```
let loc, diag_msg, log_msg =
  match cmd_loc, err_loc with
  ...
in
nodes, st, (loc, 1, diag_msg, None) :: dg, ((1, log_msg), loc) :: logs
```

src/lsp/lp_doc.ml
```diff
@@ -132,20 +132,26 @@ let process_cmd _file (nodes,st,dg,logs) cmd =
 
   | Cmd_Error(loc, msg) ->
```
fblanqui: please rename loc and msg into err_loc and err_msg respectively to understand the code more easily

src/lsp/lp_doc.ml
```diff
@@ -132,20 +132,26 @@ let process_cmd _file (nodes,st,dg,logs) cmd =
...
     let nodes = { cmd; exec = false; goals = [] } :: nodes in
-    let cmd_loc, loc, diag, log = match cmd_loc, loc with
+    let cmd_loc, diag, log = match cmd_loc, loc with
     | Some l, Some Some l' ->
         if l.fname = l'.fname then
           (* if error in the same file, use the precise location *)
```
fblanqui: if the error is in the same file as the command, we set the diag/log position to the error position and add the error position in the log message

src/lsp/lp_doc.ml
```diff
@@ -132,20 +132,26 @@ let process_cmd _file (nodes,st,dg,logs) cmd =
...
-          Some l', Some l', msg, Pos.popt_to_string (Some l') ^ msg
+          Some l',
+          msg,
+          Pos.popt_to_string
+            ~print_dirname:false
+            ~print_fname:false
+            (Some l') ^ " " ^ msg
         else
           (* else, use the location of the command *)
```
fblanqui: otherwise we set the diag/log position to the command position and add the error position to both the diag and log messages

src/lsp/lp_doc.ml
```diff
@@ -132,20 +132,26 @@ let process_cmd _file (nodes,st,dg,logs) cmd =
...
-          cmd_loc, Some l', Pos.popt_to_string (Some l') ^ "\n" ^ msg
-            , Pos.popt_to_string (Some l') ^ "\n" ^ msg
+          cmd_loc,
+          Pos.popt_to_string (Some l') ^ "\n" ^ msg,
+          Pos.popt_to_string (Some l') ^ "\n" ^ msg
     (* Otherwise,
       cmd_loc doesn't change and loc is : option_default loc cmd_loc *)
```
fblanqui: this comment is no longer up to date and should be removed

src/lsp/lp_doc.ml
fblanqui: Could you please add also a comment in pure.mli to explain why Cmd_Error takes a `popt option` as argument (instead of just a popt) by referring to the comment given in error.ml (Cmd_Error inherits this from Fatal)?

src/lsp/lp_doc.ml
fblanqui: Isn't it the case that cmd_loc is never equal to None? If so, it would be nice to add:
```
| None, _ -> assert false
```

fblanqui: Similarly, isn't it the case that err_loc is never equal to None? If so, it would be nice to add:
```
| _, None -> assert false
```

src/lsp/lp_doc.ml
```diff
@@ -132,20 +132,26 @@ let process_cmd _file (nodes,st,dg,logs) cmd =
...
-    | _, Some l' -> cmd_loc, l', msg, Pos.popt_to_string (l') ^ "\n" ^ msg
-    | _, None -> cmd_loc, cmd_loc, msg, msg in
-    nodes, st, (cmd_loc, 1, diag, None) :: dg, ((1, log), loc) :: logs
+    | _, Some l' -> cmd_loc, msg, Pos.popt_to_string (l') ^ "\n" ^ msg
```
fblanqui: Please remove parentheses around l'.

fblanqui: Isn't the case that we always have l'=None here? If so, it would be better to write:
```
| Some _, Some None ->
   (* error with no position *)
   cmd_loc, msg, Pos.popt_to_string l' ^ "\n" ^ msg
```
It would be interesting to later investigate if this may actually happen.

fblanqui (changes requested):
Thanks for your PR.

src/lsp/lp_doc.ml
```diff
@@ -130,22 +130,36 @@ let process_cmd _file (nodes,st,dg,logs) cmd =
...
-    | _, Some l' -> cmd_loc, l', msg, Pos.popt_to_string (l') ^ "\n" ^ msg
-    | _, None -> cmd_loc, cmd_loc, msg, msg in
-    nodes, st, (cmd_loc, 1, diag, None) :: dg, ((1, log), loc) :: logs
+          (* otherwise we set the diag/log position to the command position
+            and add the error position to both the diag and log messages *)
+          cmd_loc,
+          Pos.popt_to_string (Some l') ^ "\n" ^ err_msg,
+          Pos.popt_to_string (Some l') ^ "\n" ^ err_msg
+    | None, _ -> assert false
+    | _, None -> assert false
```
fblanqui, line 160: Could you please move these two assert false lines at the beginning of the match?

src/lsp/lp_doc.ml
```diff
@@ -130,22 +130,36 @@ let process_cmd _file (nodes,st,dg,logs) cmd =
     in
     nodes, st, dg_proof @ dg, logs
 
-  | Cmd_Error(loc, msg) ->
+  | Cmd_Error(err_loc, err_msg) ->
     let nodes = { cmd; exec = false; goals = [] } :: nodes in
-    let cmd_loc, loc, diag, log = match cmd_loc, loc with
+    let loc, diag_msg, log_msg = match cmd_loc, err_loc with
+    | _, Some None -> cmd_loc, err_msg, err_msg
```
fblanqui, line 136: Could you please add a comment: in case there is no error position, we use the command position.

src/lsp/lp_doc.ml
```diff
@@ -205,7 +219,19 @@ let check_text ~doc =
   let logs, diags =
     match error with
     | None -> logs, diags
-    | Some(pos,msg) -> logs @ [((1, msg), Some pos)], diags @ [pos,1,msg,None]
+    | Some(pos,msg) ->
```
fblanqui, line 222: Please rename pos into err_pos, and msg into err_msg.

PR #1505, opened 2026-09-06
adds a #abstract tactic

bodeveix:
Inside `eval`, the tactic `#abstract e P t' abstracts `e` from `P` as `Q e` and calls the tactic `t` with `Q` and the proof `P => Q e` as parameters. As an example, the following code makes a rewrite using `#abstract` to instantiate `ind_eq`: 

```
symbol test (a b: τ T) (h: π (a = b)): π (f a = f b) ≔
begin
  assume a b h;
  // f a = f b is seen as p a where p x = (f x = f b)
  eval (#abstract a (f a = f b) (λ p _, #apply (ind_eq [T] [a] [b] h p)));
  reflexivity
end;
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Tactic `all_hyps t` calls parameterized tactic term t on all hypotheses ignoring failing calls.
 - Extend `print` query to the following arguments: `verbose`, `debug`, `flag`, `builtin`, `prover`, `prover_timeout`.
 - add a version number to the header of the index db file to prevent crash of LP when the structure of db changes.
-
+- tactic `#abstract e P t` abstracts e from P and calls tactic `t` with the abstraction as parameter
 ### Changed
 
 - Tactic `simplify` now fails if the goal cannot be simplified.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Extend `print` query to the following arguments: `verbose`, `debug`, `flag`, `builtin`, `prover`, `prover_timeout`.
 - add a version number to the header of the index db file to prevent crash of LP when the structure of db changes.
 - tactic `#abstract e P t` abstracts e from P and calls tactic `t` with the abstraction as parameter
+
 ### Changed
 
 - Tactic `simplify` now fails if the goal cannot be simplified.
--- a/doc/tacticals.rst
+++ b/doc/tacticals.rst
@@ -25,9 +25,8 @@ The BNF grammar of tactics is in `lambdapi.bnf <https://raw.githubusercontent.co
    builtin "Type" ≔ … // : Π [l:L], U l → TYPE, written η below
    builtin "Tactic" ≔ …; // : TYPE, written T below
    builtin "admit" ≔ …; // : T
-   builtin "abstract" ≔ … // : Π [l:L] [a:U l] (e: η a) (P:Prop), 
-                   (Π (Q:η a → Prop), (π P → π (Q e)) → Tactic) → Tactic;
-
+   builtin "abstract" ≔ … // : Π [l l':L] [a:U l] (e: η a) (P:U l'), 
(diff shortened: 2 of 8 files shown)


doc/tacticals.rst
```diff
@@ -25,6 +25,9 @@ The BNF grammar of tactics is in `lambdapi.bnf <https://raw.githubusercontent.co
    builtin "Type" ≔ … // : Π [l:L], U l → TYPE, written η below
    builtin "Tactic" ≔ …; // : TYPE, written T below
    builtin "admit" ≔ …; // : T
+   builtin "abstract" ≔ … // : Π [l:L] [a:U l] (e: η a) (P:Prop), 
+                   (Π (Q:η a → Prop), (π P → π (Q e)) → Tactic) → Tactic;
```
fblanqui: parentheses around Q:_ are uncessary

src/handle/tactic.ml
```diff
@@ -399,6 +401,30 @@ let handle (ss:Sig_state.t) (sym_pos:popt) (priv:bool)
...
             let c = get_config ss pos in
             match Hashtbl.find c s.sym_name, ts with
+            | T_abstract, [_; _; e; t; Abst (u, bi)] -> begin
+               match Eval.whnf ctx u with
+                 Prod(tte,_) -> begin
+                   let p = Rewrite.bind_pattern e t in
+                   let bi = subst bi (mk_Abst (tte,p)) in
+                   let ctx = Env.to_ctxt env in
+                   let bi = match Eval.whnf ctx bi with
+                       Abst (ty, bi) ->
```
fblanqui: please prefix all OCaml patterns by "|" (idem on line 413)

CHANGES.md
```diff
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Tactic `all_hyps t` calls parameterized tactic term t on all hypotheses ignoring failing calls.
 - Extend `print` query to the following arguments: `verbose`, `debug`, `flag`, `builtin`, `prover`, `prover_timeout`.
 - add a version number to the header of the index db file to prevent crash of LP when the structure of db changes.
-
+- tactic `#abstract e P t` abstracts e from P and calls tactic `t` with the abstraction as parameter
 ### Changed
```
fblanqui, line 29: empty line missing before ###

src/core/builtin.ml
```diff
@@ -215,6 +215,20 @@ let _ =
             arr (arr (eps (var l) (var a)) tac) tac))
   in
   register_typ "admit" tac;
+  (*
+    Π [l:L] [a:U l] (e: η a) (P:Prop),
+    (Π (Q:η a → Prop), π (P ⇒ Q e) → Tactic) → Tactic;
```
fblanqui: why restricting to Prop?

fblanqui (changes requested):
Hi JP. An entry in the user manual explaining this new tactic is missing.
Your explanations above are unclear:
- what are the types of e, P and t'?
- what are t and Q?
- why the tactic does not depend on the current goal?

I would rather expect something like `eval #abstract a f;` with
`#abstract: Π [l l':L], U l → ((U l → U l') → Tactic) → Tactic`
having the effect of abstracting `a:U l` in the current goal, giving `g':U l'`, and applying `f` to `λ a:U l, g'`.

Actually, why not simply add a new tactic `pattern a`, with `a:U l`, which would transform the current goal `g:U l'` into `(λ l (x:U l), g') l a`?

==> So I have replaced Prop by U l'. No change in the code! 

But the tactic should apply to any formula, not only the goal. So I have kept the initial behaviour. Maybe it possible to define pattern on top of it.

Other remarks:

CHANGES.md
bodeveix: OK

src/core/builtin.ml
bodeveix: OK I have introduced l' and replaced Prop by U l', ...

src/handle/tactic.ml

doc/tacticals.rst
bodeveix: OK

